PR #359, opened 2023-05-09
[JN-307, JN-308] Admin dashboard functionality (list, view, create datasets)

MatthewBemis:
This introduces some functionality into the admin UI for managing TDR datasets. The goal is to make it easier for developers and study admins to see and manage datasets for their study environments.

Other functionality I'd like to add in the future:
* Delete dataset
* View and cleanup orphaned TDR datasets (might make this a developer script since it shouldn't really happen in persistent environments. But it's common if you clear your local DB, which leaves dangling TDR datasets that conflict with new dataset creations from your local Juniper)

Screenshots:

Study environment dataset list
Screenshot 2023-05-09 at 9 43 02 AM

Dataset summary/job history
Screenshot 2023-05-09 at 9 43 15 AM

Create dataset modal (limited to superusers)
Screenshot 2023-05-09 at 9 42 44 AM

**TO TEST:**

Make sure you have all of your local environment variables set, [per these instructions](https://broadworkbench.atlassian.net/wiki/spaces/PEARL/pages/2708471863/TDR+Export#Setting-up-TDR-export-locally%3A).

1. Boot ApiAdminApp
2. Boot admin UI
3. Make sure OurHealth portal is populated
4. Navigate to the participants list, i.e. https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants
5. Click "Terra Data Repo" and verify that you can view existing datasets, create new datasets, and view the details for a dataset.

There is a bit of a quirk to this UX that I'll iron out in a followup PR: you won't immediately see the dataset in the dataset list after creation, but it should show up within 10 mins (I may tighten the poll time in a future PR since 10 mins is a bit too slow). I'd like to display some information to the user that shows in progress or failed dataset creations. This will likely involve pre-populating a partial row in our `dataset` DB table with some information to return in the `listDatasetsForStudyEnvironment` call.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/datarepo/DataRepoExportController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/datarepo/DataRepoExportController.java
@@ -1,0 +1,69 @@
+package bio.terra.pearl.api.admin.controller.datarepo;
+
+import bio.terra.pearl.api.admin.api.DatarepoApi;
+import bio.terra.pearl.api.admin.model.DatasetName;
+import bio.terra.pearl.api.admin.service.AuthUtilService;
+import bio.terra.pearl.api.admin.service.DataRepoExportExtService;
+import bio.terra.pearl.core.model.EnvironmentName;
+import bio.terra.pearl.core.model.admin.AdminUser;
+import bio.terra.pearl.core.model.datarepo.DataRepoJob;
+import bio.terra.pearl.core.model.datarepo.Dataset;
+import java.util.List;
+import javax.servlet.http.HttpServletRequest;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+
+@Controller
+public class DataRepoExportController implements DatarepoApi {
+
+  private AuthUtilService authUtilService;
+  private DataRepoExportExtService dataRepoExportExtService;
+  private HttpServletRequest request;
+
+  public DataRepoExportController(
+      AuthUtilService authUtilService,
+      DataRepoExportExtService dataRepoExportExtService,
+      HttpServletRequest request) {
+    this.authUtilService = authUtilService;
+    this.dataRepoExportExtService = dataRepoExportExtService;
+    this.request = request;
+  }
+
+  @Override
+  public ResponseEntity<Object> listDatasetsForStudyEnvironment(
+      String portalShortcode, String studyShortcode, String envName) {
+    AdminUser user = authUtilService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+
+    List<Dataset> datasets =
+        dataRepoExportExtService.listDatasetsForStudyEnvironment(
+            portalShortcode, studyShortcode, environmentName, user);
+
+    return ResponseEntity.ok().body(datasets);
+  }
+
+  @Override
+  public ResponseEntity<Object> getJobHistoryForDataset(
+      String portalShortcode, String studyShortcode, String envName, String datasetName) {
+    AdminUser user = authUtilService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+
+    List<DataRepoJob> jobHistory =
+        dataRepoExportExtService.getJobHistoryForDataset(
+            portalShortcode, studyShortcode, environmentName, datasetName, user);
+
+    return ResponseEntity.ok().body(jobHistory);
+  }
+
+  @Override
+  public ResponseEntity<Void> createDatasetForStudyEnvironment(
+      String portalShortcode, String studyShortcode, String envName, DatasetName datasetName) {
+    AdminUser user = authUtilService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+
+    dataRepoExportExtService.createDataset(
+        portalShortcode, studyShortcode, environmentName, datasetName, user);
+
+    return ResponseEntity.accepted().build();
+  }
+}
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/DataRepoExportExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/DataRepoExportExtService.java
@@ -1,0 +1,78 @@
+package bio.terra.pearl.api.admin.service;
+
+import bio.terra.pearl.api.admin.model.DatasetName;
+import bio.terra.pearl.core.model.EnvironmentName;
+import bio.terra.pearl.core.model.admin.AdminUser;
+import bio.terra.pearl.core.model.datarepo.DataRepoJob;
+import bio.terra.pearl.core.model.datarepo.Dataset;
+import bio.terra.pearl.core.model.portal.Portal;
+import bio.terra.pearl.core.model.study.StudyEnvironment;
+import bio.terra.pearl.core.service.datarepo.DataRepoExportService;
+import bio.terra.pearl.core.service.exception.PermissionDeniedException;
+import bio.terra.pearl.core.service.study.StudyEnvironmentService;
+import java.util.List;
+import org.springframework.stereotype.Service;
+
+@Service
+public class DataRepoExportExtService {
+
+  private AuthUtilService authUtilService;
+  private DataRepoExportService dataRepoExportService;
+  private StudyEnvironmentService studyEnvironmentService;
+
+  public DataRepoExportExtService(
+      AuthUtilService authUtilService,
+      DataRepoExportService dataRepoExportService,
+      StudyEnvironmentService studyEnvironmentService) {
+    this.authUtilService = authUtilService;
+    this.dataRepoExportService = dataRepoExportService;
+    this.studyEnvironmentService = studyEnvironmentService;
+  }
+
+  public List<Dataset> listDatasetsForStudyEnvironment(
+      String portalShortcode,
+      String studyShortcode,
+      EnvironmentName environmentName,
+      AdminUser user) {
+    Portal portal = authUtilService.authUserToPortal(user, portalShortcode);
+    authUtilService.authUserToStudy(user, portalShortcode, studyShortcode);
+
+    StudyEnvironment studyEnv =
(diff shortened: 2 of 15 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/service/DataRepoExportExtService.java
```diff
@@ -0,0 +1,78 @@
...
+
+  public void createDataset(
+      String portalShortcode,
+      String studyShortcode,
+      EnvironmentName environmentName,
+      DatasetName datasetName,
+      AdminUser user) {
+    if (!user.isSuperuser()) {
+      throw new PermissionDeniedException("You do not have permissions to perform this operation");
+    }
```
MatthewBemis, line 69: createDataset is being limited to superusers at the moment. The other endpoints are open to anyone who is an admin on the study.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/datarepo/DataRepoExportController.java
```diff
@@ -0,0 +1,69 @@
...
+  @Override
+  public ResponseEntity<Void> createDatasetForStudyEnvironment(
+      String portalShortcode, String studyShortcode, String envName, DatasetName datasetName) {
+    AdminUser user = authUtilService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+
+    dataRepoExportExtService.createDataset(
+        portalShortcode, studyShortcode, environmentName, datasetName, user);
+
+    return ResponseEntity.accepted().build();
```
devonbush, line 67: Agree with your comment about keeping local records of the datasets being created.  Related, this method should either return a Dataset with a status of "CREATING", or it should return the DataRepoJob

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  title: Pearl Admin API
+  title: Juniper Admin API
```
devonbush, line 3: lol, nice catch

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
```diff
@@ -102,24 +104,32 @@ public void createDatasetsForStudyEnvironments() {
...
         }
 
+        DataRepoJob job = DataRepoJob.builder()
+                .studyEnvironmentId(studyEnv.getId())
+                .status(response.getJobStatus().getValue())
+                .datasetName(datasetName)
+                .tdrJobId(response.getId())
+                .jobType(JobType.CREATE_DATASET)
+                .build();
+
```
devonbush, line 131: no reason for this method not to return the job

ui-admin/src/study/participants/datarepo/DatasetDashboard.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+const DatasetDashboard = ({ studyEnvContext }: {studyEnvContext: StudyEnvContextT}) => {
+  const { currentEnvPath } = studyEnvContext
+  const [datasetDetails, setDatasetDetails] = useState<DatasetDetails | null>(null)
+  const [datasetJobHistory, setDatasetJobHistory] = useState<DatasetJobHistory[]>([])
+  const [isLoading, setIsLoading] = useState(true)
+  const [sorting, setSorting] = React.useState<SortingState>([])
+  const datasetName = useParams().datasetName as string
+
+  const columns = useMemo<ColumnDef<DatasetJobHistory, string>[]>(() => [{
```
devonbush: You can pull this column definition out of the component, since none of the column definitions themselves are dependent on the data.  see MailingListView.tsx for an example.  This will save a useMemo call and make the definitions clearer

ui-admin/src/study/participants/datarepo/DatasetDashboard.tsx
```diff
@@ -0,0 +1,165 @@
...
+                <a href={`https://jade.datarepo-dev.broadinstitute.org/datasets/${datasetDetails?.datasetId}`}
+                  target="_blank">View dataset in Terra Data Repo <FontAwesomeIcon icon={faExternalLink}/></a>
+              </div>
+            </div>
+          </li>
+          <li className="bg-white my-3">
+            <div style={contentHeaderStyle}>
+              <h6>Export History</h6>
+            </div>
+            <table className="table table-striped">
```
devonbush: since you're just using the basics, you can replace all the table markup with a call to `basicTableLayout` in tableUtils

ui-admin/src/study/participants/datarepo/DatasetList.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+const DatasetList = ({ studyEnvContext }: {studyEnvContext: StudyEnvContextT}) => {
+  const { currentEnvPath } = studyEnvContext
+  const [showCreateDatasetModal, setShowCreateDatasetModal] = useState(false)
+  const [datasets, setDatasets] = useState<DatasetDetails[]>([])
+  const [isLoading, setIsLoading] = useState(true)
+  const [datasetsSorting, setDatasetsSorting] = React.useState<SortingState>([])
+  const { user } = useUser()
+
+  const datasetColumns = useMemo<ColumnDef<DatasetDetails, string>[]>(() => {
```
devonbush: same comment about pulling the column defs out

ui-admin/src/study/participants/datarepo/DatasetList.tsx
```diff
@@ -0,0 +1,145 @@
...
+      show={showCreateDatasetModal}
+      setShow={setShowCreateDatasetModal}/>
+    <LoadingSpinner isLoading={isLoading}>
+      <div className="col-12 p-3">
+        <ul className="list-unstyled">
+          <li className="bg-white my-3">
+            <div style={contentHeaderStyle}>
+              <h6>Datasets</h6>
+            </div>
+            <table className="table table-striped">
```
devonbush: same, can be replaced with a call to `basicTableLayout`

devonbush (approved):
Generally looks good--big step forward in visibility.  I would like the table stuff cleaned up in this PR since it's easy and will save a lot of lines of code.  But the return values can be addressed in a later PR as you figure out the data model for representing in-progress tables.

devonbush:
Opened a PR into this branch adding bare bones error alerting on creation https://github.com/broadinstitute/pearl/pull/362

ui-admin/src/study/participants/datarepo/DatasetList.tsx
MatthewBemis: Done in c30b794f7fa660fc64a29ad06b1e3583583f76a1

ui-admin/src/study/participants/datarepo/DatasetDashboard.tsx
MatthewBemis: Done in c30b794f7fa660fc64a29ad06b1e3583583f76a1

ui-admin/src/study/participants/datarepo/DatasetDashboard.tsx
MatthewBemis: Sweet, done in 64eb80a3

ui-admin/src/study/participants/datarepo/DatasetList.tsx
MatthewBemis: Done in 64eb80a3

api-admin/src/test/java/bio/terra/pearl/api/admin/service/DataRepoExportExtServiceTests.java
```diff
@@ -0,0 +1,49 @@
...
+    Assertions.assertThrows(
+        PermissionDeniedException.class,
+        () ->
+            emptyService.createDataset(
+                "someportal",
+                "somestudy",
+                EnvironmentName.sandbox,
+                new DatasetName().name("somedataset"),
+                user));
+  }
```
breilly2, line 48: Could also add a test for a user who is allowed to perform these operations, but I like the focus on the more important case of making sure that people who shouldn't have access don't. Side note: In terms of the CLIA Risk field, I think of this as the difference between "Data availability" and "Data security/privacy".

ui-admin/src/study/participants/datarepo/DatasetDashboard.tsx
```diff
@@ -0,0 +1,142 @@
...
+    getSortedRowModel: getSortedRowModel(),
+    debugTable: true
+  })
+
+  const contentHeaderStyle = {
+    padding: '1em 0 0 1em',
+    borderBottom: '1px solid #f6f6f6'
+  }
+
+  const loadData = async () => {
```
breilly2, line 72: Nit: `useEffect` functions like this are usually declared inside `useEffect`. I don't remember all of the details about why. I think technically, since it's declared outside, it should be in the `useEffect` dependency list. However, in reality, this function may be stable across re-renders... which I think in this case is going to depend on whether or not `datasetName` is stable; I imagine it is because it's coming from the query string, but depending on how routing works, that's not universally guaranteed (though it might be in this project). (It's also possible that I missed other dependencies; I only noticed `datasetName` (and `studyEnvContext` which I'm pretty sure is fine because it's a prop).) If the function is in fact stable, then the only reason to put it in `useEffect` might be to make the linter happy... which doesn't seem to be a problem in Juniper, but Terra UI has a linter rule that does an exhaustive dependency check and would complain about this.

ui-admin/src/study/participants/datarepo/DatasetList.tsx
```diff
@@ -0,0 +1,119 @@
...
+    getCoreRowModel: getCoreRowModel(),
+    getSortedRowModel: getSortedRowModel()
+  })
+
+  const contentHeaderStyle = {
+    padding: '1em 0 0 1em',
+    borderBottom: '1px solid #f6f6f6'
+  }
+
+  const loadData = async () => {
```
breilly2, line 75: Same here, as in `DatasetDashboard` but without the `datasetName` consideration.

breilly2 (approved):
Looks good to me. I agree with Devon's comments about return values, and added another nit, but it works so I don't see it as blocking.

PR #361, opened 2023-05-10
JN-36 adding withdrawal functionality

devonbush:
This adds withdrawal capability to the admin tool.  Currently non-superusers can see the functionality, but only superusers can execute it. 
image

 Withdrawal functionality has two competing concerns:
1. When an enrollee withdraws, they want their data to be totally gone, and we want to make sure it doesn't accidentally surface
2. For legal reasons, we need to keep a lot of their data (e.g. signed consents)

So, what we do is essentially turn the enrollee data into a json blob and park it in a `withdrawn_enrollee` table.  Then we hard-delete the enrollee record.  So we'll still have the enrollee data if, e.g., we need to produce a record of the signed consent.  But it won't be in a form that we have to worry about it accidentally getting included in a query.  (we had originally planned on doing a softer delete via the "withdrawn" boolean on enrollee, but on further reflection, that brought too much risk of accidental data inclusion)

This also adds an extremely trivial study environment stats endpoint (not even linked from anywhere yet) for viewing withdrawn enrollee counts

TO REVIEW:
1. start with the schema change
2. Look at the withdrawEnrollee method in WithdrawnEnrolleeService to see the withdrawal process
3. Look at WithdrawnEnrolleeServiceTests and PopulateOurhealthTests to see the consequences of withdrawal. 
4. everything else

TO TEST:
1. restart admin tool
5. run `./scripts/populate_portal.sh ourhealth`
6. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/stats`
7. notice there is now a withdrawn enrollee
8. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants`
9. confirm the withdrawn enrollee "Will drawn" does not appear
10. pick an existing participant from the list
11. Follow the https://broadworkbench.atlassian.net/wiki/spaces/PEARL/pages/2751922177/Enrollee+withdrawal wiki (although please do not delete them from the Azure tenant, or if you do, restore them afterwards)


core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
```diff
@@ -122,12 +112,10 @@ public void delete(UUID enrolleeId, Set<CascadeProperty> cascades) {
         Enrollee enrollee = dao.find(enrolleeId).get();
         StudyEnvironment studyEnv = studyEnvironmentService.find(enrollee.getStudyEnvironmentId()).get();
         /**
-         * We should not generally delete withdrawn participants, since we need to preserve audit records.
-         * However, we allow deleting withdrawn participants because those may be test participants
-         * run through a live site to test live configurations.
+         * For production environments, we only allow deletion if a withdrawal record has already been preserved
```
breilly2, line 115: Nice check!

core/src/main/resources/db/changelog/changesets/2023_05_09_enrollee_withdraw.yaml
```diff
@@ -0,0 +1,21 @@
...
+                       - column: { name: study_environment_id, type: uuid, constraints:
+                         { nullable: false, foreignKeyName: withdrawn_enrollee_study_env_id, references: study_environment(id) }
+                       }
+                       - column: { name: shortcode, type: text, constraints: { nullable: false }  }
+                       - column: { name: enrollee_data, type: text }
+                       - column: { name: user_data, type: text }
+             - dropColumn:
+                  tableName: enrollee
+                  columns:
+                       - column: { name: withdrawn }
```
breilly2, line 21: I'm wondering how we might notice someone withdrawing and then re-enrolling. It should be extremely rare, but we probably want to make it difficult. Maybe we could insert new marker records with just `withdrawn=true`.

core/src/main/java/bio/terra/pearl/core/model/participant/WithdrawnEnrollee.java
```diff
@@ -0,0 +1,25 @@
...
+ * for compliance purposes.  The withdrawal process should be treated as irreversible.
+ */
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class WithdrawnEnrollee extends BaseEntity {
+  private String shortcode;
+  private UUID studyEnvironmentId;
+  /** JSON of the Enrollee object and children that need to e saved for compliance */
```
breilly2, line 21:
```suggestion
  /** JSON of the Enrollee object and children that need to be saved for compliance */
```

core/src/main/java/bio/terra/pearl/core/service/participant/WithdrawnEnrolleeService.java
```diff
@@ -0,0 +1,89 @@
...
+    ParticipantUser user = participantUserService.find(enrollee.getParticipantUserId()).get();
+    WithdrawnEnrollee withdrawnEnrollee = WithdrawnEnrollee.builder()
+        .shortcode(enrollee.getShortcode())
+        .studyEnvironmentId(enrollee.getStudyEnvironmentId())
+        .enrolleeData(objectMapper.writeValueAsString(enrollee))
+        .userData(objectMapper.writeValueAsString(user))
+        .build();
+    withdrawnEnrollee = create(withdrawnEnrollee);
+
+    enrolleeService.delete(enrollee.getId(), CascadeProperty.EMPTY_SET);
```
breilly2, line 63: I'm really curious about whether we should also be deleting the `ParticipantUser`. I see that `withdrawFromJuniper` does this, and I get that an individual may be registered for more than one portal and enrolled in more than one study. However, I also see that you added support for `AllowedCascades.PARTICIPANT_USER` to `EnrolleeService.delete()` so I'm not sure what your intent is.

breilly2 (commented):
Some questions/suggestions before approving.

core/src/main/resources/db/changelog/changesets/2023_05_09_enrollee_withdraw.yaml
devonbush: I'm not sure what we would do with that information?  I can't think of any reason we'd want to stop that--if someone is trying to spam a study, creating multiple accounts seems easier than withdraw-rejoin

core/src/main/java/bio/terra/pearl/core/service/participant/WithdrawnEnrolleeService.java
devonbush: I also am not sure what we should do. :)  I got halfway through implementing it, and then decided "that's a decision for later" and left the simple enrollee withdraw only for now.  ParticipantUser or PortalParticipantUser deletion is easy enough to do directly from the database if we need it in the short term

core/src/main/resources/db/changelog/changesets/2023_05_09_enrollee_withdraw.yaml
breilly2: That's a good point. And I'm finding I don't feel as strongly about it as I did yesterday.

core/src/main/java/bio/terra/pearl/core/service/participant/WithdrawnEnrolleeService.java
breilly2: In the short term, i.e. OurHealth, I might expect as a user that all of my information would be deleted. If I got confirmation that I withdrew and was still able to sign in, I would be surprised.

breilly2 (approved):
Some stuff to work out in the long term, but this is great to have in place now.

core/src/main/java/bio/terra/pearl/core/service/participant/WithdrawnEnrolleeService.java
devonbush: concur, we'll want to make sure to follow the https://broadworkbench.atlassian.net/wiki/spaces/PEARL/pages/2751922177/Enrollee+withdrawal checklist so we don't forget to clear them from B2C

MatthewBemis (approved):

PR #363, opened 2023-05-10
JN-372 sonar-scanner

TomConner:
Run SonarCloud across api-admin, api-participant, core, ui-admin, ui-participant. Using sonar-scanner rather than gradle because this repo contains both Java and TypeScript.

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,8 @@
+sonar.projectKey=broadinstitute_pearl
+sonar.organization=dsp-appsec
+sonar.projectName=pearl
+sonar.sources=api-admin,api-participant,core,ui-admin,ui-participant
+sonar.sourceEncoding=UTF-8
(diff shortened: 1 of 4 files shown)


sonar-project.properties
```diff
@@ -0,0 +1,8 @@
+sonar.projectKey=broadinstitute_pearl
```
nawatts, line 1: Will this need to change if we rename the repo?

nawatts (approved):

sonar-project.properties
devonbush: I talked to Tom about this -- some stuff will need to change, but it shouldn't be a big deal

devonbush (approved):

sonar-project.properties
TomConner: ^^ that's right

TomConner:
merging to get sonar workflow

PR #366, opened 2023-05-10
[JN-355] specifying strict snakeyaml version

devonbush:
trying again, using strict version specification.


nawatts (approved):

breilly2 (approved):
🤞 that anything that uses this as a transitive dependency is okay with this.

PR #367, opened 2023-05-10
[JN-306] Only support on-demand dataset creation

MatthewBemis:
Until we have a better plan for dataset schema migration _and_ until TDR on Azure supports row upserts, we're better off removing the scheduled process that periodically creates datasets and ingests data. It's just cluttering things at the moment.

In the short-term, the higher value and simpler approach is to embrace on-demand dataset creation. This was introduced in #359, and will be fully hooked up to the TSV ingest code (introduced in #326) shortly once I wrap up schema creation. Once that's complete, study staff will be able to click a button in the admin tool and have their current data be exported to TDR.

Note that this is only one small part of JN-306.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/ScheduledDataRepoExportService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/ScheduledDataRepoExportService.java
@@ -33,44 +33,7 @@ public class ScheduledDataRepoExportService {
     }
   }
 
-  /* Create any missing datasets every hour or so. There is one dataset per study
-    environment. By decoupling dataset creation from ingest, we don't have to worry
-    about handling large async saga transactions. If a dataset has not yet been created, ingest
-    will simply be skipped until the dataset is ready. These operations happen frequently enough
-    that it shouldn't be a problem if the first round of ingest is delayed due to a missing
-    dataset: by the next round, it should be ready.
-  */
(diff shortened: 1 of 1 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/service/ScheduledDataRepoExportService.java
```diff
@@ -33,44 +33,7 @@ public void pingDataRepoStatus() {
...
-      logger.info("Ingesting datasets...");
-      dataRepoExportService.ingestDatasets();
-    } else {
-      logger.error(
-          "Error: Skipping TDR dataset ingest, as TDR has not been configured for this environment.");
-    }
-  }
-
-  @Scheduled(timeUnit = TimeUnit.MINUTES, fixedDelay = 10, initialDelay = 0)
+  @Scheduled(timeUnit = TimeUnit.MINUTES, fixedDelay = 5, initialDelay = 0)
```
MatthewBemis, line 40: I'm also dropping this to a 5 min poll time. Honestly I could probably go even lower but I want to see how 5 min feels before being too aggressive with it.

nawatts (approved):

devonbush (approved):
Nice,  I might put a comment somewhere in the file with a link to this PR to the effect of "to restore auto-sync functionality, look to this PR" just to make it easier to find when/if we want to bring it back.

PR #368, opened 2023-05-10
JN-355 excising more snake yaml dependencies

devonbush:


nawatts (approved):

breilly2 (approved):
Oh gosh. 😬 As a code author, I would cringe at a consumer doing stuff like this to something I've published. 🤞 🙈 👍

devonbush:
I'm not sure what you mean?

devonbush:
maybe I'm doing things wrong because I'm bringing maven practices over to gradle?  the spring forums say you need to use a version >=2.7.10 to be compatible with snakeyaml 2.0.  So that's what I'm trying to do.  I'm failing because somewhere deep in our plugin dependency hierarchy, someone brings in spring boot 2.7.0

PR #370, opened 2023-05-11
[JN-371] Add note about locked surveys

nawatts:
Screenshot 2023-05-11 at 1 23 25 PM

### DIFF
--- a/ui-participant/src/hub/HubPage.tsx
+++ b/ui-participant/src/hub/HubPage.tsx
@@ -179,9 +179,14 @@ function TaskGrouping({ title, tasks, enrollee, studyShortcode }: {
   title: string, tasks: ParticipantTask[],
   enrollee: Enrollee, studyShortcode: string
 }) {
+  const hasLockedTasks = tasks.some(task => !isTaskAccessible(task, enrollee))
+
   return (
     <>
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):

ui-participant/src/hub/HubPage.tsx
```diff
@@ -179,9 +179,14 @@ function TaskGrouping({ title, tasks, enrollee, studyShortcode }: {
   title: string, tasks: ParticipantTask[],
   enrollee: Enrollee, studyShortcode: string
 }) {
+  const hasLockedTasks = tasks.some(task => !isTaskAccessible(task, enrollee))
+
   return (
     <>
       <h2 className="fs-6 text-uppercase mb-0">{title}</h2>
+      {hasLockedTasks && (
+        <p className="my-2">Some surveys may be locked until other required tasks are completed.</p>
```
nawatts: The design says "may be", but since this is only displayed when there are locked tasks, maybe change this to "are"?

devonbush:
Agree with the wording change given that you've made the display conditional.  IMHO I would also mute the text with `text-muted` to match the greyed-out scheme of the locks.  But that's just me.

devonbush (approved):

PR #372, opened 2023-05-11
[JN-375] Prod and cli flag functionality

katiewelch:
This PR updates the command line options for ./sql_connect and includes defaults to connect to prod as well as dev.
All command flags (-d database -c cluster -s subscription -r resourcegroup) are now optional. Defaults are hard coded for dev and prod but can be overridden with the flag.

Positional argument for environment is required as first arg. Valid options are dev and prod (eg ./sql_connect dev or ./sql_connect prod). Each option will select the correct defaults for other variables (listed above).

**Testing**
Both Devon and I tested this locally to ensure we were able to connect to both the prod and dev databases.

**Risk**
None - this is script is just a tool for devs to connect to databases.

### DIFF
--- a/scripts/sql_connect
+++ b/scripts/sql_connect
@@ -294,18 +294,18 @@ function exec_into_pod() {
 
 function init() {
     parse_args "$@"
-    #check_required_tools || exit 1
-    #fetch_database_creds
+    check_required_tools || exit 1
+    fetch_database_creds
 }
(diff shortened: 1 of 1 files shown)


devonbush (approved):
Tested it locally, was able to get into both dev and prod

em-may (approved):

PR #374, opened 2023-05-12
[JN-376] Update user state after auto saving survey progress

nawatts:
Currently, if you start a survey and then return to the dashboard via the Dashboard link in the navbar, the survey still shows as "Not started". After refreshing the page, the survey shows as "In progress".

This updates the user state after auto saving survey progress so that the dashboard is accurate.


breilly2 (approved):
Nice that there was a comment saying exactly where the future code should live!

devonbush (approved):
Nice fix

PR #376, opened 2023-05-15
[JN-132, JN-374] Create dataset with a real schema + support dataset descriptions

MatthewBemis:
TO TEST:

1) Launch ApiAdminApi, populated with OurHealth
2) Launch admin UI
3) Create new dataset (could take up to 5 mins to show up-> improvements to this piece are in another pending PR)
4) Click to view the dataset in TDR, and verify that all of the columns are there

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
@@ -34,11 +34,11 @@ public class DataRepoClient {
         List<ColumnModel> columns = new ArrayList<>();
 
         schemaMappings.forEach((columnName, columnDataType) ->
-                columns.add(new ColumnModel().name(columnName).datatype(TableDataType.STRING).required(false))
+                columns.add(new ColumnModel().name(columnName).datatype(columnDataType).required(false))
         );
 
         DatasetSpecificationModel schema = new DatasetSpecificationModel()
-                .tables(List.of(new TableModel().name("enrollee").columns(columns))); //.primaryKey(List.of("enrollee_shortcode"))
+                .tables(List.of(new TableModel().name("enrollee").columns(columns)));
 
         DatasetRequestModel dataset = new DatasetRequestModel()
                 .name(datasetName)
--- a/core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
@@ -28,13 +28,13 @@ public class DataRepoClient {
     }
 
     //Dataset APIs
-    public JobModel createDataset(UUID spendProfileId, String datasetName, Map<String, DataValueExportType> schemaMappings) throws ApiException {
+    public JobModel createDataset(UUID spendProfileId, String datasetName, Map<String, TableDataType> schemaMappings) throws ApiException {
         DatasetsApi datasetsApi = getDatasetsApi();
 
         List<ColumnModel> columns = new ArrayList<>();
 
         schemaMappings.forEach((columnName, columnDataType) ->
-                columns.add(new ColumnModel().name(columnName.substring(0, Math.min(columnName.length(), 63))).datatype(TableDataType.STRING).required(false))
+                columns.add(new ColumnModel().name(columnName).datatype(TableDataType.STRING).required(false))
         );
 
         DatasetSpecificationModel schema = new DatasetSpecificationModel()
--- a/core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
@@ -5,7 +5,6 @@ import bio.terra.datarepo.api.JobsApi;
 import bio.terra.datarepo.api.UnauthenticatedApi;
 import bio.terra.datarepo.client.ApiException;
 import bio.terra.datarepo.model.*;
-import bio.terra.pearl.core.service.export.formatters.DataValueExportType;
 import bio.terra.pearl.core.shared.GoogleServiceAccountUtils;
 import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.GoogleCredentials;
@@ -28,9 +27,8 @@ public class DataRepoClient {
     }
 
     //Dataset APIs
-    public JobModel createDataset(UUID spendProfileId, String datasetName, List<String> columnKeys) throws ApiException {
+    public JobModel createDataset(UUID spendProfileId, String datasetName, String description, List<String> columnKeys) throws ApiException {
         DatasetsApi datasetsApi = getDatasetsApi();
-
 
         //TODO: Temporarily setting all column types of STRING. There seems to be an issue converting our DATETIME format
         //into a TDR DATETIME. This will be resolved in JN-381.
@@ -41,6 +39,7 @@ public class DataRepoClient {
 
         DatasetRequestModel dataset = new DatasetRequestModel()
                 .name(datasetName)
+                .description(description)
                 .cloudPlatform(CloudPlatform.AZURE)
                 .defaultProfileId(spendProfileId)
                 .schema(schema);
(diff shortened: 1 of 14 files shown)


core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
```diff
@@ -27,14 +28,19 @@ public DataRepoClient(Environment env) {
...
     //Dataset APIs
-    public JobModel createDataset(UUID spendProfileId, String datasetName) throws ApiException {
+    public JobModel createDataset(UUID spendProfileId, String datasetName, Map<String, TableDataType> schemaMappings) throws ApiException {
         DatasetsApi datasetsApi = getDatasetsApi();
 
-        //TODO: AR-229. Placeholder schema for now, need to determine mapping of survey schema to TDR schema.
+        List<ColumnModel> columns = new ArrayList<>();
+
+        //TODO: Temporarily setting all column types of STRING. There seems to be an issue converting our DATETIME format
+        //into a TDR DATETIME. This will be resolved in a followup ticket.
```
MatthewBemis: Will file a Jira ticket and add to the comment here

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
```diff
@@ -132,21 +114,12 @@ public void createDataset(StudyEnvironment studyEnv, String datasetName) {
...
-    public void ingestDatasets() {
-        List<Dataset> outdatedDatasets = datasetDao.findAll().stream().filter(dataset -> dataset.getLastExported().isBefore(Instant.now().minus(1, ChronoUnit.MINUTES))).toList();
-
-        logger.info("Found {} study environments requiring dataset ingest", outdatedDatasets.size());
-
-        for(Dataset dataset : outdatedDatasets) {
-            logger.info("Ingesting data for study environment ID {}", dataset.getStudyEnvironmentId());
-            ingestDataForStudyEnvironment(dataset);
-        }
-    }
```
MatthewBemis, line 144: This is all happening on-demand now, so this code was unused.

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
```diff
@@ -89,34 +89,16 @@ public List<DataRepoJob> getJobHistoryForDataset(UUID studyEnvironmentId, String
...
-        for(StudyEnvironment studyEnv : studyEnvsWithoutDatasets) {
-            Study study = studyDao.find(studyEnv.getStudyId()).orElseThrow(() -> new StudyNotFoundException(studyEnv.getStudyId()));
-            String environmentName = studyEnv.getEnvironmentName().name();
-
-            String datasetName = makeDatasetName(DEPLOYMENT_ZONE, study.getShortcode(), environmentName);
-
-            createDataset(studyEnv, datasetName);
-        }
-
-    }
```
MatthewBemis, line 110: This is all happening on-demand now, so this code was unused.

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
```diff
@@ -186,6 +159,42 @@ public void ingestDataForStudyEnvironment(Dataset studyEnvDataset) {
...
+        Map<String, TableDataType> schemaMappings = new LinkedHashMap<>();
+
+        try {
+            List<ModuleExportInfo> exportInfoList = enrolleeExportService.generateModuleInfos(exportOptions, portalStudy.getPortalId(), studyEnvironmentId);
+
+            for(ModuleExportInfo exportInfo: exportInfoList) {
+                exportInfo.getItems().forEach(itemExportInfo -> {
+                    String columnName;
+
+                    //Profile fields don't have stable IDs, so we'll use the base column key instead
```
MatthewBemis: Does this seem right?

MatthewBemis: Actually, I see a better way of doing this now.

devonbush (approved):

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
MatthewBemis: Filed as JN-381

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
```diff
@@ -60,6 +63,7 @@ public JobModel ingestDataset(UUID spendProfileId, UUID datasetId, String tableN
                 .table(tableName)
                 .profileId(spendProfileId)
                 .format(IngestRequestModel.FormatEnum.CSV)
+                .csvFieldDelimiter("\t")
```
breilly2, line 65: 🤔 CSV is literally "comma-separated values", so I'm not sure what it means to use tab as the field delimiter. 🤷‍♂️

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
```diff
@@ -132,21 +113,12 @@ public void createDataset(StudyEnvironment studyEnv, String datasetName) {
...
-        }
-    }
-
-    public String uploadCsvToAzureStorage(UUID studyEnvironmentId) {
+    public String uploadCsvToAzureStorage(UUID studyEnvironmentId, String datasetName) {
         ExportOptions exportOptions = new ExportOptions(false, false, false, ExportFileFormat.TSV, null);
 
-        String blobName = studyEnvironmentId + "_" + Instant.now() + ".csv";
+        //Even though this is actually formatted as a TSV, TDR only accepts files ending in .csv or .json.
+        //In the DataRepoClient call, we specify that the CSV delimiter is "\t", which will make it all work fine.
```
breilly2, line 121: Oh, I see. I hope there aren't corner cases that this introduces. Maybe tab-delimited CSVs are generally well-supported and accepted.

The nice thing about CSV is that cells can contain anything, including tabs and newlines. If a cell contains anything that would otherwise confuse the parser (like `,` or `\n`) then it gets double-quoted (with any embedded double-quotes escaped)... though I guess `\t` is just as good as `,` as long as the parser adapts and correctly implements the rules in the same way.

And as long as we're correctly implementing those rules when generating the input file. I remember a time when TSV generation libraries didn't deal well with embedded tabs, but maybe things have evolved since then.

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportUtils.java
```diff
@@ -0,0 +1,27 @@
...
+            case STRING -> TableDataType.STRING;
+            case NUMBER -> TableDataType.INTEGER;
+            case BOOLEAN -> TableDataType.BOOLEAN;
+        };
+    }
+
+    //Converts a Juniper column name into a TDR-accepted column name
+    public static String juniperToDataRepoColumnName(String columnName) {
+        return columnName
+                .substring(0, Math.min(columnName.length(), 63)) //TDR accepts a maximum column name length of 64 chars
```
breilly2, line 25: What happens if this results in a non-unique column name? I'm less worried if it crashes; I'm much more worried if it silently does the wrong thing. I wonder if we need to do something like Windows used to do with munging long filenames to fit in the old "8.3" filename constraint.

breilly2 (commented):
I want to know more about column name uniqueness before approving.

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportUtils.java
MatthewBemis: Very good question. TDR will reject the request:

```{"message":"Unable to create TDR dataset for study environment 529bd22a-af8c-4900-824d-bdf3f8247dd7. Error: {\"message\":\"Validation errors - see error details\",\"errorDetail\":[\"schema: 'DuplicateColumnNames' (Duplicate columns: foo)\"]}","statusCode":500}```

And just to be extra sure, I also verified that TDR still rejects the schema if two columns have the same name but different types (in case their definition of uniqueness is a bit more complex than just naming)

MatthewBemis: I can also add a code comment about this.

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
MatthewBemis: Hah yeah, I was a bit surprised by this as well. I think we'll support CSV export in Juniper at some point so that might be the cleaner way of handling this but for now it seems to work well.

breilly2 (approved):

PR #378, opened 2023-05-16
[JN-373] Add change password

breilly2:
Change password is currently implemented with a separate policy. The policy name comes from config fetched from the API, so this PR also contains those config plumbing changes.

Screen Shot 2023-05-15 at 9 09 46 PM

The conditions under which the user is prompted to sign in when trying to change their password may be a little different than what I explained in https://github.com/broadinstitute/terraform-ap-deployments/pull/1066. I'm not 100% sure what the conditions are. I'm not terribly satisfied with my level of understanding of this -- I fully expect to get bug reports that will take a lot of fiddling to iron out -- but change password is there and generally works.

To test:
1. Restart participant API and UI
2. Sign in to https://sandbox.ourhealth.localhost:3001/
3. Open the profile menu and click "Change Password"
4. Change your password and make sure you end up back at the OurHealth portal
5. Sign out and make sure you are required to use your new password to sign back in

### DIFF
--- a/ui-participant/src/authConfig.ts
+++ b/ui-participant/src/authConfig.ts
@@ -1,4 +1,4 @@
-import { WebStorageStateStore } from 'oidc-client-ts'
+import { UserManagerSettings, WebStorageStateStore } from 'oidc-client-ts'
 import { AuthProviderProps } from 'react-oidc-context'
 
 /**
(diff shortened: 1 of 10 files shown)


nawatts (approved):

ui-participant/src/authConfig.ts
```diff
@@ -7,9 +7,32 @@ import { AuthProviderProps } from 'react-oidc-context'
  *    visit: https://docs.microsoft.com/en-us/azure/active-directory-b2c/custom-policy-overview
  */
```
devonbush, line 9: this function needs a comment

ui-participant/src/authConfig.ts
```diff
@@ -7,9 +7,32 @@ import { AuthProviderProps } from 'react-oidc-context'
...
+  }
+
+  return {
+    ...oidcConfig,
+    prompt: 'login',
+    onSigninCallback
+  }
+}
+
 // TODO: This is a modified copy of code from Terra UI. It could use some clean-up.
```
devonbush: Either remove the TODO here, or mention the specific remaining cleanup.  this could also use a descriptive comment

devonbush (approved):
looks ok (agree with your comments in the other PR that this seems like B2C isn't quite letting us do what we want). some optional comments.

ui-participant/src/authConfig.ts
breilly2: Yeah, I think enough of the glaring Terra-specific stuff has been removed at this point, so this to-do is to-done.

breilly2:
@devonbush FWIW, I think it's possible to get B2C to do more of what we want, but this is a workable starting point and we can improve it down the road if we choose to.

PR #380, opened 2023-05-17
[JN-385] Upgrade node Gradle plugin

nawatts:
The Gradle plugin we use for running Node tasks is very out of date. This upgrades it to the latest version.

@devonbush I wonder if this will fix your issue with Intellij prompting to run npm install.


devonbush (approved):
cool, I also hope this helps!

breilly2 (approved):

PR #381, opened 2023-05-17
[JN-325] basic admin visibility of study metrics

devonbush:
This adds metrics functionality to the admin tool for simple visualization of enrollment, consent, etc..   (Don't worry, *most* of the 5,700 added lines in this PR are due to package.lock changes from plotly)
image

To support testing this, this also adds rudimentary capability for populating users "in the past" via specifying a `submittedHoursAgo` attribute on some populate models.  

Lots of stuff that could be refined as we find out what's useful, but this should be enough for OurHealth in the short term.

TO TEST
1. restart ApiAdminApp
2. repopulate OurHealth
3. run `npm install` from ui-admin (and note the new note in the README if the install fails)
4. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/metrics
5. confirm it looks like the screenshot above.
6. click "Copy raw data" for the top graph, and paste into a google sheet.  You should get something like the data below:
image


ui-admin/package.json
```diff
@@ -33,7 +35,8 @@
     "@types/marked": "^4.0.7",
     "@types/react": "^18.0.18",
     "@types/react-dom": "^18.0.6",
-    "@types/react-table": "^7.7.12"
+    "@types/react-table": "^7.7.12",
+    "canvas": "^2.11.2"
```
nawatts, line 40: I didn't see this used anywhere? Is it a peer dependency of plotly? If so, it would be preferable to put it in dependencies instead of dev dependencies.

nawatts (approved):

ui-admin/package.json
devonbush: yeah, this is confusing -- and it's where I really wish you could add comments in package.json.  this is only needed for plotly jest tests.  Otherwise, the test console fills up with "...canvas" is not a function warnings.  https://github.com/hustcc/jest-canvas-mock used to let you get around this more cleanly, but there are compatibility issues between that and later versions of create-react-app. So the stackoverflow consesnsus is just that it's cleaner to go ahead and install the dependency directly

README.md
```diff
@@ -108,6 +108,8 @@ In basic development mode, this will only serve the API, not the frontend assets
   npm -w ui-core run build
   HTTPS=true npm -w ui-participant start
   ```
+If you encounter an error like "No package 'pixman-1' found" on npm install, you may need additional dependencies.
+If on OS X, try running `brew install pkg-config cairo pango libpng jpeg giflib librsvg pixman`
```
breilly2, line 112: I guess these are already available on our CI environment?

breilly2 (approved):
Turned out to be really helpful to see this demoed before reviewing.

README.md
devonbush: apparently, yes :)

PR #382, opened 2023-05-17
[JN-380] Automatically share dev datasets with dev team

MatthewBemis:
Previously, only the Juniper service account would have access to the datasets created through Juniper. This has been a huge pain point in testing, and makes it so you have to backdoor yourself onto the datasets in order to demo this functionality or test it out. I've created a [juniper-dev](https://bvdp-saturn-dev.appspot.com/#groups/juniper-dev) managed group in Terra with all of the devs on it (we're all admins so feel free to add other folks). This code will only be executed in non-prod environments. Should it somehow execute in production, it will still fail because the juniper-dev group does not exist in prod (and the group can't possibly be duplicated in production due to how Sam namespaces managed groups). 

Ultimately, this will be replaced by JN-133, which will provision a Terra managed group with all of the study admins in it, and add that as a steward to the TDR dataset. This is a stopgap to make things easier to test until that is ready.

I chose to go this route instead of automatically sharing the datasets with the super admins in our DB because for most of us, the emails we use in Terra dev are different than the emails we have as super admins. We could of course align those, but this felt easier for a stop-gap solution.

This will also make testing all of my other pending PRs a lot easier.

TO TEST:

1) Create a dataset in the Juniper admin UI.
2) Verify that your dev Terra account has access by clicking the "View in Terra Data Repo" link. (currently you can expect this to take 5mins unless you tweak the `pollRunningJobs` poll time in `ScheduledDataRepoExportService.java`)

Also note that the code on `development` is way behind what I've been demoing lately, so don't expect it to be as smooth :)

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
@@ -73,6 +73,17 @@ public class DataRepoClient {
         return datasetsApi.deleteDataset(datasetId);
     }
 
+    //TODO: This is to be replaced by JN-133. This code should only ever be executed in dev.
+    public void shareWithJuniperDevs(UUID datasetId) {
+        DatasetsApi datasetsApi = getDatasetsApi();
+
+        try {
+            datasetsApi.addDatasetPolicyMember(datasetId, "steward", new PolicyMemberRequest().email("juniper-dev@dev.test.firecloud.org"));
+        } catch (ApiException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
     //Job APIs
     public JobModel getJobStatus(String jobId) throws ApiException {
         JobsApi jobsApi = getJobsApi();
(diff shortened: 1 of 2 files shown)


devonbush (approved):

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
```diff
@@ -73,6 +73,17 @@ public JobModel deleteDataset(UUID datasetId) throws ApiException {
...
     }
 
+    //TODO: This is to be replaced by JN-133. This code should only ever be executed in dev.
+    public void shareWithJuniperDevs(UUID datasetId) {
+        DatasetsApi datasetsApi = getDatasetsApi();
+
+        try {
+            datasetsApi.addDatasetPolicyMember(datasetId, "steward", new PolicyMemberRequest().email("juniper-dev@dev.test.firecloud.org"));
+        } catch (ApiException e) {
+            throw new RuntimeException(e);
```
nawatts, line 83: Out of curiosity, why catch and re-throw this as a different exception type?

nawatts (approved):

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoClient.java
MatthewBemis: I forget exactly why I started doing this with all DataRepoClient calls. I think it had something to do with the global exception handler being able to handle RuntimeExceptions so I was just taking a "shortcut" at the time for easier handling?

Definitely an area for me to clean up a bit.

PR #383, opened 2023-05-18
JN-364 fetch survey CRUD endpoints

devonbush:
Adds a basic get and create endpoint for the upcoming create survey overhaul.  

TO TEST:
1. run ./gradlew test to check the new unit tests.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
@@ -9,6 +9,8 @@ import bio.terra.pearl.core.model.survey.Survey;
 import bio.terra.pearl.core.service.exception.PermissionDeniedException;
 import bio.terra.pearl.core.service.study.StudyEnvironmentSurveyService;
 import bio.terra.pearl.core.service.survey.SurveyService;
+import java.util.List;
+import java.util.Optional;
 import org.springframework.beans.BeanUtils;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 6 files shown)


nawatts (approved):

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
```diff
@@ -27,6 +29,34 @@ public SurveyExtService(
...
 
+  public Optional<Survey> get(
+      String portalShortcode, String stableId, int version, AdminUser adminUser) {
+    Portal portal = authUtilService.authUserToPortal(adminUser, portalShortcode);
+    Optional<Survey> surveyOpt = surveyService.findByStableId(stableId, version);
+    if (surveyOpt.isEmpty()) {
+      return surveyOpt;
+    }
+    Survey survey = surveyOpt.get();
+    if (!portal.getId().equals(survey.getPortalId())) {
```
breilly2: It wasn't immediately obvious to me exactly what this was guarding. Seemed odd to have a "permission denied" error so far down in the function.

I think it's worth extracting the code that protects against mismatched portal/survey identifiers. Maybe make `surveyService.findByStableId` also require the `portalShortcode` and do the check there. Or introduce a `authUtilService.authUserToSurvey` function. Otherwise, it'll be too easy to forget to make this check.

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -124,11 +156,11 @@ paths:
         - { name: stableId, in: path, required: true, schema: { type: string } }
       requestBody:
         required: true
-        content: { application/json: { schema: {$ref: '#/components/schemas/VersionedFormDto' } } }
+        content: { application/json: { schema: {type: object} } }
```
breilly2, line 159: Are you feeling that maintaining openapi schema definitions is too much effort for the value? Or is this endpoint shifting to be more open-ended such that we can't define a concrete schema?

api-admin/src/test/java/bio/terra/pearl/api/admin/service/forms/SurveyExtServiceTests.java
```diff
@@ -1,17 +1,100 @@
...
+        .thenReturn(Optional.of(matchedSurvey));
+    Survey unmatchedSurvey =
+        Survey.builder()
+            .stableId("testMatchedToPortalUnmatched")
+            .version(1)
+            .portalId(UUID.randomUUID())
+            .build();
+    when(mockSurveyService.findByStableId(
+            unmatchedSurvey.getStableId(), unmatchedSurvey.getVersion()))
+        .thenReturn(Optional.of(unmatchedSurvey));
```
breilly2: Suggestion: I've had good experiences extracting functions for setting up mocks like this. The creation and setup of `matchedSurvey` and `unmatchedSurvey` are nearly identical, and the volume of code here detracts from the clarity if the test. This could be two lines of code:
```
    Survey matchedSurvey = configureMockSurvey("testMatchedToPortal", portal.getId());
    Survey unmatchedSurvey = configureMockSurvey("testMatchedToPortalUnmatched", UUID.randomUUID());
```
with the addition of this function:
```
  private Survey configureMockSurvey(String surveyStableId, UUID portalId) {
    Survey survey = Survey.builder()
            .stableId(surveyStableId)
            .version(1)
            .portalId(portalId)
            .build();
    when(mockSurveyService.findByStableId(survey.getStableId(), survey.getVersion()))
        .thenReturn(Optional.of(survey));
    return survey;
  }
```

breilly2 (changes requested):
The security checks are good, but I think it's worth codifying the intent more explicitly for increased clarity and easier reuse (and reminder) in future code.

api-admin/src/main/resources/api/openapi.yml
devonbush: Yeah, I don't think we're getting any value out of the schema definitions, and I find they are a huge pain to write (lack of good IDE autocomplete, having to recompile the project to see if it matches your controller).  And the reason I reverted this one is that ConsentForm and Survey have now diverged (survey has AnswerMappings and a footer property), and so it didn't make sense to have a shared DTO

breilly2 (approved):
Nice!

PR #384, opened 2023-05-18
[JN-382] Move types for landing page config to ui-core

nawatts:
Many type definitions are duplicated between ui-admin/api and ui-participant/api. This moves some of those type definitions into ui-core, starting with the types for landing page configuration. Started with those because they are the same in both places and thus easiest to understand.

Types are re-exported from ui-admin/api and ui-participant/api instead of imported directly from ui-core to leave open the ability to tailor them for the admin/participant UI (for example, if the participant API doesn't return all the fields of a certain type that the admin API does).

### DIFF
--- a/ui-admin/src/api/api.tsx
+++ b/ui-admin/src/api/api.tsx
@@ -1,3 +1,17 @@
+import { SiteContent } from '@juniper/ui-core'
+
+export type {
+  SiteContent,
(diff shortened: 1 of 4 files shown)


devonbush (approved):
This is great, love how you handled the re-export

MatthewBemis (approved):

PR #385, opened 2023-05-18
[JN-360] Model DSM kit requests and skeleton service

breilly2:
Basic model of kit request and DAO, service, etc. Gives us a starting point to build on as we refine the details.

To test:
1. Start admin API
2. Open http://localhost:8080/swagger-ui.html#/enrollee/requestKit, authenticate, and fill in parameters (suggested values attached)
3. Check the database and verify that there is a new row in the `kit_request` table

Screen Shot 2023-05-18 at 2 34 52 PM

`kitType` is an uncontrolled string at the moment, so you can use anything you like.

### DIFF
--- a/api-admin/src/main/resources/api/openapi.yml
+++ b/api-admin/src/main/resources/api/openapi.yml
@@ -308,9 +308,7 @@ paths:
         - { name: studyShortcode, in: path, required: true, schema: { type: string } }
         - { name: envName, in: path, required: true, schema: { type: string } }
         - { name: enrolleeShortcode, in: path, required: true, schema: { type: string } }
-      requestBody:
-        required: true
-        content: { application/json: { schema: { $ref: '#/components/schemas/KitRequest' } } }
+        - { name: kitType, in: query, required: true, schema: { type: string } }
       responses:
         '200':
           description: Kit request
@@ -774,11 +772,6 @@ components:
     DatasetName:
       properties:
         name:
-          type: string
-    KitRequest:
-      type: object
-      properties:
-        kitType:
           type: string
     SystemStatus:
       required: [ ok, systems ]
--- a/api-admin/src/main/resources/api/openapi.yml
+++ b/api-admin/src/main/resources/api/openapi.yml
@@ -298,6 +298,25 @@ paths:
           content: { application/json: { schema: { type: object } } }
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/portals/v1/{portalShortcode}/studies/{studyShortcode}/env/{envName}/enrollees/{enrolleeShortcode}/requestKit:
+    post:
+      summary: Request a sample kit for an enrollee
+      tags: [ enrollee ]
+      operationId: requestKit
+      parameters:
+        - { name: portalShortcode, in: path, required: true, schema: { type: string } }
+        - { name: studyShortcode, in: path, required: true, schema: { type: string } }
+        - { name: envName, in: path, required: true, schema: { type: string } }
+        - { name: enrolleeShortcode, in: path, required: true, schema: { type: string } }
+      requestBody:
+        required: true
+        content: { application/json: { schema: { $ref: '#/components/schemas/KitRequest' } } }
+      responses:
+        '200':
+          description: Kit request
+          content: { application/json: { schema: { type: object } } }
+        '500':
+          $ref: '#/components/responses/ServerError'
   /api/portals/v1/{portalShortcode}/env/{envName}/notificationConfigs/{configId}/test:
     post:
       summary: Sends a test notification for the given config and enrolleeRuleData
@@ -755,6 +774,11 @@ components:
     DatasetName:
       properties:
         name:
+          type: string
+    KitRequest:
+      type: object
+      properties:
+        kitType:
           type: string
     SystemStatus:
       required: [ ok, systems ]
--- a/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
@@ -1,0 +1,74 @@
+package bio.terra.pearl.core.service.kit;
+
+import bio.terra.pearl.core.dao.kit.KitRequestDao;
+import bio.terra.pearl.core.model.kit.KitRequest;
+import bio.terra.pearl.core.model.kit.KitRequestStatus;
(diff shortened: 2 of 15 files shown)


core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
```diff
@@ -149,7 +149,7 @@ public Enrollee create(Enrollee enrollee) {
             enrollee.setShortcode(generateShortcode());
         }
         Enrollee savedEnrollee = dao.create(enrollee);
-        logger.info("Enrollee created.  id: {}, shortcode: {}, participantUserId", savedEnrollee.getId(),
+        logger.info("Enrollee created.  id: {}, shortcode: {}, participantUserId: {}", savedEnrollee.getId(),
```
breilly2, line 152: Drive-by fixing a bug that I discovered while reading code.

core/src/test/java/bio/terra/pearl/core/service/participant/EnrolleeServiceTests.java
```diff
@@ -34,7 +34,7 @@ public void testEnrolleeCreate() {
         Enrollee savedEnrollee = enrolleeService.create(enrollee);
         DaoTestUtils.assertGeneratedProperties(savedEnrollee);
         Assertions.assertNotNull(savedEnrollee.getShortcode());
-        Assertions.assertEquals(enrollee.getParticipantUserId(), enrollee.getParticipantUserId());
+        Assertions.assertEquals(enrollee.getParticipantUserId(), savedEnrollee.getParticipantUserId());
```
breilly2, line 37: Another drive-by bug fix.

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -298,6 +298,25 @@ paths:
...
+    post:
+      summary: Request a sample kit for an enrollee
+      tags: [ enrollee ]
+      operationId: requestKit
+      parameters:
+        - { name: portalShortcode, in: path, required: true, schema: { type: string } }
+        - { name: studyShortcode, in: path, required: true, schema: { type: string } }
+        - { name: envName, in: path, required: true, schema: { type: string } }
+        - { name: enrolleeShortcode, in: path, required: true, schema: { type: string } }
+      requestBody:
```
devonbush: Since kitType is required and a string, and fundamental to the object, I'd lean towards making kitType a URL param, probably on the path.  The big advantage there is it then gives us easier logging/analytics since all API requests are logged

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -0,0 +1,74 @@
...
+                mailingAddress.getStreet1(),
+                mailingAddress.getStreet2(),
+                mailingAddress.getCity(),
+                mailingAddress.getState(),
+                mailingAddress.getPostalCode(),
+                mailingAddress.getCountry(),
+                profile.getPhoneNumber()
+        );
+    }
+
```
devonbush: nit: I wouldn't pull this out into its own method unless you expect to be doing something besides objectMapping soon.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -0,0 +1,74 @@
...
+                .status(KitRequestStatus.CREATED)
+                .build();
+        KitRequest savedKitRequest = dao.create(kitRequest);
+        logger.info("SampleKit created. id: {}, enrollee: {}", savedKitRequest.getId(), savedKitRequest.getEnrolleeId());
+        return savedKitRequest;
+    }
+
+    private PepperKitAddress makePepperKitAddress(Profile profile) {
+        MailingAddress mailingAddress = profile.getMailingAddress();
+        return new PepperKitAddress(
```
devonbush: Long order-sensitive string constructions like this make me nervous.  This makes me lean towards making PepperKitAddress a class so you can use a lombok builder.  I've had to abandon using Records elsewhere for this same reason

core/src/main/resources/db/changelog/changesets/2023_05_15_kit_request.yaml
```diff
@@ -0,0 +1,17 @@
+databaseChangeLog:
+  - changeSet:
+      id: kit_request
+      author: breilly
+      changes:
+        - createTable:
+            tableName: kit_request
+            columns:
```
devonbush, line 8: BLOCKER: we need a "creating_admin_user_id" field here for auditability.  For now, it should be non-nullable.  Later, when we add rule-based auto-kit sending, we can modify that constraint to say "must have either an admin id, or a rule_source specified"

core/src/testFixtures/java/bio/terra/pearl/core/factory/participant/EnrolleeFactory.java
```diff
@@ -46,10 +46,11 @@ public Enrollee buildPersisted(String testName) {
         return enrolleeService.create(builderWithDependencies(testName).build());
     }
 
-    public Enrollee buildPersisted(String testName, UUID studyEnvironmentId, UUID participantUserId) {
+    public Enrollee buildPersisted(String testName, UUID studyEnvironmentId, UUID participantUserId, UUID profileId) {
```
devonbush, line 49: maybe instead of adding a paramter to this function, create a new one, and have the existing one delegate?  So we don't have to specify "null" in a bunch of places where we don't care about the profile?

devonbush (changes requested):
Good start.  A bunch of suggestions that aren't necessarily blocking, and one blocker around recording the admin user who initiated the kitrequest

core/src/testFixtures/java/bio/terra/pearl/core/factory/participant/EnrolleeFactory.java
devonbush: nvm, it looks like this wasn't used elsewhere

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
breilly2: 👍 Good call.

devonbush (approved):
nice

MatthewBemis (approved):

PR #386, opened 2023-05-18
[JN-384] Save datasets immediately upon creation

MatthewBemis:
Previously Juniper would wait for the TDR job to succeed before creating a `dataset` row in the DB. This was an artifact of the old scheduled dataset creation. For on-demand dataset creation, we want to save a record immediately with a null dataset ID and a status field.

Screenshot 2023-05-18 at 3 07 57 PM

TO TEST:

Launch Juniper admin UI
Launch Juniper admin API

Navigate to study environment TDR dashboard, i.e. https://localhost:3000/ourhealth/studies/ourheart/env/live/export/dataRepo/datasets

Create a new dataset, and note that you see it show up immediately. The "View in TDR" link is hidden in the dataset details page unless it has successfully been created.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/datarepo/DatasetDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/datarepo/DatasetDao.java
@@ -29,6 +29,10 @@ public class DatasetDao extends BaseMutableJdbiDao<Dataset> {
         return findByProperty("dataset_id", dataRepoId);
     }
 
+    public Optional<Dataset> findByDatasetName(String datasetName) {
+        return findByProperty("dataset_name", datasetName);
+    }
+
     public void updateLastExported(UUID id, Instant lastExported) {
         updateProperty(id, "last_exported", lastExported);
     }
--- a/core/src/main/java/bio/terra/pearl/core/model/datarepo/Dataset.java
+++ b/core/src/main/java/bio/terra/pearl/core/model/datarepo/Dataset.java
@@ -18,6 +18,6 @@ public class Dataset extends BaseEntity {
     private UUID datasetId;
     private String datasetName;
     private String description;
-    private String status;
+    private DatasetStatus status;
     private Instant lastExported;
 }
--- a/core/src/main/java/bio/terra/pearl/core/model/datarepo/Dataset.java
+++ b/core/src/main/java/bio/terra/pearl/core/model/datarepo/Dataset.java
@@ -18,5 +18,6 @@ public class Dataset extends BaseEntity {
     private UUID datasetId;
     private String datasetName;
     private String description;
+    private String status;
     private Instant lastExported;
 }
--- a/core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
@@ -3,7 +3,6 @@ package bio.terra.pearl.core.service.datarepo;
 import bio.terra.datarepo.client.ApiException;
 import bio.terra.datarepo.model.JobModel;
 import bio.terra.datarepo.model.JobModel.JobStatusEnum;
-import bio.terra.datarepo.model.TableDataType;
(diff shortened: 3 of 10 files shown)


core/src/main/java/bio/terra/pearl/core/dao/datarepo/DatasetDao.java
```diff
@@ -29,6 +29,10 @@ public Optional<Dataset> findByDataRepoId(UUID dataRepoId) {
         return findByProperty("dataset_id", dataRepoId);
     }
 
+    public Optional<Dataset> findByDatasetName(String datasetName) {
```
MatthewBemis, line 32: Even though dataset names are unique, I would prefer to look this up by the Juniper-generated UUID of the dataset. In order to do so, I need to add a FK to the `data_repo_job` table. I'll do this in an upcoming PR.

core/src/main/resources/db/changelog/changesets/2023_05_15_dataset_creation.yaml
```diff
@@ -0,0 +1,12 @@
+databaseChangeLog:
+  - changeSet:
+      id: "dataset_creation"
+      author: mbemis
+      changes:
+        - dropNotNullConstraint:
+            columnName:  dataset_id
```
MatthewBemis, line 7: In that upcoming DB refactor I also want to rename this column to `tdr_dataset_id`. Datasets have 2 UUIDs (`id`: the juniper UUID for the record, and `dataset_id`: the TDR-generated UUID for the dataset). This is kind of confusing, but could be less confusing with better naming.

core/src/main/resources/db/changelog/changesets/2023_05_15_dataset_creation.yaml
```diff
@@ -0,0 +1,12 @@
...
+            tableName:  dataset
+        - addColumn:
+            tableName: dataset
+            columns:
+              - column: { name: status, type: text, constraints: { nullable: false }, defaultValue: "succeeded" }
```
MatthewBemis: Up until this point, we've only saved datasets that have succeeded, so this is a sane backfill value.

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
```diff
@@ -205,20 +213,23 @@ public void pollRunningJobs() {
     private void pollAndUpdateCreateJobStatus(DataRepoJob job) {
         try {
             JobStatusEnum jobStatus = dataRepoClient.getJobStatus(job.getTdrJobId()).getJobStatus();
+            Dataset existingDataset = datasetDao.findByDatasetName(job.getDatasetName()).get();
```
MatthewBemis, line 216: this can be removed with https://github.com/broadinstitute/pearl/pull/386/files#r1198205721

core/src/main/resources/db/changelog/changesets/2023_05_15_dataset_creation.yaml
devonbush: Is this status going to be a java enum?  Should this be "SUCCEEDED"?

core/src/main/java/bio/terra/pearl/core/model/datarepo/Dataset.java
```diff
@@ -17,5 +17,6 @@ public class Dataset extends BaseEntity {
     private UUID studyEnvironmentId;
     private UUID datasetId;
     private String datasetName;
+    private String status;
```
devonbush: Should we go ahead and make a a simple enum for this?

devonbush (approved):
looks good, couple of minor suggestions

nawatts (approved):

core/src/main/java/bio/terra/pearl/core/model/datarepo/Dataset.java
MatthewBemis: Made it an enum 78aa8c84

PR #387, opened 2023-05-18
[JN-382] Move survey and response types to ui-core

nawatts:
This was a bit more complicated than the last one since many of these types were inconsistent (or at least inconsistently defined) between admin and participant UIs.


ui-admin/src/study/participants/survey/SurveyFullDataView.tsx
nawatts, line 1: In ui-admin, the Answer type had stringValue, numberValue, objectValue, and booleanValue all defined. Whereas in ui-participant, each of those fields could be undefined. Adding undefined to the types required some changes here.

ui-admin/src/study/surveys/VersionSelector.tsx
```diff
@@ -27,7 +27,7 @@ export default function VersionSelector({ studyShortname, stableId, updateVersio
           { versionList.map(survey => {
             return <li key={survey.version}>
               <button className="btn btn-secondary" onClick={() => updateVersion(survey.version)}>
-                {survey.version} <span className="detail">({survey.createdAt?.substring(0, 16)})</span>
+                {survey.version} {/*<span className="detail">({survey.createdAt?.substring(0, 16)})</span>*/}
```
nawatts, line 30: createdAt was previously incorrectly typed as string instead of number. Left this commented out for now since this entire component is broken (it requests data from an API endpoint that doesn't exist).

ui-core/src/types/surveys.ts
```diff
@@ -0,0 +1,65 @@
+export type VersionedForm = {
```
devonbush: "forms.ts" might be a better name for this file since it has both surveys and consents

devonbush (approved):
one small optional suggestion

ui-participant/src/api/api.tsx
```diff
@@ -286,7 +242,7 @@ export default {
   async submitPreRegResponse({ surveyStableId, surveyVersion, preRegResponse }:
                                {
                                  surveyStableId: string, surveyVersion: number,
-                                 preRegResponse: PreregistrationResponse
+                                 preRegResponse: Partial<PreregistrationResponse>
```
nawatts, line 245: Not great (preferably this would have a more precise type), but I don't think it's worse than the current cast of a partial PreregistrationResponse to a PreregistrationResponse.

ui-participant/src/landing/registration/Preregistration.tsx
```diff
@@ -19,13 +19,12 @@ export default function PreRegistration({ registrationContext }: { registrationC
     if (!surveyModel) {
       return
     }
-    const responseDto = {
+    const responseDto: Partial<PreregistrationResponse> = {
       resumeData: getResumeData(surveyModel, null),
       answers: getSurveyJsAnswerList(surveyModel),
-      creatingParticipantId: null,
```
nawatts, line 25: This isn't a field of pre-registration responses. There is a `creatingParticipantUserId` field.

breilly2 (approved):
Nice. Interesting observation: these are a little tougher to review because a text diff doesn't really tell the whole story of what changes were needed. Makes me glad we have ui-core now before code copying got any worse!

PR #388, opened 2023-05-18
[JN-382] Move portal and study types to ui-core

nawatts:


ui-admin/src/api/api.tsx
```diff
@@ -17,11 +22,21 @@ export type {
   NavbarItem,
   NavbarItemInternal,
   NavbarItemInternalAnchor,
+  NotificationConfig,
+  Portal,
+  PortalEnvironment,
+  PortalEnvironmentConfig,
+  PortalStudy,
```
breilly2, line 29: Paying a little closer attention this time around... is exporting these types here done to keep the diff smaller? Could/should other code be importing these from `@juniper/ui-core` directly?

ui-participant/src/landing/registration/PortalRegistrationRouter.tsx
```diff
@@ -10,7 +10,7 @@ import RegistrationUnauthed from './RegistrationUnauthed'
 
 
 export type RegistrationContextT = {
-  preRegSurvey: Survey | null,
+  preRegSurvey?: Survey,
```
breilly2, line 13: This allows for `undefined` as well as `null`, right? As a reader, this does express something different to me because I see `undefined` as uninitialized and `null` as initialized but absent. That said, I'm not necessarily objecting to this change; I haven't looked to see if that's even an important distinction in this particular case.

breilly2 (approved):

ui-admin/src/api/api.tsx
devonbush: The theory is that eventually these types will diverge (albeit slightly) between participant and admin UX (there are some properties of objects that the Participant UX does not/should not know about).  So putting a layer of indirection in the imports allows that divergence to happen more reliably

ui-participant/src/landing/registration/PortalRegistrationRouter.tsx
devonbush: I think that just allows undefined, not null.  I defer to Nick on this.  I tend to like using <Object> | null for useState constructions because, like you said, it's not that it's not-thought-of, it's that it's explicitly empty

devonbush (approved):

ui-participant/src/landing/registration/PortalRegistrationRouter.tsx
nawatts: It's an [optional property](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#optional-properties), so `<type> | undefined`. This field in the registration context is copied from the portal environment so the types need to match. And it looks like on the participant UI, the portal environment doesn't have a `preRegSurvey` key if a pre-registration survey is not configured.

ui-admin/src/api/api.tsx
nawatts: What @devonbush said. There are some types where different information is returned by the admin/participant UIs. For example, the participant API returns empty lists for StudyEnvironment's configuredConsents, configuredSurveys, and notificationConfigs properties. We could improve the type safety in the participant UI by having it's api modules export a StudyEnvironment type derived from ui-core's StudyEnvironment type (for example, `type StudyEnvironment = Omit<BaseStudyEnvironment, 'configuredConsents' | 'configuredSurveys' | 'notificationConfigs'>`).

PR #390, opened 2023-05-19
[JN-386] Combine ESLint configs

nawatts:
Currently, we have the same ESLint configuration in ui-admin/.eslintrc.js and ui-participant/.eslintrc.js. This replaces those with one file at the project root. It adds `eslintConfig` to package.json files so that the configuration is applied regardless of whether ESLint is run from the project root (`npm -w ui-admin start`) or the component folder (`cd ui-admin && npm start`). And it adds a `lint` script to the project root (`npm run lint`) to run ESLint on all UI components.


devonbush (approved):

breilly2 (approved):
Awesome!

PR #391, opened 2023-05-19
[JN-386] Combine UI CI workflows

nawatts:
Now that we have the ui-core package, there's some duplication between the admin and participant UI CI workflows: they both build ui-core. We'll also want to run tests for ui-core (once we have some tests to run) but there's no reason to run those tests twice. So instead of adding a third workflow for ui-core, this combines the UI CI workflows into one, similar to how there's one CI workflow for the API.


.github/workflows/ui-ci.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+      - name: Install dependencies
+        run: npm ci
+      - name: Build
+        run: npm --workspaces run build
```
nawatts, line 25: Scripts are run in the order in which workspaces are listed in package.json, so this ensures that ui-core is built before its dependents are.

devonbush (approved):

.github/workflows/ui-ci.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Run tests
+        run: npm --workspace=ui-core --workspace=ui-admin --workspace=ui-participant test --if-present
```
nawatts, line 27: Workspaces listed out here to avoid running e2e tests (added in #392).

MatthewBemis (approved):

PR #392, opened 2023-05-19
[JN-394] Setup e2e tests

nawatts:
This configures [Playwright](https://playwright.dev/) for end to end tests. To run e2e tests, run `npm -w e2e-tests test`.

If the APIs and UI dev servers are already running, it will use those. Otherwise, it will start them for you.

In CI, it runs tests against the UI bundled into the Java app.


devonbush (approved):
Looks good -- I'm not sure why just one of the tests is failing for me locally.  Chromium and firefox pass, but webkit fails with 
```
Error: page.goto: A server with the specified hostname could not be found.
=========================== logs ===========================
navigating to "http://sandbox.ourhealth.localhost:8081/", waiting until "load"
```

devonbush:
adding `127.0.0.1 sandbox.ourhealth.localhost` to my /etc/hosts fixes the problem--Nick pointed out that it's an issue with how webkit doesn't treat localhost as a tld

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,53 @@
...
+      with:
+        node-version: 16
+    - name: Install dependencies
+      run: npm ci
+    - name: Build applications
+      run: |
+        ./gradlew :api-admin:build -x test
+        ./gradlew :api-admin:copyWebApp
+        ./gradlew :api-participant:build -x test
+        ./gradlew :api-participant:copyWebApp
```
breilly2, line 51: I imagine we're building the applications in other actions. Does this action need to be isolated or can it use artifacts produced by another action? Or is there something special about the artifacts needed for the e2e tests?

breilly2 (approved):
Looks good! I also had to run `npx playwright install` before tests would run.

.github/workflows/e2e-tests.yml
nawatts: This is currently the only place where we build the UI into the Java app. We could possibly combine all CI into one workflow and go: build JS => build Java => run e2e tests.

nawatts:
> Looks good! I also had to run npx playwright install before tests would run.

Added a postinstall script to the e2e-tests workspace to automatically run this.

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,55 @@
+name: End-to-end tests
+
+# TODO: Once tests are added, run on PRs and commits to development.
+on: workflow_dispatch
```
nawatts, line 4: Since there are no e2e tests written yet, there's no point in running this on PRs/commits.

For now, this lets us run the workflow manually.

devonbush (approved):

PR #394, opened 2023-05-23
[JN-260] Update types for navbar configuration

nawatts:
This defines a separate type for each type of navbar item (internal, internal anchor, mailing list, external) and redefines the `NavbarItem` type as a union of those individual types. This way, TypeScript can automatically narrow the type by looking at the `itemType` field, removing the need for the `isInternalLink` and `isInternalAnchorLink` helpers.

### DIFF
--- a/ui-admin/src/portal/siteContent/SiteContentView.tsx
+++ b/ui-admin/src/portal/siteContent/SiteContentView.tsx
@@ -1,9 +1,9 @@
 import React, { useState } from 'react'
-import { HtmlPage, NavbarItem, PortalEnvironment } from '../../api/api'
+import { HtmlPage, NavbarItemInternal, PortalEnvironment } from '../../api/api'
 
 const SiteContentView = ({ portalEnv }: {portalEnv: PortalEnvironment}) => {
   const selectedLanguage = 'en'
-  const [selectedNavItem, setSelectedNavItem] = useState<NavbarItem | null>(null)
+  const [selectedNavItem, setSelectedNavItem] = useState<NavbarItemInternal | null>(null)
 
   if (!portalEnv.siteContent) {
     return <div>no site content configured yet</div>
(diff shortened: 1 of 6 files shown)


ui-participant/src/App.tsx
```diff
@@ -79,9 +79,14 @@ function App() {
   let landingRoutes: JSX.Element[] = []
   if (localContent.navbarItems) {
     landingRoutes = localContent.navbarItems
-      .filter(isInternalLink)
-      .map((navItem: NavbarItemInternal, index: number) => <Route key={index} path={navItem.htmlPage.path}
-        element={<HtmlPageView page={navItem.htmlPage}/>}/>)
+      .filter((navItem: NavbarItem): navItem is NavbarItemInternal => navItem.itemType === 'INTERNAL')
```
nawatts, line 82: Still need a type predicate here to make filter's return value have the narrowed element type.

ui-participant/src/Navbar.tsx
```diff
@@ -175,10 +175,10 @@ export function CustomNavLink({ navLink }: { navLink: NavbarItem }) {
     alert(`mailing list ${navLinkObj.label}`)
   }
 
-  if (isInternalLink(navLink)) {
+  if (navLink.itemType === 'INTERNAL') {
```
nawatts, line 178: Type predicates no longer needed here.

ui-admin/src/portal/siteContent/SiteContentView.tsx
```diff
@@ -18,11 +18,13 @@ const SiteContentView = ({ portalEnv }: {portalEnv: PortalEnvironment}) => {
   return <div className="container d-flex bg-white p-3">
     <ul className="list-group">
       <li className="list-group-item" onClick={() => setSelectedNavItem(null)}>Landing page</li>
-      {localContent.navbarItems.map(navItem => <li key={navItem.label} className="list-group-item" role="button"
-        onClick={() => setSelectedNavItem(navItem)}>
```
nawatts, line 22: Improved types already found a bug. This would allow selecting any type of navbar item, but the selected navbar item was assumed to be an internal item.

ui-admin/src/portal/siteContent/SiteContentView.tsx
```diff
@@ -1,9 +1,9 @@
 import React, { useState } from 'react'
-import { HtmlPage, NavbarItem, PortalEnvironment } from '../../api/api'
+import { HtmlPage, NavbarItemInternal, PortalEnvironment } from '../../api/api'
```
devonbush, line 2: Not your change, but we should use abosulte imports except where the files are closely related
```suggestion
import { HtmlPage, NavbarItemInternal, PortalEnvironment } from 'api/api'
```

devonbush (approved):
Nice, this is a use of typescript I'm excited to get fluent in myself

ui-admin/src/portal/siteContent/SiteContentView.tsx
nawatts: I think existing code is wildly inconsistent about this. Looks like a case for [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-relative-parent-imports.md).

breilly2 (approved):

PR #397, opened 2023-05-24
[JN-388] Minor refactoring of TDR db tables

MatthewBemis:
There are two minor pieces to this refactor:

1) There are two UUIDs for datasets. One is the internal UUID that's generated by Juniper, and the other is a UUID that's generated by TDR. It's been a bit confusing to read the code because the distinction wasn't clear from the naming. Now all instances of the TDR ID are referred to as `tdrDatasetId`, and all other instances (`datasetId` or `dataset.id`) are referring to the Juniper ID. I've also tried to draw a clear line in how these are used, so the only time that we'll use the TDR ID is when we are interacting with the DataRepoClient or storing the TDR ID in the database.

2. The `data_repo_job` used to use `dataset_name` as it's reference to the `dataset` table. While `dataset_name`s are globally unique, I wanted to switch this over to using the `dataset` UUID because it's a more natural FK and will never be mutable or reused (necessary to support deleteDataset in an upcoming PR)


api-admin/src/main/java/bio/terra/pearl/api/admin/service/DataRepoExportExtService.java
```diff
@@ -52,10 +52,9 @@ public List<DataRepoJob> getJobHistoryForDataset(
     Portal portal = authUtilService.authUserToPortal(user, portalShortcode);
     authUtilService.authUserToStudy(user, portalShortcode, studyShortcode);
 
-    StudyEnvironment studyEnv =
-        studyEnvironmentService.findByStudy(studyShortcode, environmentName).get();
+    Dataset dataset = dataRepoExportService.getDatasetByName(datasetName);
```
MatthewBemis, line 55: One of the main points of this refactor is to lean more on using `datasetId`s as the lookup key, but that's a concept hidden from the user and `datasetName`s are globally unique in Juniper and TDR anyway so it's alright to do it here

ui-admin/src/study/participants/datarepo/DatasetDashboard.tsx
```diff
@@ -121,8 +121,8 @@ const DatasetDashboard = ({ studyEnvContext }: {studyEnvContext: StudyEnvContext
                   <label>Date Created:</label> { instantToDefaultString(datasetDetails?.createdAt) }
                 </div>
                 <br/>
-                { datasetDetails?.status == 'succeeded' &&
-                  <a href={`https://jade.datarepo-dev.broadinstitute.org/datasets/${datasetDetails?.datasetId}`}
+                { datasetDetails?.status == 'CREATED' &&
```
MatthewBemis, line 123: Drive-by fixing a bug

core/src/main/resources/db/changelog/changesets/2023_05_23_refactor_tdr_tables.yaml
```diff
@@ -0,0 +1,19 @@
...
+      author: mbemis
+      changes:
+        - renameColumn:
+            oldColumnName: dataset_id
+            newColumnName: tdr_dataset_id
+            tableName: dataset
+        - addColumn:
+            tableName: data_repo_job
+            columns:
+              - column: { name: dataset_id, type: uuid, constraints: { nullable: false } }
```
MatthewBemis: This change poses a problem for existing data because existing jobs will have null dataset IDs. I could add a raw SQL statement here to look up the dataset IDs from the dataset_name and backfill, but I think at this stage it's reasonable to just drop the data from the demo server. Production shouldn't have any datasets yet.

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
```diff
@@ -301,11 +307,4 @@ public boolean getServiceStatus() {
...
     }
 
-    public String makeDatasetName(String deploymentZone, String studyName, String environmentName) {
-        if(deploymentZone.equalsIgnoreCase("prod"))
-            return String.format("d2p_%s_%s", studyName, environmentName);
-        else
-            return String.format("d2p_%s_%s_%s", deploymentZone, studyName, environmentName);
-    }
-
 }
```
MatthewBemis, line 300: Unrelated, but this naming scheme isn't used anymore since we're doing on-demand which allows users to choose a name.

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
```diff
@@ -110,24 +114,26 @@ public void createDataset(StudyEnvironment studyEnv, String datasetName, String
...
     }
 
-    public String uploadCsvToAzureStorage(UUID studyEnvironmentId, String datasetName) {
+    public String uploadCsvToAzureStorage(UUID studyEnvironmentId, UUID datasetId) {
         ExportOptions exportOptions = new ExportOptions(false, false, false, ExportFileFormat.TSV, null);
 
         //Even though this is actually formatted as a TSV, TDR only accepts files ending in .csv or .json.
         //In the DataRepoClient call, we specify that the CSV delimiter is "\t", which will make it all work fine.
-        String blobName = datasetName + "_" + studyEnvironmentId + "_" + Instant.now() + ".csv";
+        String blobName = datasetId.toString() + "_" + studyEnvironmentId + "_" + Instant.now() + ".csv";
```
MatthewBemis, line 136: Keeping the theme of _"don't use `datasetName` as a foreign key"_, this is preparing for the future `deleteDataset` feature. A `datasetName` could theoretically be re-used so I don't want to cross wires with `blobName`s that look like they might belong to a current dataset but actually belong to a deleted dataset.

core/src/main/resources/db/changelog/changesets/2023_05_23_refactor_tdr_tables.yaml
```diff
@@ -0,0 +1,19 @@
...
+        - addForeignKeyConstraint:
+            baseTableName: data_repo_job
```
devonbush: is there a reason why you have this constraint declared in a separate block from the one above? 
adding `foreignKeyName: fk_dataset_job_dataset_id, references: dataset(id)` inside the "constraints" block above should accomplish the same thing and keep the schema easier to read

devonbush (approved):
Looks good, one suggestion for simplifying the upgrade script

core/src/main/resources/db/changelog/changesets/2023_05_23_refactor_tdr_tables.yaml
MatthewBemis: Update: I just deleted the existing rows in the dev database because they were worthless, and also verified that there was nothing in prod

core/src/main/java/bio/terra/pearl/core/service/datarepo/DatasetService.java
```diff
@@ -26,21 +26,17 @@ public DatasetService(DatasetDao dao, DataRepoClient dataRepoClient) {
     public void delete(Dataset dataset) {
         dao.delete(dataset.getId());
         try {
-            JobModel jobModel = dataRepoClient.deleteDataset(dataset.getDatasetId());
-            logger.info("Deleted dataset: {} (TDR job ID {})", dataset.getDatasetId(), jobModel.getId());
+            JobModel jobModel = dataRepoClient.deleteDataset(dataset.getTdrDatasetId());
+            logger.info("Deleted dataset: {} (TDR job ID {})", dataset.getId(), jobModel.getId());
         } catch (ApiException e) {
-            logger.error("Unable to delete dataset {}. Error: {}", dataset.getDatasetId(), e.getMessage());
+            logger.error("Unable to delete dataset {}. Error: {}", dataset.getId(), e.getMessage());
```
breilly2, line 33: These used to all use `dataset.getDatasetId()`. I'm guessing that you intentionally chose between `dataset.getTdrDatasetId()` and `dataset.getId()` here... and this is probably exactly the kind of thing you're clearing up with this refactor!

breilly2 (approved):
Nice refactor!

devonbush (changes requested):
code looks good -- in accordance with our new team priority on frontend testin, since this makes a frontend change, it should include at least 1 frontend test (it can be a very simple one)

core/src/main/resources/db/changelog/changesets/2023_05_23_refactor_tdr_tables.yaml
MatthewBemis: Done in 9be083d7edf1735eed7d7dbf76eeed32e2848a83

MatthewBemis:
@devonbush I added a couple of unit tests in 6690ddee685c372a6c80bfb30609df08f6f8d4ee related to the bug that I drive-by fixed in this PR. It kicked off a bit of minor refactoring of the related UI code. I also copied in some of the code you wrote for `mocking-utils` in https://github.com/broadinstitute/juniper/pull/400

devonbush (approved):
nice

PR #398, opened 2023-05-24
[JN-260] Align navbar item and button/link config

nawatts:
Navbar items and buttons/links in landing pages have similar options: internal links, external links, join mailing list buttons. However, their configuration schemas are different.

https://github.com/broadinstitute/pearl/blob/95e5d0f337532b574c9e706ea34853f90129be08/ui-core/src/types/landingPageConfig.ts#L47-L74

https://github.com/broadinstitute/pearl/blob/95e5d0f337532b574c9e706ea34853f90129be08/ui-participant/src/landing/ConfiguredButton.tsx#L7-L34

This makes configuration more difficult since you have to remember which fields to use in which case instead of always using the same fields.

This more closely aligns the two by modifying navbar item and renaming `label` to `text` and combining `anchorLinkPath` and `externalLink` into `href`.

### DIFF
--- a/core/src/main/resources/db/changelog/changesets/2023_05_24_navbar_item.yaml
+++ b/core/src/main/resources/db/changelog/changesets/2023_05_24_navbar_item.yaml
@@ -1,0 +1,22 @@
+databaseChangeLog:
+  - changeSet:
+      id: navbar_item
+      author: nwatts
+      changes:
+        - renameColumn:
+            tableName: navbar_item
+            oldColumnName: label
+            newColumnName: text
+        - dropColumn:
+            tableName: navbar_item
+            columns:
+              - column:
+                  name: external_link_url
+              - column:
+                  name: anchor_link_path
+        - addColumn:
+            tableName: navbar_item
+            columns:
+              - column:
+                  name: href
+                  type: text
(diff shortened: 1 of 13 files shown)


core/src/main/resources/db/changelog/changesets/2023_05_24_navbar_item.yaml
```diff
@@ -0,0 +1,22 @@
...
+            tableName: navbar_item
+            oldColumnName: label
+            newColumnName: text
+        - dropColumn:
+            tableName: navbar_item
+            columns:
+              - column:
+                  name: external_link_url
+              - column:
+                  name: anchor_link_path
```
nawatts, line 16: Nothing currently uses these columns, so we can drop and replace them instead of renaming / merging data.

devonbush (approved):

devonbush (changes requested):
Please add a test to Navbar.tsx (can be a trivial "does this component render at all" test) to go along with our recent discussion of "let's make all frontend PRs have at least one new test"

ui-participant/src/Navbar.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    render(<CustomNavLink navLink={navbarItem} />)
+
+    // Assert
+    const link = screen.getByText('Mailing list link')
+
+    // Act
+    await user.click(link)
+
+    // Assert
+    expect(window.alert).toHaveBeenCalled()
```
nawatts, line 92: Obviously, this isn't the desired behavior, but leaving it as-is for another ticket/PR.

devonbush (approved):

MatthewBemis (approved):

PR #399, opened 2023-05-25
[JN-381] Respect column types (mostly) + define primary key

MatthewBemis:
This gets us to a point where we're _mostly_ respecting column types. `DATETIME` is still stored as `STRING` but everything else is respected, including basic `DATE` columns. Unfortunately, the `DATETIME` format will need some coercing on our end to get it to play nicely with TDR: it looks like TDR on Azure uses the MSSQL `datetime2` format under the hood. I am wondering if the right thing to do is to create a new `ExportFileFormat` specifically for TDR?

I also added a mechanism to respect primary keys. Previously, the `enrollee` table had no primary key.

Lastly, this also extracts a bit of code to make it easier to define other tables in the future, by adding a new `DatasetTableDefinition`.


breilly2:
> TDR on Azure uses the MSSQL `datetime2` format

Is this a TDR bug? Should TDR behave differently on Azure vs GCP? Maybe it makes sense to respect native type capabilities, but maybe not.

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
```diff
@@ -167,14 +170,14 @@ public void ingestDataForStudyEnvironment(Dataset studyEnvDataset) {
...
-    public List<String> generateDatasetSchema(UUID studyEnvironmentId) {
+    public Map<String, TableDataType> generateDatasetSchema(UUID studyEnvironmentId) {
         ExportOptions exportOptions = new ExportOptions(false, false, false, ExportFileFormat.TSV, null);
 
         //Backtrack from studyEnvironmentId to get the portalId, so we can export the study environment data
         StudyEnvironment studyEnv = studyEnvironmentDao.find(studyEnvironmentId).orElseThrow(() -> new NotFoundException("Study environment not found."));
         PortalStudy portalStudy = portalStudyDao.findByStudyId(studyEnv.getStudyId()).stream().findFirst().orElseThrow(() -> new NotFoundException("Portal study not found."));
 
-        List<String> columnKeys = new ArrayList<>();
+        Map<String, TableDataType> columnKeys = new LinkedHashMap<>();
```
breilly2: I've always wished that Java collections had a map interface that expressed predictable iteration order.

I guess in this case you don't actually need map semantics (lookup by key), just a pair of values, so this _could_ be a `List` some record type that holds the two values.

Feel free to push back on this one, but the method signature change from `List` to `Map` did give me pause until I came across this line, which is _very_ important if we care about column order.

breilly2 (approved):
Looks good. One suggestion that I _think_ I would do if I were writing the code, but it's more about readability and heading off a (admittedly unlikely) potential bug down the road. I'll leave it up to you.

MatthewBemis:
@breilly2 I will check with the TDR team, but from what I quickly gathered from skimming their code, it intentionally stores as `datetime2` for Azure: https://github.com/DataBiosphere/jade-data-repo/blob/308d88da572efffbcd5130b77c90c93b4bac5557/src/main/java/bio/terra/common/SynapseColumn.java#L67

Sadly TDR doesn't provide a useful error message when this happens, so to some degree, I'm guessing a bit. All I know for sure is that the `DATETIME` conversions are what cause TDR export to fail. I will check to be sure and follow up with fixing this last data type in another ticket. And if it's something they can tweak on their end, even better

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
devonbush: I concur-- instead of a Map, this should be a List of "TdrColumn" objects, and for now we can define "TdrColumn" as a record of name and datatype, but I wouldn't be surprised if we add more later.

devonbush: For me, this isn't about ordering (it's usually bad practice to rely on column ordering in any external database-like-system), and more about readability.  (Which might be an argument that it should be a Set rather than a List, but I'll defer to your TDR expertise on which is best)

devonbush (approved):

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
MatthewBemis: I added two new record types: `TdrColumn` and `TdrTable`, and switched these over to sets. But I've kept it as a `LinkedHashSet` because I think the ordering is important. The column order in TDR seems to be dependent on the order in which the columns were initialized. So maintaining the order in our code allows us to put things like `enrollee_shortcode` and basic profile information first in the TDR UI, and it also keeps particular survey questions grouped together. Otherwise you'll just see a bunch of columns like `oh_oh_medHx.otherKidneyConditionAgeAtDx` at the beginning of the dataset which I don't think is a great UX when viewing the dataset.

I've added a code comment to explain this, but am happy to revisit the decision.

PR #400, opened 2023-05-26
JN-390 custom email functionality

devonbush:
As demoed, this adds the ability for admins to select participants from the participant list and then send them a custom email.  To accomplish this, a new "customMessages" attribute is added to Notification that allows us to save the custom text.

  For now, the customizable message keys of the "adHoc" email template are hardcoded in the ui: `adHocSubject` and `adHocMessage`    If this feature expands, we'll want to have those keys stored with descriptions on the template itself.

TO TEST:
1. restart adminApp
2. repopulate ourhealth
3. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants`
4. select one or more participants
5. click "Send Email"
6. select the "ad hoc" template from the config list
7. specify some custom message
8. click send


api-admin/src/test/java/bio/terra/pearl/api/admin/service/AuthUtilServiceTests.java
nawatts, line 1: Should these tests be renamed from authAdmin to authUser to match the functions they're testing?

core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
```diff
@@ -58,6 +71,12 @@ public String lookup(String key) {
         return "";
     }
 
+    public String getSiteLink(PortalEnvironment portalEnv, Portal portal) {
+        String href = routingPaths.getParticipantBaseUrl(portalEnv, portal.getShortcode());
+        return String.format("<a rel=\"noopener\" href=\"%s\" target=\"_blank\">%s</a>", href, href);
```
nawatts, line 76: Should this link have some friendlier text than the URL?

ui-core/src/types/study.ts
```diff
@@ -10,8 +10,8 @@ export type StudyEnvironment = {
   id: string
   environmentName: string
   studyEnvironmentConfig: StudyEnvironmentConfig
-  preEnrollSurvey: Survey
-  preEnrollSurveyId: string
+  preEnrollSurvey: Survey | null
+  preEnrollSurveyId: string | null
```
nawatts: I'm not certain, but I think these keys are not present if there's no pre-enroll survey (assuming these behave the same way as portal environment's pre-registration surveys).

```suggestion
  preEnrollSurvey?: Survey
  preEnrollSurveyId?: string
```

ui-core/src/types/study.ts
```diff
@@ -68,7 +68,10 @@ export type NotificationConfig = {
 
 export type EmailTemplate = {
   subject: string
-  body: string
+  body: string,
+  name: string,
+  stableId: string,
```
nawatts: Nit: for consistency, none of the other types in this file have trailing commas.

```suggestion
  body: string
  name: string
  stableId: string
```

nawatts (approved):

core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
devonbush: yeah, I waffled on that -- not sure what that text would be though -- I'll check if Erin has any quick thoughts

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/NotificationsController.java
```diff
@@ -69,4 +75,35 @@ public ResponseEntity<Object> test(
...
+            environmentName,
+            adHoc.enrolleeShortcodes,
+            adHoc.customMessages,
+            adHoc.notificationConfigId);
+    return ResponseEntity.ok(configUsed);
+  }
+
+  /**
+   * object for specifying an adhoc notification. If notificationConfigId is specified, a saved
+   * config will be used otherwise the notificationConfig given will be created and then used
```
breilly2: Is that true? I see `NotificationExtService.sendAdHoc` finding a config, but not creating one.

ui-admin/src/study/participants/AdHocEmailModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+      return
+    }
+    Api.sendAdHocNotification({
+      portalShortcode: studyEnvContext.portal.shortcode,
+      studyShortcode: studyEnvContext.study.shortcode,
+      envName: studyEnvContext.currentEnv.environmentName,
+      enrolleeShortcodes,
+      customMessages: { adHocMessage, adHocSubject },
+      notificationConfigId: selectedConfig.id
+    })
```
breilly2: Any reason not to call `onDismiss()` here? When the modal didn't dismiss, I thought maybe something had gone wrong.

breilly2 (approved):
Looks good. There are a few changes here related to pre-enrollment that don't look related to emails to me. Maybe part of the member's choice experiment that got caught up in this PR?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/NotificationsController.java
devonbush: good catch, I originally wrote this to support both, but liked it better always finding one

PR #401, opened 2023-05-30
[JN-362] [JN-365] Add admin UI for requesting kits and initial DSM integration

breilly2:
Definitely some rough edges here that will be smoothed out with future tickets. Notable changes:
* Kit type entity instead of string value. This allows us to describe them better and to configure studies with them. These are global config since the kit types that Juniper can use are tied to what GP is set up to handle.
* Initial UI for requesting kits and viewing kit requests
* Initial integration with DSM


core/build.gradle
```diff
@@ -23,6 +23,7 @@ dependencies {
     implementation group: 'org.jdbi', name: 'jdbi3-sqlobject', version: '3.34.0'
     implementation group: 'org.jdbi', name: 'jdbi3-postgres', version: '3.34.0'
     implementation 'org.springframework.boot:spring-boot-starter-data-jdbc:2.7.10'
+    implementation 'org.springframework.boot:spring-boot-starter-webflux:2.7.10'
```
breilly2, line 26: This is for Spring's `WebClient`, which may be a bit more fully featured than what we need, but since we're using Spring already, this is what we get out-of-the-box.

populate/src/main/java/bio/terra/pearl/populate/service/KitRequestPopulator.java
```diff
@@ -0,0 +1,53 @@
...
+import bio.terra.pearl.core.dao.kit.KitTypeDao;
+import bio.terra.pearl.core.model.kit.KitRequest;
+import bio.terra.pearl.populate.dto.kit.KitRequestDto;
+import bio.terra.pearl.populate.service.contexts.FilePopulateContext;
+import org.springframework.stereotype.Service;
+
+import java.io.IOException;
+import java.util.Optional;
+
+// TODO: finish implementing this
```
breilly2: I've been winging it with these populators. They all seem a little bit different and the semantics of what each of the method implementions should do isn't 100% clear to me. Let me know if `KitTypePopulator` looks right and I'll come back and fill this in.

populate/src/main/java/bio/terra/pearl/populate/service/StudyPopulator.java
```diff
@@ -60,6 +67,9 @@ public StudyPopulator(StudyService studyService,
         this.preEnrollmentResponseDao = preEnrollmentResponseDao;
         this.portalDiffService = portalDiffService;
         this.studyUpdateService = studyUpdateService;
+        this.portalEnvironmentService = portalEnvironmentService;
```
breilly2, line 70: Drive-by bug fix: This wasn't previously being initialized.

ui-admin/src/study/participants/KitRequests.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { useMemo, useState } from 'react'
+import { Enrollee, KitRequest } from 'api/api'
+import { StudyEnvContextT } from '../StudyEnvironmentRouter'
+import { ColumnDef, getCoreRowModel, useReactTable } from '@tanstack/react-table'
+import { basicTableLayout } from 'util/tableUtils'
+import RequestKitModal from './RequestKitModal'
+import { faPlus } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import {instantToDefaultString} from "../../util/timeUtils";
```
devonbush: change to absolute import and single quotes
```suggestion
import {instantToDefaultString} from 'util/timeUtils';
```

ui-admin/src/api/api.tsx
```diff
@@ -435,8 +452,31 @@ export default {
...
+  async createKitRequest(portalShortcode: string, studyShortcode: string, envName: string,
+                         enrolleeShortcode: string, kitType: string): Promise<string> {
+    const params = new URLSearchParams({ kitType })
+    const url =
+      `${baseStudyEnvUrl(portalShortcode, studyShortcode, envName)}/enrollees/${enrolleeShortcode}/requestKit?${params}`
+    const response = await fetch(url, { method: 'POST', headers: this.getInitHeaders() })
+    return await this.processJsonResponse(response)
+  },
+
+  async fetchKitRequests(portalShortcode: string, studyShortcode: string, envName: string,
```
devonbush: maybe rename to `fetchEnrolleeKitRequests` for clarity since we'll likely have a few of these (fetch all, fetch by study, etc..)?

ui-admin/src/study/participants/KitRequests.tsx
```diff
@@ -0,0 +1,71 @@
...
+                                        onUpdate: () => void
+                                      }) {
+  const [showRequestKitModal, setShowRequestKitModal] = useState(false)
+
+  const onSubmit = () => {
+    setShowRequestKitModal(false)
+    onUpdate()
+  }
+
+  const columns = useMemo<ColumnDef<KitRequest, string>[]>(() => [{
```
devonbush: declaring these columns outside the component function will save a useMemo call and make it clearer that the column defs don't change in response to the returned data

ui-admin/src/study/participants/KitRequests.tsx
```diff
@@ -0,0 +1,71 @@
...
+    header: 'Created',
+    accessorKey: 'createdAt',
+    accessorFn: data => instantToDefaultString(data.createdAt)
+  }, {
+    header: 'Address',
+    cell: ({ row }) => <KitRequestAddress sentToAddressJson={row.original.sentToAddress}/>
+  }, {
+    header: 'DSM Status',
+    accessorKey: 'dsmStatus'
+  }], [enrollee.kitRequests])
```
devonbush: yeah, this isn't the right dependency here -- nothing about the column definitions change with repsect to new data.  So this should be `[]` or, better yet, the declaration pulled outside the component

core/src/main/java/bio/terra/pearl/core/config/DataStudyManagerConfig.java
```diff
@@ -0,0 +1,18 @@
+package bio.terra.pearl.core.config;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.core.env.Environment;
+import org.springframework.stereotype.Component;
+
+@Component
+@Getter @Setter
+public class DataStudyManagerConfig {
```
devonbush: This is just a wrapper around env variables, and isn't (and probably shouldn't) be used anywhere outside of DataStudyManagerClient.  Omitting this class, and moving this to inside the DSM client (or maybe a protected/private helper class inside DSM client, depending on what makes testing easier).  In any case, it definitely belongs in the service/kit directory for now, rather than the general config directory

core/src/main/java/bio/terra/pearl/core/service/kit/DataStudyManagerClient.java
```diff
@@ -0,0 +1,49 @@
...
+import com.auth0.jwt.JWT;
+import com.auth0.jwt.algorithms.Algorithm;
+import org.springframework.stereotype.Component;
+import org.springframework.web.reactive.function.client.WebClient;
+import reactor.core.publisher.Mono;
+
+import java.time.Instant;
+
+@Component
+public class DataStudyManagerClient {
```
devonbush: Let's rename this and the similar files to `PepperDSM______`      "DataStudyManager", even though it's the 'proper' name of the thing, is confusing to read since those are three very general-purpose words strung together. When I first saw "DataStudyManagerConfig" I thought "whoa, we're updating how all study data is managed?"

core/src/main/java/bio/terra/pearl/core/service/kit/DataStudyManagerKitRequest.java
```diff
@@ -0,0 +1,34 @@
+package bio.terra.pearl.core.service.kit;
+
+import lombok.Builder;
+import lombok.Getter;
+
+@Getter @Builder
+public class DataStudyManagerKitRequest {
```
devonbush: since this class isn't persisted directly, could we just put a PepperKitAddress as a member and save all the field copying?

core/src/main/java/bio/terra/pearl/core/model/kit/KitRequest.java
```diff
@@ -6,14 +6,16 @@
...
 @NoArgsConstructor
 public class KitRequest extends BaseEntity {
     private UUID creatingAdminUserId;
     private UUID enrolleeId;
-    private String kitType;
+    private UUID kitTypeId;
+    private KitType kitType;
     /**
      * JSON blob of address data sent to DSM, collected from Profile/MailingAddress.
      * TODO: decide if this should be separate fields, or maybe use the postgres jsonb type
```
devonbush: was the change before, but this TODO should either reference a ticket, or be reworded to "we might"

devonbush (changes requested):
Generally solid CRUD stuff.  Some suggestions for cleanups, the ones I feel most strongly about is calling things "PepperDSM____" rather than "DataStudyManager__"

BLOCKERS:  
1.  In accordance with the new team policy, this needs frontend tests -- don't worry about extensive tests, this more about "at least one reasonable one."  You can check out my [admin email PR](https://github.com/broadinstitute/pearl/pull/400) for an example of how to do a basic "confirm if endpoint is mocked, table appears" test
2.  This needs a backend test that we can successfully link studies to kit types to kits to enrollees.  (essentially confirming that all the DAOs and DB constraints match up).  The test flow would be something like:
a.  create kit type
b. create study
c. assign kit type to study
d. list study kit types
e. create enrollee in study
f. create a kit request for them of the given type (these latter two steps might not be necessary since they're essentially being what's already done in testCreatSampleKit, but they will be necessary if we're going to add server-side logic confirm the kit type for an enrollee is allowed by the study)

devonbush (changes requested):
Generally solid CRUD stuff.  Some suggestions for cleanups, the ones I feel most strongly about is calling things "PepperDSM____" rather than "DataStudyManager__".  Also, please put screenshots in the PR!

BLOCKERS:  
1.  In accordance with the new team policy, this needs frontend tests -- don't worry about extensive tests, this more about "at least one reasonable one."  You can check out my [admin email PR](https://github.com/broadinstitute/pearl/pull/400) for an example of how to do a basic "confirm if endpoint is mocked, table appears" test
2.  This needs a backend test that we can successfully link studies to kit types to kits to enrollees.  (essentially confirming that all the DAOs and DB constraints match up).  The test flow would be something like:
a.  create kit type
b. create study
c. assign kit type to study
d. list study kit types
e. create enrollee in study
f. create a kit request for them of the given type (these latter two steps might not be necessary since they're essentially being what's already done in testCreatSampleKit, but they will be necessary if we're going to add server-side logic confirm the kit type for an enrollee is allowed by the study)

populate/src/main/java/bio/terra/pearl/populate/service/KitTypePopulator.java
```diff
@@ -0,0 +1,47 @@
+package bio.terra.pearl.populate.service;
+
+import bio.terra.pearl.core.dao.kit.KitTypeDao;
+import bio.terra.pearl.core.model.kit.KitType;
+import bio.terra.pearl.populate.service.contexts.FilePopulateContext;
+import org.springframework.stereotype.Service;
+
+import java.io.IOException;
+import java.util.Optional;
+
```
devonbush, line 10: Yup, this looks right to me!  Once you add the dsmStatusJson field, you'll want to Override the `preProcessDto` method to convert that field to the String for dsmStatus

populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/enrollees/jsalk.json
```diff
@@ -64,5 +64,13 @@
...
     ],
     "currentPageNo": 6
-  }]
+  }],
+  "kitRequestDtos": [
+    {
+      "creatingAdminUsername": "breilly@broadinstitute.org",
+      "kitTypeName": "saliva",
+      "statusName": "CREATED",
+      "dsmStatus": "{\n  \"kitId\": \"11111111-2222-3333-4444-555555555555\",\n  \"labelDate\": \"2023-05-24\",\n  \"scanDate\": \"2023-05-24\",\n  \"receiveDate\": \"2023-05-24\",\n  \"scannedByEmail\": \"\",\n  \"receivedByEmail\": \"\",\n  \"labeledByEmail\": \"\",\n  \"deactivationDate\": \"\",\n  \"deactivatedBy\": \"\",\n  \"deactivationReason\": \"\",\n  \"trackingNumbers\": [],\n  \"returnTrackingNumber\": \"\",\n  \"errors\": [\n    {\n      \"errorDate\": \"2023-05-24\",\n      \"errorMessage\": \"\"\n    }\n  ],\n  \"discardDate\": \"2023-05-24\",\n  \"discardReason\": \"\"\n}\n"
```
devonbush: To make this more palatable, I'd add a field "dsmStatusJson" to KitRequestPopDto that's a JSONNode.  then the populator can convert that field to a string.  `HtmlSectionPopDto` is another place we do that.  Then you'll be able to specify DSM status as regular json rather than this extended string.

populate/src/main/java/bio/terra/pearl/populate/dto/kit/KitRequestDto.java
```diff
@@ -0,0 +1,13 @@
+package bio.terra.pearl.populate.dto.kit;
+
+import bio.terra.pearl.core.model.kit.KitRequest;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Getter @Setter @NoArgsConstructor
+public class KitRequestDto extends KitRequest {
```
devonbush: rename this to KitRequestPopDto to clarify that it's a DTO only intended for populate files.  (which is kind of redundant given its package location, but I've found stuff like that makes find-by-name in IDEs much easier.)

devonbush (commented):
adding some comments on the populators which I missed in the first review

ui-admin/src/study/participants/KitRequests.tsx
breilly2: I swear I saw somewhere that column definitions should be memoized, but maybe I was looking at the old v7 docs instead of the v8 docs. 🤦‍♂️

core/src/main/java/bio/terra/pearl/core/config/DataStudyManagerConfig.java
breilly2: It's in config because originally I was trying to use `@ConfigurationProperties`, but that doesn't work when it's coming from `core`. I'll move it into `DataStudyManagerClient`.

core/src/main/java/bio/terra/pearl/core/service/kit/DataStudyManagerClient.java
breilly2: Fair enough. I wasn't sure how much to lean into naming things with "Pepper", and was trying to avoid the uppercase-acronym-inside-camelCase problem.

core/src/main/java/bio/terra/pearl/core/service/kit/DataStudyManagerKitRequest.java
breilly2: I'm trying to serialize two different things: the sent-to address (saved in Juniper) and the web request to DSM. The sent-to address is a subset of the fields in the overall request and is not a separate structure, thus the copying. Alternatives:

* save the entire DSM request in Juniper instead of just the address
* use a bean copying utility library to copy the properties whose names match

ui-admin/src/study/participants/KitRequests.tsx
devonbush: it's correct that you have to memoize them *if* they are within the component function.  But these don't need to be declared within the component function, they can be an external const.

populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/enrollees/jsalk.json
breilly2: Yes, that is _much_ better. Great suggestion!

ui-admin/src/api/api.tsx
```diff
@@ -435,8 +452,31 @@ export default {
...
-    const url = `${baseStudyEnvUrl(portalShortcode, studyShortcode, envName)
-    }/enrollees/${enrolleeShortcode}/changeRecords`
+    const url =
+      `${baseStudyEnvUrl(portalShortcode, studyShortcode, envName)}/enrollees/${enrolleeShortcode}/changeRecords`
+    const response = await fetch(url, this.getGetInit())
+    return await this.processJsonResponse(response)
+  },
+
+  async createKitRequest(portalShortcode: string, studyShortcode: string, envName: string,
+                         enrolleeShortcode: string, kitType: string): Promise<string> {
```
breilly2: I want to point out that I don't agree that the suggested indentation is better. I had this:
```
  async createKitRequest(portalShortcode: string, studyShortcode: string, envName: string,
                         enrolleeShortcode: string, kitType: string): Promise<string> {
    const params = new URLSearchParams({ kitType })
    const url =
    ...
```
which makes it clear that the parameter list is split across two lines.

The suggested formatting is:
```
  async createKitRequest(portalShortcode: string, studyShortcode: string, envName: string,
    enrolleeShortcode: string, kitType: string): Promise<string> {
    const params = new URLSearchParams({ kitType })
    const url =
    ...
```
which makes the second line of parameters look like the first line of the body.

An alternative would be one parameter per line:
```
  async createKitRequest(
    portalShortcode: string,
    studyShortcode: string,
    envName: string,
    enrolleeShortcode: string,
    kitType: string
  ): Promise<string> {
    const params = new URLSearchParams({ kitType })
    const url =
    ...
```
but it seems excessive to me to take 7 lines to describe a 5 line function (though I'll probably end up going with this style because I think it's the least bad that the linter doesn't complain about).

populate/src/main/java/bio/terra/pearl/populate/service/KitTypePopulator.java
breilly2: I ended up not needing a separate `KitRequestPopulator` because it's rolled into `EnrolleePopulator` to avoid duplicating profile/address information (although it is technically possible for a kit request to have a different address than the current profile address). Although now that I think about it, the current enrollee could be passed to `KitRequesntPopulator` through a custom `FilePopulateContext`... but I'm inclined to say that this works well enough for now.

core/src/main/java/bio/terra/pearl/core/service/kit/DataStudyManagerKitRequest.java
breilly2: Devon, do you have any preference for how to resolve this?

devonbush (approved):
Tests look good, thanks!   One test to keep in mind for later iterations of this kit functionality is to add some assertions to the PopulateOurHealthTest that the kit request got successfully attached to the enrollee.

core/src/main/java/bio/terra/pearl/core/service/kit/DataStudyManagerKitRequest.java
devonbush: Now that I understand what you're doing, this seems fine the way you have it.

api-admin/src/main/resources/application.yml
```diff
@@ -20,6 +20,9 @@ env:
     storageContainerName: ${STORAGE_CONTAINER_NAME:}
     billingProfileId: ${BILLING_PROFILE_ID:550212cf-45dc-478f-9fe6-9b3290e2a1fe}
     deploymentZone: ${DEPLOYMENT_ZONE:}
+  dsm:
+    basePath: ${DSM_ADDRESS:https://dsm-dev.datadonationplatform.org/ddp}
```
MatthewBemis, line 24: I've had to think about this for TDR as well: how do we ensure that our prod Juniper instance doesn't send data to dev DSM? It looks like this would silently send data to dev if we forget to explicitly set the dsm address for prod (a potentially easy thing to do)

MatthewBemis (approved):
Overall looks good, I just have one question

api-admin/src/main/resources/application.yml
breilly2: That's a great point. I think it's possible to use `application-development.yml` as an overlay when running locally, but won't be bundled in any artifact that is run in any other environment. I ran a quick experiment and it didn't work the way I wanted it to. Could research this some more, but I'm not going to spend that time right now.

breilly2: That's a great point. I think it's possible to use `application-development.yml` as an overlay when running locally, but won't be bundled in any artifact that is run in any other environment. I ran a quick experiment and it didn't work the way I wanted it to. Could research this some more, but I'm not going to spend that time right now.

PR #403, opened 2023-05-31
[JN-403] Rename some TDR settings, add global on/off switch for TDR export

MatthewBemis:
Companion PR: https://github.com/broadinstitute/terra-helmfile/pull/4227

Two objectives in this pair of PRs:

1) Get things properly set up in dev so AppSec can start evaluating the TDR integration
2) Add a global toggle for TDR export. Even though this won't work in prod because some credentials haven't been provisioned yet, I don't like the implicit nature of the toggle. At some point we should expose whether or not TDR integration is enabled and use that to disable/hide parts of the admin UI.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/ScheduledDataRepoExportService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/ScheduledDataRepoExportService.java
@@ -61,7 +61,7 @@ public class ScheduledDataRepoExportService {
             "storageAccountKey",
             "storageContainerName");
 
-    //Kill-switch for TDR export
+    // Kill-switch for TDR export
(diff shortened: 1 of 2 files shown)


nawatts (approved):

api-admin/src/main/java/bio/terra/pearl/api/admin/service/ScheduledDataRepoExportService.java
```diff
@@ -61,8 +61,13 @@ public boolean isTdrConfigured() {
             "storageAccountKey",
             "storageContainerName");
 
-    return REQUIRED_TDR_ENV_VARS.stream()
-        .allMatch(
-            envVar -> StringUtils.isNotBlank(env.getProperty(String.format("env.tdr.%s", envVar))));
+    // Kill-switch for TDR export
```
devonbush: "global on/off switch" might be a better name

devonbush (approved):

PR #406, opened 2023-05-31
[JN-400] Add createdBy field to TDR dataset

MatthewBemis:
This adds a new createdBy field to the dataset so we know who created it.

### DIFF
--- a/ui-admin/src/study/participants/datarepo/DatasetDashboard.tsx
+++ b/ui-admin/src/study/participants/datarepo/DatasetDashboard.tsx
@@ -117,6 +117,8 @@ const DatasetDashboard = ({ studyEnvContext }: {studyEnvContext: StudyEnvContext
                   <label>Description:</label> { datasetDetails?.description ?
                     datasetDetails?.description : <span className="fst-italic">N/A</span> }
                   <br/>
+                  <label>Created By:</label> { datasetDetails?.createdBy }
+                  <br/>
                   <label>Date Created:</label> { instantToDefaultString(datasetDetails?.createdAt) }
                 </div>
                 <br/>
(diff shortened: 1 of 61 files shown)


ui-admin/src/study/participants/datarepo/DatasetDashboard.tsx
```diff
@@ -117,6 +117,8 @@ const DatasetDashboard = ({ studyEnvContext }: {studyEnvContext: StudyEnvContext
                   <label>Description:</label> { datasetDetails?.description ?
                     datasetDetails?.description : <span className="fst-italic">N/A</span> }
                   <br/>
+                  <label>Created By:</label> { datasetDetails?.createdBy }
```
nawatts: `label` isn't really appropriate here since there's no form element associated with it. I think `dl` would be the better option for this list.

nawatts (approved):

nawatts:
Re: timezones, we can do what terra-ui does at set `TZ=UTC` in the test script.
https://github.com/DataBiosphere/terra-ui/blob/e2c010741f4a13686dcb73f160ba74839488e961/package.json#L74

devonbush (changes requested):
I'm glad we're adding audit history to these.  For more robustness, let's use AdminUser uuids in the database instead of usernames.  This will make the dashboard a bit less friendly, but I'm fine for now just displaying the uuid.  (we could make a DTO returned by the server for each dataset that includes the username, but I suspect that's more trouble than this is worth for now)

MatthewBemis:
@devonbush feedback resolved in 15cf3591562e4dcb52b5d779ea01bf0e9330fd9a

devonbush (approved):

PR #408, opened 2023-05-31
[JN-405] Fail PR checks on ESLint warnings

nawatts:
When building one of the UIs, ESLint is run via a webpack plugin. In CI environments (when the CI environment variable is set), create-react-app treats webpack warnings as errors. Thus, an ESLint warning will cause the publish workflow to fail.

Since #395, ESLint is run differently in PR checks. There, we run ESLint directly (in order to cover ui-core) and disable the webpack plugin. Thus, PR checks can pass on code with an ESLint warning.

This changes the `lint` script to fail on warnings so that PR checks are consistent with the publish workflow.


ui-admin/src/study/participants/datarepo/DatasetDashboard.test.tsx
nawatts, line 1: In PR checks, ESLint is also run on test files, so had to resolve a warning here.

MatthewBemis (approved):

devonbush (approved):

PR #409, opened 2023-06-01
[JN-383] Delete Datasets

MatthewBemis:
**Question about auditing:** when we delete a dataset, should we preserve the `dataset` and `data_repo_job` records in the database (i.e. soft delete?), and maybe add the ID of who initiated it? We'd still delete the dataset in TDR, but I can imagine we might want to maintain some history of the dataset w/o having to parse the logs?

______

This adds the ability to delete a dataset. Currently only usable by super users (since they're the only ones who can create datasets right now). The deletion modal follows the pattern established by the Withdraw Participant functionality.

Screenshot 2023-06-02 at 9 43 29 AM

Heads up if testing locally:

Make sure that TDR export is configured as described here (note that some things have changed recently):
https://broadworkbench.atlassian.net/wiki/spaces/PEARL/pages/2767650818/Setting+up+TDR+locally


breilly2:
Does TDR do a hard or soft delete? Thinking about it from user expectations... If I'm someone who really understands what TDR is, who uses it, for what, etc., I might be surprised if Juniper and TDR had different forms of delete. On the other hand, if I'm someone who doesn't really understand TDR, (then I probably shouldn't have permission to delete datasets, but...) then I might not understand the implications of deleting from Juniper; I might think that I'm just doing some housecleaning in Juniper and not realize how that impacts TDR and access to the dataset.

But also, in terms of auditing, it's probably good to be able to see what has happened in the past to help us get to the bottom of inevitable "missing" datasets.

devonbush (approved):
Great question about hard/soft deletes. My slight inclination would be that it should be a soft delete (essentially, adding a new dataset status "DELETED").  But I don't think we need that yet, we can wait until we get some usage on this to figure out how best to handle it

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
```diff
@@ -128,6 +130,34 @@ public void createDataset(StudyEnvironment studyEnv, String datasetName, String
...
+        Dataset dataset = datasetDao.findByDatasetName(datasetName).orElseThrow(() -> new DatasetNotFoundException("Dataset not found."));
+
+        List<DatasetStatus> deletableStatuses = List.of(DatasetStatus.CREATED, DatasetStatus.FAILED);
+        if(!deletableStatuses.contains(dataset.getStatus())) {
+            throw new DatasetDeletionException(String.format("Could not delete dataset %s. Dataset was in %s state, but must be in a terminal state.", datasetName, dataset.getStatus()));
+        }
+
+        JobModel response;
+        try {
+            response = dataRepoClient.deleteDataset(dataset.getTdrDatasetId());
```
breilly2, line 147: Was this an intentional decision to favor honoring the request to delete the dataset from TDR over recording that the gesture was made? I could imagine arranging the database transactions such that the gesture is recorded first. Not suggesting a change, just asking the question. It's really a judgement call on what should the state be if there's a failure mid-stream and what's the process for ultimately resolving it.

But also this should probably be `@Transactional`, both for good practice and because there are multiple DAO operations. (This is something that I'm realizing I might have missed in my last set of changes.)

core/src/main/java/bio/terra/pearl/core/service/datarepo/DatasetService.java
```diff
@@ -17,14 +17,20 @@
 @Service
 public class DatasetService extends CrudService<Dataset, DatasetDao> {
 
-    private DataRepoClient dataRepoClient;
+    DataRepoClient dataRepoClient;
```
breilly2: Accidental change? I don't see any usage of this that requires a change in visibility.

ui-admin/src/study/participants/datarepo/DatasetList.tsx
```diff
@@ -23,8 +23,12 @@ const datasetColumns = (currentEnvPath: string): ColumnDef<DatasetDetails>[] =>
   id: 'datasetName',
   header: 'Dataset Name',
   accessorKey: 'datasetName',
-  cell: info => <Link to={getDatasetDashboardPath(info.getValue() as unknown as string, currentEnvPath)}
-    className="mx-2">{info.getValue() as unknown as string}</Link>
+  cell: info => {
+    return info.row.original.status !== 'DELETING' ?
+      <Link to={getDatasetDashboardPath(info.getValue() as unknown as string, currentEnvPath)} className="mx-2">
+        {info.getValue() as unknown as string}
+      </Link> : <span className="mx-2">{info.getValue() as unknown as string}</span>
```
breilly2: Oof... `as unknown as string` is unsightly. I haven't tried it yet, but I recently noticed docs about "column helpers" (https://tanstack.com/table/v8/docs/guide/column-defs#column-helpers) that supposedly make types work better. I wonder if that would help in this case.

breilly2 (approved):
Mainly a couple of question comments, plus one suggestion about `@Transactional`, but overall looks good.

core/src/main/java/bio/terra/pearl/core/service/datarepo/DatasetService.java
MatthewBemis: Sorry, not sure why that changed. Fixed in 32e90a08

ui-admin/src/study/participants/datarepo/DatasetList.tsx
MatthewBemis: I didn't spend too long on it, but I wasn't able to get the column helpers to work. But I did make this a bit less ugly by accessing the typed fields through the underlying `DatasetDetails` object instead: 8b2a7ebf

core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
MatthewBemis: I don't think I see the value in recording the gesture first in this case, unless it would be needed for auditing purposes? I am also less concerned about being in a broken state where the TDR dataset is deleted but we never recorded the action in our DB. It will be recorded in other places (app logs, TDR logs), and I think the most important thing is prioritizing the fact that a user wanted to destroy the data in TDR, which will be honored.

I will put the DAO operations inside of a txn.

PR #410, opened 2023-06-02
[JN-411] Move shared survey code to ui-core

nawatts:
Currently, code for converting Juniper Surveys into SurveyJS models is duplicated in ui-admin and ui-participant. This moves that code into ui-core. It also moves the custom survey components that are currently part of ui-participant. This is necessary to have a complete survey preview in the admin UI.

In some cases, the admin and participant implementations of functions differed. In those cases, the participant UI version was used.


breilly2:
Is it expected that the top-level `package-lock.json` changed when there were no changes to the top-level `package.json`? Is it because of the change in `ui-core/package.json` in the context of npm workspaces?

nawatts:
Yes, package-lock.json is for the entire workspace (all 3 ui-* packages and the root package.json). It looks like some dependencies got moved up to the top-level node_modules folder now that they're used by ui-core.

devonbush (approved):

breilly2 (approved):

PR #411, opened 2023-06-02
[JN-378] Cleanup SurveyEditorView

nawatts:
Groundwork for survey editor.

This removes navigation and permission concerns from SurveyEditorView, moving them up to its parent views. This will make SurveyEditorView easier to test.

It also removes the broken version selection from SurveyEditorView. Currently, the version selector attempts to load versions from an endpoint that doesn't exist. It looks like the commented out code automatically "publishes" the old version when its selected. I'm thinking we may want to allow loading an old version in the editor, but only publishing it when a user clicks "Save".


devonbush (approved):
since this is the beginning of a long refactor/upgrade, I'll leave it up to you when/where are the best times/places to add tests

breilly2 (approved):

PR #412, opened 2023-06-02
[JN-409] Make completed tasks always accessible

nawatts:
Currently, if a task is completed and there are other tasks that block the hub, the completed task shows as "Locked". This makes completed tasks accessible so that the show as "Complete" and answers can be reviewed.

### DIFF
--- a/ui-core/src/types/task.ts
+++ b/ui-core/src/types/task.ts
@@ -1,14 +1,23 @@
 export type ParticipantTask = {
   id: string
+  blocksHub: boolean
   completedAt?: number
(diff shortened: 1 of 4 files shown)


devonbush (changes requested):
please add a test, thanks!

nawatts:
Added a test. Also updated Enrollee and ParticipantTask types to reflect data returned from API.

devonbush (approved):
nice, thanks for cleaning up the types too

MatthewBemis (approved):

PR #413, opened 2023-06-02
[DDO-2886] Improve SQL Connect Script Cleanup Handling

mikeflinn:
Previously if the cleanup step of a previous invocation of this script failed and the k8s namespace that had contained the jump pod used to connect to the DB still existed, then subsequent invocations would fail with `namespace already exists errors`. This change makes it so that the script is idempotent with respect to k8s namespaces, it will only create one if it doesn't exist already. But it will still attempt cleanup upon completion.

Additionally this change adds support for adding a `username` field onto the end of k8s resources this script deploys. The benefit of this twofold
1. It provides better auditing around who connected to the db and when.
2. Provides isolation of the resources deployed by this script so that two developers could connect to the db simultaneously from their local machines without stepping on each other.

I also updated the require ticket github action to accept tickets from DevOps' DDO jira project too

### DIFF
--- a/scripts/sql_connect
+++ b/scripts/sql_connect
@@ -12,7 +12,7 @@ set -o pipefail
 # DEBUG = 2
 declare -i desired_log_level=2
 declare -a valid_environment_targets=( dev prod )
-/
+
 function usage() {
     local -r usage="
     $0: connect to an Azure managed database instance with private networking through AKS cluster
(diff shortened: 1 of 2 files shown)


scripts/sql_connect
```diff
@@ -12,7 +12,7 @@ set -o pipefail
 # DEBUG = 2
 declare -i desired_log_level=2
 declare -a valid_environment_targets=( dev prod )
-/
+
```
jyang-broad, line 15: not sure what the slash does/did in an otherwise empty line, is this an intentional change?

scripts/sql_connect
```diff
@@ -300,10 +319,10 @@ function init() {
 
 function main() {
     init "$@"
+    read_user
     setup_k8s_enviornment || exit 1
```
jyang-broad: not your typo, but can you fix `enviorment`?

scripts/sql_connect
```diff
@@ -300,10 +319,10 @@ function init() {
...
     # ( If any thing from this point forward fails script must cleanup )
     trap cleanup_k8s_enviornment SIGHUP SIGINT SIGQUIT
-    read_user
     exec_into_pod
     cleanup_k8s_enviornment
```
jyang-broad: here too

jack-r-warren (approved):

jyang-broad (approved):
looks good!
moves `read_user` in main before `setup_k8s_environment` -> `create_namespace` so we can check for existing `k8s_namespace`, which uses the `username` now

scripts/sql_connect
mikeflinn: yeah it was intentional. I suspect it was a typo from a previous pr. I'm just removing it. It does nothing

devonbush (approved):

PR #414, opened 2023-06-02
JN-412 refreshing seed populate

devonbush:
This changes our approach to BaseSeedPopulator.  Previously, it was only run on fresh DBs.  Now, we will run BaseSeedPopulator on every server startup.  BaseSeedPopulator takes the role of ensuring that any essential data for server operation is in place.  This will solve a lot of configuration errors, at the cost of us having to be more careful about what we include in the base seed, and remembering that the seeded entities will always be set to their source-controlled values on server startup. 

For example, if I go rogue, and someone needs to revoke my superuser permissions in production, then setting superuser=false in the DB for my account will not be enough -- that change would get reverted next time a pod starts.  You'd need to do a release removing my superuser permission from the seed files -- those seed files are now the 'source of truth'.   

I think this cost is worth it in the short term rather than manually shepherding migration scripts, but I could be convinced otherwise.  If so, then Brian will just need to write a liquibase migration script adding the kit request types he needs.

TO TEST:
1. restart ApiAdminApp
2. confirm base seed populator runs, despite  an existing database

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/config/InitialPopulate.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/config/InitialPopulate.java
@@ -17,12 +17,8 @@ public class InitialPopulate {
   @Autowired private BaseSeedPopulator baseSeedPopulator;
 
   @EventListener(ApplicationReadyEvent.class)
-  public void populateSeedIfNeeded() throws IOException {
-    if (adminUserService.count() == 0) {
-      log.info("No admin users found, populating base");
-      baseSeedPopulator.populate("seed");
-    } else {
-      log.info("Existing admin users found, skipping seed populate");
-    }
+  public void populateSeed() throws IOException {
+    log.info("Refreshing base seed populate data");
+    baseSeedPopulator.populate();
   }
 }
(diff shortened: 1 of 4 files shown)


breilly2:
I don't think I like the approach because of the specific scenario you mentioned. The solution I had in mind was to made base seed population add-only, so it would make sure that everything exists, but won't update entities that do exist. That means that during development, you might have to wipe your database more often when iterating on new base-seed-populated files, but I think that's fine and is actually closer to what we most often want for live environments, which is the problem we're trying to solve.

At a bare minimum, I think the scenario of superuser privilege is important enough to deserve a special case in the population code, in that re-populating an _existing_ user should probably _never_ elevate privileges. That's easy enough to manage manually that I don't think we're really losing out on anything by expressly forbidding populate to do it.

nawatts:
Agreed, I'd rather have the base seed populator initialize records in the database, but then allow them to be edited as normal. Otherwise, how can a user tell if the record they are editing is part of the base seed and thus any changes they make will be undone at a seemingly arbitrary time.

Though I guess either way, records that are part of the base seed would be "special" since they can't be permanently deleted and that quality isn't visible to users.

I'm starting to lean towards using migrations for adding seed data. Then the initialization would be one time only.

devonbush:
closing in view of Brian's migration PR

PR #415, opened 2023-06-05
[JN-362] Move creation of saliva kit type from base seed populate to liquibase

breilly2:
Preparing this in case it's ultimately what we decide to do based on conversations in #414.


nawatts (approved):

devonbush (approved):
Looks good.  I'm sad that we won't be able to have a single file to look at to see, e.g. "what are all the currently supported kit types?"  But from a data management perspective agree this is probably more robust.

PR #417, opened 2023-06-05
JN-364 adding demo shell portal

devonbush:
This adds a new portal to our seed files -- "demo".  It's currently a clone of ourhealth with stableIds changed, and some theme changes so that it's easily visually distinguishable.  The short term need for this is a portal that has multiple versions of the same survey so I can test some workflow changes (will add those in a later PR along with the admin fix).  Long term, this might evolve into a more sophisticated test harness, as well as a template for somewhat-neutral sandbox portal we could give access to potential customers to play around.  (that's why it's called 'demo' rather than test)

This does not yet include support for b2c login

TO TEST:
1. run `./scripts/populate_portal demo` 
2. go to `https://sandbox.demo.localhost:3001/`. 
3. confirm you see an ourhealth-like study but with a different color scheme
image


breilly2:
@devonbush Populating demo against a blank database didn't work on my first run. It looked like some record was missing that it expected to already be there. I populated ourhealth and then populating demo worked.

devonbush:
Good catch, the creating user of the kit request was set to an ourhealth staff member.  now fixed

breilly2 (approved):

MatthewBemis (approved):

PR #419, opened 2023-06-06
[JN-378] Add basic survey editor

nawatts:
This replaces the SurveyJS editor in the admin UI with our own basic editor. Currently, it allows editing the survey as JSON and viewing a preview of the survey rendered with SurveyJS.


ui-admin/src/App.css
```diff
@@ -86,4 +85,10 @@ a {
   max-height: 90vh;
 }
 
-
+/* Have survey editor tabs fill page height */
```
devonbush, line 88: I'm surprised the bootstrap library doesn't have a way to specify these as inline styles

ui-admin/src/study/surveys/editor/SurveyPreview.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const SurveyPreview = (props: SurveyPreviewProps) => {
+  const { survey } = props
+
+  const [surveyModel] = useState(() => {
+    const fakeForm = { content: JSON.stringify(survey) } as Survey
+
+    const model = new SurveyModel(extractSurveyContent(fakeForm))
+    model.focusFirstQuestionAutomatic = false
+    model.showTitle = false
```
devonbush, line 19: showTitle = false and widthMode static are also set on the participant side -- do we ever not want these?  And if we don't, should we pull these into a "configureDefaultSurveyDisplay" utility method somewhere?

devonbush (approved):
Like the approach and structure of getting just the basic json editor in first.  Couple of questions, nothing blocking or major

ui-admin/src/App.css
nawatts: It does, but react-bootstrap doesn't expose a way to add classes to the `.tab-content` element.

ui-admin/src/study/surveys/editor/SurveyPreview.tsx
nawatts: I'm thinking `extractSurveyContent` may become a FormContent => SurveyModel function that applies this sort of shared configuration to the SurveyModel.

ui-admin/src/study/surveys/SurveyEditorView.tsx
```diff
@@ -1,63 +1,77 @@
...
+              'btn btn-primary me-md-2',
+              { disabled: !isSaveEnabled }
+            )}
+            type="button"
+            onClick={onClickSave}
+          >
             Save
           </button>
         )}
         <button className="btn btn-secondary" type="button" onClick={onCancel}>Cancel</button>
```
breilly2, line 63: I noticed that cancel is navigating back one level further than I think it should. Looks like the `onCancel` function being passed in from `SurveyView` is responsible for this. But it looks like this isn't new with this PR.

breilly2 (approved):

PR #420, opened 2023-06-06
[JN-378] Update types for survey configuration

nawatts:
Stacked on #419.

Currently, we use some types from SurveyJS for survey configuration (Question, etc). However, these are types for model classes and we're using them for the config object passed to SurveyModel's constructor.

Thus, attempting to do something like this in a test does not work because TS expects objects in the pages array to have a bunch of methods from SurveyJS' model object for pages.
```ts
const survey: JuniperSurvey = {
  pages: [ ... ]
}
```

SurveyJS does not provide types for the configuration object. The type of SurveyModel's constructor's parameter is `any`.
https://github.com/surveyjs/survey-library/blob/e5c5bf69001af854430d224d1362a96b84134884/src/survey.ts#L784

So this adds our own types for survey configuration. These are incomplete and certainly don't cover everything available from SurveyJS, but should cover a decent percentage of our current surveys. Also note that currently nothing validates that configured content is actually valid.

### DIFF
--- a/ui-admin/src/study/surveys/editor/SurveyEditor.tsx
+++ b/ui-admin/src/study/surveys/editor/SurveyEditor.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react'
 import { Tab, Tabs } from 'react-bootstrap'
 
-import { JuniperSurvey } from '@juniper/ui-core'
+import { FormContent } from '@juniper/ui-core'
 
 import { OnChangeSurvey } from './surveyEditorTypes'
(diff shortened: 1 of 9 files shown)


devonbush (approved):
Nice, yeah, it's disappointing the lack of types from SurveyJs -- it shows that the project was created without typescript first, and then they had to back in a lot of type functionality

breilly2 (approved):

nawatts:
Rebased after merging #419.

PR #421, opened 2023-06-07
[JN-378] Move SurveyModel creation into shared code

nawatts:
Addresses https://github.com/broadinstitute/juniper/pull/419#discussion_r1220332120.

Currently, ui-core contains a function `extractSurveyContent` that takes a Juniper VersionedForm and returns a configuration object to be passed to the SurveyJS SurveyModel constructor.

However, there are additional settings that we always apply to the SurveyModel after it's created (`focusFirstQuestionAutomatic = false`, etc). This moves the SurveyModel creation into ui-core so that we can avoid duplicating that configuration every time we create a SurveyModel.

### DIFF
--- a/ui-admin/src/study/participants/survey/SurveyFullDataView.tsx
+++ b/ui-admin/src/study/participants/survey/SurveyFullDataView.tsx
@@ -1,13 +1,13 @@
 import React from 'react'
-import { Model, Question, SurveyModel } from 'survey-core'
+import { Question, SurveyModel } from 'survey-core'
 
-import { extractSurveyContent } from '@juniper/ui-core'
+import { surveyJSModelFromForm } from '@juniper/ui-core'
 
 import { Answer, ConsentForm, Survey } from 'api/api'
 
 /** renders every item in a survey response */
 export default function SurveyFullDataView({ answers, survey }: {answers: Answer[], survey: Survey | ConsentForm}) {
-  const surveyJsModel = new Model(extractSurveyContent(survey))
+  const surveyJsModel = surveyJSModelFromForm(survey)
   console.log(`rendering data for survey ${survey.stableId} -- question text not yet implemented`)
   return <dl>
     {answers.map((answer, index) => <ItemDisplay key={index}
--- a/ui-core/src/surveyUtils.ts
+++ b/ui-core/src/surveyUtils.ts
@@ -1,5 +1,8 @@
 import './surveyjs'
 
+import { cloneDeep } from 'lodash'
+import { SurveyModel } from 'survey-core'
+
 import { FormContent, FormElement, VersionedForm } from './types/forms'
 
 /** Gets a flattened list of the survey elements */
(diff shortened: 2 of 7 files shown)


ui-admin/src/study/participants/survey/SurveyFullDataView.tsx
```diff
@@ -1,13 +1,13 @@
...
 
-import { extractSurveyContent } from '@juniper/ui-core'
+import { surveyJSModelFromForm } from '@juniper/ui-core'
 
 import { Answer, ConsentForm, Survey } from 'api/api'
 
 /** renders every item in a survey response */
 export default function SurveyFullDataView({ answers, survey }: {answers: Answer[], survey: Survey | ConsentForm}) {
-  const surveyJsModel = new Model(extractSurveyContent(survey))
+  const surveyJsModel = surveyJSModelFromForm(survey)
```
devonbush, line 10: nice, I like how you pulled the `new Model` into the function itself

ui-core/src/surveyUtils.ts
```diff
@@ -38,5 +37,22 @@ export function extractSurveyContent(form: VersionedForm) {
       }
     })
   }
-  return parsedSurvey
+
+  const model = new SurveyModel(formContentClone)
+
+  model.focusFirstQuestionAutomatic = false
+  model.showTitle = false
+  model.widthMode = 'static'
```
devonbush: I might pull the three lines above into a helper function "applyDefaultModelSettings()" for clarity.

devonbush (approved):
nice, thanks for pulling that out!

ui-core/src/surveyUtils.ts
nawatts: Done in https://github.com/broadinstitute/juniper/pull/421/commits/01e102b603828614101d1e32bbb2a65364633de8.

ui-core/src/surveyUtils.ts
```diff
@@ -14,22 +17,24 @@ function getFormQuestionsHelper(element: FormElement): FormElement[] {
...
-  const questionTemplates = parsedSurvey.questionTemplates
+const applyDefaultSurveyConfig = (surveyModel: SurveyModel): void => {
+  surveyModel.focusFirstQuestionAutomatic = false
+  surveyModel.showTitle = false
+  surveyModel.widthMode = 'static'
+}
+
+/** Create a SurveyJS SurveyModel from a Juniper FormContent object. */
+export const surveyJSModelFromFormContent = (formContent: FormContent): SurveyModel => {
+  const formContentClone = cloneDeep(formContent)
```
breilly2, line 28: Why does this need to be cloned?

breilly2 (approved):

ui-core/src/surveyUtils.ts
nawatts: To avoid modifying the `formContent` object passed in. Line 42 copies properties from the question template to the question.

PR #424, opened 2023-06-08
[JN-420] Fix data length mismatch when uploading blob to Azure storage container

MatthewBemis:
I just noticed this bug when trying to export data to TDR from the OurHealth sandbox environment on dev. Data ingest was failing because we couldn't write the TSV to the Azure storage container prior to ingest. It worked fine in other study environments so that pointed to it probably being a bug related to the data. Sure enough, the logs had this error: `Request body emitted 34783 bytes, more than the expected 34781 bytes` coming from `AzureBlobStorageClient.uploadBlob`.

I was able to reproduce this locally by downloading the TSV from the sandbox study environment and reading that file in as a string to pass into `AzureBlobStorageClient.uploadBlob`.

My theory is that when we convert the TSV string data into a byte array with `data.getBytes(StandardCharsets.UTF_8)`, some conversion between charsets is resulting in there being a slightly different data length than we started with. This is a problem because we are calculating the expected blob length from `data`, but then we are passing the `dataStream` in as the actual blob upload content. I did it this way because that's how the Azure Blob Client sample repo suggested doing it, but I don't see the point in handling this stuff on our own. I think the better thing to do is pass the data off as `BinaryData`, which Microsoft kindly built to handle all of this stuff automatically.

I was able to verify that this does fix the bug.


core/src/main/java/bio/terra/pearl/core/service/datarepo/DataRepoExportService.java
```diff
@@ -3,7 +3,6 @@
 import bio.terra.datarepo.client.ApiException;
 import bio.terra.datarepo.model.JobModel;
 import bio.terra.datarepo.model.JobModel.JobStatusEnum;
-import bio.terra.datarepo.model.TableDataType;
```
MatthewBemis, line 6: unrelated cleanup, just noticed it while I was working in this file to reproduce the bug

nawatts (approved):

devonbush (approved):

PR #426, opened 2023-06-09
[JN-378] Add option to ignore validation in form preview

nawatts:
This adds an options panel to the form editor preview tab. Currently, the only option is to ignore validation, which is useful for quickly paging through a survey without having to enter required fields.

### DIFF
--- a/ui-admin/src/forms/FormPreviewOptions.tsx
+++ b/ui-admin/src/forms/FormPreviewOptions.tsx
@@ -1,0 +1,36 @@
+import React from 'react'
+
+type FormPreviewOptions = {
+  ignoreValidation: boolean
+}
+
+type FormPreviewOptionsProps = {
+  value: FormPreviewOptions
+  onChange: (newValue: FormPreviewOptions) => void
+}
+
+export const FormPreviewOptions = (props: FormPreviewOptionsProps) => {
+  const { value, onChange } = props
+  return (
+    <div>
+      <div className="form-check">
+        <label className="form-check-label" htmlFor="form-preview-ignore-validation">
+          <input
+            checked={value.ignoreValidation}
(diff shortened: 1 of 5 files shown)


MatthewBemis (approved):

ui-core/src/reactUtils.ts
```diff
@@ -0,0 +1,6 @@
+import { useCallback, useState } from 'react'
+
```
devonbush, line 2: please add comment

ui-admin/src/forms/FormPreviewOptions.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+type FormPreviewOptions = {
+  ignoreValidation: boolean
+}
+
+type FormPreviewOptionsProps = {
+  value: FormPreviewOptions
+  onChange: (newValue: FormPreviewOptions) => void
+}
+
```
devonbush, line 11: please add comment

devonbush (approved):
looks good.  could use some comments.  Do you know if there's an easy way to have an eslint rule so that `const foo = <function>` statements would require a comment?  (in the same way that functions declared with `function` require a comment)?  Quick Googling didn't find any rules like that

nawatts:
Added comments in https://github.com/broadinstitute/juniper/pull/426/commits/5f6b51f109cbb6aac10eaab7d505c47b74fa766d. Off the top of my head, I'm not aware of any such ESLint rule.

nawatts:
Rebased to resolve merge conflict.

nawatts:
Oh wait, the [require-jsdoc](https://eslint.org/docs/latest/rules/require-jsdoc) rule we already use supports that by setting the `ArrowFunctionExpression` option to true.

PR #427, opened 2023-06-09
[JN-378] Add site images to admin API

nawatts:
Currently, images are broken in the form editor preview. This is because the images are requested from a site images API which does not exist in the admin API.

This copies the site images controller from the participant API to the admin API. Note that the site images API is not authenticated/authorized. This is acceptable because any uploaded images are immediately available through the public participant API anyway. It would also be a challenge to add authentication to the images API since the images are loaded by the browser, which does not send a token header with those requests.

### DIFF
--- a/ui-admin/src/forms/FormPreview.tsx
+++ b/ui-admin/src/forms/FormPreview.tsx
@@ -10,7 +10,11 @@ type FormPreviewProps = {
 export const FormPreview = (props: FormPreviewProps) => {
   const { formContent } = props
 
-  const [surveyModel] = useState(() => surveyJSModelFromFormContent(formContent))
+  const [surveyModel] = useState(() => {
+    const model = surveyJSModelFromFormContent(formContent)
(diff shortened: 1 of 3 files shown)


ui-admin/src/forms/FormPreview.tsx
```diff
@@ -10,7 +10,11 @@ type FormPreviewProps = {
 export const FormPreview = (props: FormPreviewProps) => {
   const { formContent } = props
 
-  const [surveyModel] = useState(() => surveyJSModelFromFormContent(formContent))
+  const [surveyModel] = useState(() => {
+    const model = surveyJSModelFromFormContent(formContent)
+    model.setVariable('portalEnvironmentName', 'sandbox')
```
nawatts, line 15: This variable is used in image URLs. (Though it's ignored by the API: [JN-421](https://broadworkbench.atlassian.net/browse/JN-421)).

MatthewBemis (approved):
👍

devonbush (approved):

PR #428, opened 2023-06-09
[JN-424] Require JSDoc for function expressions

nawatts:
Currently, JSDoc is only required for function declarations (`function foo() { ... }`). This also requires it for function expressions (`const foo = function() { ... }` and arrow function expressions (`const foo = () => { ... }`). It only requires JSDoc for functions exported from the module.

Since [ESLint's built in require-jsdoc rule is deprecated](https://eslint.org/blog/2018/11/jsdoc-end-of-life/), this also switches to [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,7 +25,7 @@ module.exports = {
     ecmaVersion: 2018,
     sourceType: 'module'
   },
-  plugins: ['react', 'jsx-a11y', 'import', 'jest'],
+  plugins: ['react', 'jsx-a11y', 'import', 'jest', 'jsdoc'],
   rules: {
     'array-bracket-newline': ['warn', 'consistent'],
     'array-bracket-spacing': 'warn',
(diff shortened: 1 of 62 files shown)


nawatts:
According to ESLint, there are 87 missing JSDoc comments. We'll need to add missing comments or add ignore comments/config to get CI to pass on this PR.

.eslintrc.js
```diff
@@ -104,7 +103,17 @@ module.exports = {
     'jest/prefer-to-have-length': 'warn',
 
     // TypeScript
-    '@typescript-eslint/ban-ts-comment': 'off'
+    '@typescript-eslint/ban-ts-comment': 'off',
+
+    // JSDoc
+    'jsdoc/require-jsdoc': ['warn', {
+      publicOnly: true,
```
nawatts, line 110: Only require JSDoc for functions that are exported from the module.

devonbush (approved):

JeffKorte (approved):

PR #431, opened 2023-06-13
[JN-378] Add designer tab and table of contents to form editor

nawatts:
This adds a "Designer" tab to the form editor. For now, this tab contains a table of contents for all elements in the form. The intent is that when a user selects one of the elements in the table of contents, the right panel will show a form for editing that element.

Screenshot 2023-06-13 at 12 21 52 PM

### DIFF
--- a/ui-admin/src/forms/FormTableOfContents.tsx
+++ b/ui-admin/src/forms/FormTableOfContents.tsx
@@ -18,6 +18,7 @@ type TableOfContentsEntry = {
   children: TableOfContentsEntry[]
 }
 
+/** Recursive helper for getTableOfContents. */
 const getTableOfContentsHelper = (formElement: FormElement): TableOfContentsEntry => {
   if ('type' in formElement && formElement.type === 'panel') {
     return {
@@ -34,6 +35,7 @@ const getTableOfContentsHelper = (formElement: FormElement): TableOfContentsEntr
   }
 }
 
+/** Convert a FormContent object into a TableOfContentsEntry. */
 const getTableOfContents = (formContent: FormContent): TableOfContentsEntry => {
   return {
     type: 'form',
@@ -61,10 +63,12 @@ const getTableOfContents = (formContent: FormContent): TableOfContentsEntry => {
   }
 }
 
+/** Should a table of contents entry be expandable? */
 const isEntryTypeExpandable = (entryType: string): boolean => {
   return entryType !== 'questionOrHtml'
 }
 
+/** Should a table of contents entry be selectable? */
 const isEntryTypeSelectable = (entryType: string): boolean => {
   return entryType === 'questionOrHtml'
 }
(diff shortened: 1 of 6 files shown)


ui-admin/src/forms/FormTableOfContents.tsx
```diff
@@ -0,0 +1,346 @@
...
+  | 'page'
+  | 'panel'
+  | 'questionOrHtml'
+
+type TableOfContentsEntry = {
+  name: string
+  type: TableOfContentsEntryType
+  children: TableOfContentsEntry[]
+}
+
```
devonbush, line 20: this could use a comment

ui-admin/src/forms/FormTableOfContents.tsx
```diff
@@ -0,0 +1,346 @@
...
+          }
+        }}
+      >
+        {isExpandable && <FontAwesomeIcon icon={isExpanded ? faChevronDown : faChevronRight} />}
+      </span>
+      <a
+        id={`${id}-link`}
+        role="presentation"
+        tabIndex={-1}
+        style={{
```
devonbush, line 202: I tend to like pulling style objects with more than 3-4 props outside the component into either a const or helper function so that the structure of the surrounding markup is easier to fit on one screen.  But defer to you to on best style practices for that

ui-admin/src/forms/FormTableOfContents.tsx
```diff
@@ -0,0 +1,346 @@
...
+  )
+}
+
+type FormTableOfContentsProps = {
+  formContent: FormContent
+  selectedElementName: string | undefined
+  onSelectElement: (name: string) => void
+}
+
+export const FormTableOfContents = (props: FormTableOfContentsProps) => {
```
devonbush, line 248: I'm impressed that you wrote your own tree browser including keyboard support.  Did you evaluate any open-source treeview components like https://www.npmjs.com/package/react-accessible-treeview or https://github.com/lukasbach/react-complex-tree?

devonbush (approved):
Looks good, a couple style nits.  My main question is whether we should use a 3rd-party tree library (e.g. https://github.com/lukasbach/react-complex-tree), so that down the road it will be easier to add drag-drop reordering, etc...

JeffKorte (approved):

nawatts:
I didn't try any library. I copied and adapted this from Terra's file browser. Now that I think about it, I believe our motivation for writing our own tree view there was because of how data was fetched whenever a tree item was expanded.

Of the two mentioned though, https://github.com/lukasbach/react-complex-tree looks like it doesn't apply the tree-related ARIA roles, so VoiceOver doesn't read anything when navigating around the tree. Drag and drop in the table of contents is appealing, but I think we'd have to add another more easily accessible option for organizing questions anyway (like buttons to move a question up/down), since that library's drag and drop doesn't seem to work on a phone and I'm guessing not for keyboard users either. I'm thinking of having it so that when a page or panel is selected in the table of contents, the right panel shows its list of elements and allows reordering them.

https://www.npmjs.com/package/react-accessible-treeview seems ok with VoiceOver, but since it's already done, I lean towards the greater control we have with our own components vs a library.

JeffKorte (approved):

PR #432, opened 2023-06-14
[JN-378] Revise form editor table of contents

nawatts:
Currently, the form editor stores the selected element by name. This proved to not work very well since we'll want to edit pages and panels which don't have names and updating a question by name required iterating over the form content object to find the question to update.

This changes the editor to store the path to the selected element, which we can use to easily get the selected element (`_.get(formContent, path)`) and update it (`_.set(formContent, path, newQuestion)`). It also extracts the Tree into a generic component.

### DIFF
--- a/ui-admin/src/components/Tree.tsx
+++ b/ui-admin/src/components/Tree.tsx
@@ -4,19 +4,39 @@ import { get } from 'lodash'
 import React, { Dispatch, SetStateAction, useRef, useState } from 'react'
 
 export type TreeItemT<T> = {
+  /** User facing label. */
   label: string
+
+  /** Data associated with this tree item. */
   data: T
+
+  /** Child tree items. */
   children?: TreeItemT<T>[]
 }
 
 type TreeItemProps<T> = {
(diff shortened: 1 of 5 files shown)


ui-admin/src/components/Tree.tsx
```diff
@@ -0,0 +1,256 @@
...
+  children?: TreeItemT<T>[]
+}
+
+type TreeItemProps<T> = {
+  activeDescendant: string
+  id: string
+  isItemSelected: (item: TreeItemT<T>) => boolean
+  item: TreeItemT<T>
+  level: number
+  path: string
```
devonbush, line 34: I didn't see anywhere where you documented what the path string format is (which isn't a big deal since it's internal to this component.)  But It looks like it's basic object notation?
```suggestion
  path: string // object notation path string e.g. 'pages[1].elements[3]`
```

devonbush (approved):
nice refactor, one minor suggestion

MatthewBemis (approved):

PR #433, opened 2023-06-21
[JN-433] Add name for element in OurHealth consent form

nawatts:
Discovered this while working on the table of contents for the form editor. One element in the OurHealth consent form is missing a name / stable ID. The name field is how elements are labeled in the table of contents.

### DIFF
--- a/populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/surveys/ourHealthConsent.json
+++ b/populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/surveys/ourHealthConsent.json
@@ -62,7 +62,7 @@
             },
             {
               "type": "html",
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):

nawatts:
@devonbush now that OurHealth is in production, how do we handle updates to survey content like this?

devonbush (approved):

devonbush:
The updates can go in as usual, and we'll release them to demo/prod, and run populate.  the difference is that for non-trivial updates (I think this qualifies as a trivial update), we wouldn't use "overwrite" when repopulating, which would mean this would get created as a new version in production.  Also, if we want an update to actually make it to the live environments, we'll have to use the UI to copy over the environments.

PR #434, opened 2023-06-21
[JN-378] Allow reordering elements on form pages and panels

nawatts:
This adds the ability to reorder questions/elements in forms, both at the page level and panel level. In the future, this could be improved to add drag and drop reordering.

Screenshot 2023-06-21 at 2 21 16 PM

### DIFF
--- a/ui-admin/src/forms/FormDesigner.tsx
+++ b/ui-admin/src/forms/FormDesigner.tsx
@@ -1,7 +1,10 @@
+import { get, set } from 'lodash/fp'
 import React, { useState } from 'react'
 
-import { FormContent } from '@juniper/ui-core'
+import { FormContent, FormContentPage, FormElement } from '@juniper/ui-core'
 
+import { PageDesigner } from './designer/PageDesigner'
+import { PanelDesigner } from './designer/PanelDesigner'
 import { FormTableOfContents } from './FormTableOfContents'
 
 type FormDesignerProps = {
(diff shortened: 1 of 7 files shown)


ui-admin/src/forms/FormDesigner.tsx
```diff
@@ -26,14 +28,40 @@ export const FormDesigner = (props: FormDesignerProps) => {
...
         {(() => {
           if (selectedElementPath === undefined) {
             return (
               <p className="mt-5 text-center">Select an element to edit</p>
             )
           }
 
+          const selectedElement = get(selectedElementPath, value) as FormContentPage | FormElement
+
+          if (!('type' in selectedElement) && !('questionTemplateName' in selectedElement)) {
```
devonbush, line 41: this logic feels like belongs in a helper function

ui-admin/src/forms/FormDesigner.tsx
```diff
@@ -26,14 +28,40 @@ export const FormDesigner = (props: FormDesignerProps) => {
...
+                readOnly={readOnly}
+                value={selectedElement}
+                onChange={updatedElement => {
+                  onChange(set(selectedElementPath, updatedElement, value))
+                }}
+              />
+            )
+          }
+
+          if ('type' in selectedElement && selectedElement.type === 'panel') {
```
devonbush, line 53: likewise

devonbush (approved):
nice, worked well locally when I tried it -- good straightforward UX

ui-admin/src/forms/FormDesigner.tsx
nawatts: Agreed, but I put it here because this way TypeScript handles the type narrowing and provides some type safety around deciding which component to render based on the selected element. With a type guard helper (ex: `const isPage = (element: FormElement | FormContentPage): element is FormContentPage => { ... }`, TS trusts the developer that whatever logic in the type guard is correct.

MatthewBemis (approved):

PR #436, opened 2023-06-22
[JN-406] postgres search prototype

devonbush:
This adds proof-of-concept filtering of the participant list via Postgres.  
image
I've walked Nick through the code generally, and I'll do the same with Matt.  I'm not sure the code deserves too much attention, since we'll be refining this A LOT as we add non-hardcoded facets and saved filters and whatnot.  What we should be paying the most attention to is whether all this code seems maintainable, with an eye towards when/if/how we should use elasticSearch or some other search cache instead (or likewise use a library for the facet UI).  the big advantage of this approach is it doesn't require a separate cache/infrastructure.  Which is HUGE, because it makes troubleshooting WAY easier, and we don't have to worry about sync.   the flip side, is that we have some pretty gnarly sql mapping instead.  

TO REVIEW:
1. Start with looking at enrolleeSearch.ts and the SAMPLE_FACETS -- that gives you the data structure for the facets themselves. 
2. Look at facetsToString and facetsFromString to get a sense of how we are mapping these to URL params
3. Look at EnrolleeSearchDaoTests to see the filters and how they're built on the Java side, and how they're tested.
4. Skim the rest -- it will probably be either substantially upgraded as we move this beyond POC, or scrapped in facvor of a totally different approach

TO TEST:
1. reload adminApiApp
2. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants
3. futz around with the filters, confirm they work
4. reload the page with various filters specified, confirm they persist


MatthewBemis (approved):
Props for getting this working!

I do find the SQL fragmentation a bit difficult to follow at times, but I don't really see a way to avoid that and I think the tradeoffs are still skewed in favor of this approach versus pulling in Elastic right now.

ui-admin/src/api/enrolleeSearch.ts
```diff
@@ -0,0 +1,191 @@
...
+  stableIdOptions: [
+    { value: 'oh_oh_consent', label: 'Consent' },
+    { value: 'oh_oh_basicInfo', label: 'Basics' },
+    { value: 'oh_oh_cardioHx', label: 'Cardio History' }
+  ]
+}]
+
+
+/**
+ *
```
nawatts: As many times as you've commented on my PRs about missing docstrings... 😁

ui-admin/src/api/enrolleeSearch.ts
```diff
@@ -0,0 +1,191 @@
...
+  if (facet.type === 'INT_RANGE') {
+    return new IntRangeFacetValue(facet, facetValue as IntRangeFacetValueFields)
+  } else if (facet.type === 'STABLEID_STRING') {
+    const newValues = facetValue ? (facetValue as StableIdStringArrayFacetValueFields).values.map(stableIdVal =>
+      new StableIdStringValue(stableIdVal.stableId, stableIdVal.values)
+    ) : []
+    return new StableIdStringArrayFacetValue(facet, { values: newValues })
+  }
+  return new StringFacetValue(facet, facetValue as StringFacetValueFields)
+}
```
nawatts: You could have TypeScript verify that this function handles all possible values of `facet.type` with something like this:

```suggestion
const checkExhaustive = (facetType: never) => {
  throw new Error(`Unhandled facet type ${facetType}`)
}

export const newFacetValue = (facet: Facet, facetValue?: object): FacetValue => {
  const facetType = facet.type
  if (facetType === 'INT_RANGE') {
    return new IntRangeFacetValue(facet, facetValue as IntRangeFacetValueFields)
  } else if (facetType === 'STABLEID_STRING') {
    const newValues = facetValue ? (facetValue as StableIdStringArrayFacetValueFields).values.map(stableIdVal =>
      new StableIdStringValue(stableIdVal.stableId, stableIdVal.values)
    ) : []
    return new StableIdStringArrayFacetValue(facet, { values: newValues })
  } else if (facetType === 'STRING') {
    return new StringFacetValue(facet, facetValue as StringFacetValueFields)
  }
  checkExhaustive(facetType)
}
```

nawatts: You might also be able to use [overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) to check that the types of the facet and facetValue parameters match.

ui-admin/src/api/enrolleeSearch.ts
```diff
@@ -0,0 +1,191 @@
...
+ *
+ */
+export const facetValuesFromString = (paramString: string, facets: Facet[]): FacetValue[] => {
+  const facetValues = []
+  const paramObj = JSON.parse(paramString)
+
+  for (const categoryName in paramObj) {
+    const category = paramObj[categoryName]
+    for (const keyName in category) {
+      const matchedFacet = facets.find(facet => facet.category === categoryName && facet.keyName === keyName) as Facet
```
nawatts, line 195: Is there guaranteed to be a matching facet?

ui-admin/src/study/participants/participantList/ParticipantList.tsx
```diff
@@ -106,26 +125,30 @@ function ParticipantList({ studyEnvContext }: {studyEnvContext: StudyEnvContextT
...
+        <h2 className="h4 text-center me-4">{study.name} Participants</h2>
+        <div className="d-flex align-items-center justify-content-between">
+          <div className="d-flex align-items-center">
+            <Link to={studyEnvMetricsPath(portal.shortcode, currentEnv.environmentName, study.shortcode)}
+              className="mx-2">Metrics</Link>
+            <span className="px-1">|</span>
+            <Link to={getExportDataBrowserPath(currentEnvPath)} className="mx-2">Export preview</Link>
+            <span className="px-1">|</span>
+            <button className="btn btn-secondary" onClick={() => setShowExportModal(!showExportModal)}
+              aria-label="show or hide export modal">
```
nawatts: Since it has text ("Download"), this button doesn't need an aria-label.

ui-admin/src/study/participants/participantList/facets/EnrolleeSearchFacets.tsx
```diff
@@ -0,0 +1,86 @@
...
+import IntRangeFacetView from './IntRangeFacetView'
+import StableIdStringFacetView from './StableIdStringFacetView'
+import StringFacetView from './StringFacetView'
+
+/**
+ * Renders a list of facets in an accordion.  Takes an array of facet values -- this array should represent only
+ * those facets that have user-specified non-default values.
+ */
+export default function EnrolleeSearchFacets({ facets, facetValues, updateFacetValues }:
+    {facets: Facet[], facetValues: FacetValue[], updateFacetValues: (values: FacetValue[]) => void}) {
```
nawatts: Style nit: to make function signatures like this less noisy, I've been using this style:

```suggestion
type EnrolleeSearchFacetsProps = {
  facets: Facet[]
  facetValues: FacetValue[]
  updateFacetValues: (values: FacetValue[]) => void
}

/**
 * Renders a list of facets in an accordion.  Takes an array of facet values -- this array should represent only
 * those facets that have user-specified non-default values.
 */
export default function EnrolleeSearchFacets(props: EnrolleeSearchFacetsProps) {
  const { facets, facetValues, updateFacetValues } = props
```

ui-admin/src/study/participants/participantList/facets/EnrolleeSearchFacets.tsx
```diff
@@ -0,0 +1,86 @@
...
+    return <IntRangeFacetView facetValue={facetValue as IntRangeFacetValue}
+      updateValue={updateValue}/>
+  } else if (facet.type === 'STRING') {
+    return <StringFacetView facetValue={facetValue as StringFacetValue}
+      updateValue={updateValue}/>
+  } else if (facet.type === 'STABLEID_STRING') {
+    return <StableIdStringFacetView facetValue={facetValue as StableIdStringArrayFacetValue}
+      updateValue={updateValue}/>
+  }
+  return <span>yo</span>
```
nawatts: See comment on enrolleeSearch.ts for how to make TypeScript check that all facet types have been handled.

ui-admin/src/study/participants/participantList/facets/IntRangeFacetView.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ *
+ */
+const IntRangeFacetView = ({ facetValue, updateValue }:
+                             {facetValue: IntRangeFacetValue,
+                               updateValue: (facetValue: FacetValue | null) => void}) => {
+  const min = facetValue.min ?? ''
+  const max = facetValue.max ?? ''
+  const textToVal = (text: string): number | null => {
```
nawatts: Since this function doesn't depend on anything within the component, suggest moving it out to the module level to avoid recreating it on every render.

ui-admin/src/study/participants/participantList/facets/IntRangeFacetView.tsx
```diff
@@ -0,0 +1,35 @@
...
+    const intVal = parseInt(text)
+    if (isNaN(intVal)) {
+      return null
+    }
+    return intVal
+  }
+  return <div>
+    <label>min <input type="text" size={2} value={min}
```
nawatts: type="number"?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number

nawatts (approved):

ui-admin/src/api/enrolleeSearch.ts
devonbush: oh, that's funny.  I've got a script that auto-runs --fix and flags the errors.  It must have auto-added a bunch of empty comments

ui-admin/src/api/enrolleeSearch.ts
devonbush: that's a good point about there not being any validation on the types matching.  I think that we'll want that, but it won't be via typescript, since this function is almost always being called from url parsing, where the caller won't know and can't check the types.

ui-admin/src/api/enrolleeSearch.ts
devonbush: no, good point, will add code to ignore mismatched facets

PR #439, opened 2023-06-23
[JN-378] Add basic form components

nawatts:
While working on the question designer, I found myself frequently repeating the same markup for forms. So I made some components for basic form elements: Buttons, Checkboxes, TextInputs, and Textareas.

To make them more accessible, these components use aria-disabled instead of disabled (see https://css-tricks.com/making-disabled-buttons-more-inclusive/). Since there isn't styling based on aria-disabled, this also adds a "disabled" CSS class that mimics Bootstrap's styles for the "disabled" attribute. To allow copying those styles, which use Bootstrap variables, this also changes the main stylesheet from CSS to SCSS.


devonbush (approved):
nice, agree that consistent handling of disabled inputs will be a big win

MatthewBemis (approved):

PR #442, opened 2023-06-25
JN-437: adding UX configuration of portal configs

devonbush:
This makes the portal environment config page editable by superusers.  This was prompted by me accidentally copying over the participantHostname from the sandbox environment to the live environment as part of the release Friday, which took the participant UX offline.  To reset it, I had to log into the prod DB manually.  I'll be making a later PR to make it harder to accidentally overwrite settings like this one that should not be copied between environments.  But this makes those problems easier to fix in the short term, as well as making it easier to password-protect the site quickly.

image

If the user isn't a superuser, the button is disabled.  This should definitely be migrated to use Nick's new button component once that is merged

image

TO TEST:
1. restart ApiAdminApp
2. go to https://localhost:3000/ourhealth/env/sandbox/config
3. change some settings, hit "save"
4. confirm the changes are saved, and a popup appears indicating the save was successful


ui-admin/src/App.css
```diff
@@ -18,6 +18,13 @@
   font-style: italic;
 }
 
+.btn-primary[aria-disabled="true"] {
```
nawatts: For buttons, Bootstrap has a "disabled" CSS class that will apply Bootstrap's default styles for a disabled button. I'd prefer to use it rather than have multiple different styles for disabled buttons.

ui-admin/src/portal/PortalEnvConfigView.tsx
```diff
@@ -1,41 +1,92 @@
...
+  const [config, setConfig] = useState(portalEnv.portalEnvironmentConfig)
+  const { user } = useUser()
 
+  /** update a given field in the config */
+  const updateConfig = (propName: string, value: string | boolean) => {
+    const newConfig = _cloneDeep(config)
+    // eslint-disable-next-line
+    // @ts-ignore
+    newConfig[propName] = value
+    setConfig(newConfig)
```
nawatts: I've been liking Lodash FP for data munging like this since it treats values as immutable and returns copies.

```ts
import { set } from 'lodash/fp'

...

const updateConfig = (propName: string, value: string | boolean) => {
    setConfig(set(propName, value))
}
```

nawatts (approved):

ui-admin/src/portal/PortalEnvConfigView.tsx
```diff
@@ -1,41 +1,92 @@
...
       </div>
       <div>
         <label className="form-label">
         Email source address
-          <input type="text" className="form-control" readOnly={true} value={config.emailSourceAddress}/>
+          <input type="text" className="form-control" value={config.emailSourceAddress}
+            onChange={e => updateConfig('emailSourceAddress', e.target.value)}/>
         </label>
       </div>
+      <button onClick={save}
```
nawatts: Since #439 has merged, the Button component added there may be useful here. It handles aria-disabled, disabled styles, and tooltips.

ui-admin/src/portal/PortalEnvConfigView.tsx
devonbush: oooh, that is a big improvement

nawatts: Lodash/FP takes a bit of getting used to though, since it reorders arguments from the base Lodash functions and the documentation is... sparse.
https://github.com/lodash/lodash/wiki/FP-Guide

ui-admin/src/components/forms/Button.tsx
```diff
@@ -46,6 +47,7 @@ export const Button = (props: ButtonProps) => {
         {...buttonProps}
         ref={buttonRef}
         aria-disabled={disabled}
+        style={{ pointerEvents: 'auto' }}
```
nawatts, line 50: It's not obvious why this is needed, so a comment about showing tooltips on hover for disabled buttons would be nice.

MatthewBemis (approved):

PR #443, opened 2023-06-26
[JN-414] Client-side participant filtering

MatthewBemis:
This enables the client-side filtering available in react-table. It also adds the contactEmail field to the participant view (hidden by default).

Screenshot 2023-06-26 at 2 50 15 PM

TO TEST:
1. Go to https://localhost:3000/ourhealth/env/sandbox/participants
2. Play around with the filters

### DIFF
--- a/ui-admin/src/study/participants/enrolleeView/ParticipantTaskView.tsx
+++ b/ui-admin/src/study/participants/enrolleeView/ParticipantTaskView.tsx
@@ -9,7 +9,7 @@ import {
   SortingState,
   useReactTable
 } from '@tanstack/react-table'
-import { sortableTableHeader } from 'util/tableUtils'
+import { tableHeader } from 'util/tableUtils'
 import { instantToDefaultString } from 'util/timeUtils'
 
 const columns: ColumnDef<ParticipantTask>[] = [{
@@ -56,7 +56,7 @@ const ParticipantTaskView = ({ enrollee }: {enrollee: Enrollee}) => {
     <table className="table table-striped">
       <thead>
         <tr>
-          {table.getFlatHeaders().map(header => sortableTableHeader(header))}
+          {table.getFlatHeaders().map(header => tableHeader(header, { sortable: true, filterable: false }))}
         </tr>
       </thead>
       <tbody>
--- a/ui-admin/src/study/participants/participantList/ParticipantList.test.tsx
+++ b/ui-admin/src/study/participants/participantList/ParticipantList.test.tsx
@@ -5,6 +5,8 @@ import ParticipantList from './ParticipantList'
 import { EnrolleeSearchResult } from 'api/api'
 import { mockEnrollee, mockStudyEnvContext } from 'test-utils/mocking-utils'
 import { setupRouterTest } from 'test-utils/router-testing-utils'
+import userEvent from "@testing-library/user-event";
+import { act } from "react-dom/test-utils";
 
 jest.mock('api/api', () => ({
   searchEnrollees: () => {
@@ -27,6 +29,40 @@ test('renders a participant with link', async () => {
   })
   const participantLink = screen.getByText('JOSALK')
   expect(participantLink).toHaveAttribute('href', `/${studyEnvContext.currentEnvPath}/participants/JOSALK`)
+})
+
+test('renders filters for participant columns', async () => {
+  const studyEnvContext = mockStudyEnvContext()
+  const { RoutedComponent } = setupRouterTest(<ParticipantList studyEnvContext={studyEnvContext}/>)
+  render(RoutedComponent)
+
+  //Assert that all 3 default columns have filter inputs
+  await waitFor(() => {
+    expect(screen.getAllByPlaceholderText('Search...').length).toBe(3)
+  })
+
+})
+
+test('filters participants based on shortcode', async () => {
+  const studyEnvContext = mockStudyEnvContext()
+  const { RoutedComponent } = setupRouterTest(<ParticipantList studyEnvContext={studyEnvContext}/>)
+  render(RoutedComponent)
+
+  //Assert that JOSALK is visible in the table
+  await waitFor(() => {
+    expect(screen.getByText('JOSALK')).toBeInTheDocument()
+  })
+
+  //Search for some unknown shortcode
+  await act(() =>
+    userEvent.type(screen.getAllByPlaceholderText('Search...')[0], 'UNKNOWN SHORTCODE')
+  )
+
+  //Assert that JOSALK is no longer visible in the table
+  await waitFor(() => {
+    expect(screen.queryByText('JOSALK')).not.toBeInTheDocument()
+  })
+
 })
 
 test('send email is toggled depending on participants selected', async () => {
--- a/ui-admin/src/study/participants/participantList/ParticipantList.test.tsx
+++ b/ui-admin/src/study/participants/participantList/ParticipantList.test.tsx
@@ -36,9 +36,9 @@ test('renders filters for participant columns', async () => {
   const { RoutedComponent } = setupRouterTest(<ParticipantList studyEnvContext={studyEnvContext}/>)
   render(RoutedComponent)
 
-  //Assert that all 3 default columns have filter inputs
+  //There are 3 default columns shown, 2 of which allow text search
   const searchInputs = await screen.findAllByPlaceholderText('Search...')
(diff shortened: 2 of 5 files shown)


ui-admin/src/study/participants/enrolleeView/ParticipantTaskView.tsx
```diff
@@ -56,7 +56,7 @@ const ParticipantTaskView = ({ enrollee }: {enrollee: Enrollee}) => {
     <table className="table table-striped">
       <thead>
         <tr>
-          {table.getFlatHeaders().map(header => sortableTableHeader(header))}
+          {table.getFlatHeaders().map(header => tableHeader(header, { sortable: true, filterable: false }))}
```
MatthewBemis, line 59: Configurable sorting/filtering because I didn't want to assume that we'd want filtering on every table in the UI. But this PR makes it so all you have to do is set `filterable` to `true` and add `getFilteredRowModel: getFilteredRowModel(),` to the table properties to enable filtering.

ui-admin/src/study/participants/participantList/ParticipantList.tsx
```diff
@@ -68,12 +70,17 @@ function ParticipantList({ studyEnvContext }: {studyEnvContext: StudyEnvContextT
...
     accessorKey: 'profile.givenName'
+  }, {
+    id: 'contactEmail',
+    header: 'Contact email',
+    accessorKey: 'profile.contactEmail'
   }, {
     header: 'Consented',
-    accessorKey: 'enrollee.consented',
-    cell: info => info.getValue() ? <FontAwesomeIcon icon={faCheck}/> : ''
+    accessorFn: data => data.enrollee.consented.toString(),
```
MatthewBemis: Converting this cell to text so we can filter on it. I attempted to handle this by adding a boolean dropdown filter but I couldn't get it working quickly and it made the table look horrible.

MatthewBemis: Just caught up with Erin about this. I'm gonna do one more pass at trying to improve filtering on this column

ui-admin/src/study/participants/participantList/ParticipantList.test.tsx
```diff
@@ -29,6 +31,38 @@ test('renders a participant with link', async () => {
...
 
+test('renders filters for participant columns', async () => {
+  const studyEnvContext = mockStudyEnvContext()
+  const { RoutedComponent } = setupRouterTest(<ParticipantList studyEnvContext={studyEnvContext}/>)
+  render(RoutedComponent)
+
+  //Assert that all 3 default columns have filter inputs
+  await waitFor(() => {
+    expect(screen.getAllByPlaceholderText('Search...')).toHaveLength(3)
+  })
```
nawatts: I think this could probably be simplified to use `findBy...` instead of `waitFor... getBy...`.

```suggestion
  const searchInputs = await screen.findByPlaceholderText('Search...')
  expect(searchInputs).toHaveLength(3)
```

https://testing-library.com/docs/queries/about/#types-of-queries

ui-admin/src/study/participants/participantList/ParticipantList.test.tsx
```diff
@@ -29,6 +31,38 @@ test('renders a participant with link', async () => {
...
+
+test('filters participants based on shortcode', async () => {
+  const studyEnvContext = mockStudyEnvContext()
+  const { RoutedComponent } = setupRouterTest(<ParticipantList studyEnvContext={studyEnvContext}/>)
+  render(RoutedComponent)
+
+  //Assert that JOSALK is visible in the table
+  await waitFor(() => {
+    expect(screen.getByText('JOSALK')).toBeInTheDocument()
+  })
```
nawatts: `getBy...` and `findBy...` throw if no matching element is found, so this could be simplified to:

```suggestion
  await screen.findByText('JOSALK')
```

ui-admin/src/util/tableUtils.tsx
```diff
@@ -1,23 +1,101 @@
...
+function DebouncedInput({
+  value: initialValue,
+  onChange,
+  debounce = 200,
+  ...props
+}: {
+  value: string
+  onChange: (value: string) => void
+  debounce?: number
+} & Omit<React.InputHTMLAttributes<HTMLInputElement>, 'onChange'>) {
```
nawatts: Another way to reference props for builtin components is through `JSX.InstrinsicElements`
```suggestion
} & Omit<JSX.IntrinsicElements['input'], 'onChange'>) {
```

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d952cf91cf1bf44f412df927cf48a18670c0ca5d/types/react/index.d.ts#L3253

ui-admin/src/util/tableUtils.tsx
```diff
@@ -1,23 +1,101 @@
...
+  const [value, setValue] = React.useState(initialValue)
+
+  React.useEffect(() => {
```
nawatts: Style nit: generally, the convention is to import these instead of using them from the React namespace.

ui-admin/src/util/tableUtils.tsx
```diff
@@ -1,23 +1,101 @@
...
+}) {
+  const columnFilterValue = column.getFilterValue()
+
+  return <div>
+    <DebouncedInput
+      type="text"
+      value={(columnFilterValue ?? '') as string}
+      onChange={value => column.setFilterValue(value)}
+      placeholder={`Search...`}
+      list={`${column.id} list`}
```
nawatts: What's responsible for rendering the datalist element referenced here?

ui-admin/src/util/tableUtils.tsx
```diff
@@ -1,23 +1,101 @@
...
+  </div>
+}
+
+
+/**
+ * returns a table header with optional sorting and filtering
+ * adapted from https://tanstack.com/table/v8/docs/examples/react/sorting
+ * */
+export function tableHeader<A, B>(header: Header<A, B>, options: { sortable: boolean, filterable: boolean }) {
+  return <th key={header.id}>
```
nawatts: If sortable, the th element could use `aria-sort` attributes.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort

ui-admin/src/study/participants/participantList/ParticipantList.test.tsx
MatthewBemis: Simplified in a281500b

ui-admin/src/study/participants/participantList/ParticipantList.test.tsx
MatthewBemis: Done in a281500b

ui-admin/src/util/tableUtils.tsx
MatthewBemis: Resolved in 7bfb73d61863d529779ed3379e83b923eef5c498

ui-admin/src/util/tableUtils.tsx
MatthewBemis: Ah this is a leftover from the react-table example code. They had a bunch of handling for array column types, which we don't need. Just forgot to remove this line. Done in 00d83beaf9d7a364313eced7fae712335cb0c5df

ui-admin/src/util/tableUtils.tsx
MatthewBemis: Added aria attribute in 6ca2fc30

nawatts (approved):

ui-admin/src/study/participants/participantList/ParticipantList.test.tsx
```diff
@@ -29,6 +31,35 @@ test('renders a participant with link', async () => {
...
+  //Search for some unknown shortcode
+  await act(() =>
+    userEvent.type(screen.getAllByPlaceholderText('Search...')[0], 'UNKNOWN SHORTCODE')
+  )
+
+  //Assert that JOSALK is no longer visible in the table
+  await waitFor(() => {
+    expect(screen.queryByText('JOSALK')).not.toBeInTheDocument()
+  })
+})
```
devonbush, line 61: nice test.  This could also use a positive test case -- that if you type "jos" JOSALK is still visible

devonbush (approved):
good refactoring around handling sort + filter

PR #444, opened 2023-06-26
[JN-378] Add UI for editing questions in forms

nawatts:
This adds the UI to edit questions in surveys and other forms.

This supports most (but not all) of the SurveyJS question options that we use. In particular, this is still missing the ability to make a text question use a number input and add validation.

Some example screenshots...
Screenshot 2023-06-26 at 3 47 08 PM
Screenshot 2023-06-26 at 3 47 18 PM

### DIFF
--- a/ui-admin/src/forms/designer/QuestionDesigner.tsx
+++ b/ui-admin/src/forms/designer/QuestionDesigner.tsx
@@ -10,25 +10,25 @@ import { questionTypeDescriptions, questionTypeLabels } from './questions/questi
 import { VisibilityFields } from './questions/VisibilityFields'
 
 export type QuestionDesignerProps = {
+  question: Question
   readOnly: boolean
-  value: Question
   onChange: (newValue: Question) => void
 }
 
 /** UI for editing a question in a form. */
 export const QuestionDesigner = (props: QuestionDesignerProps) => {
-  const { readOnly, value, onChange } = props
+  const { question, readOnly, onChange } = props
 
-  const isTemplated = 'questionTemplateName' in value
+  const isTemplated = 'questionTemplateName' in question
 
   return (
     <div>
-      <h2>{value.name}</h2>
+      <h2>{question.name}</h2>
 
       {!isTemplated && (
         <>
-          <p className="fs-4 mb-0">{questionTypeLabels[value.type]} question</p>
-          <p>{questionTypeDescriptions[value.type]}</p>
+          <p className="fs-4 mb-0">{questionTypeLabels[question.type]} question</p>
+          <p>{questionTypeDescriptions[question.type]}</p>
         </>
       )}
 
@@ -36,7 +36,7 @@ export const QuestionDesigner = (props: QuestionDesignerProps) => {
         <>
           <p className="fs-4 mb-0">Templated question</p>
           <p>
-            This question uses <span className="fw-bold">{value.questionTemplateName}</span> as a template.
+            This question uses <span className="fw-bold">{question.questionTemplateName}</span> as a template.
             Question settings entered here override settings from the template.
           </p>
         </>
@@ -44,33 +44,33 @@ export const QuestionDesigner = (props: QuestionDesignerProps) => {
 
       <BaseFields
         disabled={readOnly}
-        question={value}
+        question={question}
         onChange={onChange}
       />
 
       {!isTemplated && (
         <>
           {
-            (value.type === 'checkbox' || value.type === 'dropdown' || value.type === 'radiogroup') && (
+            (question.type === 'checkbox' || question.type === 'dropdown' || question.type === 'radiogroup') && (
               <>
                 <ChoicesList
-                  question={value}
+                  question={question}
                   readOnly={readOnly}
                   onChange={onChange}
                 />
                 <OtherOptionFields
                   disabled={readOnly}
-                  question={value}
+                  question={question}
                   onChange={onChange}
                 />
               </>
             )
           }
           {
-            value.type === 'checkbox' && (
+            question.type === 'checkbox' && (
               <CheckboxFields
                 disabled={readOnly}
-                question={value}
+                question={question}
                 onChange={onChange}
               />
             )
@@ -80,7 +80,7 @@ export const QuestionDesigner = (props: QuestionDesignerProps) => {
 
       <VisibilityFields
         disabled={readOnly}
-        question={value}
+        question={question}
         onChange={onChange}
       />
     </div>
--- a/ui-admin/src/forms/designer/QuestionDesigner.tsx
+++ b/ui-admin/src/forms/designer/QuestionDesigner.tsx
@@ -1,0 +1,88 @@
+import React from 'react'
+
+import { Question } from '@juniper/ui-core'
+
+import { BaseFields } from './questions/BaseFields'
+import { CheckboxFields } from './questions/CheckboxFields'
+import { ChoicesList } from './questions/ChoicesList'
+import { OtherOptionFields } from './questions/OtherOptionFields'
+import { questionTypeDescriptions, questionTypeLabels } from './questions/questionTypes'
+import { VisibilityFields } from './questions/VisibilityFields'
+
+export type QuestionDesignerProps = {
+  readOnly: boolean
+  value: Question
+  onChange: (newValue: Question) => void
+}
+
+/** UI for editing a question in a form. */
+export const QuestionDesigner = (props: QuestionDesignerProps) => {
+  const { readOnly, value, onChange } = props
+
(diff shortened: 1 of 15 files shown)


ui-admin/src/forms/designer/questions/VisibilityFields.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Checkbox } from 'components/forms/Checkbox'
+import { TextInput } from 'components/forms/TextInput'
+
+type VisibilityFieldsProps = {
+  disabled: boolean
+  question: Question
+  onChange: (newValue: Question) => void
+}
+
+/** Controls for editing other option for questions. */
```
breilly2: This looks like copy/paste from `OtherOptionFileds.tsx`. I wasn't reading exhaustively, so there may be others that I missed.

breilly2 (approved):
Looks great! Works well based on my limited testing, and the code is beautiful!

Suggestion: when changing to "JSON Editor" or "Preview" and then back to "Designer", it would be great to return to the item that was previously being viewed.

ui-admin/src/forms/designer/QuestionDesigner.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { BaseFields } from './questions/BaseFields'
+import { CheckboxFields } from './questions/CheckboxFields'
+import { ChoicesList } from './questions/ChoicesList'
+import { OtherOptionFields } from './questions/OtherOptionFields'
+import { questionTypeDescriptions, questionTypeLabels } from './questions/questionTypes'
+import { VisibilityFields } from './questions/VisibilityFields'
+
+export type QuestionDesignerProps = {
+  readOnly: boolean
+  value: Question
```
devonbush: Maybe  "question" or "targetQuestion" or "item" instead of value?  When I saw 'value' I thought of option values

ui-admin/src/forms/designer/questions/VisibilityFields.tsx
nawatts: Fixed in https://github.com/broadinstitute/juniper/pull/444/commits/f9dc871741c38f7339a97c2b554b84f7564eeb09.

ui-admin/src/forms/designer/questions/ChoicesList.tsx
```diff
@@ -0,0 +1,158 @@
...
+/** UI for editing the list of choices for a question. */
+export const ChoicesList = (props: ChoicesListProps) => {
+  const { question, readOnly, onChange } = props
+
+  const labelId = useId()
+
+  return (
+    <div className="mb-3">
+      <p className="mb-2" id={labelId}>Choices</p>
+      <ol aria-labelledby={labelId} className="list-group list-group-numbered mb-1">
```
devonbush: IMO the list numbers are distracting and might confuse users into thinking the choice numbers will appear in the participant UI.  I'd remove them

ui-admin/src/forms/designer/questions/ChoicesList.tsx
```diff
@@ -0,0 +1,158 @@
...
+              >
+                <TextInput
+                  className="form-control"
+                  disabled={readOnly}
+                  label="Text"
+                  labelClassname="mb-0"
+                  value={choice.text}
+                  onChange={value => {
+                    onChange({
+                      ...question,
```
devonbush, line 53: this pattern makes me nervous -- since you're not deep-copying question, it seems more likely that bugs related to duplicate references being passed across state updates will result if `question` acquires nested attributes.  
This also seems like it's complex enough to warrant being in a named function.  `changeChoiceText(index, value)`

ui-admin/src/forms/designer/questions/ChoicesList.tsx
```diff
@@ -0,0 +1,158 @@
...
+                  label="Value"
+                  labelClassname="mb-0"
+                  value={choice.value}
+                  onChange={value => {
+                    onChange({
+                      ...question,
+                      choices: [
+                        ...question.choices.slice(0, i),
+                        { ...question.choices[i], value },
+                        ...question.choices.slice(i + 1)
```
devonbush, line 73: same comment as above, seems like a named `changeChoiceValue(index, value)` function would make this more readable

ui-admin/src/forms/designer/questions/ChoicesList.tsx
```diff
@@ -0,0 +1,158 @@
...
+                  icon={faChevronUp}
+                  variant="light"
+                  onClick={() => {
+                    onChange({
+                      ...question,
+                      choices: [
+                        ...question.choices.slice(0, i - 1),
+                        question.choices[i],
+                        question.choices[i - 1],
+                        ...question.choices.slice(i + 1)
```
devonbush, line 93: same :)

ui-admin/src/forms/designer/questions/ChoicesList.tsx
```diff
@@ -0,0 +1,158 @@
...
+                  icon={faTimes}
+                  variant="light"
+                  onClick={() => {
+                    onChange({
+                      ...question,
+                      choices: [
+                        ...question.choices.slice(0, i),
+                        ...question.choices.slice(i + 1)
+                      ]
+                    })
```
devonbush, line 132: Yeah, seems like we have different preferred styles for when to pull things out into component-level functions.  Let me know your thoughts -- I lean towards pulling most non-trivial things into component-level functions because it (to me) makes the markup of the HTML elements easier to follow.  Let me know what your approach is (and to what extent there is a general DSP guideline on when to inline)

devonbush (changes requested):
Looks great, and worked smooth in testing.  Marking as request changes because I want to make sure we get on the same page about function inlining -- to me a lot of this was really hard to follow when skimming, but that may just mean I need to spend more time getting used to a different set of conventions around inlining

ui-admin/src/forms/designer/questions/ChoicesList.tsx
breilly2: I'm curious about your thinking here, but not quite following. I agree that the `...` spread operator can be a bit loose, but it's also very powerful, and using typescript also focuses the range of possible values in a way that makes me less nervous about it.

ui-admin/src/forms/designer/questions/ChoicesList.tsx
breilly2: FWIW, each component on its own is small enough to be digestible to me. Extracted functions could still be useful, as you pointed out elsewhere, if naming the concept would be useful to understanding the code... or to provide a testable unit, which I think is a better reason in this specific case.

nawatts: Since all these buttons do different things, I'm not sure extracting functions would be any less verbose / easier to understand. As is, the data manipulation is right next to the button label describing its intent.

ui-admin/src/forms/designer/questions/ChoicesList.tsx
nawatts: I don't think this component needs to care whether or not something else is mutating other fields. All it does is modify the choices list.

We could make this terser with Lodash/FP, but at the cost of the type safety of built in operators.

devonbush:
approving since it seems I'm outvoted by the more senior UI coders on the codestyle questions :)

devonbush (approved):

ui-admin/src/forms/designer/QuestionDesigner.tsx
nawatts: Yeah, I think this began as mimicking a form input interface with value and onChange props. But then I started using question for other stuff as well.

ui-admin/src/forms/designer/questions/ChoicesList.tsx
nawatts: I was thinking they might be useful for longer list of choices where otherwise it might require scrolling to get a sense of where the option is in the list. Also, it calls attention to the fact that the choices _are ordered_ in the participant UI, so it may be worth considering which options are presented first.

ui-admin/src/forms/designer/QuestionDesigner.tsx
nawatts: Renamed in https://github.com/broadinstitute/juniper/pull/444/commits/531132d544e52ddaed7d13986fac204fb325d63f.

ui-admin/src/forms/designer/questions/ChoicesList.tsx
nawatts: Removed in https://github.com/broadinstitute/juniper/pull/444/commits/9592dad46617a41b586d13ffab58816d46636697.

PR #449, opened 2023-06-27
[JN-378] Navigate to study environment content page when cancelling form edits

nawatts:
Currently, clicking the "Cancel" button while editing a form navigates all the way back to the study page. This changes it to navigate to the study environment page, which is the page where you see the various forms in the study environment.


devonbush (approved):
thanks for finding/fixing this, just pushed a commit on top to use the passed-in root route

MatthewBemis (approved):

PR #450, opened 2023-06-27
Remove custom btn-secondary styles

nawatts:
Possibly controversial...

The current custom btn-secondary styles aren't great. The `font-weight: bold` on hover causes the button to grow wider, which shifts other content around.

https://github.com/broadinstitute/juniper/assets/1156625/b4398aae-bcb5-47c2-bf5d-221d71fddd9b

Bootstrap's "light" or "link" variants look better suited to providing a style like the current secondary button override.
https://getbootstrap.com/docs/5.3/components/buttons/#variants

### DIFF
--- a/ui-admin/src/App.css
+++ b/ui-admin/src/App.css
@@ -7,25 +7,6 @@
   background-color: #4D72AA;
 }
 
-.btn-secondary {
-  background-color: transparent;
-  color:#4D72AA;
-  border: none;
-}
-
-.btn-secondary[aria-disabled="true"] {
-  color: #666;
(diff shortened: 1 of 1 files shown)


devonbush:
Agree--I don't like the default bootstrap secondary style because the gray looks disabled, so moving to either link/light works.  IIRC these custom secondary styles were ported from SCP which was based on the Terra style guide.  Let's run it by Erin in Slack -- she might have updated style guide guidance

devonbush:
closing pending Erin's admin redesign style/color theme decisions

PR #451, opened 2023-06-27
[JN-378] Show feedback when form content JSON is invalid

nawatts:
Currently, when invalid JSON is entered in the form editor, there is no feedback other than the disabled save button. This adds a red outline around the textarea and adds a tooltip to the save button explaining why it is disabled. It's not intended to be a final design for feedback, just an improvement on what's there now.


devonbush (approved):
nice, the red outline is a good way of showing invalid without needlessly distracting during long edits

breilly2 (approved):

PR #452, opened 2023-06-27
[JN-414 pt2] Improve UX of Consent column filter

MatthewBemis:
I got feedback from Erin that it would be a much better UX to have the consented column be a dropdown of possible options. This required extending some of the react-table functionality because it doesn't support this type of filtering natively. I've PR'd this into `mb-jn-414-search` so this particular change would be easier to review.

I also have a couple of TS/React questions within the PR.

Screenshot 2023-06-27 at 10 51 01 AM

### DIFF
--- a/ui-admin/src/util/tableUtils.tsx
+++ b/ui-admin/src/util/tableUtils.tsx
@@ -39,6 +39,7 @@ function DebouncedInput({
 
 declare module '@tanstack/table-core' {
   //Extra column metadata for extending the built-in filter functionality of react-table
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   interface ColumnMeta<TData extends RowData, TValue> {
     //Specifies the type of the column data. By default, columns will be treated as strings
     columnType?: string
--- a/ui-admin/src/util/tableUtils.tsx
+++ b/ui-admin/src/util/tableUtils.tsx
@@ -37,17 +37,6 @@ function DebouncedInput({
   )
 }
 
-declare module '@tanstack/table-core' {
-  //Extra column metadata for extending the built-in filter functionality of react-table
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  interface ColumnMeta<TData extends RowData, TValue> {
-    //Specifies the type of the column data. By default, columns will be treated as strings
-    columnType?: string
-    //Specifies the Select options if using a dropdown filter (i.e. for booleans)
-    filterOptions?: object[]
-  }
-}
-
 /**
  * returns a Filter to handle text fields
  * adapted from https://tanstack.com/table/v8/docs/examples/react/filters
@@ -261,4 +250,15 @@ export function basicTableLayout<T>(table: Table<T>) {
       })}
     </tbody>
   </table>
+}
+
+declare module '@tanstack/table-core' {
+  //Extra column metadata for extending the built-in filter functionality of react-table
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  interface ColumnMeta<TData extends RowData, TValue> {
+    //Specifies the type of the column data. By default, columns will be treated as strings
+    columnType?: string
+    //Specifies the Select options if using a dropdown filter (i.e. for booleans)
+    filterOptions?: object[]
+  }
 }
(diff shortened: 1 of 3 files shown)


ui-admin/src/util/tableUtils.tsx
```diff
@@ -45,25 +56,76 @@ function Filter<A>({
...
+          width: 200,
+          fontWeight: 'normal'
+        }),
+        menu: baseStyles => ({
+          ...baseStyles,
+          fontWeight: 'normal'
+        })
+      }}
+      value={selectedValue}
+      // @ts-ignore
```
MatthewBemis, line 83: My lack of experience with React/TypeScript is showing. This suppresses the following type error:

```
TS2322: Type '(newValue: { value: boolean; label: string;}) => void' is not assignable to type '(newValue: SingleValue<{}>, actionMeta: ActionMeta<{}>) => void'.   Types of parameters 'newValue' and 'newValue' are incompatible.     Type 'SingleValue<{}>' is not assignable to type '{ value: boolean; label: string; }'.       Type 'null' is not assignable to type '{ value: boolean; label: string; }'.
```
I understand what it's saying, but I'm not sure what TS magic to do to fix it. Any ideas? Is it because Line 75 doesn't have a type declaration so it's inferring type based off of the default state?

ui-admin/src/util/tableUtils.tsx
```diff
@@ -36,6 +37,16 @@ function DebouncedInput({
...
+declare module '@tanstack/table-core' {
+  //Extra column metadata for extending the built-in filter functionality of react-table
+  interface ColumnMeta<TData extends RowData, TValue> {
+    //Specifies the type of the column data. By default, columns will be treated as strings
+    columnType?: string
+    //Specifies the Select options if using a dropdown filter (i.e. for booleans)
+    filterOptions?: object[]
+  }
+}
+
```
MatthewBemis: I'm not certain of where this should actually go. Also the linter warns that `TData` and `TValue` are unused, but they seem to be needed to get this to extend properly?

ui-admin/src/util/tableUtils.tsx
nawatts: > Is it because Line 75 doesn't have a type declaration so it's inferring type based off of the default state?

I would guess so. My guess is that Select has a type parameter for options/value that it's inferring from the value prop. And since `selectedValue` is initialized with `{}`, it thinks that any value will be an empty object.

Does something like this fix it? Note that this would use `undefined` to represent "no selection" and there are probably places where that undefined would have to be handled.
```ts
const [selectedValue, setSelectedValue] = useState<{ value: boolean, label: string }>()
```

ui-admin/src/util/tableUtils.tsx
nawatts: Yeah, I think ESLint isn't realizing that this has a side effect of modifying the `ColumnMeta` interface. You can always disable the linter with an `// eslint-disable-next-line <rule>` or `/* eslint-disable */` comment.

That being said... is there a way to avoid modifying that interface? Does Table take a type parameter for column metadata?

MatthewBemis: I've been looking, but modifying the interface was how the developer of react-table suggested doing it: https://tanstack.com/table/v8/docs/api/core/column-def#meta

ui-admin/src/util/tableUtils.tsx
devonbush: Yes, part of the issue is that newValue can be null/undefined, so you need to handle that case (I believe it will be null if the user deselects all options)

devonbush (approved):
Looks good, pending the fixup of handling null from the onChange

ui-admin/src/util/tableUtils.tsx
nawatts: Is it possible to get rid of this ts-ignore by adding a type parameter to `useState`?

MatthewBemis: Unfortunately not. I did add the type parameter in 9f39258c. That commit also better handles the null case and gets rid of the `Any` Select option, which complicated the typing. Instead it uses the built-in `isClearable` option.

nawatts (approved):

PR #447, opened 2023-06-26
[JN-440] adding more detail to env copy screen

devonbush:
In preparation for making items individually selectable, this adds more detail to the environment copy screen.  Once OurHealth needs changes to surveys in production, fine-grained control will let us just copy over the updated surveys (after testing), without obliterating the other config settings.  This was a problem in the release Friday, where I wanted to just copy over Nick's small update to one of the ourhealth surveys, which resulted in me inadvertently nuking the participantHostname setting, causing downtime.  
The next step will be to add checkboxes next to each config item, and, except in the case where the destination environment is not yet initiailzed, they'll be unchecked by default.

image

TO TEST:
1. repopulate Ourhealth portal
2. go to https://localhost:3000/ourhealth/env/irb/diff/sandbox
3. confirm you see a ui like the above, where individual surveys are listed (previously, it just listed total counts)

PR #453, opened 2023-06-27
[JN-445] Prevent ESLint from autofixing missing JSDoc

nawatts:
Prompted by https://github.com/broadinstitute/juniper/pull/447#discussion_r1244037290

ESLint will "helpfully" fix missing JSDoc comments by adding an empty JSDoc comment. This disables the `require-jsdoc` rule when running in fix mode, which prevents auto-fixing it and leaves it for the developer to fix.


devonbush (approved):
my hero!

breilly2 (approved):

PR #454, opened 2023-06-28
[JN-443] Fix a bug causing numeric fields to not be written to the database

breilly2:
To test, enter a number in a numeric field and make sure it is saved. For example:
1. In the participant UI, fill out enough surveys to unlock the "Family History" survey.
2. In the Family History survey, select "Mother" for the first question of the second page.
3. In the field asking for age, enter a number (less than 50).
4. Unfocus the field and wait a few seconds for autosave (opening the network tool window helps here) or click through the remaining pages and complete the survey.
5. Open the Family History survey again and check that the value you entered is re-populated in the form (or check in the admin).


devonbush:
Yikes, great catch.  We'll want to release this tomorrow.

devonbush (approved):

devonbush:
looking in prod, 2 participants might have been impacted by this

nawatts (approved):

PR #457, opened 2023-06-28
[JN-378] Allow deleting elements from pages and panels

nawatts:
This allows deleting elements from pages and panels. While we're at it, it also updates buttons in ElementList to use the IconButton component.

### DIFF
--- a/ui-admin/src/forms/designer/ElementList.test.tsx
+++ b/ui-admin/src/forms/designer/ElementList.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@testing-library/react'
+import { act, render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import React from 'react'
 
@@ -43,7 +43,7 @@ describe('ElementList', () => {
 
     // Act
     const moveBarUpButton = screen.getAllByLabelText('Move this element before the previous one')[1]
-    await user.click(moveBarUpButton)
+    await act(() => user.click(moveBarUpButton))
 
     // Assert
     expect(onChange).toHaveBeenCalledWith([
@@ -65,7 +65,7 @@ describe('ElementList', () => {
     ])
 
     const moveBarDownButton = screen.getAllByLabelText('Move this element after the next one')[1]
-    await user.click(moveBarDownButton)
+    await act(() => user.click(moveBarDownButton))
 
     // Assert
     expect(onChange).toHaveBeenCalledWith([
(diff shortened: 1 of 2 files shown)


devonbush (approved):
Looks good.  When testing, the page jitters when I first move my mouse over a move choice up/down button.  It looks like it's very briefly adding both horizontal and vertical scrollbars, and then reverting them as something renrenders/recomputes based on the tooltip.  Not a big deal, but just FYI if it's easy to fix by changing the tooltip position/dimensions

nawatts:
I'm not seeing anything like that. Could you post a video?

ui-admin/src/forms/designer/ElementList.test.tsx
```diff
@@ -43,7 +43,7 @@ describe('ElementList', () => {
 
     // Act
     const moveBarUpButton = screen.getAllByLabelText('Move this element before the previous one')[1]
-    await user.click(moveBarUpButton)
+    await act(() => user.click(moveBarUpButton))
```
breilly2, line 46: I tried to do a little learning about `act()` and most of what I came up with was either [legacy documentation](https://legacy.reactjs.org/docs/test-utils.html#act) or articles ([1](https://javascript.plainenglish.io/you-probably-dont-need-act-in-your-react-tests-2a0bcd2ad65c) [2](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#wrapping-things-in-act-unnecessarily)) saying that `act()` isn't necessary. Is what I found accurate or is there some other documentation/article you know of that I could learn from?

breilly2 (approved):

ui-admin/src/forms/designer/ElementList.test.tsx
nawatts: After switching to the Button component, this showed a warning about something not wrapped in `act`. It looked like it was related to the tooltip state. I added this and didn't investigate further.

nawatts: But yeah, generally, act vs no act has been frustrating and often seems inconsistent. I _think_ act is only necessary when the action triggers something async. The second article you referenced (["Common mistakes with React Testing Library"](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#wrapping-things-in-act-unnecessarily)) is what I've been going off of.

devonbush:
@nawatts  here's a video of the jitter

[buttonJitter.webm](https://github.com/broadinstitute/juniper/assets/2800795/5b5144bd-750b-473e-bf6e-031a7fb1a739)

nawatts:
Ooh. In the question choices list. I was looking at the elements in a page/panel list.

But then... I don't see any jitter in the question choices list either.

PR #459, opened 2023-06-28
[JN-447]  Making admin date display more robust

devonbush:
this adds better handling of null birthdates, which would previously show "Invalid date" on the admin tool.  This also fixes up the type signatures -- I realized the array Date() constructor we were using (and that automatically handled the zero->one conversion) is actually undocumented?    Fun fact, `new Date(1996, 5, 16) != new Date([1996, 5, 16])`. 

TO TEST:
1. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHBASC  
2. confirm bithdate field is empty
3. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHSALK
4. confirm birthdate shows as 11/12/1979

### DIFF
--- a/ui-admin/src/util/timeUtils.tsx
+++ b/ui-admin/src/util/timeUtils.tsx
@@ -6,7 +6,7 @@ export function instantToDefaultString(instant?: number) {
   return new Date(instant * 1000).toLocaleString()
 }
 /** renders a java LocalDate as a date string.  this should be sensitive to the user's locale */
-export function dateToDefaultString(date?: number[]) {
+export function dateToDefaultString(date: number[] | undefined) {
   if (typeof date === 'undefined' || date.length != 3) {
     return ''
   }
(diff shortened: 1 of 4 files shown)


ui-admin/src/util/timeUtils.tsx
```diff
@@ -6,8 +6,12 @@ export function instantToDefaultString(instant?: number) {
   return new Date(instant * 1000).toLocaleString()
 }
 /** renders a java LocalDate as a date string.  this should be sensitive to the user's locale */
-export function dateToDefaultString(date: number) {
-  return new Date(date).toLocaleDateString()
+export function dateToDefaultString(date?: number[]) {
```
nawatts: Suggest declaring this type as `number[] | undefined` instead of making the argument optional. As is, TS will allow calling the function without an argument (`dateToDefaultString()`), which I assume is not intended.

```suggestion
export function dateToDefaultString(date: number[] | undefined) {
```

nawatts (approved):

breilly2 (approved):

PR #462, opened 2023-06-28
[JN-378] Allow creating panels from existing questions

nawatts:
This allows creating a panel from existing questions within a page.

Along the way, realized that the FormPanel type needed an update... FormPanels should not contain nested panels.

https://github.com/broadinstitute/juniper/assets/1156625/efd9a5b6-fac3-4369-99e8-b14fee20615d

### DIFF
--- a/ui-admin/src/forms/designer/NewPanelForm.tsx
+++ b/ui-admin/src/forms/designer/NewPanelForm.tsx
@@ -1,0 +1,64 @@
+import { concat, flow, uniq, without } from 'lodash/fp'
+import React, { useState } from 'react'
+
+import { FormPanel } from '@juniper/ui-core'
+
+import { Button } from 'components/forms/Button'
+import { Checkbox } from 'components/forms/Checkbox'
+
+type NewPanelFormProps = {
+  availableElements: FormPanel['elements']
+  onCreate: (newPanel: FormPanel) => void
+}
+
+/** UI for creating a new panel. */
+export const NewPanelForm = (props: NewPanelFormProps) => {
+  const { availableElements, onCreate } = props
+
+  const [selectedElements, setSelectedElements] = useState<string[]>([])
+
+  return (
+    <>
+      <fieldset className="pb-3 border-bottom mb-3">
+        <legend className="form-label fs-6 mb-2">Select elements to include</legend>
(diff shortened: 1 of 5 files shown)


ui-admin/src/forms/designer/PageDesigner.tsx
```diff
@@ -14,16 +30,59 @@ export type PageDesignerProps = {
...
 
+  const [showCreatePanelModal, setShowCreatePanelModal] = useState(false)
+
   return (
     <div>
       <h2>Page</h2>
+
+      <div className="mb-3">
+        <Button
+          disabled={readOnly || value.elements.filter(canBeIncludedInPanel).length === 0}
```
devonbush, line 41: I would pull `value.elements.filter(canBeIncludedInPanel)` into a variable computed up above.  Also, if it's not too hard, I would include items that are already in another panel in the available questions, so it's easier to rearrange items between panels, rather than having to delete/recreate panels.

Related, this might speak to the need for a QuestionSelector component that takes a current page/panel  and shows the items on that page/panel to choose from, but also allows picking questions from elsewhere in the survey, to allow more complex rearrangement.  (definitely outside the scope of this PR)

devonbush (approved):

ui-admin/src/forms/designer/NewPanelForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+        {availableElements.map(element => {
+          return (
+            <div key={element.name} className="mb-2">
+              <Checkbox
+                checked={selectedElements.includes(element.name)}
+                label={element.name}
+                onChange={checked => {
+                  if (checked) {
+                    setSelectedElements(flow(concat([element.name]), uniq))
```
breilly2, line 33: I always forget about the form of state setter that takes a function to transform the previous state. 👍

breilly2 (approved):

ui-admin/src/forms/designer/PageDesigner.tsx
nawatts: 👍 I'll leave this is as for now because I think this will soon change when exploring your QuestionSelector idea.

PR #464, opened 2023-06-29
JN-442 Fixing survey ordering on new versions

devonbush:
`surveyOrder`, along with some other properties, was being elided by the controller thanks to some outdated API DTOs.  My opinion is still that defining DTOs in the api isn't worth the trouble given that we don't have 3rd party API consumers.  Unless we're going to handwrite typesafe DTO converters (which we might eventually want to do when we need to support really robust APIs), they don't give us any extra guarantees of safety and/or appropriateness, just an extra layer for bugs.
 
TO TEST:
1. restart ApiAdminApp, repopulate ourhealth
2. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/surveys/oh_oh_famHx?readOnly=false, and make a trivial change
3. save the survey
4. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox -- confirm the surveys still appear in the correct order (Family History should still appear beneath Other Medical History)

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConfiguredConsentController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConfiguredConsentController.java
@@ -1,7 +1,6 @@
 package bio.terra.pearl.api.admin.controller.forms;
 
 import bio.terra.pearl.api.admin.api.ConfiguredConsentApi;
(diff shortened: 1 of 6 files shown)


breilly2 (approved):
I re-tested the scenarios I tested with to find the bug and everything looks good now. Thanks!

MatthewBemis (approved):

PR #465, opened 2023-06-29
[JN-440] fine-grained environment copying

devonbush:
This adds per-config variable and per-survey/consent checkboxes for copying over configuration from one environment to another.  It also adds some sensible defaults -- if the target environment isn't initialized, everything will be selected by default.  Otherwise, everything will be unselected, since it's generally preferable to not copy things by accident than to accidentally copy over a draft that isn't IRB approved.  Still tons of room to improve the UX on this page (e.g. select all, clear all buttons, tooltips, confirm dialogs), but since this is just an internal tool for now (and we probably won't actually use it until August when the next round of OurHealth changes goes through, this should be good enough.

image

TO TEST:
1. repopulate OurHealth (just to make sure you're starting fresh)
2. go to https://localhost:3000/ourhealth/env/irb/diff/sandbox
3. confirm you see a screen showing the differences between the uninitialized sandbox and initialized IRB environment, all boxes should be checked
4. click "Copy changes"
5. click into the sandbox environment, and update some surveys or change some configs
6. go back to https://localhost:3000/ourhealth/env/irb/diff/sandbox
7. confirm you see a screen showing the changes you made, with options to copy them (all deselected)
8. select some of the chagnes
9. hit "Copy changes"
10. go into the IRB environment and confirm the changes you selected are applied, and other changes are not


ui-admin/src/api/api.tsx
```diff
@@ -183,17 +183,19 @@ export type StudyEnvironmentChange = {
...
+} | {
+  changed: false
 }
 
 export type ConfigChange = {
   propertyName: string,
-  oldValue: object,
-  newValue: object
+  oldValue: object | boolean | string,
+  newValue: object | boolean | string
```
nawatts: Might be worth defining a type for these values.

```ts
export type ConfigChangeValue = object | boolean | string

export type ConfigChange = {
  propertyName: string
  oldValue: ConfigChangeValue
  newValue: ConfigChangeValue
}
```

ui-admin/src/portal/publish/PortalEnvDiffProvider.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+/**
+ * loads a diff between two environments, based on the passed-in environment and an environment name in a url param
+ * also contains logic for updating an environment with a changeset
+ */
+const PortalEnvDiffProvider = ({ portal, portalEnv, updatePortal }: PortalEnvDiffProviderProps) => {
+  const params = useParams()
+  const sourceEnvName: string | undefined = params.sourceEnvName
+  const [isLoading, setIsLoading] = useState(true)
+  const [diffResult, setDiffResult] = useState<PortalEnvironmentChange | null>(null)
```
nawatts: IDK if null vs undefined is important here, but React's types provide a convenience helper for state of type `T | undefined` that's a little bit more concise.

```ts
// diffResult is typed PortalEnvironmentChange | undefined
const [diffResult, setDiffResult] = useState<PortalEnvironmentChange>()
```

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b384d9e82713dc8400a1289eeb9be7f2fc467390/types/react/index.d.ts#L934-L941

ui-admin/src/portal/publish/PortalEnvDiffView.tsx
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * loads and displays the differences between two portal environments
+ * */
+export default function PortalEnvDiffView(
+  { changeSet, portal, portalEnv, applyChanges, sourceName }: EnvironmentDiffProps) {
+  const [selectedChanges, setSelectedChanges] = useState<PortalEnvironmentChange>(getDefaultPortalEnvChanges(changeSet))
+
+  const updateSelectedStudyEnvChanges = (update: StudyEnvironmentChange) => {
+    const matchedIndex = selectedChanges.studyEnvChanges
+      .findIndex(change => change.studyShortcode === update.studyShortcode)
```
nawatts, line 91: Will a match always be found? Or does this need to handle if `matchedIndex` is -1?

ui-admin/src/portal/publish/diffComponents.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { render, screen } from '@testing-library/react'
+import { fireEvent, render, screen } from '@testing-library/react'
```
nawatts: Generally, `@testing-library/user-event` is preferred over `fireEvent`.
https://testing-library.com/docs/user-event/intro/#differences-from-fireevent

There's some places I've used `fireEvent.change` because it was more convenient when testing controlled components vs adding a test wrapper that handled state updates.

ui-admin/src/portal/publish/diffComponents.tsx
```diff
@@ -24,77 +24,169 @@ export const VersionChangeView = ({ record }: {record: VersionedEntityChange}) =
...
   if (!configChanges.length) {
     return <span className="fst-italic text-muted">no changes</span>
   }
+  /** handles add/remove from the selected changes */
+  const updateSelection = (propertyName: string, selected: boolean) => {
+    const selectionIndex = selectedChanges.findIndex(change => change.propertyName === propertyName)
+    if (selected && selectionIndex < 0) {
+      updateSelectedChanges([
+        ...selectedChanges,
+        configChanges.find(change => change.propertyName === propertyName) as ConfigChange
```
nawatts: I think we should update ESLint configuration to allow non-null assertions and use them in cases like this. I think that would make it clearer that this line is assuming something is not null vs casting another type to ConfigChange. 

https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-

ui-admin/src/portal/publish/diffComponents.tsx
```diff
@@ -24,77 +24,169 @@ export const VersionChangeView = ({ record }: {record: VersionedEntityChange}) =
...
+      ])
+    }
+    if (!selected && selectionIndex >= 0) {
+      const newArr = [...selectedChanges]
+      newArr.splice(selectionIndex, 1)
+      updateSelectedChanges(newArr)
```
nawatts, line 49: Not requesting this change since it's more of a style preference, but if you want to do this without the intermediate variable and the in-place update, this could also be written as:

```suggestion
      updateSelectedChanges([
        ...selectedChanges.slice(0, selectionIndex),
        ...selectedChanges.slice(selectionIndex + 1)
     ])
```

ui-admin/src/portal/publish/diffComponents.tsx
```diff
@@ -24,77 +24,169 @@ export const VersionChangeView = ({ record }: {record: VersionedEntityChange}) =
...
+export const ConfigChangeListView = <T extends Configable>
+  ({ configChangeList, renderItemSummary, selectedChanges, setSelectedChanges }:
+                                            ConfigChangeListViewProps<T>) => {
   if (!configChangeList.addedItems.length &&
     !configChangeList.removedItems.length && !configChangeList.changedItems.length) {
     return <span className="fst-italic text-muted">no changes</span>
   }
+
+  /** returns a new array with 'item' inserted or removed */
+  const makeModifiedArray = <R, >(array: R[], item: R, matchIndex: number, isAdd: boolean): R[] => {
```
nawatts: Using an options object to get named parameters might make this more readable. The `isAdd` parameter in particular is obscure when reading the function calls.
https://www.codereadability.com/what-are-javascript-options-objects/
https://alexkondov.com/should-you-pass-boolean-to-functions/

nawatts (approved):

ui-admin/src/portal/publish/PortalEnvDiffProvider.tsx
devonbush: I like using null instead of undefined in cases like this because it makes it a bit easier to see what's going on in the debugger (if I see something is `null` I know it's at least being set somewhere).  But it's just a small personal preference.  If React leans the other way, we should go with that

ui-admin/src/portal/publish/PortalEnvDiffView.tsx
devonbush: correct, a match will always be found (or if there isn't, it's because of a bug)

ui-admin/src/portal/publish/diffComponents.tsx
devonbush: yeah, I saw that syntax on your recent panel question-order modification PR.  I think you're right that it's probably a bit cleaner.

ui-admin/src/portal/publish/diffComponents.tsx
devonbush: good point.  Maybe I'll extract two helper functions `addToArray` and `removeFromArray`

ui-admin/src/portal/publish/PortalEnvDiffProvider.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const navigate = useNavigate()
+
+  const applyChanges = (changeSet: PortalEnvironmentChange) => {
+    Api.applyEnvChanges(portal.shortcode, portalEnv.environmentName, changeSet).then(result => {
+      Store.addNotification(successNotification(`${portalEnv.environmentName} environment updated`))
+      const updatedPortal = _cloneDeep(portal)
+      const envIndex = updatedPortal.portalEnvironments.findIndex(env => env.environmentName === result.environmentName)
+      updatedPortal.portalEnvironments[envIndex] = result
+      updatePortal(updatedPortal)
```
breilly2, line 33: Does this trace back to a useful implementation? I couldn't find one, and I suspect that the lack of one is causing a behavior I noticed. Immediately after cloning, if I click on the "Content" link for the Irb environment, the loaded content isn't updated. Refreshing the page fixes it. This was obvious to me when copying the first time and initializing the Irb environment, but it would be much less obvious if the changes were more subtle.

breilly2 (approved):
One thing I noticed that might be a bug, but generally does work based on my testing. Should either fix the bug or file a ticket for it.

ui-admin/src/portal/publish/PortalEnvDiffProvider.tsx
devonbush: correct, that behavior isn't implemented yet.  Was waiting on the admin tool design rearchitecture to figure our what kind of state management we want

PR #466, opened 2023-06-30
JN-450: Fixing survey rerender bug

devonbush:
there was a bug where moving quickly through a survey would sometimes cause fields to be erased while you were typing them.  This was caused by a rerender trigged by the `updateEnrollee` call after each in-progress answer save.  SurveyJS doesn't save answers to state until the field has lost focus, so if a rerender happens while a user is typing, the typing is lost.  This fixes the issue by making the updateEnrollee call not trigger a rerender.  
This also simplifies the useSurveyJsModel hook to avoid additional rerenders there.  While I couldn't write a full automated regression test for the issue, I did start adding tests to the participant survey view components.

TO TEST:
1. log in as consented@test.com to participant portal
2. start filling out basics
3. on the second page of the survey, fill out one address text field, then quickly start filling in another.  Confirm your typing is not suddenly erased in the second field
4. generally click around the survey and refresh the page use back button to confirm I haven't introduced any other regressions

### DIFF
--- a/ui-participant/src/hub/survey/SurveyView.test.tsx
+++ b/ui-participant/src/hub/survey/SurveyView.test.tsx
@@ -1,0 +1,39 @@
+import React from 'react'
+
+import { generateThreePageSurvey } from 'test-utils/test-survey-factory'
+import { PageNumberControl, useSurveyJSModel } from 'util/surveyJsUtils'
+import { render, screen } from '@testing-library/react'
+import { SurveyFooter } from './SurveyView'
+import { usePortalEnv } from 'providers/PortalProvider'
+import { Survey } from '@juniper/ui-core'
+
+jest.mock('providers/PortalProvider', () => ({ usePortalEnv: jest.fn() }))
+beforeEach(() => {
+  // @ts-expect-error TS doesn't realize this function is mocked
+  usePortalEnv.mockReturnValue({
+    portalEnv: {
+      environmentName: 'sandbox'
+    }
+  })
+})
+
+const FooterTestComponent = ({ pageNum, survey }: {pageNum: number, survey: Survey}) => {
+  const pager: PageNumberControl = { pageNumber: pageNum, updatePageNumber: () => 1 }
+  const { surveyModel } = useSurveyJSModel(survey, null,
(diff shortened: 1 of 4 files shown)


MatthewBemis (approved):

ui-participant/src/hub/survey/SurveyView.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { PageNumberControl, useSurveyJSModel } from 'util/surveyJsUtils'
+import { render, screen } from '@testing-library/react'
+import { SurveyFooter } from './SurveyView'
+import { usePortalEnv } from 'providers/PortalProvider'
+import { Survey } from '@juniper/ui-core'
+
+jest.mock('providers/PortalProvider', () => ({ usePortalEnv: jest.fn() }))
+beforeEach(() => {
+  // @ts-expect-error TS doesn't realize this function is mocked
+  usePortalEnv.mockReturnValue({
```
breilly2: I see other places in the code using `as jest.Mock` to handle this.
```suggestion
  (usePortalEnv as jest.Mock).mockReturnValue({
```

ui-participant/src/hub/survey/SurveyView.tsx
```diff
@@ -82,46 +82,48 @@ function RawSurveyView({ form, enrollee, resumableData, pager, studyShortcode, t
...
+        profile: response.profile
       }
-      return freshSurveyModel
+      /**
+       * CAREFUL -- we're updating the enrollee object so that if they navigate back to the dashboard, they'll
+       * see this survey as 'in progress' and capture any profile changes.
+       * However, we don't want to trigger a rerender, because that will wipe out any answers that the user has
+       * typed but are still in focus.  SurveyJS does not write answers to data/state until the question loses focus.
+       * So we use a 'updateWithoutRerender' flag on update Enrollee, this works since there are no currently
+       * visible components that use the enrollee object--otherwise they would not be refreshed
```
breilly2, line 119: Yikes! Definitely a worthwhile warning!

breilly2 (approved):
Thank you for fixing this!

PR #468, opened 2023-06-30
[JN-378] Add stub for validating form content

nawatts:
Currently, nothing validates that edited form content actually conforms to the FormContent type or other requirements (like question title shouldn't be an empty string).

This allows saving invalid form content and also invalidates assumptions made by the editor UI. For example, replacing form content in the JSON editor with an empty object crashes the form designer.

This adds a stub function for validation and wires it into the form editor.

### DIFF
--- a/ui-admin/src/forms/FormContentEditor.tsx
+++ b/ui-admin/src/forms/FormContentEditor.tsx
@@ -7,6 +7,7 @@ import { FormDesigner } from './FormDesigner'
 import { OnChangeFormContent } from './formEditorTypes'
 import { FormContentJsonEditor } from './FormContentJsonEditor'
 import { FormPreview } from './FormPreview'
+import { validateFormContent } from './formContentValidation'
 
 type FormContentEditorProps = {
   initialContent: string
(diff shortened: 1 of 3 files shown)


ui-admin/src/forms/FormContentEditor.tsx
```diff
@@ -43,7 +44,12 @@ export const FormContentEditor = (props: FormContentEditorProps) => {
             value={editedContent}
             onChange={newContent => {
               setEditedContent(newContent)
-              onChange(true, newContent)
+              try {
+                validateFormContent(newContent)
```
nawatts, line 48: We could possibly have two different validation functions to make this more efficient. Here, we at least know that `newContent` conforms to the FormContent type and we're validating things like non-empty question titles. Whereas in the JSON editor, the user could put in any value (the type is actually `unknown`) and we have to do many more checks.

MatthewBemis (approved):

devonbush (approved):
Since it's your last day on the team and this is a small change, and you did earn kudos this week, we'll waive the 'all frontend PRs must have a new test' requirement for this one :)

PR #469, opened 2023-06-30
[JN-444] Fix "Welcome to study" banner not appearing

MatthewBemis:
`hubUpdate` is not always defined when this page renders, so this adds a hook to update `displayedHubMessage` if/when it is eventually defined. I suspect this bug was also impacting other banners as well.

Screenshot 2023-06-30 at 12 56 31 PM

TO TEST:

1) Enroll in a study as a participant
2) You should now see the welcome banner display.

### DIFF
--- a/ui-participant/src/hub/HubPage.test.tsx
+++ b/ui-participant/src/hub/HubPage.test.tsx
@@ -1,0 +1,70 @@
+import { render, screen } from '@testing-library/react'
+import React from 'react'
+import HubPage from './HubPage'
+import { setupRouterTest } from '../test-utils/router-testing-utils'
+
+
+jest.mock('../providers/PortalProvider', () => {
+  return {
+    usePortalEnv: () => ({
+      portal: {
+        id: 'portal-id',
+        name: 'Test Portal',
+        shortcode: 'TESTPORTAL',
+        portalEnvironments: [],
+        portalStudies: [{
+          study: {
+            name: 'Test Study',
+            shortcode: 'STUDYSHORTCODE',
+            studyEnvironments: [{
+              id: 'test-study-env-id'
+            }]
+          }
+        }]
+      }
+    })
+  }
+})
+
+jest.mock('../providers/UserProvider', () => {
+  return {
+    useUser: () => ({
+      enrollees: [{
+        id: 'enrollee-id',
+        shortcode: 'ENSHORTCODE',
+        consented: false,
+        studyEnvironmentId: 'test-study-env-id',
+        participantTasks: [{
+          id: 'task-id',
+          blocksHub: true,
+          createdAt: 0,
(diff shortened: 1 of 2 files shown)


nawatts (approved):

devonbush (changes requested):
Good find for the cause.  Rather than adding a useEffect, I think it would be cleaner to replace the current `displayedHubMessage` useState with a boolean `showMessage` useState.  So then rather than clearing the message itself when the user dismisses the alert, we just toggle a boolean.  I think having the extra state for displayedHubMessage was a holdover from when there were actions on this page that could alter the message, rather than just show/hide it

All frontend changes now have to have at least one automated test.  Since we currently don't have any automated tests on HubPage, I'm fine if the test you write is just a more basic rendering test for the page as a whole

MatthewBemis:
@devonbush Makes sense, I appreciate the historical context. Switched to a boolean toggle in a13051c59d2f7c0f81fb3b598bf9b2c176eebaa2

I'll add some form of test coverage shortly.

ui-participant/src/hub/HubPage.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+          status: 'NEW',
+          studyEnvironmentId: 'test-study-env-id',
+          taskType: 'CONSENT',
+          targetName: 'Test Survey',
+          targetStableId: 'test_survey_id',
+          targetAssignedVersion: 0,
+          taskOrder: 0
+        }]
+      }]
+    })
```
devonbush, line 52: using the mockEnrollee function will help make this test more robust against future changes tot he enrollee schema.

```suggestion
    useUser: () => ({
      enrollees: [{
        ...mockEnrollee(),
        consented: false,
        studyEnvironmentId: 'test-study-env-id',
        participantTasks: [{
          id: 'task-id',
          blocksHub: true,
          createdAt: 0,
          enrolleeId: 'enrollee-id',
          portalParticipantUserId: 'portal-participant-user-id',
          status: 'NEW',
          studyEnvironmentId: 'test-study-env-id',
          taskType: 'CONSENT',
          targetName: 'Test Survey',
          targetStableId: 'test_survey_id',
          targetAssignedVersion: 0,
          taskOrder: 0
        }]
      }]
    })
```

devonbush (approved):
Nice, I like those provider mocks you set up -- we might want to extract those into utility functions as they get more used.  One more optional suggestion

ui-participant/src/hub/HubPage.test.tsx
MatthewBemis: It looks like this would require moving `mocking-utils` out of `ui-admin` and into `ui-core`. I think it's worth having shared test utils across the participant and admin UIs but from a quick glance it looks like it's too large of a refactor to have in this PR.

PR #470, opened 2023-06-30
JN-444 more initial screen issues

devonbush:
In looking at Matt's PR, I realized there was another problem with the initial hub screen -- we were flashing a list of studies before the proper activity list immediately after people joined, because of a delay between when we routed to the hub, and when we updated the state of the enrollee.  This adds a new option to the `updateEnrollee` function, 'afterFn' which allows the caller to specify a function to be executed only after the enrollee update is complete.  so for both sign up and survey completion, we can delay navigating back to the hub until we've finished handling all the state updates.

TO TEST:
1. go to participant site 
2. Put a breakpoint on line 24 of HubPage.tsx (the return statement)
3. sign up a new enrollee
4. when the breakpoint is hit, continue, and then look at the hub, it should show the list of surveys/consent.   It should not show "List of studies to join".

### DIFF
--- a/ui-participant/src/providers/UserProvider.tsx
+++ b/ui-participant/src/providers/UserProvider.tsx
@@ -14,16 +14,13 @@ const anonymousUser: User = {
   username: 'anonymous'
 }
 
-type UpdateEnrolleeOptions = {
-  afterFn?: () => void // function to call after the update is complete (such as to do a navigate)
-}
 export type UserContextT = {
   user: User,
   enrollees: Enrollee[],  // this data is included to speed initial hub rendering.  it is NOT kept current
   loginUser: (result: LoginResult, accessToken: string) => void,
   loginUserInternal: (result: LoginResult) => void,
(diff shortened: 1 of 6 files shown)


ui-participant/src/providers/UserProvider.tsx
```diff
@@ -99,6 +103,9 @@ export default function UserProvider({ children }: { children: React.ReactNode }
         enrollees: updatedEnrollees
       }
     })
+    if (afterFn) {
+      window.setTimeout(afterFn, 0)
+    }
```
nawatts: To make the interface of this function a bit more conventional, I'd suggest returning a Promise instead of taking a callback.

```suggestion
    return new Promise(resolve => {
      window.setTimeout(resolve, 0)
    })
```

And call it like:
```ts
updateEnrollee(...).then(() => {
  ...
})
```

nawatts (approved):

MatthewBemis (approved):

PR #472, opened 2023-07-06
[JN-451] Add ability to create a new survey

MatthewBemis:
* Adds a new endpoint to configure a survey for a study environment
* Adds new modal for creating a survey in the study admin UI: 
Screenshot 2023-07-06 at 5 07 03 PM

Note that this is the bare minimum needed to unblock survey creation. There are other bits of configuration that we will eventually want to surface to the user: whether or not the survey is required, what the eligibility rules are, survey order, etc.

TO TEST:

1. Boot up Admin API + Admin UI
2. Go to the survey list in the study content manager and add a new survey
3. Verify that a blank survey gets created and it takes you to the survey builder

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/SurveyController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/SurveyController.java
@@ -4,6 +4,7 @@ import bio.terra.pearl.api.admin.api.SurveyApi;
 import bio.terra.pearl.api.admin.service.AuthUtilService;
 import bio.terra.pearl.api.admin.service.forms.SurveyExtService;
 import bio.terra.pearl.core.model.admin.AdminUser;
+import bio.terra.pearl.core.model.survey.StudyEnvironmentSurvey;
 import bio.terra.pearl.core.model.survey.Survey;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import javax.servlet.http.HttpServletRequest;
@@ -36,14 +37,12 @@ public class SurveyController implements SurveyApi {
   }
 
   @Override
-  public ResponseEntity<Object> create(String portalShortcode, String stableId, Object body) {
+  public ResponseEntity<Object> create(String portalShortcode, String studyShortCode, Object body) {
     AdminUser adminUser = requestService.requireAdminUser(request);
 
     Survey survey = objectMapper.convertValue(body, Survey.class);
-    if (!stableId.equals(survey.getStableId())) {
-      throw new IllegalArgumentException("survey parameters don't match");
-    }
-    Survey savedSurvey = surveyExtService.create(portalShortcode, survey, adminUser);
+    StudyEnvironmentSurvey savedSurvey =
+        surveyExtService.create(portalShortcode, studyShortCode, survey, adminUser);
     return ResponseEntity.ok(savedSurvey);
   }
 
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/SurveyController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/SurveyController.java
@@ -4,7 +4,6 @@ import bio.terra.pearl.api.admin.api.SurveyApi;
 import bio.terra.pearl.api.admin.service.AuthUtilService;
 import bio.terra.pearl.api.admin.service.forms.SurveyExtService;
 import bio.terra.pearl.core.model.admin.AdminUser;
-import bio.terra.pearl.core.model.survey.StudyEnvironmentSurvey;
 import bio.terra.pearl.core.model.survey.Survey;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import javax.servlet.http.HttpServletRequest;
@@ -37,12 +36,11 @@ public class SurveyController implements SurveyApi {
   }
 
   @Override
-  public ResponseEntity<Object> create(String portalShortcode, String studyShortCode, Object body) {
+  public ResponseEntity<Object> create(String portalShortcode, Object body) {
     AdminUser adminUser = requestService.requireAdminUser(request);
 
     Survey survey = objectMapper.convertValue(body, Survey.class);
-    StudyEnvironmentSurvey savedSurvey =
-        surveyExtService.create(portalShortcode, studyShortCode, survey, adminUser);
+    Survey savedSurvey = surveyExtService.create(portalShortcode, survey, adminUser);
     return ResponseEntity.ok(savedSurvey);
   }
 
--- a/api-admin/src/main/resources/api/openapi.yml
+++ b/api-admin/src/main/resources/api/openapi.yml
@@ -129,14 +129,13 @@ paths:
           content: { application/json: { schema: {type: object} } }
         '500':
           $ref: '#/components/responses/ServerError'
-  /api/portals/v1/{portalShortcode}/studies/{studyShortcode}/surveys:
+  /api/portals/v1/{portalShortcode}/surveys:
     post:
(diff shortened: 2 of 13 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/SurveyController.java
```diff
@@ -36,13 +36,10 @@ public ResponseEntity<Object> get(String portalShortcode, String stableId, Integ
...
 
   @Override
-  public ResponseEntity<Object> create(String portalShortcode, String stableId, Object body) {
+  public ResponseEntity<Object> create(String portalShortcode, Object body) {
     AdminUser adminUser = requestService.requireAdminUser(request);
 
     Survey survey = objectMapper.convertValue(body, Survey.class);
-    if (!stableId.equals(survey.getStableId())) {
-      throw new IllegalArgumentException("survey parameters don't match");
-    }
```
MatthewBemis, line 45: I don't think there's any reason to have `stableId` as a path parameter in the endpoint given that the survey is fully defined in the request body.

ui-admin/src/study/surveys/CreateSurveyModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+          allowParticipantReedit: true,
+          allowParticipantStart: true,
+          id: '',
+          prepopulate: false,
+          recurrenceIntervalDays: 0,
+          recur: false,
+          studyEnvironmentId: studyEnvContext.currentEnv.id,
+          survey: createdSurvey,
+          surveyId: createdSurvey.id,
+          surveyOrder: studyEnvContext.currentEnv.configuredSurveys.length
```
MatthewBemis, line 47: As we continue to build out the survey builder, these options should eventually be exposed to the study admins for customization.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/SurveyController.java
devonbush: yeah, this is another case where I might have gone overboard on url params -- the nice thing about having them is that it gives us extra logging for free, but I think you're right that it's overkill here

ui-admin/src/portal/PortalProvider.tsx
```diff
@@ -7,13 +7,15 @@ import { NavBreadcrumb } from 'navbar/AdminNavbar'
 
 export type PortalContextT = {
   updatePortal: (portal: Portal) => void
+  reloadPortal: (shortcode: string, onReload: () => void) => void,
```
devonbush: This should return a promise rather than take a callback for easier use
```suggestion
  reloadPortal: (shortcode: string) => Promise<Portal>,
```

ui-admin/src/portal/PortalProvider.tsx
```diff
@@ -54,21 +57,28 @@ function RawPortalProvider({ shortcode, children }:
     setPortalState(updatedPortal)
   }
 
-  useEffect(() => {
+  /** grabs the latest from the server and updates the portal object */
+  function reloadPortal(shortcode: string, onReload: () => void) {
     Api.getPortal(shortcode).then(result => {
```
devonbush:
```suggestion
  function reloadPortal(shortcode: string): Promise<Portal> {
    return Api.getPortal(shortcode).then(result => {
```

devonbush (approved):
Looks good, but definitely change the reload method to return a promise rather than using callbacks

devonbush:
Also, consider merging in latest development, and adding an InfoPopup next to the stableId field with help text about what the stableId is

devonbush:
when I click create survey, it takes me to the right page, but I suspet there's a timing issue.  It shows 
image
if I reload the page, then it's fine

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -129,14 +129,13 @@ paths:
           content: { application/json: { schema: {type: object} } }
         '500':
           $ref: '#/components/responses/ServerError'
-  /api/portals/v1/{portalShortcode}/surveys/{stableId}:
+  /api/portals/v1/{portalShortcode}/surveys:
     post:
```
breilly2, line 133: I think this is appropriate. The other option to consider would be keeping `{stableId}` and using `PUT`, which would be nicely RESTful.

ui-admin/src/study/surveys/CreateSurveyModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const createdSurvey = await Api.createNewSurvey(studyEnvContext.portal.shortcode,
+      {
+        createdAt: 0, id: '', lastUpdatedAt: 0, version: 1,
+        content: '{"pages":[]}', name: surveyName, stableId: surveyStableId
+      }).catch(e =>
+      Store.addNotification(failureNotification(`Error creating survey: ${e.message}`))
+    ) as VersionedForm
+
+    try {
+      await Api.createConfiguredSurvey(studyEnvContext.portal.shortcode,
```
breilly2, line 33: Should this be a separate API call or should it be combined with `createNewSurvey`? What happens if `createNewSurvey` succeeds but `createConfiguredSurvey` fails? I suspect it would put the application in a weird state that may not be easily represented to the user through the admin UI, which leads me to think that this should be a single transactional operation.

breilly2 (approved):
I'm approving as-is, but I think it's worth considering whether `createNewSurvey` and `createConfiguredSurvey` should be merged into a single API call (recognizing that some day we may want to allow creating a survey and configuring it for multiple studies, but that will require more UI/UX than I feel we need today).

ui-admin/src/study/surveys/CreateSurveyModal.tsx
MatthewBemis: If `createConfiguredSurvey` fails, it will leave a blank survey that lives at the portal-level and isn't visible to the user.

I had originally implemented this as one single endpoint in an earlier commit, but after reading a lot of the existing code it felt a bit messy to me. I view the act of creating a survey and configuring a survey as two distinct operations. Surveys are created at the portal-level, presumably so they can be re-used by other studies in the portal. Configuring a survey is per-study environment. There were 3 options:

1. Combine the operations into one larger endpoint, easiest on the client but muddies concepts in the service code.
2. Separate the endpoints, but combine the operations in the UI so the user doesn't know the difference. Puts more burden on the client.
3. Expose to the user in the UI that these are two separate actions. Puts more burden on the user and requires more UX thought: how would we expose creating surveys at the portal-level? Is that something that we ultimately want to do to allow for survey re-use?

I went with 2 because it felt like the right balance and I tend to favor writing more precise endpoints, and it also felt like the easiest to reason about. The downside is that the error handling in the UI is trickier. I also like that it leaves the door open for us to switch to the third option in the future.

MatthewBemis:
@devonbush are you testing with the latest commit? There was a much earlier commit on this branch where that would happen but it should be fixed

ui-admin/src/portal/PortalProvider.tsx
MatthewBemis: Got rid of callback in 96d4f97f

ui-admin/src/study/surveys/CreateSurveyModal.tsx
breilly2: Agreed that we may need #3 in the future. The downside of straddling #1 and #3 is that, if it ends up happening to someone that the operation only partially completes, it will require manual intervention to fix; they can't try again because the stable ID is already used, and they don't have a UI to do just the 2nd part themselves. The risk is probably small and it may never happen, but small risks like this will build up and incur real costs if we have too many of them. I think it's worth at least being aware of and deliberate about the accumulation of these small risks.

PR #473, opened 2023-07-10
[JN-455] basic participant notes

devonbush:
This adds basic text note creation to the admin tool.  Notes are not yet editable.  Later work might include:
1. markdown for notes
2. editing or deleting notes (and maintaining audit trail)
3. Refining the "Profile & Notes" enrollee view screen to be more compact
4. UX for adding notes associated with kits (the backend for supporting that is in place, but no UX)

Notable additions support of this feature include:
1.  adding a new provider to load admin users so that usernames could be associated with notes. Eventually, there'll be a whole bunch of "background stuff to load on sign-in" like other usernames, kit types, etc... and we'll consolidate that into a single endpoint
2. Added a "TimeShiftable" interface for entities that can be populated in the past.

TO TEST:
1. redeploy AdminApiApp, repopulate ourhealth
2. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHSALK. 
3. confirm you see two notes at the bottom of the page
image
4. Add a new note
5. reload page, confirm note persists


api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeExtService.java
```diff
@@ -79,34 +78,18 @@ public WithdrawnEnrollee withdrawEnrollee(AdminUser user, String enroleeShortcod
...
-    Enrollee enrollee = authAdminUserToEnrollee(adminUser, enrolleeShortcode);
+    Enrollee enrollee = authUtilService.authAdminUserToEnrollee(adminUser, enrolleeShortcode);
     return kitRequestService.requestKit(adminUser, enrollee, kitTypeName);
   }
 
   public Collection<KitRequest> getKitRequests(AdminUser adminUser, String enrolleeShortcode) {
-    Enrollee enrollee = authAdminUserToEnrollee(adminUser, enrolleeShortcode);
+    Enrollee enrollee = authUtilService.authAdminUserToEnrollee(adminUser, enrolleeShortcode);
     return kitRequestService.getKitRequests(adminUser, enrollee);
   }
```
devonbush, line 94: this enrolleeService is starting to get a little too kitchen-sink.   Pulling the auth method into main authUtils makes it easier to make other services related to enrollee operations

core/src/main/java/bio/terra/pearl/core/model/participant/Enrollee.java
```diff
@@ -39,5 +37,7 @@ public class Enrollee extends BaseEntity {
     @Builder.Default
     private Set<ParticipantTask> participantTasks = new HashSet<>();
     @Builder.Default
-    private Set<KitRequest> kitRequests = new HashSet<>();
+    private List<ParticipantNote> participantNotes = new ArrayList<>();
```
devonbush, line 40: I originally made most collections Sets because Sets play much nicer with Hibernate.  I think we should start moving these Collections on our models to Lists since that's what our DB calls return, and will allow us to track and return sorted collections.

breilly2: Yeah, this is always an ORM balancing act. `Set` makes sense in the object model to represent bags of distinct objects (no duplicates). However, database queries return rows in an order, even if that order is undefined (or even non-deterministic across executions of the same query). Therefore, using `List` is accurate, but sometimes implies too much; yes, the items are in an order, but that order may not have any significance. I try to avoid using `List` when the order is not deliberate and significant.

If we're going to use `List`, I think we should make an effort to fetch the elements in a consistent order and document it.

devonbush: Agree about consistent ordering.  What I also like about List is that it reminds us that we always send them to the client as a list, and so they're going to be put in *an* order, so we might as well make it one we control

ui-admin/src/study/participants/enrolleeView/ParticipantNoteView.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+type ParticipantNoteViewProps = {
+  enrollee: Enrollee,
+  note: ParticipantNote,
+  currentEnvPath: string,
+  users: AdminUser[]
+}
+/** renders a single note, with the admin user name and a link to the kit request, if it exists */
+export function ParticipantNoteView({ enrollee, note, users, currentEnvPath }: ParticipantNoteViewProps) {
+  const matchedUser = users.find(user => user.id === note.creatingAdminUserId)
```
breilly2, line 18: Could we resolve usernames when fetching the notes from the database and avoid the need for the `AdminUserProvider`?

breilly2 (approved):
Looks good. I do wonder if we need the complexity of `AdminUserProvider` as opposed to just resolving the username when fetching the notes.

ui-admin/src/study/participants/enrolleeView/ParticipantNoteView.tsx
devonbush: yup, that's one valid approach (I know it's what you did with kit types).  The drawback is that it couples the API backend more tightly to the frontend, and can result in a lot of duplicate fetching (e.g. we might need to fetch admin users along with kits, notes, tasks, etc...  So that's why I proposed we move instead to an architecture where there's an initial background load of all the supporting collections (users, kit types, etc...), and then the frontend can decide when it wants to display additional detail about something.  Both work, and we can wait until we've got a couple more cases like this to make a decision

MatthewBemis (approved):

ui-admin/src/study/participants/enrolleeView/ParticipantNoteView.tsx
breilly2: I know this is already merged, but just to clarify my thoughts about this...

Resolving server-side does couple the controller (and perhaps service, depending on division of responsibilities) to the UI, but controllers exist to serve interfaces so I don't think that's inherently bad. Besides, resolving client-side couples the UI to the database model. Pre-fetching can also be inefficient (the client might fetch data it never needs), and it feels to me a bit like premature optimization. In general, I'm more worried about identifying and fixing subtle bugs with cached, non-refreshed state in the client than I am in finding ways to optimize queries when we need to improve performance.

PR #475, opened 2023-07-11
JN-470 surfacing frontend auth errors

devonbush:
image

This standardizes error messaging for most (but not all) participant API calls.  the UX isn't great, but it's a lot better than the generic "an error has occurred" messages we were doing before.

TO TEST:
1. login to the participant UI.
2. Simulate that you have logged out/user not found.  The easiest way I found to do this was editing lines 37-39 of RequestUtilService to always throw the exception.  
3. Attempt to do an action
4. confirm you see an alert like the screenshot above

### DIFF
--- a/ui-participant/src/api/api.tsx
+++ b/ui-participant/src/api/api.tsx
@@ -155,7 +155,7 @@ export default {
     if (response.ok) {
       return obj
     }
-    return Promise.reject(response)
+    return Promise.reject(obj)
   },
 
   async getConfig(): Promise<Config> {
@@ -327,7 +327,10 @@ export default {
     }
     const result = await fetch(url, {
       method: 'POST',
-      headers: this.getInitHeaders(),
+      headers: {
+        ...this.getInitHeaders(),
+        Authorization: 'foo'
+      },
       body: JSON.stringify(response)
     })
     return await this.processJsonResponse(result)
--- a/ui-participant/src/api/api.tsx
+++ b/ui-participant/src/api/api.tsx
@@ -8,7 +8,7 @@ import {
   StudyEnvironmentSurvey,
   SurveyResponse
 } from '@juniper/ui-core'
-import {defaultApiErrorHandle} from "../util/error-utils";
+import { defaultApiErrorHandle } from 'util/error-utils'
 
 export type {
   Answer,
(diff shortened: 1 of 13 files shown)


ui-participant/src/api/api.tsx
```diff
@@ -150,12 +151,16 @@ export default {
...
     const obj = await response.json()
     if (response.ok) {
       return obj
     }
-    return Promise.reject(response)
+    if (opts.alertErrors) {
+      defaultApiErrorHandle(obj)
+    }
+    return Promise.reject(obj)
   },
```
breilly2, line 164: When adding additional options, all will have to be specified, but you may want to only specify one and let the others keep their defaults. Something like this will allow for fully optional opts with the correct defaults:
```suggestion
  async processJsonResponse(response: Response, opts?: { alertErrors?: boolean }) {
    const defaultOpts = { alertErrors: true }
    const { alertErrors } = { ...defaultOpts, ...opts }
    
    const obj = await response.json()
    if (response.ok) {
      return obj
    }
    if (alertErrors) {
      defaultApiErrorHandle(obj)
    }
    return Promise.reject(obj)
  },
```

ui-participant/src/hub/survey/SurveyView.tsx
```diff
@@ -101,9 +102,11 @@ function RawSurveyView({ form, enrollee, resumableData, pager, studyShortcode, t
       surveyId: form.id,
       complete: activeResponse?.complete ?? false
     } as SurveyResponse
+    // only log & alert if this is the first autosave problem to avoid spamming logs & alerts
+    const alertErrors =  !lastAutoSaveErrored.current
     Api.submitSurveyResponse({
       studyShortcode, stableId: form.stableId, enrolleeShortcode: enrollee.shortcode,
-      version: form.version, response: responseDto, taskId
+      version: form.version, response: responseDto, taskId, alertErrors
```
breilly2, line 109: Oof... yeah, that could get ugly. Good call!

breilly2 (approved):

MatthewBemis (approved):

ui-participant/src/api/api.tsx
devonbush: cool, will use that once this inevitably grows :)

PR #477, opened 2023-07-12
[JN-417] Implement basic kit management UI

breilly2:
Also adds random kit requests to the bulk test user creation feature.

To test:
1. Wipe and repopulate your database (or try without, but there might be issues with previously-populated data... which I don't like so I'd actually like to know if you're able to test this without wiping).
2. Start the admin UI and go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits
3. Play around with the filtering, sorting, and column visibility.
4. Click the "Kits" tab to see the one default populated kit request.
5. Click back over to the "Enrollees" tab, select the `OHSALK` row, click "Send sample collection kit", and click the "Request Kit" button.
6. Click back over to the "Kits" tab to see that there are now 2 kit requests for `OHSALK`.
7. In another window, go to http://localhost:8080/swagger-ui.html#/populate/bulkPopulateEnrollees and fill in `ourhealth`/`sandbox`/`ourheart`/`20` to generate 20 random users.
8. Back in the admin UI, click back over to the "Enrollees" tab to see the generated enrollees. All should have the basics and cardiometabolic medical history surveys completed. A few may oddly not be consented (5% chance). Many will have an existing kit request (80% chance).
9. Click back over to the "Kits" tab to see the generated kits. Most will be "Prepared", fewer will be "Sent", and fewer still will be "Returned".
10. Try out the status filter.

image
image
image


core/src/main/java/bio/terra/pearl/core/model/kit/KitRequestStatus.java
```diff
@@ -5,7 +5,7 @@
 /**
  * High-level, Juniper-centric kit status. Only needs to be specific enough for Juniper to decide how to interact with
  * Pepper. This is not intended to mirror all possible states in Pepper. For that, see
- * {@link bio.terra.pearl.core.service.kit.PepperDSMKitStatus#currentStatus}.
+ * {@link bio.terra.pearl.core.service.kit.PepperKitStatus#currentStatus}.
```
breilly2, line 8: This is just a rename refactor. In retrospect, I wish that I had left this change out of this PR. But now that you know about it, it should be pretty easy to filter out.

populate/src/main/resources/seed/portals/ourhealth/participants/jsalk.json
```diff
@@ -9,10 +9,10 @@
     "sexAtBirth": "male",
     "mailingAddress": {
       "street1": "32 Elm Street",
-      "city": "New York",
-      "state": "NY",
+      "city": "Cambridge",
+      "state": "MA",
       "country": "US",
-      "postalCode": "10005"
+      "postalCode": "02142"
```
breilly2, line 15: Updating `JSALK` to have a real, valid mailing address. This isn't strictly necessary for this PR, but will be helpful for testing live integration with Pepper since Pepper will reject invalid addresses.

Hopefully this change won't break anything (other than a test that I already updated). However, it won't update the demo server database unless we re-populate which we probably don't want to do wholesale, so I'm not sure of the best way to handle that.

populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/enrollees/seed.json
```diff
@@ -15,6 +15,32 @@
       {"questionStableId":"qualified", "booleanValue": true}
     ]
   },
+  "participantTaskDtos": [
```
breilly2, line 18: Bulk users had survey responses, but no corresponding tasks.

ui-admin/src/study/kits/KitEnrolleeSelection.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const onSubmit = async (kitType: string) => {
+    const enrolleesSelected = Object.keys(rowSelection)
+      .filter(key => rowSelection[key])
+      .map(key => enrollees[parseInt(key)].shortcode)
+
+    // This iteration should be happening server-side: JN-460
+    for (const shortcode of enrolleesSelected) {
+      await Api.createKitRequest(
+        portal.shortcode, study.shortcode, currentEnv.environmentName, shortcode, kitType)
+    }
```
breilly2, line 69: Matt, this client-side orchestration goes against what I was suggesting on your recent PR, so I created JN-460 to clean that up and move it server-side. At least in this case, users would be able to see which kits were created and which weren't and could re-submit any that failed; users could recover without us having to do any support work.

ui-admin/src/study/kits/KitList.tsx
```diff
@@ -0,0 +1,133 @@
...
+import LoadingSpinner from 'util/LoadingSpinner'
+import { basicTableLayout } from 'util/tableUtils'
+import { instantToDateString, isoToInstant } from 'util/timeUtils'
+
+const pepperStatusToHumanStatus = (dsmStatus?: string): string => {
+  const statusMap: Record<string, string> = {
+    'CREATED': 'Created',
+    'LABELED': 'Prepared',
+    'SCANNED': 'Sent',
+    'RECEIVED': 'Returned'
```
breilly2, line 20: On the left are my guesses as to what Pepper might send for `currentStatus` (unless we decide to do away with that field since it's just derived from the individual activity dates). On the right are my guesses as to what this really means from the point of view of a study admin. For example, `SCANNED` means that the kit has been scanned on the way out... effectively having been shipped to the participant while `RECEIVED` means that GP has gotten the kit back, i.e. "returned" from the participant (as opposed to "received" by the participant).

breilly2: Also, I realize this is not very DRY. This UI is definitely going to be refined in the near future which could affect the de-dupe efforts, so I'm leaving this as-is for now.

populate/src/main/resources/seed/portals/ourhealth/participants/jsalk.json
devonbush: there's a populate endpoint that accepts an individual enrollee file, so you could just use that to repopulate just jsalk -- I'll be doing the same with participant notes

populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/enrollees/jsalk.json
```diff
@@ -86,11 +87,11 @@
         "returnTrackingNumber": "",
         "errors": [
```
devonbush: what's the reason for the empty error object?

populate/src/main/java/bio/terra/pearl/populate/service/EnrolleePopulator.java
```diff
@@ -413,13 +419,60 @@ public void bulkPopulateEnrollees(String portalShortcode, EnvironmentName envNam
                 popDto.setSubmittedHoursAgo(PopulateUtils.randomInteger(0, 480)); //add some jitter to when they joined, to make graphs/views more interesting
                 popDto.setShortcode(PopulateUtils.randomShortcode(""));
```
devonbush, line 424: this is a long section, let's pull into helper function "populateKitRequests(popDto)" or somesuch

populate/src/main/java/bio/terra/pearl/populate/service/EnrolleePopulator.java
```diff
@@ -413,13 +419,60 @@ public void bulkPopulateEnrollees(String portalShortcode, EnvironmentName envNam
...
+                recent = recent.minus(PopulateUtils.randomInteger(1, 72), HOURS);
+                pepperStatus.setReceiveDate(formatter.format(recent));
+            case "SCANNED":
+                recent = recent.minus(PopulateUtils.randomInteger(10, 30), DAYS);
+                pepperStatus.setScanDate(formatter.format(recent));
+            case "LABELED":
+                recent = recent.minus(PopulateUtils.randomInteger(5, 7), DAYS);
+                pepperStatus.setLabelDate(formatter.format(recent));
+            case "CREATED":
+                // Unfortunately, we can't do much here because we don't allow modifying createdAt
```
devonbush: IMPORTANT: good news -- we do allow that in populate contexts!  Add a method to `EnrolleePopulateDao` that directly changes the date.  See, e.g. `changeParticipantNoteTime`

core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeDao.java
```diff
@@ -117,6 +117,18 @@ public Enrollee loadForAdminView(Enrollee enrollee) {
         return enrollee;
     }
 
+    public List<Enrollee> findForKitManagement(UUID studyEnvironmentId) {
```
devonbush, line 141: IMPORTANT: this is going to hit performance constraints quickly, since it's doing 2 sequential queries per enrollee, and is unpaged.  This can be more performant and more modular if you create a method 
``` 
public void attachTasksToEnrollees(List<Enrollee> enrollees) {
   List<UUID> enrolleeIds = enrolles.map ...
   List<ParticipantTask> tasks = participantTaskDao.findByEnrolleeIds()
   Map<UUID, enrollee> enrolleeMap = <<map enrollees by their ids>>
   for (task) {
       enrolleeMap(task.getEnrolleeId).participantTasks.add(task);
    }
}
```
and then do the same for kitRequests.
Then the service method can just call the findByStudy, attachKits, and attachTasks methods, so the DAO doesn't have to be aware of what our Kit UI expects

core/src/main/java/bio/terra/pearl/core/dao/kit/KitRequestDao.java
```diff
@@ -28,15 +34,24 @@ public List<KitRequest> findByEnrollee(UUID enrolleeId) {
      */
     public List<KitRequest> findIncompleteKits(UUID studyEnvironmentId) {
```
devonbush: IMPORTANT: the DAO shouldn't need to be aware of business-level concerns like what statuses are incomplete.  You've already written the query to be general-purpose, so let's just make this "findKitsByStatuses" and let the service layer handle passing the right statuses

core/src/main/java/bio/terra/pearl/core/dao/kit/KitRequestDao.java
```diff
@@ -15,6 +15,12 @@ public KitRequestDao(Jdbi jdbi) {
         super(jdbi);
     }
 
+    private final String BASE_QUERY_BY_STUDY =
+            " select " + prefixedGetQueryColumns("kit") + " from " + tableName + " kit "
+            + " join enrollee on kit.enrollee_id = enrollee.id "
+            + " join study_environment on enrollee.study_environment_id = study_environment.id "
+            + " where study_environment.id = :studyEnvironmentId ";
+
```
devonbush, line 23: appreciate that you pulled this out.  I suspect we'll only have more of this sort of query-combining-and-sharing, and so I've got in the back of my mind that we'll eventually want to adopt a formal tool (e.g. JOOQ) that lets us compose sql queries more robustly

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -871,6 +871,37 @@ paths:
                 format: binary
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/portals/v1/{portalShortcode}/studies/{studyShortcode}/env/{envName}/enrolleesForKitManagement:
```
devonbush: the "forKitManagement" suffix seems odd.  
eventually, we'll want a convention for specifying which subcollections should be included with these sorts of queries.  e.g. 
`api/portals/.../enrollees?includes=KitRequest;ParticipantTask. `

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -871,6 +871,37 @@ paths:
...
+        - { name: portalShortcode, in: path, required: true, schema: { type: string } }
+        - { name: studyShortcode, in: path, required: true, schema: { type: string } }
+        - { name: envName, in: path, required: true, schema: { type: string } }
+      responses:
+        '200':
+          description: enrollee list
+          content: { application/json: { schema: { type: object } } }
+        '500':
+          $ref: '#/components/responses/ServerError'
+  /api/portals/v1/{portalShortcode}/studies/{studyShortcode}/env/{envName}/kitsForKitManagement:
```
devonbush: let's just call this one `kits`

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -871,6 +871,37 @@ paths:
                 format: binary
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/portals/v1/{portalShortcode}/studies/{studyShortcode}/env/{envName}/enrolleesForKitManagement:
+    get:
+      summary: Gets a list of enrollees with tasks and kit requests
+      tags: [ kit ]
```
devonbush: I'd lean towards putting this in the Enrollee controller, since it returns enrollees and doesn't have any special kit-filtering logic.  But maybe this endpoint will later filter by kit properties?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/kit/KitController.java
```diff
@@ -0,0 +1,55 @@
...
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+
+    var enrollees =
+        enrolleeExtService.findForKitManagement(
+            adminUser, portalShortcode, studyShortcode, environmentName);
+    return ResponseEntity.ok(enrollees);
+  }
+
+  @Override
+  public ResponseEntity<Object> listKitsForKitManagement(
```
devonbush:
```suggestion
  public ResponseEntity<Object> kitsByStudyEnvironment(
```

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/kit/KitController.java
```diff
@@ -0,0 +1,55 @@
...
+      String portalShortcode, String studyShortcode, String envName) {
+    AdminUser adminUser = authUtilService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+
+    var kits =
+        kitExtService.getKitRequestsForStudyEnvironment(
```
devonbush:
```suggestion
        kitExtService.getKitRequestsByStudyEnvironment(
```

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/kit/KitController.java
devonbush: we generally try to use the naming conventions `findFor<purpose>`   and `findBy<criteria>`. so this is a 'by' instead of a 'for'

api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
```diff
@@ -0,0 +1,40 @@
...
+  public KitExtService(
+      AuthUtilService authUtilService,
+      KitRequestService kitRequestService,
+      StudyEnvironmentService studyEnvironmentService) {
+    this.authUtilService = authUtilService;
+    this.kitRequestService = kitRequestService;
+    this.studyEnvironmentService = studyEnvironmentService;
+  }
+
+  public Collection<KitRequest> getKitRequestsForStudyEnvironment(
```
devonbush:
```suggestion
  public Collection<KitRequest> getKitRequestsByStudyEnvironment(
```

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -98,27 +102,49 @@ public static PepperKitAddress makePepperKitAddress(Profile profile) {
...
+     * Fetch all sample kits for a study environment.
+     */
+    public Collection<KitRequest> getSampleKitsForStudyEnvironment(StudyEnvironment studyEnvironment) {
+        var allKitTypes = kitTypeDao.findAll();
+        var kitTypeMap = allKitTypes.stream().collect(Collectors.toMap(KitType::getId, Function.identity()));
+        var kits = dao.findByStudyEnvironment(studyEnvironment.getId());
+        kits.forEach(kit -> {
+            kit.setKitType(kitTypeMap.get(kit.getKitTypeId()));
+            try {
+                kit.setPepperStatus(objectMapper.readValue(kit.getDsmStatus(), PepperKitStatus.class));
```
devonbush: We generally try to avoid parsing json strings server-side, since they're going to just get re-stringified anyway on the way to the client--do we need to parse them here, or can we move that logic to the client?

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -98,27 +102,49 @@ public static PepperKitAddress makePepperKitAddress(Profile profile) {
...
      */
-    public Collection<KitRequest> getKitRequests(AdminUser adminUser, Enrollee enrollee) {
+    public Collection<KitRequest> getKitRequests(Enrollee enrollee) {
         return dao.findByEnrollee(enrollee.getId());
     }
 
+    /**
+     * Fetch all sample kits for a study environment.
+     */
+    public Collection<KitRequest> getSampleKitsForStudyEnvironment(StudyEnvironment studyEnvironment) {
```
devonbush:
```suggestion
    public Collection<KitRequest> getSampleKitsByStudyEnvironment(StudyEnvironment studyEnvironment) {
```

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/kit/KitController.java
```diff
@@ -0,0 +1,55 @@
...
+      KitExtService kitExtService,
+      HttpServletRequest request) {
+    this.authUtilService = authUtilService;
+    this.enrolleeExtService = enrolleeExtService;
+    this.kitExtService = kitExtService;
+    this.request = request;
+  }
+
+  @Override
+  public ResponseEntity<Object> listForKitManagement(
```
devonbush:
```suggestion
  public ResponseEntity<Object> enrolleesWithKits(
```

ui-admin/src/study/kits/KitList.tsx
```diff
@@ -0,0 +1,133 @@
...
+      portal.shortcode, study.shortcode, currentEnv.environmentName)
+    setKits(kits)
+    setIsLoading(false)
+  }
+
+  useEffect(() => {
+    loadKits()
+  }, [])
+
+  // Inspired by https://stackoverflow.com/a/70171464/244191
```
devonbush: this seems like it should be in a utility class somewhere with more detailed explanation of what it's trying to do

ui-admin/src/study/kits/KitEnrolleeSelection.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+    setShowRequestKitModal(false)
+    loadEnrollees()
+  }
+
+  const numSelected = Object.keys(rowSelection).length
+  const enableActionButtons = numSelected > 0
+
+  const hasEnrolleeCompletedSurvey = (enrollee: Enrollee, survey: Survey): boolean => {
```
devonbush: instead of making a computed column for this, since the operation is non-trivial, I'd lean towards making a "transformEnrollees()" function that you call in the `loadEnrollees` function to calculate all these boolean column values, and then the columns can just be standard booleans

ui-admin/src/study/kits/KitEnrolleeSelection.tsx
```diff
@@ -0,0 +1,176 @@
...
+      checked={table.getIsAllRowsSelected()} indeterminate={table.getIsSomeRowsSelected()}
+      onChange={table.getToggleAllRowsSelectedHandler()}/>,
+    cell: ({ row }) => (
+      <div className="px-1">
+        <IndeterminateCheckbox
+          checked={row.getIsSelected()} indeterminate={row.getIsSomeSelected()}
+          onChange={row.getToggleSelectedHandler()} disabled={!row.getCanSelect()}/>
+      </div>
+    )
+  }, {
```
devonbush, line 112: might not be in the scope of this PR, but this table could really use to be sortable by enrollee creation date

ui-admin/src/study/kits/KitManager.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return <div className="ParticipantList container pt-2">
+    <div className="row">
+      <div className="col-12">
+        <h2 className="h5">{study.name} Sample Kits</h2>
+        <Tabs
+          activeKey={activeTab ?? undefined}
+          mountOnEnter
+          unmountOnExit
+          onSelect={setActiveTab}
+        >
```
devonbush, line 27: IMPORTANT:  (probably the closest to a blocker).  These tabs need to be routable -- e.g `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/byKit`
and 
`https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/byEnrollee`
or somesuch.

This can be done by making "activeTab" a parameter passed into KitManager from KitRouter

devonbush (changes requested):
Lots of nifty functionality here.  bunch of comments, I've put "IMPORTANT" on the ones I care the most about, since none of them are per-se a blocker, but want to make sure we get on the same page (or you tell me it would take way too long to address) on at least most of them.  
Tested it locally, works well.  I didn't repopulate my DB, and the only thing I noticed that might have been amiss was that JSALK had a kit status of "unknown"

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/kit/KitController.java
breilly2: I will update. I instinctively avoided "by" because I often use that when describing things that are maps (e.g. `enrolleeByStudy`).

api-admin/src/main/resources/api/openapi.yml
breilly2: I'll move it since that makes the most sense after renaming it.

api-admin/src/main/resources/api/openapi.yml
breilly2: With the method rename, this becomes `enrolleesWithKits`, which I think is fine for now. If we implement fetch groups in the future, I think it'll be easy to notice `enrolleesWithKits` as a candidate for rolling into a fetch-group-enabled endpoint.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
breilly2: Hmm... good question. This is probably the result of things changing during development that I didn't revisit during clean-up.

Thinking through changing this...

I could claim that this is protecting type safety in the UI code. Client-side, this is going to be deserialized to a [`PepperKitStatus`](https://github.com/broadinstitute/juniper/blob/1dd13e483c8ba1687655405dd3471a37e06914fe/ui-admin/src/api/api.tsx#L162). I'm not sure what would happen if the JSON didn't match that shape. Typescript won't care because we don't have the type information at runtime (which brings into question the validity of type safety here).

Alternatively, we could pass through `pepperStatusJson` and make the parsing explicit in the UI. That has the benefit of there being a natural place to handle unexpected JSON.

No matter what, I stand by the comment that we should be defensive about unexpected JSON from Pepper and make sure that Juniper degrades gracefully instead of being completely broken.

@devonbush, which approach do you prefer?

populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/enrollees/jsalk.json
breilly2: I think I was just capturing the complete shape of this object. I'm not sure why I put the placeholder dates in before, but now that I've gotten rid of them, I'll remove the empty error object.

ui-admin/src/study/kits/KitManager.tsx
breilly2: Yes, this is probably the thing I was least comfortable about not addressing before opening the PR. Well, that and neglecting UI tests ☹️. In both cases, I decided to get the PR going instead of delaying it further. I should be able to fix this today.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
devonbush: If we expect bad json from Pepper (and we're not filtering it out on receipt) then we'll want to surface that in the UI, and you're right that that means initially treating the raw field as "unknown" in Typescript until proven otherwise so that our JS doesn't error out.  I lean towards keeping our backend APIs simple "here's what's in the database" and letting our frontend do validation/parsing so that we get maximum visibility into what's going on, and it's faster to iterate on and debug.  But either way works

devonbush (approved):

ui-admin/src/study/kits/KitEnrolleeSelection.tsx
breilly2: Some calculation can be done ahead of time, but still need to use `accessorFn` to do a looking by survey since the list of surveys is dynamic based on the study.

ui-admin/src/study/kits/KitList.tsx
breilly2: I found a much simpler solution: `_.countBy`. 🤦‍♂️

ui-admin/src/study/kits/KitList.tsx
```diff
@@ -0,0 +1,122 @@
...
+    accessorKey: 'pepperStatus.currentStatus',
+    accessorFn: data => pepperStatusToHumanStatus(data?.pepperStatus?.currentStatus),
+    meta: {
+      columnType: 'string',
+      filterOptions: [
+        { value: true, label: 'Created' },
+        { value: true, label: 'Prepared' },
+        { value: true, label: 'Sent' },
+        { value: true, label: 'Received' }
+      ]
```
MatthewBemis: I don't think this column filter is working as you're intending. Currently it shows up as a search bar where you can type in the status, but given the use of `filterOptions` here I am assuming you want a dropdown. When I wrote the code for the custom dropdown filter, I only implemented it for `boolean`. 

I just pushed this commit to another branch that extends it to handle `string` columns as well, if you'd like to pull it into this PR: https://github.com/broadinstitute/juniper/commit/49a607428d4d7e0462f7753e000cbec5f2869d7c

MatthewBemis (approved):
This looks good to me. I have one suggestion about fixing the column filter for kit status, which I think is worth fixing, but :+1: regardless

ui-admin/src/study/kits/KitList.tsx
breilly2: Oh, nice! Yes, this is a place where I had to make a decision to let it be imperfect rather than spend more time on it. This _might_ be less relevant depending on Erin's direction with designs, but I'll definitely update it now if it's easy to do.

PR #478, opened 2023-07-13
JN-456: Enabling live-editing of site content

devonbush:
Enables quick-and-dirty demo of live editing of SiteContent.  

https://localhost:3000/ourhealth/env/sandbox/siteContent

Saving is currently not supported, nor is anything beyond simple text edits within the json blobs.  this is only intended for demo purposes.   this moves the landing page components from ui-participant into UI core.  Another thought was having ui-admin depend on ui-participant, but Nick and I discussed and that would probably be unwieldy from a build artifact point of view.   Most of the change is just the dependency wrangling to sort that out.  

Later steps (which might not happen right away depending on demand), include:
1. pulling config fields into separate editable components
2. enabling actual saving
3. add/remove/reorder sections
4. add/remove/reorder navbar items
5. configuring footer, navlogo, and other properties of sitecontent

TO TEST:
1. you'll probably need to stop all ui projects, run `npm install`, and then restart them all (first building ui-core)
2. click through https://sandbox.ourhealth.localhost:3001 to confirm stuff still generally works
3. go to https://localhost:3000/ourhealth/env/sandbox/siteContent.  
4. edit some text, confirm the updates show in the preview panels


ui-admin/package.json
```diff
@@ -57,5 +57,11 @@
   },
   "eslintConfig": {
     "extends": "../.eslintrc.js"
+  },
+  "jest": {
+    "moduleNameMapper": {
+      "react-markdown": "<rootDir>/../node_modules/react-markdown/react-markdown.min.js",
```
nawatts, line 63: This seems unnecessary, since it's resolving react-markdown from node_modules as it would be normally.

ui-core/src/participant/landing/ConfiguredButton.tsx
```diff
@@ -2,7 +2,7 @@ import classNames from 'classnames'
 import React from 'react'
 import { Link } from 'react-router-dom'
 
-import { requireOptionalString, requirePlainObject, requireString } from 'util/validationUtils'
+import { requireOptionalString, requirePlainObject, requireString } from '../../participant/util/validationUtils'
```
nawatts: These paths seem unnecessarily long. Couldn't this be `../util/validationUtils`?

ui-core/src/types/landingPageConfig.ts
```diff
@@ -50,25 +50,28 @@ export type NavbarItem =
   | NavbarItemMailingList
   | NavbarItemExternal
 
-export type NavbarItemInternal = {
-  itemType: 'INTERNAL'
+export type BaseNavBarItem = {
```
nawatts: Suggest not exporting this to avoid the temptation to use it elsewhere. Elsewhere, you almost always want a union of the different NavBarItem types vs this base type.

ui-admin/package.json
devonbush: I couldn't get tests to pass without it.  They fail with
```
Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

/Users/dbush/code/pearl/node_modules/react-markdown/index.js:6
    export {uriTransformer} from './lib/uri-transformer.js'
    ^^^^^^

    SyntaxError: Unexpected token 'export'

       8 | var _classnames = _interopRequireDefault(require("classnames"));
       9 | var _react = _interopRequireDefault(require("react"));
    > 10 | var _reactMarkdown = _interopRequireDefault(require("react-markdown"));
```

ui-core/src/participant/landing/ConfiguredButton.tsx
devonbush: probably leftover from moving them around.

ui-admin/package.json
nawatts: Ah, looks like that package exports ES modules and Jest needs CommonJS.

ui-admin/src/portal/siteContent/SiteContentView.tsx
```diff
@@ -1,51 +1,119 @@
...
+  pageOpts.unshift({ label: 'Landing page', value: null })
+
+  return <div className="d-flex bg-white p-3">
     <div className="ps-3">
-      <h2 className="h5">{renderedTitle}</h2>
+      <div className="d-flex mb-3 w-100">
+        <div>
+          <Select options={pageOpts} value={selectedNavOpt}
+            onChange={e => setSelectedNavOpt(e ?? landingPageOption)}/>
+        </div>
```
breilly2, line 83: Nit: There should be some min width on the drop-down. When something like "FAQ" is selected, it's tough to read the other options.

ui-admin/src/test-utils/mock-site-content.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/** mock single page section */
+export const mockHtmlSection = (): HtmlSection => {
+  return {
+    id: 'fakeId',
+    sectionType: 'HERO_CENTERED',
+    sectionConfig: JSON.stringify({
+      title: 'about us',
+      blurb: 'we are the best'
```
MatthewBemis, line 44: 💯

ui-core/src/participant/landing/sections/ParticipationDetailTemplate.tsx
```diff
@@ -3,15 +3,16 @@ import React from 'react'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faClock } from '@fortawesome/free-regular-svg-icons'
 
-import { SectionConfig } from 'api/api'
-import { getSectionStyle } from 'util/styleUtils'
-import { withValidatedSectionConfig } from 'util/withValidatedSectionConfig'
-import { requireOptionalString } from 'util/validationUtils'
+import { SectionConfig } from '../../../types/landingPageConfig'
```
MatthewBemis, line 6: Some of the import paths in this file and others can be shortened. I don't have strong feelings about this but since I've seen it come up in other PRs, is our style preference to always use the shortened path?

MatthewBemis (approved):
Overall, looks good to me. Tested it out a bit and things work as expected. 

_Non-blocker:_

This may be planned for a future ticket, but links in the previews do not route to the participant UI, they route to the relative path within the admin UI, which always throws an error

breilly2 (approved):
Thank you for fixing the issue with joining the mailing list that we discussed over Slack. One nit, but otherwise looks very good!

Interesting observation: I like the layout you chose with the editable JSON on the left and the preview on the right. In most cases, there is a correlation between the length of the JSON and the height of the content, so the text areas tend to be reasonably appropriately sized, which is pretty cool.

devonbush:
yeah, fixing the routing links to not go to bad places when they click on buttons will be a separate piece of work

PR #479, opened 2023-07-14
[JN-453] Add new questions to a survey

MatthewBemis:
This adds the ability to add new questions to a survey page. Not included in this PR is the autofill "scratch box" functionality. That will be coming in the next PR.

Untitled

Screenshot 2023-07-14 at 12 43 38 PM

TO TEST:

Play around with the question designer. Make sure that switching between types works correctly and that the questions are saved correctly by viewing the generated JSON in the JSON Editor, as well as the final result in the Survey Preview tab.

### DIFF
--- a/ui-admin/src/forms/designer/NewQuestionForm.tsx
+++ b/ui-admin/src/forms/designer/NewQuestionForm.tsx
@@ -8,15 +8,16 @@ import { TextInput } from '../../components/forms/TextInput'
 
 type NewQuestionFormProps = {
     onCreate: (newQuestion: Question) => void
+    readOnly: boolean
 }
 
 /** UI for creating a new question. */
 export const NewQuestionForm = (props: NewQuestionFormProps) => {
-  const { onCreate } = props
+  const { onCreate, readOnly } = props
   const [selectedQuestionType, setSelectedQuestionType] = useState<QuestionType>('text')
   const [questionName, setQuestionName] = useState<string>('')
 
-  const emptyQuestions: Record<QuestionType, Question> = {
+  const baseQuestions: Record<QuestionType, Question> = {
     checkbox: {
       type: 'checkbox',
       name: questionName,
@@ -52,12 +53,7 @@ export const NewQuestionForm = (props: NewQuestionFormProps) => {
     }
   }
 
-  const [question, setQuestion] = useState<Question>(emptyQuestions[selectedQuestionType])
-
-  // //TODO: probably don't need this as an effect
-  // useEffect(() => {
-  //   setQuestion(emptyQuestions[selectedQuestionType])
-  // }, [selectedQuestionType])
+  const [question, setQuestion] = useState<Question>(baseQuestions[selectedQuestionType])
 
   return (
     <>
@@ -74,12 +70,12 @@ export const NewQuestionForm = (props: NewQuestionFormProps) => {
               }}
             />
           </div>
-          <label className="form-label" htmlFor="text-question-input-type">Question type</label>
+          <label className="form-label" htmlFor="questionType">Question type</label>
           <select id="questionType" className="form-select" value={selectedQuestionType}
             onChange={e => {
               const newQuestionType = e.target.value as QuestionType
               setSelectedQuestionType(newQuestionType)
-              setQuestion(emptyQuestions[newQuestionType])
+              setQuestion(baseQuestions[newQuestionType])
             }}>
             <option value="text">Text</option>
             <option value="checkbox">Checkbox</option>
@@ -93,7 +89,7 @@ export const NewQuestionForm = (props: NewQuestionFormProps) => {
         { selectedQuestionType && <QuestionDesigner
           question={question}
           showName={false}
-          readOnly={false} //maybe pass this in from parent component, but you can't really get here if it's read-only
+          readOnly={readOnly}
           onChange={updatedElement => {
             setQuestion(updatedElement)
           }}
--- a/ui-admin/src/forms/designer/NewQuestionForm.tsx
+++ b/ui-admin/src/forms/designer/NewQuestionForm.tsx
@@ -4,7 +4,7 @@ import { Question, QuestionType } from '@juniper/ui-core'
 
 import { Button } from 'components/forms/Button'
 import { QuestionDesigner } from './QuestionDesigner'
-import { TextInput } from '../../components/forms/TextInput'
+import { TextInput } from 'components/forms/TextInput'
 
 type NewQuestionFormProps = {
     onCreate: (newQuestion: Question) => void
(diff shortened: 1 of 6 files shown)


ui-admin/src/forms/designer/NewQuestionForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+    readOnly: boolean
+}
+
+/** UI for creating a new question. */
+export const NewQuestionForm = (props: NewQuestionFormProps) => {
+  const { onCreate, readOnly } = props
+  const [selectedQuestionType, setSelectedQuestionType] = useState<QuestionType>('text')
+  const [questionName, setQuestionName] = useState<string>('')
+
+  const baseQuestions: Record<QuestionType, Question> = {
```
devonbush, line 20: I think this code will get cleaner if we explicitly make an object for the question types:
```
type QuestionTypeDef = {
  name: string
  displayName: string
  hasChoices: boolean
}
const baseQuestionTypeDefs: QuestionTypeDef[] = [{
  name: 'checkbox',
  hasChoices: true,
  displayName: 'Checkbox'
} ...
}]

const baseQuestions = baseQuestionTypeDefs.map(typeDef => ({
   type: typeDef.name,
   name: questionName,
   title: '',
   choices: typeDef.hasChoices ? [] : undefined 
})  

Then you can likewise use a .map call to generate the type options

File: ui-admin/src/forms/designer/NewQuestionForm.tsx [hunk 2]
```diff
@@ -0,0 +1,109 @@
+import React, { useState } from 'react'
+
+import { Question, QuestionType } from '@juniper/ui-core'
+
+import { Button } from 'components/forms/Button'
+import { QuestionDesigner } from './QuestionDesigner'
+import { TextInput } from '../../components/forms/TextInput'
```
devonbush:
```suggestion
import { TextInput } from 'components/forms/TextInput'
```

devonbush (approved):
Couldn't play around with it because my loaner laptop is refusing to install docker, so I can't get anything to run.  thanks for including the screenshots, though, looks great.  My main thought is that we want to start being robust about how we represent Question type definitions, since that will eventually dictate what users can/can't configure for a question.  Looking through the SurveyJS API, I don't think they have a singular QuestionType definition (they have https://surveyjs.io/form-library/documentation/api-reference/question but that's more of a base class, than a type def), so I think we'll want to roll our own so we're controlling exactly what we do/don't support

devonbush:
Finally got a chance to check this out - really like how you added all the descriptive helper text -- made it much easier to navigate

devonbush:
It doesn't have to be in this PR, but we really should enforce non-null name/stableIds -- I accidetnally created one with a blank name, and it was almost impossible to find

ui-admin/src/forms/designer/NewQuestionForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const [question, setQuestion] = useState<Question>(baseQuestions[selectedQuestionType])
+
+  return (
+    <>
+      <div className="text-align-right">
+        <div className="mb-3">
+          <div className="mb-3">
+            <TextInput
+              description='The stable identifier for the survey question'
+              label='Question name'
```
breilly2: I think this label should include "stable identifier". I ignored the description (which might actually be a little too low-contrast; I haven't checked it with a11y tooling) and didn't realize that that's what I was choosing.

ui-admin/src/forms/designer/NewQuestionForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+              value={questionName}
+              onChange={value => {
+                setQuestionName(value)
+                setQuestion({ ...question, name: value })
+              }}
```
breilly2, line 70: Some uniqueness validation here would be helpful. I did eventually get an error for a duplicate name, but only when trying to save the survey.

ui-admin/src/forms/designer/NewQuestionForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+            />
+          </div>
+          <label className="form-label" htmlFor="questionType">Question type</label>
+          <select id="questionType" className="form-select" value={selectedQuestionType}
+            onChange={e => {
+              const newQuestionType = e.target.value as QuestionType
+              setSelectedQuestionType(newQuestionType)
+              setQuestion(baseQuestions[newQuestionType])
+            }}>
```
breilly2, line 79: Switching between `checkbox`, `dropdown`, and `radiogroup` should preserve the configured options, or at least warn that the options will be discarded. I think it's natural to play around with options in a dialog like this to see what's possible and eventually settle back on what was initially chosen, and losing work when doing this is not great. Even better would be to always remember the list of options so that you could switch from, e.g., `dropdown` to `text` and back again without losing work.

I'm also losing "Question text", "Require response", and "Conditionally show this question" when changing the question type, and those should all be valid no matter what type of question it is.

breilly2, line 79: Bug: I'm also seeing that, even though the "Description" input doesn't get blanked out when changing question type, it doesn't actually get saved.

breilly2 (changes requested):
A couple of nice-to-haves, but also an important bug that should be fixed around capturing "Question text" and "Description" after changing question type.

ui-admin/src/forms/designer/NewQuestionForm.tsx
MatthewBemis: This will be handled as a part of https://broadworkbench.atlassian.net/browse/JN-461

ui-admin/src/forms/designer/NewQuestionForm.tsx
MatthewBemis: Updated label in f2b09a51

I just checked the contrast for the description text with Axe Devtools and it passed contrast requirements

ui-admin/src/forms/designer/NewQuestionForm.tsx
MatthewBemis: Agh you're right. Just took a closer look and it seems like there's a bug with descriptions in other parts of the survey editor as well. I'll track it down.

MatthewBemis: 7c9ae04a7377586921c0b138b2d58d47c238c9c5 should fix it

ui-admin/src/forms/designer/NewQuestionForm.tsx
breilly2: Cool. Probably just my brain filtering out the description given the relative size and contrast compared to the label.

ui-admin/src/forms/designer/NewQuestionForm.tsx
breilly2: Cool. That at least makes the behavior predictable, in that I'm not seeing text in the input that doesn't get saved.

breilly2 (approved):
Thank you for fixing the bug. I do think that the form should be less destructive when changing question type, and should remember as much as possible when toggling between question types (which someone might do while exploring and deciding what they want), but that can be a future enhancement.

MatthewBemis:
Going to merge this as-is so Erin can play around with it, but will improve the destructive nature of question-switching in a future PR.

PR #481, opened 2023-07-19
[JN-477] admin help for export

devonbush:
This adds basic help docs (copied over from Pepper) for the download capability.  The existence of help docs within our app is probably temporary.  Long-term, we'll want to use ZenDesk or some other solution that lets non-engineers write/edit the docs.   I'll put in a request to comms to see about getting that set up.  In the meantime, this will give us a place to park user-visible documentation.  Also, having it be in-app makes the release/review process much simpler -- the docs get auto-released with the features, rather than having to be reviewed and released separately.

image

image

TO TEST:
1. Go to https://localhost:3000/ourhealth/env/live/participants
2. click "download"
3. click the link to the help page
4. confirm the help page shows up in a new tab


ui-admin/src/help/ExportHelp.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <ul>
+        <li>
+          <b>.xlsx</b> will create an Excel spreadsheet of the participant data. Empty cells will represent
+                    null values.
+        </li>
+        <li>
+          <b>.tsv</b> (tab-delimited values). Will export a tab-delimited file. This may be useful in
+                    environments where Excel is unavailable, or if the
+                    number of columns to be exported exceeds 16K. In order to have data be compliant, double-quotes will
+                    be replaced by single quotes, and any values including
```
breilly2, line 22: It isn't really necessary to replace double quotes with single quotes. It's unlikely that the meaning of any responses will be different, but I have to confess to being a little uneasy about any modification of responses.

ui-admin/src/study/participants/export/ExportDataControl.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+  expect(screen.getByText('Excel (.xlsx)')).toBeInTheDocument()
+})
+
+test('help page loads', async () => {
+  const { RoutedComponent } = setupRouterTest(
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    <ExportDataControl studyEnvContext={mockStudyEnvContext()} show={true} setShow={() => {}}/>)
+  render(RoutedComponent)
+  userEvent.click(screen.getByText('help page'))
+  waitFor(() => expect(screen.getByText('Participant List Export Info')).toBeInTheDocument())
```
breilly2, line 24: Interesting. I didn't expect to not need any code to recognize the content opened in a new window.

breilly2 (approved):

ui-admin/src/help/ExportHelp.tsx
devonbush: you're right that they can be escaped, but other applications handling of escaped quotes in csv files varies enough that it was decided for Pepper that this strategy would result in less user issues (and they can always grab the .xlsx format if quote fidelity is important)

ui-admin/src/study/participants/export/ExportDataControl.test.tsx
devonbush: I'm guessing that the headless browser just ignores the target="_blank" and opens it in the same window

MatthewBemis (approved):

PR #482, opened 2023-07-19
[JN-490] avoiding error on dupe mailing list signup

devonbush:
TO TEST:
1. restart ApiParticipantApp 
2. go to https://sandbox.ourhealth.localhost:3001
3. join the mailing list with an address
4. reload the page
5. join the mailing list again with the same address
6. confirm it succeeds
7. go to https://localhost:3000/ourhealth/env/sandbox/mailingList -- confirm the address only appears once


breilly2 (approved):
This is a good thing to fix. Another option would be to remove any uniqueness checks/constraints and let the study staff (possibly with the help of the admin tool) handle deduplication. Ignoring the second sign-up will discard any difference in name. It's subtle, but if I signed up a while ago and didn't remember, then signed up again and got an email with a different name (perhaps because I changed my name), I'd be very suspicious about why my old name was used.

MatthewBemis (approved):

PR #485, opened 2023-07-25
[JN-453] Preserve fields when switching question types in new question form

MatthewBemis:
This improves the behavior of the `NewQuestionForm` added in https://github.com/broadinstitute/juniper/pull/479

Specifically, it fixes the problem where switching the question type would reset all of the fields, even if the question types had those fields in common. (described in https://github.com/broadinstitute/juniper/pull/479#discussion_r1266734590)

To do this, we keep a `Question` object in state and accumulate all of the possible fields. When the question is saved, we remove any extraneous fields that do not belong to the final selected question type (i.e. if you started as a `TextQuestion`, switched to the `DropdownQuestion` type, added a few choices, but then decided you wanted to go to `DropdownQuestion`, we would omit the `choices` property right before saving)

I intend to add unit tests for this, but want feedback on the approach before investing more time into it.

TO TEST:

1. Start filling out fields for any question type
2. Switch to a question type that has some fields in common with the first type (for example, TextQuestion -> RadiogroupQuestion)
3. Verify that the common fields are still filled out
4. Fill out some fields that are unique to RadiogroupQuestion (`choices` is one of them)
5. Switch back to the original question type (or you can also test by switching to a third question type)
6. Verify that the common fields are still filled out
7. Save the question
8. Verify in the JSON preview that the extraneous fields were not saved and all of the relevant fields were persisted

### DIFF
--- a/ui-admin/src/forms/designer/NewQuestionForm.tsx
+++ b/ui-admin/src/forms/designer/NewQuestionForm.tsx
@@ -41,11 +41,8 @@ export const NewQuestionForm = (props: NewQuestionFormProps) => {
             onChange={e => {
               const newQuestionType = e.target.value as QuestionType
               setSelectedQuestionType(newQuestionType)
-              const baseQuestion = baseQuestions[newQuestionType]
-              console.log(question)
-              console.log(baseQuestion)
               // @ts-ignore
-              setQuestion({ ...baseQuestion, ...question, type: newQuestionType })
+              setQuestion({ ...baseQuestions[newQuestionType], ...question, type: newQuestionType })
             }}>
             <option value="text">Text</option>
             <option value="checkbox">Checkbox</option>
--- a/ui-admin/src/forms/designer/NewQuestionForm.tsx
+++ b/ui-admin/src/forms/designer/NewQuestionForm.tsx
@@ -17,9 +17,9 @@ type NewQuestionFormProps = {
 export const NewQuestionForm = (props: NewQuestionFormProps) => {
   const { onCreate, readOnly } = props
   const [selectedQuestionType, setSelectedQuestionType] = useState<QuestionType>('text')
-  const [questionName, setQuestionName] = useState<string>('')
 
   const [question, setQuestion] = useState<Question>(baseQuestions[selectedQuestionType])
+  const { name: questionName } = question
 
   return (
     <>
@@ -31,7 +31,6 @@ export const NewQuestionForm = (props: NewQuestionFormProps) => {
               label='Question stable ID'
               value={questionName}
               onChange={value => {
-                setQuestionName(value)
                 setQuestion({ ...question, name: value })
               }}
             />
--- a/ui-admin/src/forms/designer/NewQuestionForm.tsx
+++ b/ui-admin/src/forms/designer/NewQuestionForm.tsx
@@ -5,6 +5,8 @@ import { Question, QuestionType } from '@juniper/ui-core'
 import { Button } from 'components/forms/Button'
 import { QuestionDesigner } from './QuestionDesigner'
 import { TextInput } from 'components/forms/TextInput'
+import { baseQuestions } from './questions/questionTypes'
+import _ from 'lodash'
 
 type NewQuestionFormProps = {
     onCreate: (newQuestion: Question) => void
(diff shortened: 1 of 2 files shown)


ui-admin/src/forms/designer/NewQuestionForm.tsx
```diff
@@ -75,7 +41,8 @@ export const NewQuestionForm = (props: NewQuestionFormProps) => {
             onChange={e => {
               const newQuestionType = e.target.value as QuestionType
               setSelectedQuestionType(newQuestionType)
-              setQuestion(baseQuestions[newQuestionType])
+              // @ts-ignore
+              setQuestion({ ...baseQuestions[newQuestionType], ...question, type: newQuestionType })
```
MatthewBemis: `ts-ignore`d because adding the `type` field here gives: 

`TS2322: Type '"checkbox" | "dropdown" | "medications" | "radiogroup" | "signaturepad" | "text"' is not assignable to type '"text"'.   Type '"checkbox"' is not assignable to type '"text"'.`

Not totally sure how to fix this. TypeScript really thinks that this should only ever be a `TextQuestion` and I don't know how to convince it otherwise

ui-admin/src/forms/designer/NewQuestionForm.tsx
```diff
@@ -17,42 +19,6 @@ export const NewQuestionForm = (props: NewQuestionFormProps) => {
   const [selectedQuestionType, setSelectedQuestionType] = useState<QuestionType>('text')
   const [questionName, setQuestionName] = useState<string>('')
 
-  const baseQuestions: Record<QuestionType, Question> = {
```
MatthewBemis, line 20: Moved to `questionTypes.tsx` (and modified a bit to represent all possible fields)

ui-admin/src/forms/designer/NewQuestionForm.tsx
nawatts: The only thing I can come up with is that this could be slightly improved to provide some type safety by casting the result instead of using `ts-ignore`.

```suggestion
              setQuestion({ ...baseQuestions[newQuestionType], ...question, type: newQuestionType } as Question)
```

MatthewBemis: Good call, done in 64ffcd5a

ui-admin/src/forms/designer/questions/questionTypes.tsx
```diff
@@ -17,3 +17,71 @@ export const questionTypeDescriptions: Record<QuestionType, string> = {
...
+  checkbox: {
+    type: 'checkbox',
+    name: '',
+    title: '',
+    description: '',
+    isRequired: false,
+    choices: [],
+    showNoneItem: undefined,
+    noneText: undefined,
+    noneValue: undefined,
```
breilly2, line 32: Purely a side note, but it's odd to me that dropdown and radio questions do not offer a "none" option.

ui-admin/src/forms/designer/NewQuestionForm.tsx
```diff
@@ -17,42 +19,6 @@ export const NewQuestionForm = (props: NewQuestionFormProps) => {
   const [selectedQuestionType, setSelectedQuestionType] = useState<QuestionType>('text')
   const [questionName, setQuestionName] = useState<string>('')
```
breilly2: This looks like duplicated state to me. I think this could be replaced with `const { name: questionName } = question` (below the `useState<Question>` line).

breilly2 (approved):
Nice! This feels like the right balance given the language constraints.

A couple of things crossed my mind that I ultimately couldn't think of better solutions for:
* It would be nice if `_.pick`ing the relevant question fields to submit could be based on the type, but we don't have that type information at runtime, so having prototypical examples of each question seems like the next best thing.
* I worried a little bit about the code distance between the question `type`s and the prototypical values in `baseQuestions` (not just in different files but in different modules) because these things are likely to change together. However, the other mental model I have for this is that `baseQuestions` represents the UI form which might be intentionally different from the underlying model. With that framing, the code distance becomes more of a design feature than a flaw.
* I wondered if it would make sense to have a `QuestionForm` type that explicitly held the union of all of the fields, making it closer to a model of the UI than a model of the question data that needs to be submitted. However, that runs into the issue of `_.pick`ing the relevant values to submit and, again, we don't know the relevant keys without the prototypical values, at which point a `QuestionForm` becomes redundant.

ui-admin/src/forms/designer/questions/questionTypes.tsx
nawatts: Dropdown and radio questions can have a "none" option in their regular options. In checkbox questions, this "special" none option clears all other checkboxes.

nawatts (approved):

ui-admin/src/forms/designer/NewQuestionForm.tsx
MatthewBemis: Good eye, fixed in 0ecef11f

PR #487, opened 2023-07-26
[JN-463] Support adding new question templates

MatthewBemis:
This adds the ability to A) create new question templates and B) create new questions that use those question templates (see videos below)

TO TEST:

1. Try creating new question templates of different question types
2. Try creating new questions based on those templates. Make sure that the optional overrides work correctly
3. When creating a new question, try switching back and forth between using a template and not using a template. Make sure that the appropriate fields are disabled when a question template is selected, and that when saving a question afterwards, the correct fields are saved in the JSON
4. Verify that the questions show up correctly in the survey preview

UPDATE: there is one bug that I noticed after requesting reviews. If you select a question type, and then select a question template, it doesn't re-render the question type dropdown after setting the type to `undefined`. Not sure why this is happening. The underlying state is correct, so it's mainly a rendering problem. Looking into it...

Adding a new question template:

https://github.com/broadinstitute/juniper/assets/7257391/2ecfcba7-95c4-4f85-8b77-9d5f440ecae0

Using a question template to construct a new question:

https://github.com/broadinstitute/juniper/assets/7257391/c7748c0c-c125-4307-93e3-904860249e83


breilly2:
For the bug you mentioned, I think it would make sense to set the question type to the type matching the template (instead of `undefined`) at the same time as disabling the question type selection.

ui-admin/src/forms/designer/NewQuestionForm.tsx
```diff
@@ -6,22 +6,26 @@ import { Button } from 'components/forms/Button'
 import { QuestionDesigner } from './QuestionDesigner'
 import { TextInput } from 'components/forms/TextInput'
 import { baseQuestions } from './questions/questionTypes'
-import _ from 'lodash'
+import _, { isEmpty } from 'lodash'
```
breilly2: May as well use `_.isEmpty` instead of importing it explicitly... or go through the file and find what lodash functions are being used and import them explicitly. I'd rather either of those than living half way in between.

Side note: I've been meaning to ask someone if there is an advantage to importing individual functions instead of the entire library. I could guess that it might be to reduce the compiled bundle size, but there are other ways to do that; Terra UI uses "tree shaking" somewhere in the build that, from what I understand, effectively removes dead code.

breilly2 (approved):

ui-admin/src/forms/designer/NewQuestionForm.tsx
devonbush: I believe we also have tree shaking that means that the import style should have no impact on bundle size.  In general, I prefer specific imports since it makes the dependencies clearer at the top of the file, but it's not a strong preference.

ui-admin/src/forms/designer/NewQuestionForm.tsx
```diff
@@ -40,8 +44,35 @@ export const NewQuestionForm = (props: NewQuestionFormProps) => {
...
+                setFreetextMode(false)
+                setSelectedQuestionTemplateName(newValue ? newValue : undefined)
+                setQuestion(unset('type', { name: questionName, questionTemplateName: newValue?.value }) as Question)
+              }}
+            />
+            <p
+              className="form-text"
+              id="questionTemplateDesc"
+            >
+              {'Select a question template to pre-populate the question fields'}
```
devonbush: This doesn't need to be wrapped
```suggestion
              Select a question template to pre-populate the question fields
```

ui-admin/src/forms/designer/QuestionTemplatesDesigner.tsx
```diff
@@ -1,8 +1,13 @@
-import React from 'react'
+import React, { useState } from 'react'
 
 import { FormContent } from '@juniper/ui-core'
 
 import { QuestionTemplateList } from './QuestionTemplateList'
+import { Button } from '../../components/forms/Button'
```
devonbush:
```suggestion
import { Button } from 'components/forms/Button'
```

devonbush (approved):
Tested it locally, seems good.  A couple minor suggestions

ui-admin/src/forms/designer/QuestionTemplatesDesigner.tsx
MatthewBemis: I apologize for how many times you've had to make this suggestion on my PRs 🤦‍♂️

PR #488, opened 2023-07-27
[JN-453] Add freetext/scratchbox option for question builder

MatthewBemis:
Screen recording:

https://github.com/broadinstitute/juniper/assets/7257391/c0513f71-094d-4886-9742-9bf545878e66

### DIFF
--- a/ui-admin/src/forms/designer/NewQuestionForm.test.tsx
+++ b/ui-admin/src/forms/designer/NewQuestionForm.test.tsx
@@ -17,10 +17,10 @@ describe('NewQuestionForm', () => {
 
   test('updates to the appropriate QuestionDesigner when a new question type is selected', async () => {
     //Arrange
+    const user = userEvent.setup()
     render(<NewQuestionForm onCreate={() => jest.fn()} readOnly={false}/>)
 
     //Act
-    const user = userEvent.setup()
     const questionTypeSelect = screen.getByLabelText('Question type')
     await act(() => user.selectOptions(questionTypeSelect, 'checkbox'))
 
@@ -31,10 +31,10 @@ describe('NewQuestionForm', () => {
 
   test('renders freetext input', async () => {
     //Arrange
+    const user = userEvent.setup()
     render(<NewQuestionForm onCreate={() => jest.fn()} readOnly={false}/>)
 
     //Act
-    const user = userEvent.setup()
     const questionTypeSelect = screen.getByLabelText('Question type')
     const freetextModeCheckbox = screen.getByLabelText('Enable freetext mode')
     await act(() => user.click(freetextModeCheckbox))
@@ -46,10 +46,10 @@ describe('NewQuestionForm', () => {
 
   test('updates to the appropriate QuestionDesigner based on freetext input', async () => {
     //Arrange
+    const user = userEvent.setup()
     render(<NewQuestionForm onCreate={() => jest.fn()} readOnly={false}/>)
 
     //Act
-    const user = userEvent.setup()
     const questionTypeSelect = screen.getByLabelText('Question type')
     const freetextModeCheckbox = screen.getByLabelText('Enable freetext mode')
     await act(() => user.click(freetextModeCheckbox))
--- a/ui-admin/src/forms/designer/NewQuestionForm.test.tsx
+++ b/ui-admin/src/forms/designer/NewQuestionForm.test.tsx
@@ -51,9 +51,9 @@ describe('NewQuestionForm', () => {
 
     //Act
     const questionTypeSelect = screen.getByLabelText('Question type')
-    fireEvent.change(questionTypeSelect, {target: {value: 'freetext'}})
+    fireEvent.change(questionTypeSelect, { target: { value: 'freetext' } })
     const freetextInput = screen.getByLabelText('Freetext') as HTMLInputElement
-    userEvent.pointer({target: freetextInput, offset: 0, keys: '[MouseLeft]'})
+    userEvent.pointer({ target: freetextInput, offset: 0, keys: '[MouseLeft]' })
     userEvent.keyboard('This is a question!\nThis is an option!\nThis is another option!')
 
     //Assert
--- a/ui-admin/src/forms/designer/NewQuestionForm.test.tsx
+++ b/ui-admin/src/forms/designer/NewQuestionForm.test.tsx
@@ -1,7 +1,8 @@
 import React from 'react'
-import { render, screen, fireEvent } from '@testing-library/react'
+import { render, screen, fireEvent, waitFor } from '@testing-library/react'
 import { NewQuestionForm } from './NewQuestionForm'
 import { questionTypeDescriptions } from './questions/questionTypes'
+import userEvent from '@testing-library/user-event'
 
 describe('NewQuestionForm', () => {
   test('renders the default view for a new question', () => {
@@ -13,7 +14,7 @@ describe('NewQuestionForm', () => {
     const questionTypeSelect = screen.getByLabelText('Question type')
     expect(questionStableIdInput).toBeInTheDocument()
     expect(questionTypeSelect).toBeInTheDocument()
-    expect((questionTypeSelect as HTMLSelectElement).value).toBe('text')
+    expect((questionTypeSelect as HTMLSelectElement).value).toBe('placeholder')
   })
 
   test('updates to the appropriate QuestionDesigner when a new question type is selected', () => {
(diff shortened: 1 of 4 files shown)


ui-admin/src/forms/designer/NewQuestionForm.test.tsx
```diff
@@ -13,7 +14,7 @@ describe('NewQuestionForm', () => {
     const questionTypeSelect = screen.getByLabelText('Question type')
     expect(questionStableIdInput).toBeInTheDocument()
     expect(questionTypeSelect).toBeInTheDocument()
-    expect((questionTypeSelect as HTMLSelectElement).value).toBe('text')
+    expect((questionTypeSelect as HTMLSelectElement).value).toBe('placeholder')
```
nawatts:
```suggestion
    expect(questionTypeSelect).toHaveValue('placeholder')
```

`@testing-library/jest-dom` (imported in setupTests) has a lot of matchers that make assertions like this a bit clearer.

https://github.com/testing-library/jest-dom#table-of-contents
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-the-wrong-assertion

ui-admin/src/forms/designer/NewQuestionForm.test.tsx
```diff
@@ -28,4 +29,35 @@ describe('NewQuestionForm', () => {
...
+    render(<NewQuestionForm onCreate={() => jest.fn()} readOnly={false} />)
+
+    //Act
+    const questionTypeSelect = screen.getByLabelText('Question type')
+    fireEvent.change(questionTypeSelect, { target: { value: 'freetext' } })
+
+    //Assert
+    const freetextInput = screen.getByLabelText('Freetext')
+    expect((questionTypeSelect as HTMLSelectElement).value).toBe('freetext')
+    expect(freetextInput as HTMLInputElement).toBeInTheDocument()
```
nawatts: This is superfluous as `getBy...` queries will throw an error if they don't find a match.

ui-admin/src/forms/designer/NewQuestionForm.test.tsx
```diff
@@ -28,4 +29,35 @@ describe('NewQuestionForm', () => {
...
     expect(screen.getByText(questionTypeDescriptions.checkbox)).toBeInTheDocument()
   })
+
+  test('renders freetext input', () => {
+    //Arrange
+    render(<NewQuestionForm onCreate={() => jest.fn()} readOnly={false} />)
+
+    //Act
+    const questionTypeSelect = screen.getByLabelText('Question type')
+    fireEvent.change(questionTypeSelect, { target: { value: 'freetext' } })
```
nawatts: Suggest using `@testing-library/user-event` over `fireEvent`.
```ts
...
it('does something', async () => {
  const user = userEvent.setup()
  ...
  await user.type(questionTypeSelect, 'freetext')
```

https://testing-library.com/docs/user-event/intro#differences-from-fireevent
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event

ui-admin/src/forms/designer/NewQuestionForm.test.tsx
```diff
@@ -28,4 +29,35 @@ describe('NewQuestionForm', () => {
...
+  test('updates to the appropriate QuestionDesigner based on freetext input', async () => {
+    //Arrange
+    render(<NewQuestionForm onCreate={() => jest.fn()} readOnly={false}/>)
+
+    //Act
+    const questionTypeSelect = screen.getByLabelText('Question type')
+    fireEvent.change(questionTypeSelect, { target: { value: 'freetext' } })
+    const freetextInput = screen.getByLabelText('Freetext') as HTMLInputElement
+    userEvent.pointer({ target: freetextInput, offset: 0, keys: '[MouseLeft]' })
+    userEvent.keyboard('This is a question!\nThis is an option!\nThis is another option!')
```
nawatts: Most `userEvent` methods return Promises and should be awaited

```suggestion
    await userEvent.pointer({ target: freetextInput, offset: 0, keys: '[MouseLeft]' })
    await userEvent.keyboard('This is a question!\nThis is an option!\nThis is another option!')
```

https://testing-library.com/docs/user-event/pointer
https://testing-library.com/docs/user-event/keyboard

Awaiting these methods may allow getting rid of the `waitFor`s below.

Also wondering if this combination of `pointer` and `keyboard` could be replaced by [`type`](https://testing-library.com/docs/user-event/utility#type).

ui-admin/src/forms/designer/NewQuestionForm.tsx
```diff
@@ -16,9 +18,10 @@ type NewQuestionFormProps = {
 /** UI for creating a new question. */
 export const NewQuestionForm = (props: NewQuestionFormProps) => {
   const { onCreate, readOnly } = props
-  const [selectedQuestionType, setSelectedQuestionType] = useState<QuestionType>('text')
 
-  const [question, setQuestion] = useState<Question>(baseQuestions[selectedQuestionType])
+  const [question, setQuestion] = useState<Question>(baseQuestions['text'])
+  const [selectedQuestionType, setSelectedQuestionType] = useState<QuestionType | undefined>(undefined)
+  const [freetext, setFreetext] = useState<string | undefined>(undefined)
```
nawatts: It's a matter of style preference, but React's types have a shorthand for this. `useState<T>()` is equivalent to `useState<T | undefined>(undefined)`.

```suggestion
  const [selectedQuestionType, setSelectedQuestionType] = useState<QuestionType>()
  const [freetext, setFreetext] = useState<string>()
```

ui-admin/src/forms/designer/NewQuestionForm.tsx
```diff
@@ -36,22 +39,54 @@ export const NewQuestionForm = (props: NewQuestionFormProps) => {
             />
           </div>
           <label className="form-label" htmlFor="questionType">Question type</label>
-          <select id="questionType" className="form-select" value={selectedQuestionType}
+          <select id="questionType" defaultValue="placeholder" className="form-select" value={selectedQuestionType}
```
nawatts: It's a little unusual to mix `defaultValue` (usually used for uncontrolled components) and `value` (which makes a component controlled).
https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components

I think I would suggest removing `defaultValue` here and using a string state variable (initialized to `placeholder`) for the value of this select menu.

As is, it looks like selecting `freetext` here gets into a weird state where `selectedQuestionType` is no longer valid until something is entered in the free text field.

MatthewBemis:
Unit tests for the NewQuestionForm component have been cleaned up in 8a2e49f1

ui-admin/src/forms/designer/NewQuestionForm.test.tsx
```diff
@@ -1,31 +1,63 @@
...
+  test('updates to the appropriate QuestionDesigner when a new question type is selected', async () => {
     //Arrange
-    render(<NewQuestionForm onCreate={() => jest.fn()} readOnly={false} />)
+    render(<NewQuestionForm onCreate={() => jest.fn()} readOnly={false}/>)
 
     //Act
+    const user = userEvent.setup()
     const questionTypeSelect = screen.getByLabelText('Question type')
-    fireEvent.change(questionTypeSelect, { target: { value: 'checkbox' } })
+    await act(() => user.selectOptions(questionTypeSelect, 'checkbox'))
```
nawatts, line 25: Generally, `userEvent` methods should handle `act` internally. 
```suggestion
    await user.selectOptions(questionTypeSelect, 'checkbox')
```

MatthewBemis: I added the `act` because I was getting this console error when running the test locally (the test still passed regardless):

```
  console.error
    Warning: An update to NewQuestionForm inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
        at NewQuestionForm (/Users/mbemis/Code/broad/pearl/ui-admin/src/forms/designer/NewQuestionForm.tsx:20:11)
```

ui-admin/src/forms/designer/NewQuestionForm.test.tsx
```diff
@@ -1,31 +1,63 @@
...
+    const questionTypeSelect = screen.getByLabelText('Question type')
+    await act(() => user.selectOptions(questionTypeSelect, 'freetext'))
+
+    //Assert
+    screen.getByLabelText('Freetext')
+    expect(questionTypeSelect).toHaveValue('freetext')
+    expect(screen.getByText(questionTypeDescriptions.text)).toBeInTheDocument() //initial question type is text
+  })
+
+  test('updates to the appropriate QuestionDesigner based on freetext input', async () => {
```
breilly2, line 47: Something to think about in the future, especially if we continue to evolve the freetext parsing, might be to extract a function that takes the freetext and returns the resulting attribute bag. Unit tests could then easily be written against that function, and then the UI tests could check that the right UI is rendered for various attribute bags. That technique could increase test coverage while avoiding an explosion of permutations between the parsing function and UI rendering.

breilly2 (approved):
👍 Looks good. Definitely a time saver!

There are a couple of tough edge cases to think about as this feature evolves:
1. If you select freetext, enter some text, and then select freetext again, the freetext area is cleared, though the question type doesn't actually change and no data is lost.
2. There are some other quirks around changing back and forth between freetext and other input types. For example, switching to freetext when you have some bits already filled out gives you an empty freetext input. Then, entering anything into that freetext input erases any already entered content in favor of what it detected.
3. Entering freetext is a necessarily destructive operation. Having the input inline doesn't quite fit my mental model of how the data flows. It's basically a way to do bulk input (copy/paste), but the form itself is the source of truth. Put another way, we can't (today) run the function in reverse where we take the question properties and create a lossless text representation (other than JSON). It might be a more natural user experience to have the text input be modal to make it clear that it is destructive. We could give hints about how it's going to interpret the content (question type, list of options, etc).

MatthewBemis:
I improved the freetext stuff a bit in 8b175f0f

The freetext mode option is now separated from the question type selector. This makes it easier to handle the state and we can ensure that `selectedQuestionType` is always a valid `QuestionType`. The freetext is also preserved when switching between question types.

Screenshot 2023-07-31 at 5 29 04 PM

ui-admin/src/forms/designer/NewQuestionForm.test.tsx
nawatts: Just out of curiosity, I wonder if this changes if `userEvent.setup` is called before rendering the component.

> We recommend invoking [userEvent.setup()](https://testing-library.com/docs/user-event/setup) before the component is rendered.
https://testing-library.com/docs/user-event/intro#writing-tests-with-userevent

nawatts (approved):

ui-admin/src/forms/designer/NewQuestionForm.test.tsx
MatthewBemis: I tried it out and I still got the same warning, but I just pushed a commit to at least move the `userEvent.setup` call before the `render`, per the recommendation

PR #494, opened 2023-08-08
[JN-508] Fix navbar dropdown

MatthewBemis:
See https://broadworkbench.atlassian.net/browse/JN-508

Downsizing the window for ourhealthstudy.org collapses the navbar items into a hamburger menu. The hamburger button was no longer clickable, meaning users couldn't log in unless their browser was a certain minimum width. This most likely affected mobile users.

This seems to have been fallout from https://github.com/broadinstitute/juniper/pull/478, which changed the React Bootstrap versions. Isolated by testing the before and after of that commit.

### DIFF
--- a/ui-participant/src/Navbar.tsx
+++ b/ui-participant/src/Navbar.tsx
@@ -68,7 +68,7 @@ export default function Navbar(props: NavbarProps) {
       <button
         aria-controls={dropdownId} aria-expanded="false" aria-label="Toggle navigation"
         className="navbar-toggler"
-        data-bs-toggle="collapse" data-bs-target={`#${CSS.escape(dropdownId)}`}
+        data-bs-toggle="collapse" data-bs-target={`#${dropdownId}`}
         type="button"
       >
         <span className="navbar-toggler-icon"/>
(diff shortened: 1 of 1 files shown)


ui-participant/src/Navbar.tsx
```diff
@@ -68,7 +68,7 @@ export default function Navbar(props: NavbarProps) {
       <button
         aria-controls={dropdownId} aria-expanded="false" aria-label="Toggle navigation"
         className="navbar-toggler"
-        data-bs-toggle="collapse" data-bs-target={`#${CSS.escape(dropdownId)}`}
+        data-bs-toggle="collapse" data-bs-target={`#${dropdownId}`}
```
MatthewBemis, line 72: My guess is that the newer version of Bootstrap handles the escaping now, but I haven't found proof. I am wondering if there are other parts of the code that would need to be updated, if true (I haven't checked yet)

devonbush: It's the opposite -- I updated us to a later version and it broke things.  React `useId` deliberately includes weird characters that then throw off bootstrap.  So in addition to getting rid of the escaping, we should switch line 60 above to use lodash unique id instead `_uniqueId('navDropdown')`

devonbush (approved):

devonbush:
https://blog.openreplay.com/understanding-the-useid-hook-in-react/

breilly2 (approved):

PR #495, opened 2023-08-08
[JN-505] Save survey drafts to local storage

MatthewBemis:
After editing content, autosave is triggered every 60s:
Screenshot 2023-08-11 at 8 15 11 AM

Hitting the `Cancel` button in the survey editor prompts the user with a few options. `Exit & save draft` does one last save to local storage in case there were edits since the last auto-save occurred:
Screenshot 2023-08-11 at 8 23 32 AM

Loading the survey editor with an unsaved survey warns the user that they're editing a draft:
Screenshot 2023-08-11 at 8 15 31 AM

**Future improvements, out of scope:**

* Preventing version collisions in the backend with optimistic locking

**Scenarios to test:**

General behavior:
* Drafts do not start saving until you start making edits (and `Saving draft...` notice is visible for 2 seconds)
* Survey content and timestamp are updated correctly every time an autosave occurs
* Start editing a survey in a non-incognito browser window, and then in an incognito browser log in, make some different changes, and save to the DB. This will increment the survey version. Refresh the page as the original user, and the draft should *not* be loaded because the versions do not match

Reloading from a survey draft:
* Popup is shown to user alerting them that they are working off of a draft. Timestamp is provided for additional context
* Refreshing a survey successfully loads it from local storage
* `Save` button should be clickable, requiring no additional edits to enable. Clicking `Save` correctly persists to the DB

`Cancel` button behavior:
* Modal pops up with 3 options: `Exit & save draft`, `Exit & discard draft`, and `Cancel`. Clicking each option works as expected (verify by inspecting local storage)


ui-admin/src/study/surveys/SurveyEditorView.tsx
```diff
@@ -12,6 +13,25 @@ type SurveyEditorViewProps = {
...
+  return <Modal show={true} onHide={onDismiss}>
+    <Modal.Header>
+      <Modal.Title>Survey Draft Loaded</Modal.Title>
+    </Modal.Header>
+    <Modal.Body>
+      <p>Previously unsaved changes to this survey have been automatically loaded.
+        Be sure to save the survey in order to publish your changes.</p>
+    </Modal.Body>
+  </Modal>
+}
```
MatthewBemis: I ran out of time before going OOO tomorrow but I intend to break this out into it's own component and add a red button that allows the user to discard the draft if they don't want it. Right now they're forced to use it unless they're savvy enough to go into local storage to destroy the value

ui-admin/src/study/surveys/SurveyEditorView.tsx
```diff
@@ -12,6 +13,25 @@ type SurveyEditorViewProps = {
   onSave: (update: { content: string }) => Promise<void>
 }
 
+type Draft = {
+  content: string
+  date: number
```
MatthewBemis, line 18: `date` is currently unused. I want to use this info in the un-finished modal to tell the user when the draft is from so they can get some sort of basic context

ui-admin/src/study/surveys/SurveyEditorView.tsx
```diff
@@ -21,18 +41,67 @@ const SurveyEditorView = (props: SurveyEditorViewProps) => {
...
+        localStorage.setItem(FORM_DRAFT_KEY, JSON.stringify({ content: editedContentRef.current, date: saveDate }))
+        //Saving a draft happens so quickly that the "Saving draft..." message isn't even visible to the user.
+        //Set a timeout to show it for 2 seconds so the user knows that their drafts are being saved.
+        setTimeout(() => {
+          setSavingDraft(false)
+        }, 2000)
+      }
+    }
+
+    const draftSaveInterval = setInterval(saveToLocalStorage, 30000) //save draft every 30 seconds
```
MatthewBemis, line 80: Recommend setting this to ~5,000 when testing locally

ui-admin/src/forms/designer/utils/formDraftUtils.tsx
```diff
@@ -0,0 +1,42 @@
+import { VersionedForm } from '@juniper/ui-core'
+
+export type FormDraft = {
+  content: string
+  date: number
+}
+
+/** returns a form draft key for local storage */
+export function getFormDraftKey({ form }: { form: VersionedForm }) {
+  return `surveyDraft_${form.id}_${form.version}`
```
MatthewBemis: note to self: minor detail, but rename this and a few other places to `formDraft_...` for consistency and correctness

devonbush: I would also construct the key as form.stableId + form.version so that it's more human-readable, as opposed to using the id

ui-admin/src/forms/designer/utils/formDraftUtils.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
```
devonbush, line 12: I like how you've pulled all these functions out so it will be easier when we eventually move to having drafts saved server-side

ui-admin/src/study/surveys/SurveyEditorView.tsx
```diff
@@ -21,24 +24,66 @@ const SurveyEditorView = (props: SurveyEditorViewProps) => {
...
+      if (draftRef.current) {
+        saveDraft({
+          formDraftKey: FORM_DRAFT_KEY,
+          draft: draftRef.current,
+          setSavingDraft
+        })
+      }
+    }
+
+    const draftSaveInterval = setInterval(saveToLocalStorage, 60000)
```
devonbush: best practice is to avoid using setInterval, and instead set up a timeout chain -- this prevents race conditions where the next action starts before the previous one finishes, and minimizes the impact if the cleanup fails.  Checkout lines 132-144 of SurveyView.tsx in ui-participant for an example

ui-admin/src/study/surveys/SurveyEditorView.tsx
```diff
@@ -21,24 +24,66 @@ const SurveyEditorView = (props: SurveyEditorViewProps) => {
...
+  //for them to know if the version of the survey they're seeing are from a draft or are actually published.
+  const [showLoadedDraftModal, setShowLoadedDraftModal] = useState(!!getDraft({ formDraftKey: FORM_DRAFT_KEY }))
+  const [showDiscardDraftModal, setShowDiscardDraftModal] = useState(false)
+
+  const [draft, setDraft] = useState<FormDraft | undefined>(
+    !readOnly ? getDraft({ formDraftKey: FORM_DRAFT_KEY }) : undefined)
+
+  const isSaveEnabled = !!draft && isEditorValid && !saving
+
+  const draftRef = useRef<FormDraft | undefined>()
```
devonbush: having both draft as a state and a ref is confusing. I think you'll be able to avoid it if you use a somewhat non-intuitive mode of React's useState hooks where you call setDraft with a function.  Then React will always pass that function the latest value of the state thing, so you don't have to use a ref.  This is counter-intutitive because you'll be calling setDraft even when you don't want to change the value, but it's actually the best practice. 
```
  setDraft(currentDraft => {
    // currentDraft is guaranteed to be up-to-date 
    // do some stuff
    return currentDraft  // returning what you were passed in means the underlying value won't change
})
```

devonbush (changes requested):
This is good stuff--really appreciate the thoughtful approach around the loading and cancel dialogs -- it's giving a lot of useful information.  
Codewise, you should definitely upgrade the timeout infrastructure around drafts to match what we do for autosaving survey responses for participants.  And also investigate whether there a way to avoid having both a ref and a state variable holding the the draft .  
Functionality-wise, every 60 seconds seems a bit too far apart if someone is making changes in the doc.  that will cover the case where they idletimeout, but not if they inadvertently navigate away.  Unless/until we implement an "are you sure you want to leave?" alert, I'd suggest reducing the interval to every 10 seconds.  Related, it shouldn't be hard/expensive to do a string compare of the content to determine if subsequent saves are needed, so you can avoid flashing "saving..." if they haven't made any changes recently.

ui-admin/src/forms/designer/modals/DiscardLocalDraftModal.tsx
```diff
@@ -0,0 +1,54 @@
...
+        }}
+      >
+        Exit & save draft
+      </Button>
+      <Button
+        variant="danger"
+        onClick={() => {
+          deleteDraft({ formDraftKey })
+          onDismiss()
+          onExit()
```
breilly2, line 39: I'm always waffling between having modals perform actions vs invoking callback functions; it usually comes down to the specific situation and whether or not the modal is being reused in different contexts. I find it interesting that, in this case, both techniques are being used at the same time. I'm curious to know what your thoughts were around the decision for each of these (including `onSaveDraft`).

ui-admin/src/study/surveys/SurveyEditorView.tsx
MatthewBemis: Done in 8a38de640d3aea7ba1dc038d6fcf7aee2346a04a

ui-admin/src/study/surveys/SurveyEditorView.tsx
breilly2: Would that work? My guess is that the ref is to have the latest draft in global state where the autosave callback can fetch it. Otherwise, the autosave callback will capture the initial draft and never see any edits. I'm not sure how to get around that other than registering a new callback any time state changes, which would interfere with the timing loop.

ui-admin/src/study/surveys/SurveyEditorView.tsx
```diff
@@ -1,9 +1,13 @@
...
+import React, { useRef, useState } from 'react'
 
 import { VersionedForm } from 'api/api'
 
 import { Button } from 'components/forms/Button'
 import { FormContentEditor } from 'forms/FormContentEditor'
+import LoadedLocalDraftModal from 'forms/designer/modals/LoadedLocalDraftModal'
+import DiscardLocalDraftModal from 'forms/designer/modals/DiscardLocalDraftModal'
+import { deleteDraft, FormDraft, getDraft, getFormDraftKey, saveDraft } from 'forms/designer/utils/formDraftUtils'
+import { useAutosaveEffect } from '@juniper/ui-core/build/autoSaveUtils'
```
breilly2, line 10: IntelliJ refuses to like this. It's probably just my local environment -- it runs fine and IntelliJ is perfectly happy with the same import in ui-participant's SurveyView.tsx -- but I figured I'd mention it in case anyone else is seeing it.

breilly2 (commented):
Looking good. Adding some comments. May approve as-is depending on the result of the duplicate state conversation.

ui-admin/src/study/surveys/SurveyEditorView.tsx
devonbush: yeah, that's the idea -- you can't pass the draft itself because the callback will capture it and then get stale, but you can pass the setDraft function.  The set[X] function is guaranteed to always return the latest version whenever it is called, and set[X] itself doesn't change between renders

breilly2: Oh, I think I see. You're suggesting using the state setter to get the latest state value. Sneaky.

MatthewBemis: Ah TIL, that's an interesting way to do this. I took a shot at it in c11751e893365e1b22b13acc316deb2955753cd5

Also handles draft comparison to avoid useless autosaves

devonbush (approved):

ui-admin/src/study/surveys/SurveyEditorView.tsx
breilly2: Update: IntelliJ is happy now. 🤷‍♂️

breilly2 (approved):
I retested the problem cases we discovered over the past couple of days and everything seems to be working now. Very nice!

PR #499, opened 2023-08-10
[JN-510] persisting answer mappings along with surveys

devonbush:
This fixes a bug where answer mappings (the things that map some survey questions to profile fields) weren't being copied forward to new versions of the survey.  This is because the API service expects that the new version will specify any answer mappings, but the answer mappings for the existing survey were never being sent to the frontend.  To solve this, as well as to move our admin tool into more modular loading patterns, the survey editing component now loads the survey itself, along with the answer mappings, rather than relying on the initial portal blob fetch.    

This also does a drive-by fix of the "cancel" button navigation on the survey editing page

TO TEST:
1. repopulate ourhealth (just to make sure you don't have stale survey copies floating around)
2. in your DB, run `select * from answer_mapping;` and observe there are 9 rows.
3. edit the 'basics' survey `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms/surveys/oh_oh_basicInfo?readOnly=false`
4. save it
5. confirm the version is updated
6. in your DB, run `select * from answer_mapping;` and confirm there are now 18 rows.


ui-admin/src/study/surveys/SurveyView.tsx
```diff
@@ -53,30 +54,62 @@ function RawSurveyView({ studyEnvContext, survey, readOnly = false }:
...
+  if (!appliedVersion) {
+    return <span>The survey {stableId} is not already configured for this environment
+      -- you must specify a version</span>
   }
-  return <RawSurveyView studyEnvContext={studyEnvContext} survey={survey} readOnly={isReadOnly}/>
+  const { isLoading, survey } = useLoadedSurvey(portal.shortcode, stableId, appliedVersion)
+
+  return <>
+    { isLoading && <LoadingSpinner/> }
+    { !isLoading && <RawSurveyView studyEnvContext={studyEnvContext} survey={survey!} readOnly={isReadOnly}/> }
```
MatthewBemis: I'm mainly asking this for my own TypeScript learning: the non-null assertion on `survey` looks safe because the `useLoadedSurvey` implementation seems to guarantee that `isLoading` will never be `false` if `survey` is `undefined`, but is there a reason to not do the below suggestion instead? Is this a case where we view TypeScript more as a tool and don't want it to dictate/get in the way too much if we know better than it does? I'm still trying to dial-in on when circumventing TypeScript checks are appropriate.

```suggestion
    { !isLoading && survey && <RawSurveyView studyEnvContext={studyEnvContext} survey={survey} readOnly={isReadOnly}/> }
```

MatthewBemis (approved):
This looks good to me, just one question that's more of a curiosity. Thanks for starting the work on survey versions.

ui-admin/src/study/surveys/SurveyView.tsx
devonbush: good question, I could see that going either way.  I wrote it the way I did because  the parallel is/not loading structure, was what came to mind first, but I think either way would be appropriate.  Your suggestion is probably better since it's safer against upstream changes to useLoadedSurvey

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyService.java
```diff
@@ -139,6 +139,13 @@ public Survey createNewVersion(UUID portalId, Survey survey) {
         return create(newSurvey);
     }
 
+    public void attachAnswerMappings(Survey survey) {
+        survey.getAnswerMappings().clear();
+        survey.getAnswerMappings().addAll(
+                answerMappingDao.findBySurveyId(survey.getId())
```
breilly2: Why `clear` and `addAll` instead of `setAnswerMappings`? I don't think there's anything special about the collection instances (like there could be in a Hibernate/JPA application).

breilly2 (approved):
Looks good. Just one question (which might just be me blanking on something).

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyService.java
devonbush: yup, that is definitely my hibernate muscle-memory showing

PR #503, opened 2023-08-14
JN-518: fixing portal switching bug

devonbush:
Using the study selector to move between studies in two different portals was causing a couple of different problems -- there can be a race condition where if the study router updates before the portal router, the participant list will fail to load since it's trying to load the study from the wrong portal.  There was also just a straight-up stupid bug in StudyRouter that's been uncaught for months since we never pay attention to portals with more than one study.   In lieu of unit tests, I've updated our eslint to catch those fat-fingered equal signs.

TO TEST:
1. populate ourhealth and hearthive
2. go to https://localhost:3000/hearthive/studies/pacing/env/live/participants
3. use the study selector, and select the OurHealth study
4. confirm no error appears, and the participant list ui switches to OurHealth

### DIFF
--- a/ui-admin/src/login/IdleStatusMonitor.tsx
+++ b/ui-admin/src/login/IdleStatusMonitor.tsx
@@ -118,7 +118,7 @@ const InactivityTimer = ({ maxIdleSessionDuration, idleWarningDuration, doSignOu
 
   useEffect(() => {
     const targetEvents = ['click', 'keydown']
-    const updateLastActive = () => lastRecordedActivity.current = Date.now()
+    const updateLastActive = () => { lastRecordedActivity.current = Date.now() }
 
     if (!lastRecordedActivity.current) {
       updateLastActive()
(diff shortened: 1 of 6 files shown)


ui-admin/src/login/IdleStatusMonitor.tsx
```diff
@@ -118,7 +118,7 @@ const InactivityTimer = ({ maxIdleSessionDuration, idleWarningDuration, doSignOu
 
   useEffect(() => {
     const targetEvents = ['click', 'keydown']
-    const updateLastActive = () => lastRecordedActivity.current = Date.now()
+    const updateLastActive = () => { lastRecordedActivity.current = Date.now() }
```
breilly2, line 121: :( I wish the eslint check didn't require the extra ceremony around inline functions like this... especially in cases like this where typescript is saying that the given function returns `void` anyway.

breilly2 (approved):

MatthewBemis (approved):

PR #506, opened 2023-08-14
[JN-519] Fix "subsequent saves to surveys are lost"

MatthewBemis:
Currently fighting with jest mocks and the unit test is taking a bit longer to get working than I had hoped. Here's the fix for now and we can opt to merge it ahead of the unit test if we'd like.

To test:

Repro the case described in https://broadworkbench.atlassian.net/browse/JN-519


devonbush (approved):

breilly2 (approved):
Not clear to me why the old code behaved the way it did, but that's definitely some sketchy state updating. New code looks better.

PR #507, opened 2023-08-14
JN -513: adding new logos to ourhealth content

devonbush:
Adds the logos for Stanford and HMS to the home page.  Uses the bootstrap 'row-gap' class to handle the case where the flex styling wraps

image

TO TEST:
1. repopulate ourhealth
2. goto https://sandbox.ourhealth.localhost:3001/
3. confirm you see all four logos


breilly2:
I'm seeing no vertical padding between rows of logos.
Screenshot 2023-08-14 at 4 48 50 PM
My window is less wide than yours, so it's wrapping the logos differently, but I still see the same lack of vertical padding when I make my window wider.

breilly2:
Some interesting behavior when changing screen width, particularly right around 991px. Actually seems reasonably responsive for each width, even if it's a little odd to watch things jump around during resizing.

https://github.com/broadinstitute/juniper/assets/19228696/f517644e-26e5-4626-9190-1bf4b4614e33

Then finally, somewhere around 550px when the logos all stack vertically, they finally have some nice vertical spacing.

devonbush:
Interesting -- looks like the `row-gap` property isn't being honored in those screenshots.  Can you double-check that your auto-rebuild of ui-core triggered after you pulled the branch, and got picked up by ui-participant?  I wonder if the change to the underlying component adding that CSS class didn't get picked up?

breilly2:
That was it. Sorry for the false alarm.

breilly2 (approved):

MatthewBemis (approved):

devonbush:
Yeah, that ui-core rebuild trips me up all the time

PR #511, opened 2023-08-15
[JN-476] Added robust handling of errors from Pepper

breilly2:
Includes additional tests, including integration tests against dev pepper (which currently must be invoked manually).

To test:
1. Run all tests
2. Run integration tests: `./gradlew integrationTest`; note: the existing test intentionally makes a Pepper request that should fail to avoid accumulation of test data in Pepper

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeController.java
@@ -73,7 +73,7 @@ public class EnrolleeController implements EnrolleeApi {
       KitRequest sampleKit = enrolleeExtService.requestKit(adminUser, enrolleeShortcode, kitType);
       return ResponseEntity.ok(sampleKit);
     } catch (PepperException e) {
-      log.warn("Error requesting sample kit from Pepper", e);
+      log.error("Error requesting sample kit from Pepper", e);
       // In the case of a PepperException, we can do better than this because we'll likely know what
       // Pepper was unhappy
       // about, such as an address failing to validate.
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeController.java
@@ -7,14 +7,19 @@ import bio.terra.pearl.core.model.EnvironmentName;
 import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.kit.KitRequest;
 import bio.terra.pearl.core.model.participant.Enrollee;
+import bio.terra.pearl.core.service.kit.PepperException;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import java.util.UUID;
 import javax.servlet.http.HttpServletRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
 
 @Controller
 public class EnrolleeController implements EnrolleeApi {
+  Logger log = LoggerFactory.getLogger(EnrolleeController.class);
+
   private AuthUtilService authUtilService;
   private EnrolleeExtService enrolleeExtService;
   private HttpServletRequest request;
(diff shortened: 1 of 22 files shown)


core/build.gradle
```diff
@@ -46,12 +46,24 @@ dependencies {
...
 
     testFixturesImplementation 'org.springframework.boot:spring-boot-starter-test:2.7.10'
     testFixturesImplementation 'org.apache.commons:commons-text:1.10.0'
     testFixturesImplementation 'com.fasterxml.jackson.core:jackson-databind:2.14.1'
 }
 
 test {
-    useJUnitPlatform ()
+    useJUnitPlatform {
+        // By default, exclude `@IntegrationTest`s
```
breilly2, line 58: By default, I didn't want to run tests against a live external system, but I did want to have ready-to-run Pepper tests. I added `@IntegraitonTest` to mark tests with an `integration` tag so we could easily filter them in/out.

core/src/main/java/bio/terra/pearl/core/service/kit/LivePepperDSMClient.java
```diff
@@ -27,32 +29,39 @@ public LivePepperDSMClient(PepperDSMConfig pepperDSMConfig,
                                WebClient.Builder webClientBuilder,
                                ObjectMapper objectMapper) {
         this.pepperDSMConfig = pepperDSMConfig;
-        this.webClient = webClientBuilder
-                .baseUrl(this.pepperDSMConfig.getBasePath())
```
breilly2, line 31: Not locking-in `baseUrl` here makes it _much_ easier to configure differently for tests (i.e., to point at a local mock server).

populate/src/main/resources/seed/portals/demo/studies/heartdemo/enrollees/jsalk.json
```diff
@@ -79,7 +79,7 @@
   "kitRequestDtos": [
     {
       "creatingAdminUsername": "staff@heartdemo.org",
-      "kitTypeName": "saliva",
+      "kitTypeName": "SALIVA",
```
breilly2, line 82: Prior mis-configuration that will need to be updated in demo/prod.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeController.java
```diff
@@ -67,7 +72,11 @@ public ResponseEntity<Object> requestKit(
     try {
       KitRequest sampleKit = enrolleeExtService.requestKit(adminUser, enrolleeShortcode, kitType);
       return ResponseEntity.ok(sampleKit);
-    } catch (JsonProcessingException e) {
+    } catch (PepperException e) {
+      log.warn("Error requesting sample kit from Pepper", e);
+      // In the case of a PepperException, we can do better than this because we'll likely know what
```
devonbush, line 77: This seems like it should be logged as an error, rather than a warning, since it means we couldn't fulfill the main objective of the service call.

populate/src/main/resources/seed/portals/demo/studies/heartdemo/enrollees/jsalk.json
devonbush: I think your migration script will handle that -- the kit type name isn't stored directly anywhere except in the kit type table, right?

devonbush (approved):
couple questions, nothing blocking.  Overall looks good

core/src/test/java/bio/terra/pearl/core/service/kit/LivePepperDSMClientIntegrationTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Transactional
+    @IntegrationTest
+    public void testSendKitRequestParsesInvalidAddressError() throws Exception {
+        // Arrange
+        var enrollee = enrolleeFactory.buildPersisted("testSendKitRequestParsesInvalidAddressError");
+        var kitRequest = kitRequestFactory.buildPersisted("testSendKitRequestParsesInvalidAddressError", enrollee.getId());
+        var address = PepperKitAddress.builder()
+                .firstName("Juniper")
+                .lastName("Testerson")
```
MatthewBemis, line 27: Happy to see that the Testerson family has made it to Juniper 🙂

MatthewBemis (approved):

populate/src/main/resources/seed/portals/demo/studies/heartdemo/enrollees/jsalk.json
breilly2: Correct. My comment here predates the addition of the migration script (fast-follow commit to handle Github Action failures), so I agree that this should be all set.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeController.java
breilly2: Works for me. I was thinking of error as something critically wrong with Juniper itself, whereas here Juniper is fine, it's just something about the connection with Pepper that didn't work. I'm happy with either so I'll make the change.

PR #512, opened 2023-08-16
[JN-525] site content populate

devonbush:
Previously, there was no way to *just* update website content via the API, you had to update the whole portal.  This adds an API endpoint for accessing the siteContentPopulator directly.  It also adds a UX for accessing populate functions (visible only to superusers), so that we don't have to futz with the swagger API.

image

This PR only includes populating the new content--it does not handle attaching it to an environment.  That will be a later PR (which will include general version-selecting of siteContent)

I probably already got carried away too much on this ticket, so I had to stop, but there are definitely some things I hope we get to later, including:
1. Making it possible to overwrite a SiteContent object even while it's attached to a portal (similar to how we can do that for surveys) -- this is useful in development to quickly test changes without nuking your whole portal every time
2. some convenience improvements for the UX -- filenames should auto-suggest based on conventions, portal shortcodes could be prepopulated, etc.
3. a simplified "reset demo" button, available only in the demo environment, for Jeff/Mounica to press that resets everything

TO TEST: 

1. redeploy adminApiApp
2. login to the admin tool
3. choose "Populate" from the superuser functions
4. choose "Site content" from the nav
5. enter "ourhealth"  for portal shortcode
6. select "No" for overwrite
7. enter "portals/ourhealth/siteContent/siteContent.json"
8. presss "Populate"
9. confirm the UX displays a success message
10. in your database, `select id, stable_id, version from site_content` to confirm the new version was created

### DIFF
--- a/api-admin/src/main/resources/api/openapi.yml
+++ b/api-admin/src/main/resources/api/openapi.yml
@@ -753,6 +753,24 @@ paths:
                 type: object
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/internal/v1/populate/siteContent/{portalShortcode}:
+    post:
+      summary: Populates a siteContent to a given portal
+      tags: [ populate ]
+      operationId: populateSiteContent
+      parameters:
+        - { name: portalShortcode, in: path, required: true, schema: { type: string } }
+        - { name: filePathName, in: query, required: true, schema: { type: string } }
+        - { name: overwrite, in: query, required: false, schema: { type: boolean } }
+      responses:
+        '201':
+          description: siteContent object
+          content:
+            application/json:
+              schema:
+                type: object
+        '500':
+          $ref: '#/components/responses/ServerError'
   /api/internal/v1/populate/survey/{portalShortcode}:
     post:
       summary: Populates a survey to a given portal
--- a/populate/src/main/java/bio/terra/pearl/populate/service/StudyPopulator.java
+++ b/populate/src/main/java/bio/terra/pearl/populate/service/StudyPopulator.java
@@ -174,7 +174,7 @@ public class StudyPopulator extends BasePopulator<Study, StudyPopDto, PortalPopu
             StudyEnvironment destEnv = portalDiffService.
                     loadStudyEnvForProcessing(existingStudy.getShortcode(), EnvironmentName.sandbox);
             PortalEnvironment destPortalEnv = portalEnvironmentService
-                    .findOne(context.getPortalShortcode(), context.getEnvironmentName()).get();
+                    .findOne(context.getPortalShortcode(), EnvironmentName.sandbox).get();
             try {
                 var studyEnvChange = portalDiffService.diffStudyEnvs(existingStudy.getShortcode(),
                         sourceEnv, destEnv);
--- a/ui-admin/src/populate/PopulateControls.tsx
+++ b/ui-admin/src/populate/PopulateControls.tsx
@@ -1,72 +1,60 @@
-import React, {useState} from 'react'
-import {Link, NavLink, Outlet, Route, Routes} from "react-router-dom";
-import PopulatePortalControl from "./PopulatePortalControl";
-import Select from "react-select";
-import PopulateSurveyControl from "./PopulateSurveyControl";
-import PopulateSiteContentControl from "./PopulateSiteContent";
+import React, { useState } from 'react'
+import { Button } from 'components/forms/Button'
+import LoadingSpinner from 'util/LoadingSpinner'
 
-export default function PopulateControls({portalShortcode}: {portalShortcode?: string}) {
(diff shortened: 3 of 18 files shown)


api-admin/src/main/resources/api/openapi.yml
```diff
@@ -753,6 +753,24 @@ paths:
                 type: object
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/internal/v1/populate/siteContent/{portalShortcode}:
```
MatthewBemis, line 756: I realize this path is consistent with the other populate endpoints, but I think this would make more sense as `/api/internal/v1/populate/portals/{portalShortcode}/siteContent`

ui-admin/src/populate/PopulateControls.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { useState } from 'react'
+import { Button } from 'components/forms/Button'
+import LoadingSpinner from 'util/LoadingSpinner'
+
+/** renders a boolean dropdown for setting whether the populate is an overwrite command */
```
MatthewBemis:
```suggestion
/** renders a boolean radiogroup for setting whether the populate is an overwrite command */
```

ui-admin/src/populate/PopulateControls.tsx
```diff
@@ -0,0 +1,60 @@
...
+export const OverwriteControl = ({ text, isOverwrite, setIsOverwrite }:
+    {text: React.ReactNode, isOverwrite: boolean, setIsOverwrite: (bool: boolean) => void}) => {
+  return <div className="py-2">
+    <span className="fw-bold">Overwrite</span><br/>
+    <label className="me-3">
+      <input type="radio" name="humanReadable" value="false" checked={!isOverwrite}
```
MatthewBemis: Are the `humanReadable` names here and below leftover from copy/paste?

MatthewBemis (approved):
Just a few minor comments, nothing blocking. Didn't have an opportunity to test it out thoroughly, but I wanted to make sure you had the approval before I'm OOO.

populate/src/main/java/bio/terra/pearl/populate/service/StudyPopulator.java
```diff
@@ -174,7 +174,7 @@ protected Study createOrUpdate(Study existingStudy, StudyPopDto popDto, PortalPo
             StudyEnvironment destEnv = portalDiffService.
                     loadStudyEnvForProcessing(existingStudy.getShortcode(), EnvironmentName.sandbox);
             PortalEnvironment destPortalEnv = portalEnvironmentService
-                    .findOne(context.getPortalShortcode(), context.getEnvironmentName()).get();
+                    .findOne(context.getPortalShortcode(), EnvironmentName.sandbox).get();
```
breilly2, line 177: Did you mean to commit this change? Are we losing any functionality by hardwiring this to `sandbox` instead of getting it from the passed-in `context`?

ui-admin/src/populate/PopulateSiteContent.tsx
```diff
@@ -0,0 +1,42 @@
...
+            Uploads the SiteContent file to the database.  Does NOT attach it to a study.
+    </p>
+    <div className="d-flex flex-column row-gap-2">
+      {shortcodeControl}
+      {fileNameControl}
+      <OverwriteControl isOverwrite={isOverwrite} setIsOverwrite={setIsOverwrite}
+        text={<span>
+                    If no, the site content will
+                    be populated to the next available version number for its stable id.<br/>
+                    If yes, existing site content with the same stableId/version will be deleted`
```
breilly2: This was confusing to me on first read. What do you think of something like:
```suggestion
                    If no, a new version will
                    be created for its stableId.<br/>
                    If yes, the existing stableId/version will be replaced.`
```
(Similar rewording probably applies to other populate functions.)

breilly2 (approved):
Generally looks good. The most important question I have is about hardwiring to `sandbox`, but if that's intended then this is good with me.

api-admin/src/main/resources/api/openapi.yml
devonbush: I could see that -- the endpoints were originally structured this way because it was thought that multiple portals might share a siteContent (or parts of one).  We haven't yet found the use case for that though, and in the meantime, we'll leave as is

populate/src/main/java/bio/terra/pearl/populate/service/StudyPopulator.java
devonbush: Yes, this is deliberate.  See the comment on line 170 -- if we're not in overwrite mode, we only ever want to change the sandbox.  Updates to to irb/live environments must be done manually through the UX

PR #513, opened 2023-08-17
[JN-525] selecting site content version

devonbush:
This is step 2 of 3 in getting the ability to work with siteContents effectively.  This step adds the ability to select a version from the site content editor screen. This enables, e.g. going back to edit a past version.  We still do not have the ability to save changes or attach them to an environment (that will be step 3).

image

TO TEST:
0. repopulate ourhealth, and redeploy adminApiApp
1.  make a small change to the content in `populate/src/main/resources/seed/portals/ourhealth/siteContent/siteContent.json`
2. go to `https://localhost:3000/populate/siteContent`
3. fill in 'ourhealth' as the portal shortcode (note the filename now auto-populates)
4. press "populate"
5. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/siteContent`
6. Click the 'select' button next to the `Website content version 1` 
7. confirm the dialog appears as above
8. confirm 'version 2' is available to be selected
9. selected version 2, and press 'edit version 2'
10. confirm the content updates and you can see your change


populate/src/main/resources/seed/portals/ourhealth/siteContent/siteContent.json
```diff
@@ -13,7 +13,7 @@
       "sectionDtos": [{
         "sectionType": "HERO_WITH_IMAGE",
         "sectionConfigJson": {
-          "title": "Help us understand cardiovascular disease risk among South Asian populations.",
+          "title": "Help us 333 understand cardiovascular disease risk among South Asian populations.",
```
breilly2: Probably didn't mean to keep this change. :)

ui-admin/src/populate/PopulateControls.tsx
```diff
@@ -24,32 +24,25 @@ export const OverwriteControl = ({ text, isOverwrite, setIsOverwrite }:
...
-    fileName,
-    fileNameControl: <label className="form-label">
+export const FileNameControl = ({ fileName, setFileName }: {fileName: string, setFileName: (n: string) => void}) => {
+  return <label className="form-label">
             File path (from <code>/populate/src/main/resources/seed</code>)
-      <input type="text" value={fileName} className="form-control"
-        onChange={e => setFileName(e.target.value)}/>
-    </label>
-  }
+    <input type="text" value={fileName} className="form-control"
```
breilly2, line 30: Any reason not to use `TextInput` here?

Also, I could see this as a thin wrapper around a `TextInput` and its `label`, in which case its props API could be the same as `TextInput`, specifically directly passing through `value` and `onChange`. It's probably a toss-up, but all else being equal, I try to favor following existing patterns.

Same for `PortalShortcodeControl` below.

ui-admin/src/populate/PopulateControls.tsx
```diff
@@ -24,32 +24,25 @@ export const OverwriteControl = ({ text, isOverwrite, setIsOverwrite }:
 /**
  * renders a control for specifying a file name --this could be updated to handle the conventions of the various paths
  */
-export const useFileNameControl = () => {
-  const [fileName, setFileName] = useState('')
-  return {
-    fileName,
-    fileNameControl: <label className="form-label">
+export const FileNameControl = ({ fileName, setFileName }: {fileName: string, setFileName: (n: string) => void}) => {
```
breilly2, line 27: Nice. I think these make much more sense as components than hooks.

breilly2 (approved):
Other than the insert of "333", looks good to me. Only a minor question/suggestion.

ui-admin/src/populate/PopulateControls.tsx
devonbush: yeah, I was palying around with the hooks to see if it could make things more self-contained.  answer: no. :)

ui-admin/src/portal/siteContent/SiteContentVersionSelector.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const [versionList, setVersionList] = useState<SiteContent[]>([])
+  const [isLoading, setIsLoading] = useState(true)
+  const [selectedVersion, setSelectedVersion] = useState<SiteContent>()
+  const selectId = useId()
+  const versionOpts = versionList.map(content => ({
+    label: <span>
+            Version <strong>{ content.version }</strong>
+      <span className="text-muted fst-italic ms-2">
+                ({instantToDefaultString(content.createdAt)})
+      </span>
```
MatthewBemis, line 28: Definitely stealing this for the survey version selector. I had previously just shown version numbers but I like the context that the created date adds, and it'll be nice to have a consistent UX. I'll see if it makes sense to break this out into a re-usable component in my PR.

ui-admin/src/portal/siteContent/SiteContentView.tsx
```diff
@@ -112,7 +140,26 @@ const SiteContentView = ({ portalEnv, portalShortcode }: {portalEnv: PortalEnvir
     submitMailingListContact: () => Promise.resolve({})
   }
 
-  return <InitializedSiteContentView siteContent={portalEnv.siteContent} previewApi={previewApi}/>
+  const loadSiteContent = async (stableId: string, version: number) => {
+    setIsLoading(true)
+    Api.getSiteContent(portalShortcode, stableId, version).then(response => {
+      setSiteContent(response)
+      setIsLoading(false)
```
MatthewBemis, line 147: Right now this will spin forever if the API errors out. I know we show the error notification but I don't think we should give the impression that we're still attempting to load content.

Error handling is probably an area that could use some improvement and consistency across the admin tool, so this isn't blocking for this PR.

MatthewBemis (approved):

ui-admin/src/portal/siteContent/SiteContentView.tsx
devonbush: Good point -- we have a standard loading spinner, but we don't have a standard "empty page" display.  I'll see if I can put something in the next sprint for that

devonbush: JN-537 created

PR #514, opened 2023-08-21
JN-525 site content save

devonbush:
The "save" button now works on the website editor so you can save changes made in the UX.  Hitting save creates a new version of the site content, and then also updates the environment to point to the new version.  As we've talked about before, we eventually may want some 'save as draft' capability, but that's outside the scope of this PR.

In testing, I found that we were missing non-null constraints on siteContent stableId&version.  So that's fixed :)

This also does some drive-by cleanup of environment diffing routes so that the portal environment controller namespace was clearer for the new controller methods needed

TO TEST:
1. go to  https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/siteContent
2. make some changes to the site content
3. hit 'save'
4. confirm the changes are saved as a new version, and you see success notifications telling you the version was created *and* the environment was updated
5. reload the page, confirm your changes persist
6. go to https://localhost:3000/ourhealth/studies/ourheart/publishing
7. confirm you see your latest version as the sandbox version


api-admin/src/main/java/bio/terra/pearl/api/admin/service/siteContent/SiteContentExtService.java
```diff
@@ -31,6 +31,14 @@ public Optional<SiteContent> get(
     return Optional.empty();
   }
 
+  public SiteContent create(
+      String portalShortcode, String stableId, SiteContent siteContent, AdminUser user) {
+    Portal portal = authUtilService.authUserToPortal(user, portalShortcode);
+    siteContent.setPortalId(portal.getId());
+    siteContent.setStableId(stableId);
```
breilly2: Is it possible that `siteContent` will already have `stableId` set?

ui-admin/src/portal/siteContent/SiteContentLoader.tsx
```diff
@@ -0,0 +1,84 @@
...
+    } catch {
+      Store.addNotification(failureNotification('save failed'))
+      return
+    }
+    try {
+      let updatedEnv = {
+        ...portalEnv,
+        siteContentId: newVersion.id
+      }
+      updatedEnv = await Api.updatePortalEnv(portalShortcode, portalEnv.environmentName, updatedEnv)
```
breilly2, line 55: Suggestion: We could have an API endpoint that both creates a new version and updates the portal env. We would have more control over the transaction and be resilient to client connection issues. As is, if the first call succeeds and the second fails, unless there's a UI allowance for just the `updatePortalEnv` part, then we'd probably have to intervene to fix it.

breilly2 (approved):
Overall looks good. A couple of non-blocking comments.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/siteContent/SiteContentExtService.java
devonbush: it's possible, but overwriting it is fine.  the way to interpret this endpoint is "make a new version of this stableId, using the passed-in stieContent". So it might be that someone copied a sitecontent from another portal that has a different stableId, and then they're asking to save it as a new stableId (and it might have a different portalId too).   

Throughout this application, API endpoints are generally structured so that stableIds and portalShortcodes are on the path, so that it's easier to trace what's happening via access logs. This has the drawback here (and other places) of redundant fields in payloads.

ui-admin/src/portal/siteContent/SiteContentLoader.tsx
devonbush: The thought here is (especially since we don't have autosave yet for siteContents) make sure the save succeeds.  I'd rather have the user later have to use the version selector to switch to a version that was saved-but-not-connected than have their whole save fail, and then they'll lose work unless they can immediately solve why the environment update fails

devonbush:
(force push was a rebase of development to avoid merge conflicts)

MatthewBemis (approved):

PR #516, opened 2023-08-22
[JN-474] Fetch status from Pepper (DSM) by kit ID or study

breilly2:
This may not complete [JN-474](https://broadworkbench.atlassian.net/browse/JN-474) since I haven't yet been able to do any integration testing against Pepper. However, I'm getting anxious to get this code merged. It shouldn't require much if any additional changes, though I'll probably update the integration test before calling JN-474 done; expect one more PR for this ticket.

To test:
At this point, just run the test suite and see that it passes. Once we have more integration and UI built out, we'll be able to do better testing.

[JN-474]: https://broadworkbench.atlassian.net/browse/JN-474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentDao.java
@@ -106,24 +106,4 @@ public class StudyEnvironmentDao extends BaseMutableJdbiDao<StudyEnvironment> {
     public void deleteByStudyId(UUID studyId) {
         deleteByProperty("study_id", studyId);
     }
-
-    /**
-     * Returns the StudyEnvironment matching the given Pepper study name and environment name.
-     */
-    public Optional<StudyEnvironment> findByPepperStudyName(String pepperStudyName, EnvironmentName environmentName) {
-        var sql = """
-                    select %1$s
-                    from study_environment
-                    join study on study.id = %2$s.study_id
-                    where study.name = :pepperStudyName
-                    and %2$s.environment_name = :environmentName
-                    """
-                .formatted(prefixedGetQueryColumns(getTableName()), getTableName());
-        return jdbi.withHandle(handle ->
-                handle.createQuery(sql)
-                        .bind("pepperStudyName", pepperStudyName)
-                        .bind("environmentName", environmentName)
-                        .mapTo(clazz)
-                        .findOne());
-    }
 }
--- a/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentDao.java
@@ -107,7 +107,10 @@ public class StudyEnvironmentDao extends BaseMutableJdbiDao<StudyEnvironment> {
         deleteByProperty("study_id", studyId);
     }
 
-    public StudyEnvironment findByPepperStudyName(String pepperStudyName, EnvironmentName environmentName) {
+    /**
+     * Returns the StudyEnvironment matching the given Pepper study name and environment name.
+     */
+    public Optional<StudyEnvironment> findByPepperStudyName(String pepperStudyName, EnvironmentName environmentName) {
         var sql = """
                     select %1$s
                     from study_environment
@@ -121,6 +124,6 @@ public class StudyEnvironmentDao extends BaseMutableJdbiDao<StudyEnvironment> {
                         .bind("pepperStudyName", pepperStudyName)
                         .bind("environmentName", environmentName)
                         .mapTo(clazz)
-                        .one());
+                        .findOne());
     }
 }
--- a/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentDao.java
@@ -106,4 +106,21 @@ public class StudyEnvironmentDao extends BaseMutableJdbiDao<StudyEnvironment> {
     public void deleteByStudyId(UUID studyId) {
         deleteByProperty("study_id", studyId);
     }
+
+    public StudyEnvironment findByPepperStudyName(String pepperStudyName, EnvironmentName environmentName) {
+        var sql = """
+                    select %1$s
+                    from study_environment
+                    join study on study.id = %2$s.study_id
+                    where study.name = :pepperStudyName
+                    and %2$s.environment_name = :environmentName
+                    """
+                .formatted(prefixedGetQueryColumns(getTableName()), getTableName());
+        return jdbi.withHandle(handle ->
+                handle.createQuery(sql)
+                        .bind("pepperStudyName", pepperStudyName)
+                        .bind("environmentName", environmentName)
+                        .mapTo(clazz)
+                        .one());
+    }
 }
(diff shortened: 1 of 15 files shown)


core/src/main/java/bio/terra/pearl/core/service/kit/LivePepperDSMClient.java
```diff
@@ -131,17 +157,17 @@ private <T> Function<String, Mono<T>> deserializeTo(Class<T> clazz) {
...
-    private String basePath;
-    private String secret;
-
-    public PepperDSMConfig(Environment environment) {
-        this.basePath = environment.getProperty("env.dsm.basePath");
-        this.secret = environment.getProperty("env.dsm.secret");
+    @Component
+    @Getter
+    @Setter
+    public static class PepperDSMConfig {
```
breilly2, line 168: Moving this to a `static` inner class avoids a visibility warning. This is how it should have been originally... I was just a little out of practice.

core/src/main/java/bio/terra/pearl/core/service/kit/PepperDSMClient.java
```diff
@@ -19,5 +18,5 @@ public interface PepperDSMClient {
      */
     String sendKitRequest(Enrollee enrollee, KitRequest kitRequest, PepperKitAddress address);
     PepperKitStatus fetchKitStatus(UUID kitRequestId);
-    Collection<PepperKitStatus> fetchKitStatusByStudy(UUID studyId);
+    Collection<PepperKitStatus> fetchKitStatusByStudy(String pepperStudyName);
```
breilly2: Pepper wants a study name, not the Juniper study UUID.

core/src/test/java/bio/terra/pearl/core/service/kit/LivePepperDSMClientTest.java
```diff
@@ -60,23 +63,6 @@ void startup(TestInfo info) throws Exception {
         mockWebServer.shutdown();
     }
 
-    @Transactional
-    @Test
-    public void testSuccessResponseFromPepper() throws Exception {
```
breilly2, line 65: Renamed to `testSendKitRequest` and moved to be with the other success tests.

core/src/test/java/bio/terra/pearl/core/service/kit/LivePepperDSMClientTest.java
```diff
@@ -130,13 +116,86 @@ public void test5xxResponseFromPepper() throws Exception {
...
+                .build();
+        mockPepperResponse(HttpStatus.OK, objectMapper.writeValueAsString(kitStatus));
+
+        // Act
+        var response = client.sendKitRequest(enrollee, kitRequest, address);
+
+        // Assert
+        assertThat(response, containsString(kitStatus.getKitId()));
+        assertThat(response, containsString("New"));
+        verifyRequestForPath("/shipKit");
```
breilly2, line 135: The addition of this line is the only change to the body of this test (renamed from `testSuccessResponseFromPepper`).

core/src/test/java/bio/terra/pearl/core/service/kit/StubPepperDSMClientTest.java
```diff
@@ -0,0 +1,83 @@
...
+import bio.terra.pearl.core.service.study.StudyEnvironmentService;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasProperty;
+
+class StubPepperDSMClientTest extends BaseSpringBootTest {
```
breilly2, line 20: Might seem like overkill to have a test for a stub implementation, but it was a stepping stone to updating `PepperDSMClient.fetchKitStatusByStudy` to take a study name instead of a UUID.

core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentDao.java
```diff
@@ -106,4 +106,24 @@ public StudyEnvironment loadWithAllContent(StudyEnvironment studyEnv) {
     public void deleteByStudyId(UUID studyId) {
         deleteByProperty("study_id", studyId);
     }
+
+    /**
+     * Returns the StudyEnvironment matching the given Pepper study name and environment name.
+     */
+    public Optional<StudyEnvironment> findByPepperStudyName(String pepperStudyName, EnvironmentName environmentName) {
```
devonbush: Two thoughts:
1. This should be renamed to `findByStudyName(String studyName, ...)`. since that's what it's doing.  it's a service-level concern that pepperStudyName happens to coincide with study name
2. This should probably use study shortcode instead of name.  shortcode is guaranteed to be globally unique, whereas two studies can have the same name

core/src/main/java/bio/terra/pearl/core/model/study/Study.java
```diff
@@ -16,6 +16,7 @@
 public class Study extends BaseEntity {
     private String name;
     private String shortcode;
+    private String pepperStudyName;
```
devonbush: oh, this is an actual field -- did you mean to have the dao method search on this instead?

core/src/main/java/bio/terra/pearl/core/service/kit/PepperDSMClient.java
devonbush: how does this name get set?

populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/study.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "OurHealth",
   "shortcode": "ourheart",
+  "pepperStudyName": "juniper-ourhealth",
```
devonbush: this should use the study shortcode -- rather than the portal shortcode.  I wonder if the right move is to have the pepper study name always be `juniper-${studyShortcode}`.  That way we don't need an extra DB field, DSM gets an easy distinction, and we can hardcode the prefix to ensure consistency

devonbush (changes requested):
Nice -- agree that getting this code in will be good.  I'd like to learn more about the pepperStudyName constraints.  I also suspect there's a bug in the DAO

core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentDao.java
breilly2: Good catch! Yes, there is a bug in the DAO _and_ in the test. My intent was to use `pepperStudyName`, so this DAO method name is correct, but the implementation is wrong. However, based on further discussion with Pegah, it looks like we're going to use `shortcode` (or a simple derivative of `shortcode`) instead, so no extra field needed, and this DAO method will go away because we already have `findByStudy`.

populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/study.json
breilly2: That's a perfect middle-ground. We're going with `juniper-${studyShortcode}`.

devonbush (approved):

MatthewBemis (approved):

PR #517, opened 2023-08-22
[JN-452] Form version selector

MatthewBemis:
This adds a new "History" button next to the form version in the editor, allowing you to open previous versions in a new tab in the editor. Previous versions open in "Preview" mode.

Video of the version selector in use (now out of date, doesn't show open in new browser tab option):

https://github.com/broadinstitute/juniper/assets/7257391/a4b8c886-bd0e-421a-bd41-3abf103de006

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
@@ -44,6 +44,9 @@ public class SurveyExtService {
   public List<Survey> listVersions(String portalShortcode, String stableId, AdminUser adminUser) {
     Portal portal = authUtilService.authUserToPortal(adminUser, portalShortcode);
     // TODO: do we need to auth survey to portal here?
+    //This is used to power the version selector in the admin UI. It's not necessary to return the surveys
+    //with any content or answer mappings- the response will be too large. Instead, just get the individual
+    //versions as content is needed.
     List<Survey> surveys = surveyService.findByStableIdNoContent(stableId);
     return surveys;
   }
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
@@ -41,6 +41,13 @@ public class SurveyExtService {
     return survey;
   }
 
+  public List<Survey> listVersions(String portalShortcode, String stableId, AdminUser adminUser) {
+    Portal portal = authUtilService.authUserToPortal(adminUser, portalShortcode);
+    // TODO: do we need to auth survey to portal here?
+    List<Survey> surveys = surveyService.findByStableIdNoContent(stableId);
+    return surveys;
+  }
+
   public Survey create(String portalShortcode, Survey survey, AdminUser adminUser) {
     Portal portal = authUtilService.authUserToPortal(adminUser, portalShortcode);
     List<Survey> existing = surveyService.findByStableId(survey.getStableId());
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
@@ -49,7 +49,8 @@ public class SurveyExtService {
     // to return the surveys with any content or answer mappings, the response will
     // be too large. Instead, just get the individual versions as content is needed.
     List<Survey> surveys = surveyService.findByStableIdNoContent(stableId);
-    return surveys;
+    List<Survey> surveysInPortal = surveys.stream().map().filter(survey -> portal.getId().equals(survey.getPortalId())).toList();
+    return surveysInPortal;
   }
 
   public Survey create(String portalShortcode, Survey survey, AdminUser adminUser) {
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
@@ -43,17 +43,12 @@ public class SurveyExtService {
 
   public List<Survey> listVersions(String portalShortcode, String stableId, AdminUser adminUser) {
     Portal portal = authUtilService.authUserToPortal(adminUser, portalShortcode);
-    authSurveyToPortal(portal, stableId, 1); // TODO: this assumes there's always a v1. Bad?
-
     // This is used to populate the version selector in the admin UI. It's not necessary
     // to return the surveys with any content or answer mappings, the response will
     // be too large. Instead, just get the individual versions as content is needed.
     List<Survey> surveys = surveyService.findByStableIdNoContent(stableId);
     List<Survey> surveysInPortal =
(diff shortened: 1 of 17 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
```diff
@@ -41,6 +41,17 @@ public Survey get(String portalShortcode, String stableId, int version, AdminUse
     return survey;
   }
 
+  public List<Survey> listVersions(String portalShortcode, String stableId, AdminUser adminUser) {
+    Portal portal = authUtilService.authUserToPortal(adminUser, portalShortcode);
+    authSurveyToPortal(portal, stableId, 1); // TODO: this assumes there's always a v1. Bad?
```
MatthewBemis: I can't imagine that there's any reason to check that all survey versions are accessible, so checking the first version seems like it should be enough. But I'm not sure if it's dangerous to assume that there's always a v1.

Perhaps a better route is to slightly change the behavior of `authSurveyToPortal`, so it doesn't require a specific version number.

breilly2: Interesting problem. `authSurveyToPortal` just loads the survey to make sure it exists, verifying that the portal/stableId/version combination is valid. You _could_ add a variant of `authSurveyToPortal` that checks if there are _any_ survey versions, but that would eventually call `SurveyDao.findByStableId` which would fetch _all_ the data... and all this to `findByStableIdNoContent` below because you don't need all of the data.

Maybe we need a `boolean` `existsByStableId` to use for auth checks.

breilly2:
I noticed something during testing that may not be from this PR but is instead making it more apparent. If I create a new survey, it saves a new blank survey as v1 (which is fortunate for this PRs assumption that there's always a v1 😉). However, when opening the history selector, v1 does not have a date/time stamp. Looking in the database, `created_at` and `last_updated_at` are `1969-12-31 19:00:00.000000`, so I'm guessing those fields aren't being initialized and/or saved when creating a new survey.

Created https://broadworkbench.atlassian.net/browse/JN-540.

breilly2 (approved):
I think this looks good. Other than manual DB intervention, I couldn't find a case where there wouldn't be a v1 of a survey.

A killer feature would be a button to copy a version being previewed into the editor, which could then be tweaked and saved as a new version.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
devonbush: Yeah, let's move the portal id check after the findByStableId so we can check all the returned surveys -- it'll be more correct and more performant.  Suggestion below

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
```diff
@@ -41,6 +41,17 @@ public Survey get(String portalShortcode, String stableId, int version, AdminUse
...
+
+    // This is used to populate the version selector in the admin UI. It's not necessary
+    // to return the surveys with any content or answer mappings, the response will
+    // be too large. Instead, just get the individual versions as content is needed.
+    List<Survey> surveys = surveyService.findByStableIdNoContent(stableId);
+    return surveys;
```
devonbush:
```suggestion
    List<Survey> surveys = surveyService.findByStableIdNoContent(stableId);
    List<Survey> surveysInPortal = surveys.stream().map().filter(survey -> portal.getId().equals(survey.getPortalId())).toList();
    return surveysInPortal;
```

core/src/main/java/bio/terra/pearl/core/dao/survey/SurveyDao.java
```diff
@@ -24,11 +24,20 @@ public Optional<Survey> findByStableIdWithMappings(String stableId, int version)
         return surveyOpt;
     }
 
+    public List<Survey> findByStableIdNoContent(String stableId) {
+        List<Survey> surveys = jdbi.withHandle(handle ->
```
devonbush, line 28: since this is the second place we're fetching stuff without content, it might be worth pulling out a helper `getNonContentColumns()` method that returns a list of Strings for the columns these methods are interested in.

devonbush (approved):
Looks good.  Agree with Brian that the next step will be giving users some way to revert to prior versions.   

Two suggestions, definitely update the portal->survey auth check, and then the other is a maybe-nice-to-have

ui-admin/src/test-utils/mocking-utils.tsx
```diff
@@ -57,6 +57,28 @@ export const mockSurvey: () => Survey = () => ({
...
+  {
+    id: 'surveyId1',
+    stableId: 'survey1',
+    version: 2,
+    content: '{}',
+    name: 'Survey number one',
+    lastUpdatedAt: 0,
+    createdAt: 0
+  }
+])
```
devonbush: Using the existing mockSurvey() function will make it clearer which parameters you care about in this list. 

```suggestion
export const mockSurveyVersionsList: () => Survey[] = () => ([
  {
     ...mockSurvey(),
     id: 'surveyId1',
     stableId: 'survey1',
     version: 1
  },
  {
     ...mockSurvey(),
     id: 'surveyId2',
     stableId: 'survey1',
     version: 2
  }
])
```

ui-admin/src/study/surveys/VersionSelector.tsx
```diff
@@ -1,41 +1,109 @@
...
+        onClick={() => {
+          if (selectedVersion) {
+            loadVersion(selectedVersion)
+          }
+          setShow(false)
+        }}
+      >
+        View preview
+      </Button>
+      <Button
```
devonbush: I think it would be more friendly to have this be a link, rather than a button, since it's taking you somewhere

```
<a href={`${studyEnvFormsPath(
          studyEnvContext.portal.shortcode,
          studyEnvContext.study.shortcode,
          studyEnvContext.currentEnv.environmentName
      )}/surveys/${stableId}/${selectedVersion}?readOnly=true`}
        className="btn btn-secondary"
        aria-disabled={!selectedVersion}
         style={{pointerEvents: selectedVersion ? undefined : 'none'}}
        onClick={() => setShow(false)}
        target="_blank"
      >
        Open read-only editor <FontAwesomeIcon icon={faArrowRightFromBracket}/>
      </a>
```

devonbush (approved):
Looks good, a couple of non-blocking suggestions

breilly2 (approved):
Looks good!

One thing I noticed: When there's only one version, it's still available to select in the version selector, even though it is also the one currently being edited.

One concern: Encouraging users to open more windows increases the chances that they might have changes in two windows at the same time. The read-only view can be escaped by using the site navigation to return to the same survey. At some point, we should Implement the check that makes sure the version being edited is still the latest version in the database before saving changes. Fortunately, the auto-save feature will recover edits across tabs! Though the content from one tab will probably overwrite the content from another tab instead of keeping the two tabs in sync... something we might want to build some resilience for.

MatthewBemis:
@breilly2 leaving the current version as selectable in the dropdown was intentional. I thought it might be useful for users because it could allow them to reference the version that they're actively editing on top of.

Good point re: the concerns. I'll write up a ticket to capture these!

PR #518, opened 2023-08-22
[JN-538] Add error boundaries to form editor tabs

MatthewBemis:
To test: use the JSON editor to add an empty page with no elements in it.

Screenshot 2023-08-22 at 4 54 40 PM

### DIFF
--- a/ui-admin/src/forms/FormContentEditor.test.tsx
+++ b/ui-admin/src/forms/FormContentEditor.test.tsx
@@ -3,6 +3,7 @@ import React from 'react'
 
 import { FormContentEditor } from './FormContentEditor'
 
+//This is valid JSON, but invalid survey JSON
 const formContent: string = JSON.stringify({
   title: 'Test survey',
   pages: [
--- a/ui-admin/src/forms/FormContentEditor.test.tsx
+++ b/ui-admin/src/forms/FormContentEditor.test.tsx
@@ -12,7 +12,7 @@ const formContent: string = JSON.stringify({
 })
 
 describe('FormContentEditor', () => {
-  it('should trap errors in an ErrorBoundary', () => {
+  it('should trap FormDesigner errors in an ErrorBoundary', () => {
     // Arrange
     const { container } = render(<FormContentEditor
       initialContent={formContent} readOnly={false} onChange={jest.fn()}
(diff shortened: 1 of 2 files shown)


ui-admin/src/forms/FormContentEditor.test.tsx
```diff
@@ -0,0 +1,28 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+
+import { FormContentEditor } from './FormContentEditor'
+
+//This is valid JSON, but invalid survey JSON
```
MatthewBemis, line 6: Right now the JSON editor will prevent you from saving or leaving the editor tab if the JSON is invalid. I think the next step to improve this would be to give that same treatment to valid JSON that isn't valid _survey_ JSON.

MatthewBemis: Filed as https://broadworkbench.atlassian.net/browse/JN-539

Didn't want to put it into this PR because I think it's a bit more effort to make sure the edge cases are covered, and the priority with this PR is to prevent irrecoverable states.

ui-admin/src/forms/FormContentEditor.tsx
```diff
@@ -8,6 +8,7 @@ import { OnChangeFormContent } from './formEditorTypes'
 import { FormContentJsonEditor } from './FormContentJsonEditor'
 import { FormPreview } from './FormPreview'
 import { validateFormContent } from './formContentValidation'
+import ErrorBoundary from 'util/ErrorBoundary'
```
MatthewBemis, line 11: Nothing in this file actually changed other than wrapping the components in ErrorBoundary.

View without whitespace changes https://github.com/broadinstitute/juniper/pull/518/files?w=1

devonbush (approved):

devonbush:
nice, thanks for the quick fix

breilly2 (approved):
Nice! I like that the JSON editor is still available even if the designer and preview are unhappy.

breilly2:
@MatthewBemis @devonbush Is it expected that I'll still see this in dev?

Screenshot 2023-08-23 at 9 14 26 AM

If I dismiss that overlay, then I see what Matt included in the screenshot.

MatthewBemis:
@breilly2 I think that overlay is only shown when `NODE_ENV` is set to a development value. In prod and dev you shouldn't see these overlays. But I double-checked this by reproducing the "Cannot read properties of undefined" error in dev, and I got the same blank screen that Shriie had reported. The behavior of the error overlays aren't impacted by this PR.

PR #519, opened 2023-08-23
[JN-359] kit request UX update

devonbush:
UX update for the kit request page.  The main update is upgrading our table filters to take initial values, so that the table is pre-filtered to be only consented participants who have completed required activities who not already had a kit request.  This also adds a new synthetic participant who has completed required activities but hasn't gotten a kit.

image

This also does a drive-by of setting the default sort of the participant list to be by enrollee creation date.

TO TEST:
1. repopulate ourhealth
2. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/eligible`
3. confirm you see the table with filters prepopulated, and one participant "OHKITS" listed


ui-admin/src/study/kits/KitEnrolleeSelection.tsx
```diff
@@ -75,29 +81,20 @@ export default function KitEnrolleeSelection({ studyEnvContext }: { studyEnvCont
...
-              { value: false, label: 'Not Completed' }
-            ]
-          },
-          filterFn: 'equals',
-          cell: ({ row: { original: enrollee } }) => enrollee.taskCompletionStatus[survey.stableId]
-            ? <FontAwesomeIcon icon={faCheck}/>
-            : ''
-        }
-      })
+  const requiredSurveys = currentEnv.configuredSurveys.filter(survey => survey.required)
```
breilly2, line 85: This is such a 🤦‍♂️ moment. I don't know what I was thinking `required` meant in the survey model. Maybe that the survey is required for the user's data to be included in the dataset? I guess I'm still wondering if we want to model different requirements for being included in the dataset and kit eligibility.

breilly2 (approved):
Looks good. Thanks for helping out with this!

ui-admin/src/study/kits/KitEnrolleeSelection.tsx
devonbush: Yeah, I think eventually we'll need to have to support an explicit "eligibleForKit" rule string, since folks will want to filter on state geography and lots of other stuff.  but for now it's a nice coincidence that kit eligibility == finished required surveys

MatthewBemis (approved):

PR #520, opened 2023-08-23
[JN-460] Move bulk kit request batches from UI to server

breilly2:
Pretty straightforward change from `n` API calls to `1`. To test:
1. Go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/byEnrollee
2. Select multiple rows, click "Send sample collection kit" and "Request Kit"
3. Open the Kits tab (https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/byKit) and see that there are new kits (assuming a freshly populated database, there will be 1 more than the number of enrollees you selected)

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeController.java
@@ -5,9 +5,7 @@ import bio.terra.pearl.api.admin.service.AuthUtilService;
 import bio.terra.pearl.api.admin.service.enrollee.EnrolleeExtService;
 import bio.terra.pearl.core.model.EnvironmentName;
 import bio.terra.pearl.core.model.admin.AdminUser;
-import bio.terra.pearl.core.model.kit.KitRequest;
 import bio.terra.pearl.core.model.participant.Enrollee;
-import bio.terra.pearl.core.service.kit.PepperException;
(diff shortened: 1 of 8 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeController.java
```diff
@@ -75,8 +75,7 @@ public ResponseEntity<Object> requestKit(
     } catch (PepperException e) {
       log.error("Error requesting sample kit from Pepper", e);
       // In the case of a PepperException, we can do better than this because we'll likely know what
-      // Pepper was unhappy
-      // about, such as an address failing to validate.
```
breilly2, line 79: Some auto-reformatter kept splitting the first comment line at 80 characters. Didn't notice this in my prior PR.

devonbush (approved):
Nice.  I'd lean towards moving all the kit-related methods from the EnrolleeController and EnrolleeExtService into the KitController and KitExtService now that you have those.  The singular kit request methods are enrollee-scoped, but they resonate more strongly to me as "kit-related" and so that would be the first place I'd look for them

breilly2:
@devonbush Yes, I agree 100%. I got mixed up with that myself. I thought there was an argument at one point for putting them in an Enrollee class instead of a Kit class, but I'm probably mis-remembering that.

MatthewBemis (approved):

PR #521, opened 2023-08-23
[JN-459] Transform kit status column into tabs

breilly2:
https://broadworkbench.atlassian.net/browse/JN-459

Rough sketch closer to Erin's design sketch, where instead of a status column in the requested kits list, there are separate tabs for each status for seeing a list of those kits.

image

To test:
1. Start up the admin API and go to http://localhost:8080/swagger-ui.html#/populate/bulkPopulateEnrollees
2. Authorize and invoke `bulkPopulateEnrollees` with `portalShortcode`= `ourhealth`, `envName`= `sandbox`, `studyShortcode`= `ourheart`, and `numEnrollees`= something like `30`.
3. Start the admin UI and go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/byKit
4. Note the tabs for different statuses, with the count of kits in each status, and that clicking on the tabs sets the default column visibility to a set of columns that are appropriate for that status.


ui-admin/src/study/kits/KitList.tsx
```diff
@@ -1,33 +1,96 @@
...
+const pepperStatusToHumanStatus = (pepperStatus?: PepperKitStatus): string => {
+  const tab = statusTabs.find(tab => tab.status === pepperStatus?.currentStatus)
+  return tab?.label || pepperStatus?.currentStatus || '(unknown)'
 }
 
 /** Loads sample kits for a study and shows them as a list. */
 export default function KitList({ studyEnvContext }: { studyEnvContext: StudyEnvContextT }) {
   const { portal, study, currentEnv } = studyEnvContext
   const [isLoading, setIsLoading] = useState(true)
+  const [activeTab, setActiveTab] = useState<string | null>(statusTabs[0].status)
```
devonbush: this should be a url param so the tabs are linkable.  I could see this either being a path param or a query param

ui-admin/src/study/kits/KitList.tsx
```diff
@@ -42,16 +105,43 @@ export default function KitList({ studyEnvContext }: { studyEnvContext: StudyEnv
     loadKits()
   }, [studyEnvContext.study.shortcode, studyEnvContext.currentEnv.environmentName])
 
+  const kitsByStatus = _groupBy(kits, kit => {
+    return statusTabs.find(tab => tab.status === kit.pepperStatus?.currentStatus)?.status || 'ERROR'
+  })
+
   return <LoadingSpinner isLoading={isLoading}>
-    <KitListView kits={kits} studyEnvContext={studyEnvContext}/>
+    <Tabs
```
devonbush: I don't know of any reason to use bootstrap tabs if you're going to do mountOnEnter/unMountOnExit.  is there an accessibility benefit?  Otherwise, the code will be cleaner to just manage it ourselves

ui-admin/src/study/kits/KitList.tsx
```diff
@@ -42,16 +105,43 @@ export default function KitList({ studyEnvContext }: { studyEnvContext: StudyEnv
...
+      activeKey={activeTab ?? undefined}
+      mountOnEnter
+      unmountOnExit
+      onSelect={setActiveTab}
+    >
+      { statusTabs.map(tab => {
```
devonbush: yeah, a big benefit of not using bootstrap tabs is getting to avoid the 'map' here.  You can just render the tab you want, which means you won't need to use the lodash groupby

ui-admin/src/study/kits/KitList.tsx
```diff
@@ -42,16 +105,43 @@ export default function KitList({ studyEnvContext }: { studyEnvContext: StudyEnv
     loadKits()
   }, [studyEnvContext.study.shortcode, studyEnvContext.currentEnv.environmentName])
 
+  const kitsByStatus = _groupBy(kits, kit => {
```
devonbush, line 119: Rather than filtering here, we could let the table do the filtering for us.  That way we can give users the option to remove the filter and see all kits in a single list too if they want (or that could be another tab)

devonbush (changes requested):
nice, great to see this looking closer to the design.  Blocker is that the tabs need to be routable, other things are just suggestions

ui-admin/src/study/kits/KitList.tsx
breilly2: I think I see what you're suggesting... use the built-in react-table filtering, but hard code the filters into the "tab" UI instead of using the column filter control.

One other thing that we get out of this `_groupBy` is lists that we can show those counts in the tabs. Unless we can also get that from react-table somehow, I think we'll still need to use `_countBy` or something like it... plus, we'd need to make sure that logic perfectly matches react-table's filtering logic for the counts to be accurate.

devonbush: good point on the counts, that makes sense

breilly2:
Tabs are now routable. I do mostly like the other ideas suggested, but will address those in a follow-up PR.

MatthewBemis (approved):

ui-admin/src/study/kits/KitList.tsx
```diff
@@ -42,16 +103,59 @@ export default function KitList({ studyEnvContext }: { studyEnvContext: StudyEnv
...
+            <div className="py-3 px-5">
+              {kits?.length} {tab.label}
+            </div>
+          </NavLink>
+        })}
+      </div>
+      <Routes>
+        <Route index element={<Navigate to={statusTabs[0].label.toLowerCase()} replace={true}/>}/>
+        { statusTabs.map(tab => {
+          return <Route key={tab.label} path={tab.label.toLowerCase()} element={
```
devonbush: it feels backwards to base route paths off of labels.  If you don't want to create a separate "path" property for each tab, you could run the dependency in the other direction, and have the tabs have a "key" and then the label is the key title cased.

ui-admin/src/study/kits/KitList.tsx
```diff
@@ -1,27 +1,88 @@
...
+    label: 'Issues',
+    additionalColumns: [
+      'pepperStatus_labelDate', 'pepperStatus_trackingNumber',
+      'pepperStatus_scanDate', 'pepperStatus_returnTrackingNumber',
+      'pepperStatus_receiveDate',
+      'pepperStatus'
+    ]
+  }
+]
+
```
devonbush, line 83: a brief comment explaining what the intended visibility state is would be helpful

devonbush (approved):
looks good, thanks for making the tabs routable.  a couple style suggestions

PR #524, opened 2023-08-24
[JN-535] edit notification configs

devonbush:
This enables some properties of notification configs to be editable via UI.  Not the email templates, but the param values like max number of reminders.  (The backend supports creating a new email template, but updating the frontend was outside the scope of this ticket).  This also upgrades the notifcation config UX to match the recently restyled populate UX.

image

TO TEST:
1. go to http://localhost:3000/ourhealth/studies/ourheart/env/sandbox/notificationContent/configs
2. choose a config
3. update the Max reminders value
4. hit save
5. confirm the updated value is saved, and persists on refresh

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
@@ -1,30 +1,44 @@
 package bio.terra.pearl.api.admin.service.notifications;
 
 import bio.terra.pearl.api.admin.service.AuthUtilService;
+import bio.terra.pearl.core.dao.publishing.PortalEnvironmentChangeRecordDao;
 import bio.terra.pearl.core.model.EnvironmentName;
 import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.notification.NotificationConfig;
 import bio.terra.pearl.core.model.portal.Portal;
+import bio.terra.pearl.core.model.portal.PortalEnvironment;
+import bio.terra.pearl.core.model.publishing.PortalEnvironmentChange;
+import bio.terra.pearl.core.model.publishing.StudyEnvironmentChange;
 import bio.terra.pearl.core.model.study.PortalStudy;
 import bio.terra.pearl.core.model.study.StudyEnvironment;
 import bio.terra.pearl.core.service.notification.NotificationConfigService;
+import bio.terra.pearl.core.service.portal.PortalEnvironmentService;
+import bio.terra.pearl.core.service.publishing.PortalDiffService;
 import bio.terra.pearl.core.service.study.StudyEnvironmentService;
 import java.util.List;
+import java.util.UUID;
+
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
 
 @Service
 public class NotificationConfigExtService {
   private NotificationConfigService notificationConfigService;
   private AuthUtilService authUtilService;
   private StudyEnvironmentService studyEnvironmentService;
+  private PortalEnvironmentService portalEnvironmentService;
+  private PortalEnvironmentChangeRecordDao portalEnvironmentChangeRecordDao;
 
   public NotificationConfigExtService(
-      NotificationConfigService notificationConfigService,
-      AuthUtilService authUtilService,
-      StudyEnvironmentService studyEnvironmentService) {
+          NotificationConfigService notificationConfigService,
+          AuthUtilService authUtilService,
+          StudyEnvironmentService studyEnvironmentService, PortalEnvironmentService portalEnvironmentService,
+          PortalEnvironmentChangeRecordDao portalEnvironmentChangeRecordDao) {
     this.notificationConfigService = notificationConfigService;
     this.authUtilService = authUtilService;
     this.studyEnvironmentService = studyEnvironmentService;
+    this.portalEnvironmentService = portalEnvironmentService;
+    this.portalEnvironmentChangeRecordDao = portalEnvironmentChangeRecordDao;
   }
 
   public List<NotificationConfig> findForStudy(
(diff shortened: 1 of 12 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
```diff
@@ -1,30 +1,41 @@
...
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
 
 @Service
 public class NotificationConfigExtService {
   private NotificationConfigService notificationConfigService;
   private AuthUtilService authUtilService;
   private StudyEnvironmentService studyEnvironmentService;
+  private PortalEnvironmentService portalEnvironmentService;
+  private PortalEnvironmentChangeRecordDao portalEnvironmentChangeRecordDao;
```
breilly2: Unexpected DAO dependency, but it looks like it's not even being used.

ui-admin/src/study/notifications/NotificationConfigView.tsx
```diff
@@ -29,77 +31,116 @@ export default function NotificationConfigView({ studyEnvContext }: {studyEnvCon
...
-          </div>
         </div>
-        }
         <div>
-          <label className="form-label">Delivery
-            <Select options={deliveryTypeOptions}
-              value={deliveryTypeOptions.find(opt => opt.value === config.deliveryType)}/>
+          <label className="form-label">Repeat reminder after:
+            <div className="d-flex align-items-center">
+              <input className="form-control me-2" type="text" value={config.reminderIntervalMinutes}
```
nawatts: Could use `type="number"` here to get some stepper buttons.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number

ui-admin/src/study/notifications/NotificationConfigView.tsx
```diff
@@ -29,77 +31,116 @@ export default function NotificationConfigView({ studyEnvContext }: {studyEnvCon
...
+                onChange={e => setConfig(
+                  { ...config, reminderIntervalMinutes: parseInt(e.target.value) || 0 }
+                )}
+              />
+              minutes</div>
           </label>
         </div>
+        <div>
+          <label className="form-label">Max reminders:
```
nawatts: Some labels have a `:`, some don't. Is there a method to that?

ui-admin/src/study/notifications/NotificationContent.tsx
```diff
@@ -1,30 +1,68 @@
...
-            </Link>
-          </div>
-        </li>
-        ) }
-      </ul>
-      <button className="btn btn-secondary" onClick={() => alert('not yet implemented')}>
-        <FontAwesomeIcon icon={faPlus}/> Add
-      </button>
+  const navigate = useNavigate()
+  const [prevEnv, setPrevEnv] = useState<string>(currentEnv.environmentName)
```
nawatts: I think this name is unclear. Not sure if this is "previous environment" or "preview environment".

nawatts (approved):

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -476,6 +476,26 @@ paths:
           content: { application/json: { schema: { type: object } } }
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/portals/v1/{portalShortcode}/studies/{studyShortcode}/env/{envName}/notificationConfigs/{configId}:
+    patch:
```
breilly2, line 480: I'm not sure that `PATCH` is the right verb given that `configId` is a UUID and the effect is to create a new entity with a new UUID. I guess the alternative would be a `POST` to a generic update endpoint. Neither is really ideal, though I'm maybe 51% in favor of `POST` because it avoids me potentially making the wrong assumption about it's behavior when I see `PATCH`.

breilly2 (approved):
One quick thing to fix and one thought/suggestion, but otherwise looks good.

api-admin/src/main/resources/api/openapi.yml
nawatts: Technically, [PUT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT) might be most appropriate here.

breilly2: `PUT` should be idempotent. The client would have to choose the UUID `configId` to make it idempotent.

devonbush:
oops, sorry, I must have fat-fingered github and added Nick as a reviewer by mistake on this.  didn't mean to do that, but thanks for the great review, Nick!

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
devonbush: yup, leftover form refactor

api-admin/src/main/resources/api/openapi.yml
nawatts: Since the endpoint's URL includes `configId`, I assumed that was the ID of the entity being created/updated. Does this endpoint create an entity of another type that's related to the notification config with `configId`? If that's the case then, yeah, POST seems appropriate for this endpoint.

devonbush: I agree there's no clean answer to this -- the technical  operation is PATCH + POST, but we want to do them together because we don't want the study to not have a config if the PATCH succeeds but the POST fails.  And from a behavior perspective, PUT is also appropriate `replaces a representation of the target resource with the request payload`. Maybe I'll go with PUT since that's less common and so will clue the consumer that this is a complex operation

devonbush: On second thought, reading the docs on PUT, Brian, you're right they emphasize idempotency.  So I'm going to stick with PATCH, since that's the operation that's happening to the config identified by id in the URL (it's having it's `active` field set to false).

ui-admin/src/study/notifications/NotificationConfigView.tsx
devonbush: colons were leftover from a previous layout where they were in a horizontal row.  I think they're redundant with a vertical layout

PR #526, opened 2023-08-25
[JN-480] enabling mailing list contact delete

devonbush:
Enables deletion from the MailingList, along with creating an audit trail of the delete operation.  I'm not going to lie, I made the backend endpoint first, and then realized that the frontend UX made way more sense as multiselect, so there's a bit of a hack where the frontend just calls delete over and over again to do multi-deletes.  But given how rarely this feature is expected to be used, I don't think it's worth going back and changing the backend to support bulk deletes.  

This has some schema changes to DataChangeRecord to support auditing participant data that isn't directly tied to a PortalParticipantUser.   I suspect soon we'll want to add an explicit "operationType" enum to DataChangeRecord (CREATE, EDIT, DELETE), but for now I'm just marking DELETE as the fieldName.  Again, I didn't want to do a lot for a very out-of-critical-path feature.

image

TO TEST:
1. go to the sandbox mailingList for ourhealth https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/mailingList
2. select one or more entries
3. select "Remove"
4. confirm the deletion
5. observe the entries are deleted and the table is updated.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/MailingListExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/MailingListExtService.java
@@ -5,24 +5,38 @@ import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.portal.MailingListContact;
 import bio.terra.pearl.core.model.portal.Portal;
 import bio.terra.pearl.core.model.portal.PortalEnvironment;
+import bio.terra.pearl.core.model.workflow.DataChangeRecord;
+import bio.terra.pearl.core.service.CascadeProperty;
+import bio.terra.pearl.core.service.exception.PermissionDeniedException;
 import bio.terra.pearl.core.service.portal.MailingListContactService;
 import bio.terra.pearl.core.service.portal.PortalEnvironmentService;
+import bio.terra.pearl.core.service.workflow.DataChangeRecordService;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.List;
+import java.util.UUID;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
 
 @Service
 public class MailingListExtService {
   private AuthUtilService authUtilService;
   private MailingListContactService mailingListContactService;
   private PortalEnvironmentService portalEnvironmentService;
+  private DataChangeRecordService dataChangeRecordService;
+  private ObjectMapper objectMapper;
 
   public MailingListExtService(
       AuthUtilService requestUtilService,
       MailingListContactService mailingListContactService,
-      PortalEnvironmentService portalEnvironmentService) {
+      PortalEnvironmentService portalEnvironmentService,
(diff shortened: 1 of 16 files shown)


ui-admin/src/portal/MailingListView.tsx
```diff
@@ -78,33 +83,77 @@ export default function MailingListView({ portalContext, portalEnv }:
...
 
       {basicTableLayout(table)}
+      { showDeleteConfirm && <Modal show={true} onHide={() => setShowDeleteConfirm(false)}>
+        <Modal.Body>
+          <div>Do you want to delete the <strong>{ numSelected }</strong> selected entries?</div>
+
+          <div className="pt-3">This operation CANNOT BE UNDONE.</div>
+        </Modal.Body>
+        <Modal.Footer>
+          <button type="button" className="btn btn-primary" onClick={performDelete}>
```
MatthewBemis:
```suggestion
          <button type="button" className="btn btn-danger" onClick={performDelete}>
```

Since this is a destructive action, I think it would be appropriate to style this as `danger`

MatthewBemis (approved):
Looks good. Just one suggestion

api-admin/src/main/java/bio/terra/pearl/api/admin/service/MailingListExtService.java
```diff
@@ -32,4 +46,30 @@ public List<MailingListContact> getAll(
...
+              .modelName(MailingListContact.class.getSimpleName())
+              .responsibleAdminUserId(user.getId())
+              .portalEnvironmentId(portalEnv.getId())
+              .newValue(null)
+              .oldValue(objectMapper.writeValueAsString(contact))
+              .build();
+      dataChangeRecordService.create(changeRecord);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException("could not create audit trail", e);
+    }
```
breilly2, line 72: I expected to see this in `MailingListContactService`. We probably don't want to `delete` without writing a change record, so we may as well limit the possibility of that happening.

breilly2 (approved):
One suggestion, but I'll leave it up to you.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/MailingListExtService.java
devonbush: I thought about that -- the trouble is that we sometimes DO need to delete without audit trail, like when we're doing populates.  And right now the expectations of the core services is "I just do whatever CRUD operations you tell me to".    So I put it here for now, you're right that this isn't yet the optimal configuration.

breilly2: That's what I figured. I just feel like we keep pushing business logic out into additional layers. DAOs are simple, so I figured Services should be smarter. But if Services are just simple layers over DAOs then I'm not sure what value they add. And if ExtServices start holding the business logic, then it's doing that as well as handling authorization... but I thought the goal of the ExtServices was to be simple proxies to Services that added authorization.

My final argument on the specific point of populate is that that seems like enough of a special case that it could use DAOs directly instead of having to go through Services.

PR #529, opened 2023-08-29
[JN-539] Form JSON editor validation

MatthewBemis:
This detects a bunch of different cases:

Invalid JSON:
Screenshot 2023-08-30 at 2 18 08 PM

Invalid question template references:
Screenshot 2023-08-30 at 2 17 58 PM

Elements with missing types:
Screenshot 2023-08-30 at 2 18 31 PM

Elements with missing names:
Screenshot 2023-08-30 at 2 18 46 PM

Elements with duplicate names:
Screenshot 2023-08-30 at 2 19 17 PM

Forms without pages:
Screenshot 2023-08-30 at 2 21 18 PM

Pages without elements:
Screenshot 2023-08-30 at 2 21 29 PM

Panels without elements:
Screenshot 2023-08-30 at 2 21 37 PM

TO TEST:

* Try out all of the above test cases
* Make sure that you can correct the errors and everything still saves correctly
* Make sure that the Designer and Preview tabs are disabled while there are errors, and that the Save button is disabled
* Make sure that you don't encounter any of these errors while playing around in the Designer tab (there are enough guardrails in place, so you shouldn't be able to trigger any of these errors)

### DIFF
--- a/ui-admin/src/study/surveys/SurveyEditorView.tsx
+++ b/ui-admin/src/study/surveys/SurveyEditorView.tsx
@@ -34,7 +34,7 @@ const SurveyEditorView = (props: SurveyEditorViewProps) => {
   const FORM_DRAFT_KEY = getFormDraftKey({ form: currentForm })
   const FORM_DRAFT_SAVE_INTERVAL = 10000
 
-  const [isEditorValid, setIsEditorValid] = useState(true)
+  const [validationError, setValidationError] = useState<string>()
   const [saving, setSaving] = useState(false)
   const [savingDraft, setSavingDraft] = useState(false)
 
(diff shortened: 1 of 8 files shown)


ui-admin/src/study/surveys/SurveyEditorView.tsx
```diff
@@ -101,6 +103,28 @@ const SurveyEditorView = (props: SurveyEditorViewProps) => {
           </h5>
         </div>
         { savingDraft && <span className="detail me-2 ms-2">Saving draft...</span> }
+        { validationErrors && !isEmpty(validationErrors) &&
```
devonbush: based on my read of the docs, isEmpty returns true for null/undefined
```suggestion
        { !isEmpty(validationErrors) &&
```

devonbush (commented):
Looks great -- tested locally.  Love the warnings when template names don't match.  Probably the biggest future win would be adding a caret position indicator (like in your IDE) that tells them where the caret is so they can narrow down the location faster.  But the feedback on errors is so fast that might not even be necessary

devonbush (approved):

breilly2 (approved):
Looks good! Easy to identify where additional validation logic can be added in the future.

PR #530, opened 2023-08-29
[JN-517] Suggest a value for question choices

MatthewBemis:
https://github.com/broadinstitute/juniper/assets/7257391/9541225c-3ff7-4184-8eb8-d4618dce5dde

### DIFF
--- a/ui-admin/src/forms/designer/questions/ChoicesList.test.tsx
+++ b/ui-admin/src/forms/designer/questions/ChoicesList.test.tsx
@@ -78,6 +78,28 @@ describe('ChoicesList', () => {
     })
   })
 
+  it('automatically generates values based on text', () => {
+    // Arrange
+    const onChange = jest.fn()
+    render(<ChoicesList question={question} readOnly={false} onChange={onChange} />)
+
+    const barChoice = screen.getAllByRole('listitem')[1]
+
+    // Act
+    const barLabelInput = getByLabelText(barChoice, 'Text')
+    fireEvent.change(barLabelInput, { target: { value: 'This is a test question' } })
+
+    // Assert
+    expect(onChange).toHaveBeenCalledWith({
+      ...question,
+      choices: [
+        { value: 'foo', text: 'Foo' },
+        { value: 'thisIsATestQuestion', text: 'This is a test question' },
+        { value: 'baz', text: 'Baz' }
+      ]
+    })
+  })
+
   it('allows reordering choices', async () => {
     // Arrange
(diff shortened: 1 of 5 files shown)


ui-admin/src/forms/designer/questions/ChoicesList.test.tsx
```diff
@@ -78,6 +78,28 @@ describe('ChoicesList', () => {
...
+  it('automatically generates values based on text', () => {
+    // Arrange
+    const onChange = jest.fn()
+    render(<ChoicesList question={question} readOnly={false} onChange={onChange} />)
+
+    const barChoice = screen.getAllByRole('listitem')[1]
+
+    // Act
+    const barLabelInput = getByLabelText(barChoice, 'Text')
+    fireEvent.change(barLabelInput, { target: { value: 'This is a test question' } })
```
devonbush, line 90: `fireEvent` is now deprecated.  New tests should use userEvent, which provides more realistic simulation of actual user input via a browser.  See https://testing-library.com/docs/user-event/intro


```suggestion
    userEvent.type(barLabelInput, 'This is a test question' )
```

devonbush (approved):
nice! tested locally, works great

breilly2:
I'm not sure that I like an edited value being auto-updated when editing the text. I might choose a value different than the suggestion and then reword or fix a typo in the text and not realize that it updated the value. I think this is particularly dangerous if the survey has been saved, especially if there are already responses. This feature is a great accelerator for the straightforward case of creating new items, but once an item exists, I think we should turn off the automatic updating.

MatthewBemis:
@breilly2 that's a great point, I'll tweak the behavior of this to not override values that have already been populated

JeffKorte:
Looks good!  Excited to get this into production.

JeffKorte (approved):

MatthewBemis:
@breilly2 I've updated the behavior a bit in this commit: 87a18302d8a9088809336e3eb010fc3e5b4de85b

The idea is that the autofill is only enabled if the question is a new question and the user hasn't tweaked the choice value to differ from what would've been autofilled.

We discussed a bit over Slack DM, but to make the context public, it was pretty difficult to differentiate between all of the various cases when the autofill should be enabled in a way that didn't require adding a bunch of new parent component state or refactoring the onChange handlers up through the entire survey editor.

I think this still gets us most of the value because it still enables the feature for people who are building out brand new questions.

breilly2 (approved):
Looks good!

PR #531, opened 2023-08-29
[JN-548] Add Download JSON button to form editor

MatthewBemis:
Screenshot 2023-08-30 at 10 12 06 AM

To test:

* Click the button and verify that it downloads properly formatted JSON


breilly2 (approved):
Looks good. One suggestion is to add some indication, maybe including a timestamp, to the filename when downloading edits that haven't been saved. Otherwise, the filename could be read as being the definitive content for that version.

devonbush (approved):
Agree with Brian's suggestion --another option is that if it has been modified, just slap a `-draft` or `-modified` to the end of the filename.

MatthewBemis:
Updated filenames to differentiate between final and draft versions:

Screenshot 2023-09-08 at 11 10 23 AM

During demo last week we had discussed burying this in an "Advanced" dropdown menu, but Erin suggested holding off on that for now, so merging as-is.

PR #532, opened 2023-09-01
JN-552 published version names

devonbush:
This adds "published_version" to all versioned entities.  This is a 'cosmetic' version intended to help make the published dataset cleaner.  The UI display of the published version is very rough, but that can be refined later, the important thing for this PR is to get the data infrastructure in place.  For now, the published version is auto-generated.  The critical thing is to never rely on the published version for application logic.  That will ensure that published and unpublished surveys/content behave the same, and therefore the sandboxes will accurately mirror production.

Along the way, I upgraded our publishing services to make them more typesafe and add proper integration tests in place of the prior mocked-out test. 

TO REVIEW:
1. The real action is the `assignPublishedVersionIfNeeded` methods in PublishingUtils.java, and how they are called from StudyPublishingService and PortalPublishingService.  Start with those methods and move outwards.
2. The majority of this PR is a bunch of boring refactoring (adding generics, pulling out a base class, extracting utility methods, etc...) that can be skimmed. 

TO TEST:
1. repopulate OurHealth
4. Go to https://localhost:3000/ourhealth/studies/ourheart/publishing 
5.  copy the sandbox environment to the irb environment.  
6. copy the irb environment to the live environment
7. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms/preEnroll/oh_oh_preenroll?readOnly=false
8. Confirm the survey now shows a published version
image


breilly2:
Testing notes: On the publish page, the publishing of the content isn't immediately reflected. I noticed when publishing from sandbox to IRB and captured a screenshot when publishing to live:

Screenshot 2023-09-01 at 9 51 55 AM

After reloading the page, the result of publishing was visible:

Screenshot 2023-09-01 at 9 52 10 AM

Similarly, viewing the edited survey in the environment that was published to continues to show the old version. Again, reloading fixed this.

core/src/main/java/bio/terra/pearl/core/dao/BaseVersionedJdbiDao.java
```diff
@@ -20,6 +22,7 @@ public List<T> findByStableId(String stableId) {
         return findAllByProperty("stable_id", stableId);
     }
 
+    /** gets 1 plus the previous highest version, or 1 if no published version already exists */
```
breilly2:
```suggestion
    /** gets 1 plus the previous highest version, or 1 if no version already exists */
```
Or some other minor change so that it's not exactly the same as the docs for `getNextPublishedVersion`.

core/src/main/java/bio/terra/pearl/core/model/Versioned.java
```diff
@@ -1,7 +1,16 @@
 package bio.terra.pearl.core.model;
 
+/**
+ * Versioned is used for customer-editable documents.  All documents must have a stableId and version, and the stableID-version
+ * pair must be unique.  stableId-publishedVersion pairs must also be unique, but not every document is required to have a publishedVersion
+ *
+ * PublishedVersion is only intended as a cosmetic feature for users to "tag" documents.  All application logic and data fetching should be based on
+ * version.  this ensures that the application will operate the same in the sandbox environment as it does in live environments.
+ */
```
breilly2, line 9: We talked about source control analogues in person. Another Analogue I wanted to mention is Confluence URLs. For example, in `https://broadworkbench.atlassian.net/wiki/spaces/PEARL/pages/2842230785/Code+coverage`, `2842230785` is the stable document ID while `Code+coverage` is the title. The page title can be changed, thus changing the URL, and it will still load the same document. https://broadworkbench.atlassian.net/wiki/spaces/PEARL/pages/2842230785 always works to load the document, and Confluence doesn't even care if the title is or ever was correct, so https://broadworkbench.atlassian.net/wiki/spaces/PEARL/pages/2842230785/foo also works! In both cases, the browser is redirected to the document with the current title in the URL.

core/src/main/java/bio/terra/pearl/core/service/VersionedEntityService.java
```diff
@@ -1,9 +1,44 @@
...
+    }
 
-public interface VersionedEntityService<T> {
-    Optional<T> findByStableId(String stableId, int version);
+    @Transactional
+    public void assignPublishedVersion(UUID id) {
+        T entity = dao.find(id).get();
+        if (entity.getPublishedVersion() != null) {
+            // this survey already has a published version, do not reassign
+            return;
```
breilly2, line 37: This feels like an unexpected case. Are there sufficient checks upstream of this that we know we're saving a new version that we can make this an error? Or at least log a warning that that this is an unexpected event? In the `assignPublishedVersionIfNeeded` functions that call this, what are the situations in which this is needed vs not needed, and why wouldn't the caller know for sure?

breilly2 (approved):
Mostly one question/concern in `VersionedEntityService`, which might be fine or might benefit from a few more code comments, but I won't block approval on it. Otherwise looks good. I'm glad this strategy worked out! I much prefer this mental model to removing intermediate versions when publishing.

core/src/main/java/bio/terra/pearl/core/service/VersionedEntityService.java
devonbush: it's up to the caller whether this is unexpected.   Where this might occur is if OurHealth releases v1 of a survey to live, then v2, and then decides to revert back to v1 of the survey again.  When they copy irb->live to make the old version live again, the survey will already have a published version, and so no assignment is needed.  So no error handling is needed there.  You are totally right that if we ever let humans get their hands on assigning versions, we'll want to report stuff to the UI

MatthewBemis (approved):

PR #533, opened 2023-09-01
JN-550 sandbox safety

devonbush:
This puts surveys and site contents into read-only mode if a non-sandbox environment is selected.  It also enhances the service-layer safeguards so that now not even superusers can save to non-sandbox environments.  We can always update the DB directly, and this add even more misclick-safety

TO TEST:
1. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms/preEnroll/oh_oh_preenroll`
2. make a small change in the survey
3. observe the save button is enabled
4. use the environment selector to go to 'irb' or 'live
5. confirm the survey is put into read-only mode


ui-admin/src/portal/siteContent/SiteContentEditor.tsx
```diff
@@ -101,7 +103,10 @@ const SiteContentEditor = (props: InitializedSiteContentViewProps) => {
         <button className="btn btn-secondary" onClick={() => alert('not yet implemented')}>
           <FontAwesomeIcon icon={faPlus}/> Add page
         </button>
-        <button className="btn btn-primary ms-auto" onClick={() => createNewVersion(workingContent)}>Save</button>
+        {!readOnly && <Button className="ms-auto" variant="primary"
+          onClick={() => createNewVersion(workingContent)}>
+          Save
+        </Button>}
```
breilly2, line 109: I think this should also hide the Cancel button. If there's no Save button, then there's no context for what one might be cancelling, and clicking the Cancel button confusingly (I think) navigates back to the sandbox.

Also, the Cancel button strangely jumps over to the left when Save is suppressed.

breilly2:
Testing note: When changing to IRB or live after making an edit, I think it's confusing to continue showing the edit. It makes it look like the change exists in those environments. Even more confusingly, switching back to sandbox then erases those changes without warning.

breilly2 (approved):
I ran into some weirdness in testing, but this is an important thing to fix and this PR is an improvement.

MatthewBemis (approved):
:+1:

This is unrelated to the PR, but while testing I noticed that the environment switcher doesn't handle uninitialized environments very well. In the Survey Editor this leads to JS errors, and in the Site Content editor this leads to content from a different environment being displayed.

Filed as https://broadworkbench.atlassian.net/browse/JN-560

devonbush:
thanks for the great testing -- one thing that this has made me think about is whether toggling the environment should add an entry to the history state.  e.g. if you switch to the irb env, when you hit "back" do you want to go back to sandbox, or back to whatever page you were on before.  My sense is the latter

ui-admin/src/portal/siteContent/SiteContentEditor.tsx
devonbush: yeah, this is good feedback -- we've been using cancel as a cancel/back action.  But you're right that the behavior of the 'back' in this instance isn't clear

breilly2:
@devonbush I've had the same thought with routable tabs on a page. While I like having them be routable, I think they probably shouldn't add to history state (currently they do). As for the environment selector, I think it largely depends on other aspects of the UX. For example, since nav items are separately and differently opinionated about the default environment, I'm inclined to agree with you. If, however, the that weren't the case and if the environment selector were more closely visually linked to the side nav, then maybe they should have separate history state.

PR #534, opened 2023-09-01
JN-482: adding admin survey printing

devonbush:
This adds a new print/download button to all admin survey pages.  Clicking the button brings up a full-screen view of the document, and auto-prompts the printing dialog.  
image

image

Most of the effort on this PR went to print style refinements, like page breaks and radio button appearance.  Chrome works ok, but I couldn't get page breaks to render properly in safari.   Likewise, I suspect soon we'll want to generate these pdfs server-side in bulk so they can be sent in batches alongside medical record requests.  But that's a problem for another day that will depend on whether we support admins providing a separate pdf template (as pepper does now), or whether we always auto-generate the pdfs from surveyjs configs.

TO TEST:
1. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHSALK/consents/oh_oh_consent` 
2. select "print/download'
3. confirm modal pops up, and print dialog appears.
4. confirm form appears ok in print preview

### DIFF
--- a/ui-admin/src/study/participants/enrolleeView/EnrolleeView.tsx
+++ b/ui-admin/src/study/participants/enrolleeView/EnrolleeView.tsx
@@ -80,7 +80,7 @@ export default function EnrolleeView({ enrollee, studyEnvContext, onUpdate }:
     <div className="row mt-2">
       <div className="col-12">
         <div className="d-flex">
-          <div className="participantTabs" style={{ minWidth: '280', maxWidth: '280px' }}>
+          <div style={{ minWidth: '280', maxWidth: '280px' }}>
             <ul className="list-group">
               <li className="list-group-item">
                 <NavLink to="profile" className={getLinkCssClasses}>Profile &amp; Notes</NavLink>
--- a/ui-admin/src/study/participants/enrolleeView/EnrolleeView.tsx
+++ b/ui-admin/src/study/participants/enrolleeView/EnrolleeView.tsx
@@ -156,19 +156,18 @@ export default function EnrolleeView({ enrollee, studyEnvContext, onUpdate }:
                 <Route path="profile" element={<EnrolleeProfile enrollee={enrollee}
                   studyEnvContext={studyEnvContext}
                   onUpdate={onUpdate}/>}/>
-                <Route path="consents" element={<div>consents</div>}/>
-                { currentEnv.preEnrollSurvey && <Route path="preRegistration" element={
+                { currentEnv.preEnrollSurvey && <Route path="preRegistration/*" element={
                   <PreEnrollmentView preEnrollSurvey={currentEnv.preEnrollSurvey}
                     preEnrollResponse={enrollee.preEnrollmentResponse}/>
                 }/> }
                 <Route path="surveys">
-                  <Route path=":surveyStableId" element={<EnrolleeSurveyView enrollee={enrollee}
+                  <Route path=":surveyStableId/*" element={<EnrolleeSurveyView enrollee={enrollee}
                     responseMap={responseMap}/>}/>
                   <Route path="*" element={<div>Unknown participant survey page</div>}/>
                 </Route>
                 <Route path="tasks" element={<ParticipantTaskView enrollee={enrollee}/>}/>
                 <Route path="consents">
-                  <Route path=":consentStableId" element={<EnrolleeConsentView enrollee={enrollee}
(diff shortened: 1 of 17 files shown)


ui-admin/src/study/participants/enrolleeView/EnrolleeView.tsx
```diff
@@ -80,7 +80,7 @@ export default function EnrolleeView({ enrollee, studyEnvContext, onUpdate }:
     <div className="row mt-2">
       <div className="col-12">
         <div className="d-flex">
-          <div className="participantTabs" style={{ minWidth: '280', maxWidth: '280px' }}>
+          <div style={{ minWidth: '280', maxWidth: '280px' }}>
```
devonbush, line 83: drive-by removal of unused class

breilly2:
Everything looks the same to me in Safari.
Oddly, PgUp/PgDn don't work (doesn't seem to be browser-dependent).

ui-admin/src/study/participants/survey/PrintFormModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+type DownloadFormViewProps = {
+    answers: Answer[],
+    resumeData?: string,
+    survey: Survey | ConsentForm,
+    onDismiss: () => void
+}
+
+
+/** renders the form in a fullscreen modal, and pops up the print dialog */
+const PrintFormModal = ({ survey, answers, resumeData, onDismiss }: DownloadFormViewProps) => {
```
breilly2: I wonder if this should be a routable page instead of a modal. The modal might be what's getting in the way of PgUp/PgDn not working. Also, since the modal takes up the whole screen and doesn't look like a modal, I'm tempted to think that the back button should go back to the survey view, not back to wherever I was before opening the survey.

However, I did just instinctively hit `esc` to close the print view and was happy that it worked. Maybe it should stay as a modal but make it just barely less than the full screen so that it looks more like a modal?

breilly2 (approved):
Nice! I notice that, in browser, the form entries appear disabled, but they look less ghostly in the actual PDF.

ui-admin/src/study/participants/survey/PrintFormModal.tsx
devonbush: yeah, I'd like to get more of a sense of how this will be used.  My spidey sense is that this functionality is very temporary -- either admin users will almost never use it, in which case it doesn't matter except as an occasional nice-to-have, or we will have a customer that needs it for all participants (like the Pepper studies that auto-send consent pdfs to various third parties) in which case we'll want to invest in some sort of Nodejs backend that can generate these pdfs server-side

devonbush: good idea on making it routable -- done in latest commit

ui-admin/src/study/participants/survey/PrintFormModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+/** renders the form in a fullscreen modal, and pops up the print dialog */
+const PrintFormModal = ({ survey, answers, resumeData }: DownloadFormViewProps) => {
+  const surveyJsData = makeSurveyJsData(resumeData, answers, undefined)
+  const surveyJsModel = surveyJSModelFromForm(survey)
+  surveyJsModel.data = surveyJsData.data
+  configureModelForPrint(surveyJsModel)
+  useEffect(() => {
+    waitForImages().then(() => { window.print() })
+  }, [surveyJsModel])
```
MatthewBemis, line 32:
```suggestion
  useEffect(() => {
    waitForImages().then(() => { printForm() })
  }, [surveyJsModel])

  function printForm() {
    const initialPageTitle = document.title
    //The filename defaults to whatever the page title is, so this temporarily sets it to the form name
    document.title = `${survey.name}.pdf`
    window.print()
    document.title = initialPageTitle
  }
```

Not sure if this is too hacky, but this would be a decent UX improvement IMO. Right now the downloaded form is titled "Juniper.pdf" by default.

MatthewBemis (approved):
Tested it out, looks good. One suggestion but I leave it totally up to you.

ui-admin/src/study/participants/survey/PrintFormModal.tsx
devonbush: great idea, done!

PR #535, opened 2023-09-03
JN-537 adding api wrapper utils

devonbush:
Adds some hooks and utils for standard loading and error handling for our admin pages.  I've updated a representative sample of existing components with the new utility functions.  If y'all like this direction, I'll apply them to all the other relevant places where we call APIs.

TO TEST:
1. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/mailingList
2. confirm it loads, behaves correctly on environment switch, and updates after deleting an entry.
3. update MailingListController's get() method to throw an error (e.g. by appending a random string to the portal shortcode)
4. reload the mailing list view page.
5. confirm you see an error alert
image


breilly2:
Possible unrelated mailing list UI bug (I've only tested so far, not looked at the code): after deleting someone from the mailing list, download doesn't work. It looks like the selection model still contains the deleted entity, so even with no selection, it says "1 of 1 selected". If I select the remaining row -- either individually or with select-all -- it says "2 of 1 selected".

ui-admin/src/util/api-utils.tsx
```diff
@@ -0,0 +1,70 @@
...
+  useEffect(() => {
+    reload()
+  }, deps)
+  return { isLoading, isError, reload }
+}
+
+/**
+ * utility function for wrapping an Api call in loading and error handling
+ */
+export const doApiLoad = async (loadingFunc: () => Promise<unknown>,
```
breilly2: I think I'm recognizing that `useLoadingEvent` is for loading data for view pages while `doApiLoad` is for calling create/update endpoints that also happen to return the new updated data. Is that correct? If so, a few thoughts:
* Can `ParticipantList` use `useLoadingEffect` instead of `doApiLoad`? I attempted this and it seemed to work.
* It feels a little odd to me to pass state mutators to a function in this way. Can `doApiLoad` return the state that it would maintain if it were a hook instead of taking callback functions.

ui-admin/src/api/api.tsx
```diff
@@ -349,7 +349,7 @@ export default {
     if (response.ok) {
       return obj
     }
-    return Promise.reject(response)
+    return Promise.reject(obj)
```
breilly2, line 352: It's hard to know what impacts this change will have, especially since the value is interpreted by each client of `Api` rather than within `Api` itself. Have you explored all of the usages to make sure that there aren't callers relying on this being `response`?

ui-admin/src/portal/MailingListView.tsx
```diff
@@ -83,17 +83,10 @@ export default function MailingListView({ portalContext, portalEnv }:
...
-      setIsLoading(false)
-    }).catch((e: Error) => {
-      alert(`error loading mailing list ${  e}`)
-      setIsLoading(false)
-    })
-  }
+  const { isLoading, reload } = useLoadingEffect(async () => {
+    const result = await Api.fetchMailingList(portalContext.portal.shortcode, portalEnv.environmentName)
+    setContacts(result)
+  }, [portalContext.portal.shortcode, portalEnv.environmentName])
```
breilly2, line 89: I think this looks really nice! It's a common pattern, so definitely a good call to extract it.

ui-admin/src/study/participants/participantList/ParticipantList.tsx
```diff
@@ -130,15 +129,11 @@ function ParticipantList({ studyEnvContext }: {studyEnvContext: StudyEnvContextT
...
         facetValues)
       setParticipantList(response)
-    } catch (e) {
-      Store.addNotification(failureNotification('Error loading participants'))
-    }
-    setIsLoading(false)
+    }, { setIsLoading })
   }
 
   useEffect(() => {
```
breilly2: This is what I wrote in an attempt to replace this `useEffect` (and the `searchEnrollees` function that only it uses) with `useLoadingEffect`:
```
  const { isLoading } = useLoadingEffect(async () => {
    const response = await Api.searchEnrollees(portal.shortcode, study.shortcode, currentEnv.environmentName,
      facetValues)
    setParticipantList(response)
  }, [portal.shortcode, study.shortcode, currentEnv.environmentName])
```
Does that look to you like it would work?

ui-admin/src/study/surveys/PreEnrollView.tsx
```diff
@@ -28,18 +29,17 @@ function RawPreEnrollView({ studyEnvContext, survey, readOnly = false }:
   /** saves as a new version and updates the study environment accordingly */
   async function createNewVersion({ content: updatedContent }: { content: string }): Promise<void> {
     survey.content = updatedContent
-    try {
+    doApiLoad(async () => {
```
breilly2, line 32: Just based on this diff, it looks like the amount of code here is exactly the same, just replacing `try`/`catch` with `doApiLoad`. Are we gaining any new behavior from `doApiLoad` that makes this change more of a win?

ui-admin/src/user/CreateUserModal.tsx
```diff
@@ -22,16 +23,12 @@ const CreateUserModal = ({ onDismiss, portals, userCreated }:
   const selectedPortalOpt = portalOpts.find(portalOpt => portalOpt.value === portalShortcode)
 
   const createUser = async () => {
-    setIsLoading(true)
-    try {
+    doApiLoad(async () => {
```
breilly2, line 26: Ah, interesting. Unlike in `PreEnrollView`, it looks like this use of `doApiLoad` eliminates the need to `catch` and `Store.addNotification` here.

breilly2 (commented):
Definitely some good stuff here. I do have some questions about the patterns that I'm interested to explore more before spreading through the rest of the codebase.

ui-admin/src/api/api.tsx
devonbush: that would be part of the full PR

ui-admin/src/study/surveys/PreEnrollView.tsx
devonbush: here, no, just better standardization

ui-admin/src/util/api-utils.tsx
devonbush: Yes, using useLoadingEffect should work for participant list.  

Hmm... doApiLoad could return a Promise<{obj, isLoaded, isError}>. So it would replace
``` 
  doApiLoad(() => {
      const response = await Api.loadStuff()
      setLoadedStuff(response)
  }, {setIsLoading})
```
with
```
  setIsLoading(true)
  const { response, isError, isLoading} = await doApiLoad(() => {
           return Api.loadStuff()
           })
   setLoadedStuff(response)
   setIsLoading(isLoading)
```

I think you're right the latter is a little 'cleaner' in terms of signatures, but requires more discipline to use, because it's easier to leave off one of the setIsLoading

MatthewBemis (commented):
I like it, seems relatively easy to work with. 👍 for standardization and better error handling.

breilly2:
We chatted about this some last week, including getting some input from Nick. I think that `useLoadingEffect` is a definite win. I'm less certain about the pattern around `doApiLoad`, but I agree that it is an improvement... though it's still easy to forget to use the loading state (`PreEnrollView` still doesn't).

Definitely need to complete the audit of any impact of the change from `return Promise.reject(response)` to `return Promise.reject(obj)` before merging.

ui-admin/src/util/api-utils.tsx
```diff
@@ -0,0 +1,70 @@
...
+  opts: {
+        setIsLoading?: (isLoading: boolean) => void,
+        setIsError?: (isError: boolean) => void,
+        customErrorMsg?: string
+    } = {}) => {
+  if (opts.setIsLoading) { opts.setIsLoading(true) }
```
breilly2: There's a nice shorthand for these:
```suggestion
  opts.setIsLoading?.(true)
```

devonbush:
confirm mailing list bug is unrelated...

devonbush:
Added a test, and upgraded a lot more of our pages to use the new paradigms.  Definitely not an exhaustive upgrade, just tried to hit the highlights of most of our commonly used pages so that the pattern is validated across most of our pages.

breilly2 (approved):
I'm still a little uncomfortable about not fully auditing the impact of changing the object passed to `Promise.reject()`, but I'll leave that decision up to you.

MatthewBemis (approved):

devonbush:
I'm ok with that -- I tested the important common error case (failed login) and that behaved as expected.  I think exhaustively testing/auditing every other server call isn't worth the effort given how poor the existing error handling was

PR #536, opened 2023-09-03
JN-551 fixing participant list table styling

devonbush:
Previously, when the table was too wide to fit in the page, it would flow onto a white background, looking funny.  Now the parent container expands with to to contain it in a gray background.  the navbar user and help menus are given 'sticky' position so that they flow naturally as the user scrolls.    (Note this does not solve the pre-existing problem that the nav goes away at small screen sizes -- that's a fix for another ticket)

image

This PR also has two other ride-along changes:
1. the participant list table is upgraded to use table utils
2. the column show/hide feature is put in a proper modal, so it's not so annoying to dismiss, and has better accessibility affordances.

TO TEST:
1. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants`
2. use the show/hide control to toggle on all columns
3. confirm that even if the table forces horizontal scroll, it stays within the page-styled gray background

### DIFF
--- a/ui-admin/src/util/tableUtils.tsx
+++ b/ui-admin/src/util/tableUtils.tsx
@@ -3,6 +3,8 @@ import { CellContext, Column, flexRender, Header, RowData, Table } from '@tansta
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faCaretDown, faCaretUp, faCheck, faColumns } from '@fortawesome/free-solid-svg-icons'
 import Select from 'react-select'
+import Modal from 'react-bootstrap/Modal'
+import { Button } from '../components/forms/Button'
 
 /**
  * Returns a debounced input react component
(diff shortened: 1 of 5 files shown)


MatthewBemis (approved):

ui-admin/src/util/tableUtils.tsx
```diff
@@ -216,42 +218,51 @@ export function IndeterminateCheckbox({
  * */
 export function ColumnVisibilityControl<T>({ table }: {table: Table<T>}) {
   const [show, setShow] = useState(false)
-  return <div className="position-relative ms-auto">
+  return <div className="ms-auto">
     <button className="btn btn-secondary" onClick={() => setShow(!show)} aria-label="show or hide columns">
```
breilly2, line 222: It would be nice if this could also be `sticky`.

breilly2 (approved):
Looks good. We'll have to apply this to other pages that use potentially wide tables. Is there any way to extract or lift the styling so we get it by default?

ui-admin/src/util/tableUtils.tsx
devonbush: yeah, I think that's a good question for Erin about what elements stay sticky and what scrolls with the page.  I'll go ahead and merge this, and then get feedback at demo

PR #538, opened 2023-09-06
JN-533 submitted surveys go back to first page

devonbush:
On submission events, this resets the page number to 1, so that if the user returns, they'll go back to the beginning.  

TO TEST:
1. sign into participant view as jsalk@test.com
2. fill out an uncompleted survey
3. hit submit
4. from the dashboard, go back to that survey
5. confirm you get taken back tot he first page


breilly2 (approved):
Looks good. I also tested making sure that the current page is respected for an "incomplete" survey, both on the first pass through and if going back in and editing, in which case autosave does nicely save the active page at the time of autosave.

MatthewBemis (approved):

PR #540, opened 2023-09-07
[JN-461] Designer validation

MatthewBemis:
This adds very simple unenforced validation to a few areas in the survey question designer.

Screenshot 2023-09-15 at 11 57 05 AM

TO TEST:

Try adding a new question or editing an existing one. Required fields should have a red asterisk next to the field label and the input area should by styled with a warning if it's empty. Once some text is entered, the input area should return to normal styling.

_______

~~This is super rough, but this one has been slow to implement so I wanted to start getting feedback before I put more time into it. There aren't any tests yet, and I want to do a bit of refactoring for the `onChange` stuff which would allow me to push the overall validation check down into the `FormDesigner` component where it belongs. But for now, you get the idea.~~

~~The main challenge for implementing this was the designer tab doesn't have a concept of saving an individual question. You can freely click around the table of contents without actually committing to your changes. I think that's good from a UX perspective, because it reduces the number of clicks required of the user, but this means that we have to disable the use of the table of contents until the user has fixed any errors. Is this too restrictive/burdensome? Is the validation warning itself too loud?~~

~~Another issue is that the validation checks only occur when the `QuestionDesigner` is rendered for a question. This means that if you close the page out and load from a draft, any validation errors are lost (it doesn't re-validate the entire survey). I'm thinking of this type of validation as "best effort" rather than airtight, because none of these validation errors are actually going to cause problems when the survey is rendered. But it at least encourages decent field hygiene.~~

~~Final thought: I'm wondering if overall this is more trouble than it's worth right now. Even without this, we're still preventing all of the breaking cases (like invalid survey json, and requiring stable IDs for new questions)~~

~~https://github.com/broadinstitute/juniper/assets/7257391/5584c733-3389-446a-9e22-2cf4e41a9d8c~~

### DIFF
--- a/ui-admin/src/components/forms/TextInput.tsx
+++ b/ui-admin/src/components/forms/TextInput.tsx
@@ -5,6 +5,7 @@ import InfoPopup from './InfoPopup'
 export type TextInputProps = Omit<JSX.IntrinsicElements['input'], 'onChange'> & {
   infoContent?: React.ReactNode,
   description?: string
+  required?: boolean
   label: string
   labelClassname?: string,
   onChange?: (value: string) => void
@@ -12,7 +13,7 @@ export type TextInputProps = Omit<JSX.IntrinsicElements['input'], 'onChange'> & 
 
 /** A text input with label and description. */
 export const TextInput = (props: TextInputProps) => {
-  const { infoContent, description, label, labelClassname, ...inputProps } = props
+  const { infoContent, description, required, label, labelClassname, ...inputProps } = props
   const { className, disabled, id, value, onChange } = inputProps
 
   const generatedId = useId()
@@ -26,6 +27,7 @@ export const TextInput = (props: TextInputProps) => {
         htmlFor={inputId}
       >
         {label}
+        {required && <span className="text-danger">*</span>}
       </label>
       {infoContent &&
         //@ts-ignore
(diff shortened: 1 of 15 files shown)


devonbush:
I think this is definitely going in the right direction.  I like the design of the red-bordered text fields.  I think you're right that replacing the table of contents with the message is overkill though. I might just let people click on it to navigate around anyway.  If we really want to disable navigating, maybe just put a transparent overlay over it, and then if the y try to click it, then show a popup saying to fix the current question first.  That way the user doesn't have the jarring experience of seeing the table of contents disappear every time they make a change

breilly2:
I agree that it is very jarring to replace the ToC with a validation error message. I think the input styling is enough of a first step, especially if we treat this as in incremental improvement instead of a guarantee that we'll prevent anything invalid from being saved.

MatthewBemis:
I appreciate the feedback. I'll update to get rid of the blocking error message and proceed with just the input styling as a first step.

ui-admin/src/components/forms/TextInput.tsx
```diff
@@ -26,6 +27,7 @@ export const TextInput = (props: TextInputProps) => {
         htmlFor={inputId}
       >
         {label}
+        {required && <span className="text-danger">*</span>}
```
breilly2, line 30: I questioned if the `*` should actually be in the label, but it looks like the answer is "yes": https://www.deque.com/blog/anatomy-of-accessible-forms-required-form-fields/#:~:text=Providing%20Star%20(Asterisk)%20Symbol,is%20accessible%20to%20assistive%20technologies.

breilly2 (approved):

devonbush (approved):

PR #542, opened 2023-09-12
[JN-563] Setup cunningh as a superuser and update README

denniscunningham:
Not sure if there are any tests I should run, or other steps I should take prior to the PR, so just let me know.

TO TEST:
1. Verify that README text is legible and accurate.
2. Reseed DB and confirm cunningh is in `admin_user` table.


devonbush (approved):
looks good, thanks again for your patience with the setup instructions!

MatthewBemis (approved):

PR #545, opened 2023-09-12
JN-562 -- better handling of invisible question clearing

devonbush:
This updates our participant survey-saving process to always send a diff of answers that have changed, even on form submission.  Previously, on submission, we would resend all answers.  While this was desirable when autosave was new and untrusted, it leads to a lot of corner cases as we have to support two types of answer processing, and edge cases where surveyJS's 'clearOnSubmit' property for hidden questions would cause values that had been cleared for the user to not be deleted from the server.   Also, moving to always sending diffs will avoid lots of synchronous editing issues when we start enabling admin survey editing.

The change is pretty simple--the core is just changing the submit function in SurveyView.tsx to use `answers: getUpdatedAnswers` instead of getting the full answer list. 

For the curious, I first went down the path of supporting both patch and post submissions but just with enhanced server-side diffing and auditing.  You can see the code on the `db-survey-answer-clear` branch.  There's some cool stuff there like a base DataAuditedService that I hope to pull into future work, but it just wasn't worth the time for this change.  

TO TEST:
1. repopulate ourhealth
2.sign in to participant UI as jsalk@test.com
3. start filling out the family history survey, answering a few questions, and then go back and say "not at all" for the first question
4. click 'submit'
5. in the admin tool, at `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHSALK/surveys/oh_oh_famHx` confirm none of the answers other than the first remain

### DIFF
--- a/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/survey/SurveyResponseController.java
+++ b/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/survey/SurveyResponseController.java
@@ -49,7 +49,7 @@ public class SurveyResponseController implements SurveyResponseApi {
   }
 
   @Override
-  public ResponseEntity<Object> response(
+  public ResponseEntity<Object> update(
       String portalShortcode,
       String envName,
       String studyShortcode,
@@ -62,7 +62,7 @@ public class SurveyResponseController implements SurveyResponseApi {
     EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
     SurveyResponse responseDto = objectMapper.convertValue(body, SurveyResponse.class);
     HubResponse hubResponse =
-        surveyResponseExtService.submitResponse(
+        surveyResponseExtService.updateResponse(
             user, portalShortcode, environmentName, responseDto, enrolleeShortcode, taskId);
     return ResponseEntity.ok(hubResponse);
   }
--- a/api-participant/src/main/java/bio/terra/pearl/api/participant/service/SurveyResponseExtService.java
+++ b/api-participant/src/main/java/bio/terra/pearl/api/participant/service/SurveyResponseExtService.java
@@ -47,7 +47,7 @@ public class SurveyResponseExtService {
         studyEnv.getId(), stableId, version, enrollee, taskId);
   }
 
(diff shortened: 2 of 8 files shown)


breilly2:
If the goal here is to avoid the case where there is inconsistent survey response state, i.e. answers for a page that should be hidden, this doesn't quite get there.

1. Fill out the family history survey answering "None at all" for the first question.
2. Complete the survey.
3. Re-enter the survey and change the first answer to "Some".
4. Fill in something on page 2.
5. Go pack to page 1 and change the first answer back to "None at all".
6. Wait for the auto-save.
7. _Don't_ complete the survey. Instead just go back to the dashboard.

The result is that the admin view shows a completed survey with inconsistent answers.

Side note: It felt strange to me to go into the admin UI and be able to see auto-saved responses for surveys not marked as complete. We've probably discussed this before, and maybe we do want to show these, but I think we should indicate that somehow.

Another thought is that, if a participant changes responses to a completed survey but doesn't click "Complete" again, maybe we want to auto-save the survey as not complete. That way admins could see that state and understand why the responses may be inconsistent with conditional questions and participants could see the survey in their dashboard return to an in-progress state.

devonbush:
thanks for the comments and testing, Brian.  You're right that in addition to these code changes, the surveys that want to leverage these changes will need to set `"clearInvisibleValues": "onHiddenContainer"` We've instructed Shriie to apply that to the family history survey in prod.  I haven't yet applied it to the OH populated surveys, because I'm planning to do a big copy of all her changes to our populate files once they go to IRB, and didn't want merge conflict.  Also, we might just want to make that setting the default.

For your second and third points, yup, there's lots of possible paths we could take.  For now, the direction was "Submit should only impact the task, values are good as soon as they are entered".   Agree there's risk of confusion either way -- in one case participants might make changes, not realize they were saved, and exit the form thinking that counts as canceling.  In another, they make a change, don't realize they have to go all the way to the end and hit "submit", and so their survey gets stuck in limbo.  We're leaning more towards the google-docs-style "changes are changes as soon as they are made" and you're right there's a lot of complexity and nuance in how we should present that to participants and admin

MatthewBemis (approved):
👍

breilly2 (approved):

PR #549, opened 2023-09-13
JN-540 . fix createdAt and lastUpdated timestamp values on new survey creation

ssettipalli:
Let the backend service set createdAt and lastUpdated timestamps(instant) during new survey creation.
front-end/client passed values are ignored

QA
Add a new survey from admin tools
Click “Add Survey” and fill in a name/stableId

in the juniper database, run select id, name, created_at, last_updated_at from survey; and make sure current date/timestamp is populated.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
@@ -9,7 +9,6 @@ import bio.terra.pearl.core.model.survey.Survey;
 import bio.terra.pearl.core.service.exception.NotFoundException;
 import bio.terra.pearl.core.service.study.StudyEnvironmentSurveyService;
 import bio.terra.pearl.core.service.survey.SurveyService;
-import java.time.Instant;
 import java.util.List;
 import java.util.Optional;
 import org.springframework.beans.BeanUtils;
@@ -56,9 +55,6 @@ public class SurveyExtService {
     }
     survey.setPortalId(portal.getId());
     survey.setVersion(1);
-    Instant now = Instant.now();
-    survey.setCreatedAt(now);
-    survey.setLastUpdatedAt(now);
     return surveyService.create(survey);
   }
 
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
@@ -9,6 +9,7 @@ import bio.terra.pearl.core.model.survey.Survey;
 import bio.terra.pearl.core.service.exception.NotFoundException;
 import bio.terra.pearl.core.service.study.StudyEnvironmentSurveyService;
 import bio.terra.pearl.core.service.survey.SurveyService;
+import java.time.Instant;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


devonbush:
Looks good, please add an automated test.  I think you'll want to add something like the following in SurveyExtServiceTests
```
@Test
  public void createAssignsDateAndVersion() {
    AdminUser user = AdminUser.builder().superuser(true).build();
    when(mockAuthUtilService.authUserToPortal(user, "foo"))
        .thenReturn(new Portal());

    Survey survey = Survey.builder().stableId("createAssignsDate").name("createAssignsDate Test").build();
    surveyExtService.create("foo", survey, user);
    
    ArgumentCaptor<Survey> argumentCaptor = ArgumentCaptor.forClass(Survey.class);
    Mockito.verify(mockSurveyService).create(argumentCaptor.capture());
    assertThat(
        argumentCaptor.getValue().getCreatedAt(), greaterThan(Instant.now().minusMillis(200)));
      // assert other properties
  }
```

ssettipalli:
> Looks good, please add an automated test. I think you'll want to add something like the following in SurveyExtServiceTests
> 
> ```
> @Test
>   public void createAssignsDateAndVersion() {
>     AdminUser user = AdminUser.builder().superuser(true).build();
>     when(mockAuthUtilService.authUserToPortal(user, "foo"))
>         .thenReturn(new Portal());
> 
>     Survey survey = Survey.builder().stableId("createAssignsDate").name("createAssignsDate Test").build();
>     surveyExtService.create("foo", survey, user);
>     
>     ArgumentCaptor<Survey> argumentCaptor = ArgumentCaptor.forClass(Survey.class);
>     Mockito.verify(mockSurveyService).create(argumentCaptor.capture());
>     assertThat(
>         argumentCaptor.getValue().getCreatedAt(), greaterThan(Instant.now().minusMillis(200)));
>       // assert other properties
>   }
> ```

@devonbush 
Thanks for the suggestion. 
I was going to update a existing test to check for these date values and missed it. 
Updated existing test case. Lets see if this good enough rather than a separate test case.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
```diff
@@ -55,6 +56,9 @@ public Survey create(String portalShortcode, Survey survey, AdminUser adminUser)
     }
     survey.setPortalId(portal.getId());
     survey.setVersion(1);
+    Instant now = Instant.now();
```
ssettipalli: Same changes in SurveyController.create() works too but having the changes here in Service seemed cleaner and covers any other service calls other than front-end.

devonbush:
Ah, thanks for checking.  The test that you updated actually isn't testing your updated Service.  This is where it gets confusing that SurveyExtService (the bit that handles authorization and logic for api admins) is different than SurveyService (the base underlying CRUD service).  But now that I think about it, setting the dates is probably better done in the base SurveyService anyway.   
`DaoTestUtils.assertGeneratedProperties` already checks that the created_at date is being set.  So you'll just want to create a new test that confirms the date is set even when the initial date is null (note that Survey objects by default get their createdAt date set to now in BaseEntity, so for the test to be robust, you'll need to explicitly set the date to null)

devonbush (approved):
nice!

MatthewBemis (approved):

breilly2 (approved):

PR #551, opened 2023-09-15
JN-573 global participant search bar

devonbush:
#### DESCRIPTION

Searches substrings of given name, family name, email or shortcode (case-insensitive).  This is a no-frills implementation, so it does not auto-update as the user types, nor does it do name sounds like.  We can add those later if desired. 

image

 This is implemented as a "Facet" so that it can plug into our existing facet infrastructure that enables url persistence and other goodies.  Now that I'm seeing how the SQLFacet infrastructure is evolving, I think a later PR (when we upgrade to support searching by answer values) will refactor the backend search into a 3-part process so that facets have more comprehensible methods like "getFilteredEnrolleeIds" and "getAssociatedCollections" methods, rather than the current single-query approach which might be more database efficient, but leads to some very opaque implementation methods returning sql-ish stuff.

1. redeploy ApiAdminApp
2. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants
3. search for participants by name/shortcode/email
4. reload the page, confirm search and keyword persists


core/src/main/java/bio/terra/pearl/core/service/participant/search/facets/sql/KeywordFacetSqlGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+        return IntStream.range(0, facetValue.getValues().size())
+                .mapToObj(index -> {
+                    String paramName = EnrolleeSearchUtils.getSqlParamName("keyword", facetValue.getKeyName(), index);
+                    return """
+                            (profile.given_name ilike :%s 
+                              OR profile.family_name ilike :%s
+                              OR profile.contact_email ilike :%s
+                              OR enrollee.shortcode ilike :%s)
+                            """
+                            .formatted(paramName, paramName, paramName, paramName);
```
breilly2: I recently learned about referencing formatting arguments positionally which allows you to avoid repeating the value in cases like this:
```suggestion
                    return """
                            (profile.given_name ilike :%1$s 
                              OR profile.family_name ilike :%1$s
                              OR profile.contact_email ilike :%1$s
                              OR enrollee.shortcode ilike :%1$s)
                            """
                            .formatted(paramName);
```

breilly2 (changes requested):
Looks good. One testing nit: when using the environment drop-down or clicking the "participants" breadcrumb, we should either re-apply the filter or clear the filter text input:

https://github.com/broadinstitute/juniper/assets/19228696/724d29c6-86ec-4fea-b75e-5af86c70505a

breilly2 (approved):

MatthewBemis (approved):
Looks good!

PR #552, opened 2023-09-18
JN-467 add url to JS errors

devonbush:
#### DESCRIPTION

We saw a new JS error in the logs about circular references in JSON.  I'd like to know what page the user received it on, so we're including the URL in error logs now.

1.  edit a participant UI component to throw an error
2. visit the participant UI and trigger that component
3. confirm the JS error is logged via the network tab, and the current href is sent.

### DIFF
--- a/ui-participant/src/util/loggingUtils.ts
+++ b/ui-participant/src/util/loggingUtils.ts
@@ -39,7 +39,7 @@ export const logError = (detail: ErrorEventDetail, stackTrace: string) => {
   Api.log({
     eventType: 'ERROR',
     eventName: 'jserror',
-    eventDetail: JSON.stringify(detail) + '\n' + window.location.href,
+    eventDetail: `${JSON.stringify(detail)}\n${window.location.href}`,
     stackTrace
   })
 }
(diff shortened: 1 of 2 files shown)


ui-participant/src/util/loggingUtils.ts
```diff
@@ -39,7 +39,7 @@ export const logError = (detail: ErrorEventDetail, stackTrace: string) => {
   Api.log({
     eventType: 'ERROR',
     eventName: 'jserror',
-    eventDetail: JSON.stringify(detail),
+    eventDetail: `${JSON.stringify(detail)}\n${window.location.href}`,
```
MatthewBemis, line 42: Should this be its own attribute? Maybe `eventLocation`? As-is this might make it more difficult to categorize or filter error events.

devonbush: yeah, I think you're right -- I was too lazy to make it on this PR since it was just a quick one-off.  Can I take an IOU and I'll add that column when we add this error logging to the admin tool?  Then we can think about whether we also want to add a 'source' column for participant/admin/server

MatthewBemis: works for me 👍

MatthewBemis (approved):

breilly2 (approved):

PR #556, opened 2023-09-20
[JN-474] [JN-366] Integrate with the latest updates to DSM's kit status responses

breilly2:
https://broadworkbench.atlassian.net/browse/JN-474

Also includes minor improvements to kit UI for https://broadworkbench.atlassian.net/browse/JN-366

DSM now includes its understanding of the current status of a kit request in its status responses. 

To facilitate testing and demoing, there is also a new superuser-only endpoint to invoke the bulk status update that normally only runs once a day (late at night). To minimize the potential impact on DSM, the endpoint targets a specific study instead of refreshing the status for all in-progress kits in all studies.

#### DESCRIPTION

**IMPORTANT** This PR switches the default Pepper client code from a stub implementation to a live implementation that actually talks to DSM.

Screenshot 2023-09-20 at 11 48 20 AM

1. Start the admin API and repopulate your local database (`OHKITS`'s address has been updated to a valid address)
2. Start the admin UI
3. Go to the list of users eligible for kits in the sandbox environment (https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/eligible)
4. Select `OHKITS`, click "Send sample collection kit", and click "Request Kit"
5. Click over to the "Requested". You should see a new kit filed under the "Created" tab
6. Click on the `OHKITS` link to view the kit details for that participants kits (https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHKITS/kitRequests). Click the (i) in the DSM Status column and notice the raw status JSON from DSM, particularly the `dsmShippingLabel` attribute.
7. Go to https://dsm-dev.datadonationplatform.org/ and sign in with your Broad Google account. Check with Pegah if you need permissions, or have someone who does have permissions perform the steps in DSM. These next steps are acting in the role of someone from GP who ships and receives kits.
8. Select the "Our Heart" study, then under "Samples", click "Kits without labels"
9. Filter the list by Shipping ID using the `dsmShippingLabel` from the JSON status. Notice that the kit that was just created shows up.
10. Since creating labels often fails in dev. Instead, go back to the "Samples" nav and click "Final Scan"
11. For "Kit Label", enter some unique value; I like to use a date/time-based number. For "DSM Label", enter the `dsmShippingLabel` from earlier.
12. Click "Save Scan Pairs", then click it again so you get a "Scanned successfully" message.
13. Go back to the kit details for `OHKITS` in Juniper (https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHKITS/kitRequests) and reload. Notice that the kit status has not changed.
14. Go to http://localhost:8080/swagger-ui.html#/kit/refreshKitStatuses, auth, and run for ourhealth/ourheart to batch-update kit statuses from DSM. This is the process that normally happens overnight.
15. Again, reload https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHKITS/kitRequests and notice that the status has changed from "Kit Without Label" to "Sent".

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/kit/PepperDSMClientProvider.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/kit/PepperDSMClientProvider.java
@@ -20,7 +20,7 @@ public class PepperDSMClientProvider {
     @Primary
     @Bean
     public PepperDSMClient getPepperDSMClient() {
-        return getStubPepperDSMClient();
+        return getLivePepperDSMClient();
     }
 
     @Bean
--- a/core/src/main/java/bio/terra/pearl/core/service/kit/PepperDSMClientProvider.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/kit/PepperDSMClientProvider.java
@@ -6,10 +6,14 @@ import org.springframework.context.annotation.Primary;
 
 @Configuration
 public class PepperDSMClientProvider {
+    private final LivePepperDSMClient.PepperDSMConfig pepperDSMConfig;
     private final LivePepperDSMClient livePepperDSMClient;
     private final StubPepperDSMClient stubPepperDSMClient;
 
-    public PepperDSMClientProvider(LivePepperDSMClient livePepperDSMClient, StubPepperDSMClient stubPepperDSMClient) {
+    public PepperDSMClientProvider(LivePepperDSMClient.PepperDSMConfig pepperDSMConfig,
+                                   LivePepperDSMClient livePepperDSMClient,
+                                   StubPepperDSMClient stubPepperDSMClient) {
+        this.pepperDSMConfig = pepperDSMConfig;
         this.livePepperDSMClient = livePepperDSMClient;
         this.stubPepperDSMClient = stubPepperDSMClient;
     }
@@ -20,7 +24,9 @@ public class PepperDSMClientProvider {
     @Primary
     @Bean
     public PepperDSMClient getPepperDSMClient() {
-        return getLivePepperDSMClient();
+        return pepperDSMConfig.useLiveDsm() ?
+                getLivePepperDSMClient() :
+                getStubPepperDSMClient();
     }
 
     @Bean
--- a/core/src/main/java/bio/terra/pearl/core/service/kit/PepperKitStatus.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/kit/PepperKitStatus.java
@@ -1,6 +1,7 @@
 package bio.terra.pearl.core.service.kit;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.*;
 import lombok.experimental.SuperBuilder;
 
@@ -12,9 +13,15 @@ import lombok.experimental.SuperBuilder;
 public class PepperKitStatus {
     private String juniperKitId;
     private String currentStatus;
+    private String dsmShippingLabel;
+    @JsonProperty("participantId")
+    private String dsmParticipantId;
     private String labelDate;
+    private String labelByEmail;
     private String scanDate;
+    private String scanByEmail;
     private String receiveDate;
+    private String trackingScanBy;
     private String trackingNumber;
     private String returnTrackingNumber;
     private String errorMessage;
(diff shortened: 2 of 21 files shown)


core/src/main/java/bio/terra/pearl/core/service/kit/PepperDSMClientProvider.java
```diff
@@ -20,7 +20,7 @@ public PepperDSMClientProvider(LivePepperDSMClient livePepperDSMClient, StubPepp
     @Primary
     @Bean
     public PepperDSMClient getPepperDSMClient() {
-        return getStubPepperDSMClient();
+        return getLivePepperDSMClient();
```
breilly2: Update coming to make this configuration-based so it can default to stub for local dev and live for demo.

devonbush:
When I try to send a kit request from my local dev, I get `Unable to parse response from Pepper as bio.terra.pearl.core.service.kit.PepperErrorResponse: Could not verify auth0 token`. I have `DSM_JWT_SIGNING_SECRET` set to [redacted] as per the readme -- is that meant to be replaced by something?

breilly2:
The secret is correct. I could not reproduce the error myself. I think we should check the DSM logs to see if there's any more detail there about why it wasn't happy with the JWT.

core/src/main/java/bio/terra/pearl/core/service/kit/PepperKitStatus.java
```diff
@@ -12,9 +13,15 @@
 public class PepperKitStatus {
     private String juniperKitId;
     private String currentStatus;
+    private String dsmShippingLabel;
+    @JsonProperty("participantId")
```
devonbush: this could use a comment explaining why we want to export this as participantId -- I suspect because it's just giving us a Participant.id that we gave to it?  But I thought DSM was always operating on our Enrollee shortcodes (or ids).  Should DSM know that two kit requests for different enrollees in different studies actually correspond to the same Juniper participant?

ui-admin/src/api/api.tsx
```diff
@@ -177,12 +179,11 @@ function parsePepperKitStatus(json: string | undefined): PepperKitStatus {
...
       const pepperStatus = JSON.parse(json)
-      if (pepperStatus.kitId && pepperStatus.currentStatus) {
-        return {
-          ...emptyPepperKitStatus,
-          ..._pick(pepperStatus,
-            'kitId', 'currentStatus', 'labelDate', 'scanDate', 'receiveDate', 'trackingNumber', 'returnTrackingNumber')
-        }
+      return {
+        ...emptyPepperKitStatus,
+        ..._pick(pepperStatus,
```
devonbush, line 184: nice, _pick is something that I need to get in the habit of using

ui-admin/src/study/participants/KitRequests.tsx
```diff
@@ -56,6 +40,26 @@ export default function KitRequests({ enrollee, studyEnvContext, onUpdate }:
     onUpdate()
   }
 
+  const columns: ColumnDef<KitRequest, string>[] = [{
```
devonbush: style nit -- I think this can and should be declared outside of the component -- it doesn't change based on the data

devonbush (approved):
Looks good.  A couple questions, nothing blocking -- congrats on end-to-end up and running!

MatthewBemis (approved):
My testing was fairly limited while I wait for access to the study in DSM, but I know you demo'd this, so testing it out is less critical. I appreciate the thorough testing instructions in this PR, it'll be useful to come back to. We discussed over Slack that the DSM stub needs to be updated but I won't block on that since the important part is the end-to-end flow!

core/src/main/java/bio/terra/pearl/core/service/kit/PepperKitStatus.java
breilly2: This is a great question because it highlights the need for some documentation in these files. I'm in the process of adding that.

> why we want to export this as participantId

It's the other way around. In the JSON we receive from DSM, this is called `participantId`. Not knowing what to expect the value to be, I named the property `dsmParticipantId`. However, I can now see that this is echoing the enrollee ID passed to DSM via `PepperDSMKitRequest.juniperParticipantId`. Therefore, removing the `dsm` prefix (and therefore the `@JsonProperty` annotation) makes sense.

Alternatively, we could keep `@JsonProperty("participantId")` and rename the field to `enrolleeShortcode` (along with a similar change in `PepperDSMKitRequest`) which would make these files the definitive place where differences in vocabulary between Juniper and Pepper are defined. In fact, I think I like that the best. @devonbush, what do you think?

I think your remaining questions are explained by that difference in vocabulary; DSM is saying "participantId" while Juniper is using enrollee shortcode.

PR #557, opened 2023-09-20
[JN-515] Support removing and deleting surveys

MatthewBemis:
#### DESCRIPTION

This adds the option to remove or delete surveys. You can only perform these actions in sandbox environments. If you want to remove a survey from an IRB or Live environment, you have to go through the proper publishing flow by first removing it from sandbox and promoting it to the higher environments.

New three-dots popout menu next to surveys: 
Screenshot 2023-09-20 at 3 35 34 PM

Two options provided:
Screenshot 2023-09-20 at 3 35 42 PM

Remove survey modal:
Screenshot 2023-09-20 at 3 35 48 PM

Delete survey modal:
Screenshot 2023-09-20 at 3 35 59 PM

* Confirm that the three dots menu is only shown in the sandbox view
* Remove a survey from a sandbox environment. Confirm that it still exists in IRB and Live (assuming it was published there in the first place). Confirm that you can remove the survey from IRB and Live by following the publishing flow.
* Make sure that deletes work correctly. Try deleting a survey that's used in IRB or Live- confirm that it throws an error.
* Create a brand new survey that's only used in sandbox. Create a new version of this survey. Confirm that you can delete it and that all versions are deleted from the DB.


api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConfiguredSurveyController.java
```diff
@@ -61,4 +61,15 @@ public ResponseEntity<Object> create(
...
+  @Override
+  public ResponseEntity<Void> remove(
+      String portalShortcode, String studyShortcode, String envName, UUID configuredSurveyId) {
+    AdminUser adminUser = requestService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+
+    surveyExtService.removeConfiguredSurvey(
+        portalShortcode, studyShortcode, environmentName, configuredSurveyId, adminUser);
+    return ResponseEntity.noContent().build();
+  }
```
MatthewBemis, line 74: I'm still in the process of adding backend test coverage for the remove and delete code, but that shouldn't block getting started on reviews, imo

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
```diff
@@ -58,6 +64,59 @@ public Survey create(String portalShortcode, Survey survey, AdminUser adminUser)
     return surveyService.create(survey);
   }
 
+  public void delete(String portalShortcode, String surveyStableId, AdminUser adminUser) {
+    Portal portal = authUtilService.authUserToPortal(adminUser, portalShortcode);
+    List<Survey> existingVersions = surveyService.findByStableId(surveyStableId);
+    List<UUID> existingVersionIds = existingVersions.stream().map(BaseEntity::getId).toList();
+
+    if (existingVersionIds.size() == 0) {
+      throw new NotFoundException("Survey not found in portal " + portal.getName());
```
devonbush: You're not limiting your search to the portal (which is correct).
```suggestion
      throw new NotFoundException("Survey not found");
```
Also, you might as well move this check up a line -- just check the size of the survey list returned from the DB

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
```diff
@@ -58,6 +64,59 @@ public Survey create(String portalShortcode, Survey survey, AdminUser adminUser)
...
+    // Resolve all of the study environments that contain the configured surveys
+    List<StudyEnvironment> referencingStudyEnvs =
+        referencingSurveys.stream()
+            .map(
+                configuredSurvey -> {
+                  return studyEnvironmentService
+                      .find(configuredSurvey.getStudyEnvironmentId())
+                      .get();
+                })
+            .toList();
```
devonbush: Instead of getting the environments one-by-one, make a list of all the studyEnv ids, then grab them with findAll()
```suggestion
    List<StudyEnvironment> referencingStudyEnvs =
       studyEnvironmentService.findAll(
            referencingSurveys.stream()
                  .map( configuredSurvey -> StudyEnvironmentSurvey::getStudyEnvironmentId).toList()
       );
```

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
```diff
@@ -58,6 +64,59 @@ public Survey create(String portalShortcode, Survey survey, AdminUser adminUser)
...
+    if (referencingStudyEnvs.stream()
+        .anyMatch(r -> !EnvironmentName.sandbox.equals(r.getEnvironmentName()))) {
+      throw new IllegalArgumentException(
+          "Survey is referenced by a non-sandbox environment and cannot be deleted");
+    } else {
+      // At this point, only sandbox environments reference the survey, so we can unlink
+      // the survey from each env to prepare for full deletion.
+      referencingSurveys.forEach(
+          configuredSurvey -> {
+            studyEnvironmentSurveyService.delete(
```
devonbush: I'd add a "deleteBySurveyId" method to StudyEnvironmentSurveyService & Dao, leveraging the "deleteByProperty" method in BaseJdbiDao, and then just call that

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
```diff
@@ -58,6 +64,59 @@ public Survey create(String portalShortcode, Survey survey, AdminUser adminUser)
...
+                configuredSurvey.getId(), CascadeProperty.EMPTY_SET);
+          });
+
+      // Delete the survey
+      for (Survey survey : existingVersions) {
+        if (portal.getId().equals(survey.getPortalId())) {
```
devonbush: I think this check should be moved up to the beginning -- the original list of referencingSurveys should be filtered by the Portal.  As you have it, if a version of the survey is owned by another portal, it will be detached but not deleted, which I don't think is what we want

core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
```diff
@@ -45,6 +45,10 @@ public List<StudyEnvironmentSurvey> findAllByStudyEnvIdWithSurvey(UUID studyEnvI
         return studyEnvSurvs;
     }
 
+    public List<StudyEnvironmentSurvey> findAllStudyEnvsWithSurveyId(UUID surveyId) {
```
devonbush: This isn't returning StudyEnvironments, so I think you can simplify the name
```suggestion
    public List<StudyEnvironmentSurvey> findBySurveyId(UUID surveyId) {
```

core/src/main/java/bio/terra/pearl/core/service/study/StudyEnvironmentSurveyService.java
```diff
@@ -28,4 +28,9 @@ public Optional<StudyEnvironmentSurvey> findBySurvey(UUID studyEnvId, String sta
         // environment.  For now, just pick one
         return configs.stream().findFirst();
     }
+
+    public List<StudyEnvironmentSurvey> findAllStudyEnvsWithSurveyId(UUID surveyId) {
```
devonbush: same, rename to findBySurveyId

devonbush (changes requested):
Generally looks solid - one logic gap in how surveys not associated with the given portal are handled.  I'd also recommend stealing the Terra "more actions" button style
image

Also, "Detach" might be a more accurate word than "remove" for where you aren't deleting, just taking it off the environment

breilly2:
This may be a pre-existing issue, but this change makes it apparent. I removed the Mental Health study from sandbox and then copied IRB to sandbox. When returning to the forms and surveys, the Mental Health survey was still missing from the sandbox. A page reload fixes it, so it seems like there's some cached data there that isn't being refreshed after the environment copy. Same thing happens when publishing forward from sandbox to IRB.

Fortunately, the Publish Content screen seems to load enough context to be aware of the real state of things. I can do one copy and then use the target of that copy as the source of another copy and it knows about the survey there.

breilly2:
Attempting to delete a survey from sandbox that has responses results in the survey being removed from sandbox but also and "error deleting survey" notification. I didn't copy it to another environment (so I could actually delete it) and now I think I don't have any way of getting it back. I also don't see the survey or responses when viewing the participant who answered the survey or in the data export.

Side note: To test this, I had to register a new user and go through the required surveys. I don't think we have a function to assign a new survey to an existing user, which is a function that I suspect we may want at some point.

breilly2 (approved):
Agree with Devon's findings. Otherwise code looks good. I found some edge cases during testing that I added as top-level comments on this PR. They don't seem particularly crazy important to me so I won't block this PR, but others might have other opinions.

devonbush:
agree, we're missing a master "view all surveys" UX where you could get detached surveys back.  But that's a ticket for another day :)

devonbush (approved):
I'd like that portalId check to be moved up to the beginning, but I can live without it

MatthewBemis:
@devonbush sorry I thought I pushed that, it was sitting on my local branch :) 

39bd493ca83636976c71837727fc7247523849dd

PR #558, opened 2023-09-20
[JN-569] Autofill survey stable ID

MatthewBemis:
#### DESCRIPTION

This adds autofill for survey stable IDs during creation

Create a new survey and verify that as you type a name, the survey stable ID is automatically generated.
Confirm that once you make a single edit to the stable ID, it no longer autofills based on changes to the name field.

### DIFF
--- a/ui-admin/src/study/surveys/CreateSurveyModal.test.tsx
+++ b/ui-admin/src/study/surveys/CreateSurveyModal.test.tsx
@@ -42,4 +42,27 @@ describe('CreateSurveyModal', () => {
     const createButton = screen.getByText('Create')
     expect(createButton).toBeEnabled()
   })
+
+  test('should autofill the stable ID as the user fills in the survey name', async () => {
+    //Arrange
+    const user = userEvent.setup()
+    const studyEnvContext = mockStudyEnvContext()
+    const { RoutedComponent } = setupRouterTest(<CreateSurveyModal
+      studyEnvContext={studyEnvContext}
+      isReadOnlyEnv={false}
+      show={true}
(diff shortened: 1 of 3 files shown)


devonbush (approved):
nice -- I like the extra affordance for manually changing the stableID

breilly2 (approved):
Looks fine if this is what we really want. I believe these stable IDs need to be globally unique, and I think it's likely that people will encounter conflicts in the future. Prefixes would be nice. I guess it doesn't really matter who gets to be the first to own the `medications` stable ID, but it does seem to break with convention of using portal/study prefixes which could be misleading/confusing down the road. Yes, survey editors can change it, but if it's auto-filled for you, there's little incentive to do so.

MatthewBemis:
@breilly2 you're right, ideally it'd be something like this:
Screenshot 2023-09-21 at 4 53 03 PM

We'd probably have to store the portal and study prefixes in the DB somewhere and retrieve those. I'll hold off on merging this for the moment, it wouldn't be much effort to future-proof this to avoid headaches/confusion down the line.

MatthewBemis:
I'm going to merge this because it's a UX improvement over what we have, and doesn't make the issue of uniqueness _worse_ than it already is. I filed that as a future improvement: [JN-596](https://broadworkbench.atlassian.net/browse/JN-596)

[JN-596]: https://broadworkbench.atlassian.net/browse/JN-596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #559, opened 2023-09-21
JN-474 admin kit refresh

devonbush:
#### DESCRIPTION
This adds a "refresh" button on the kit list page to refresh statuses.  this also opens up the status refresh endpoint to all admins, instead of just superusers, as I imagine OurHealth will want to be able to refresh statuses when handling support.
(along the way, this does some touchup styling to the nav tabs)

image

1. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/requested/created`
2. click 'refresh'
3. confirm kit display is refreshed, and a successful call is made to  `POST /api/portals/v1/ourhealth/studies/ourheart/env/sandbox/kits/refreshKitStatuses`
4. (follow the instructions in https://github.com/broadinstitute/juniper/pull/556 if you want to see statuses change)

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
@@ -6,7 +6,6 @@ import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.kit.KitRequest;
 import bio.terra.pearl.core.model.participant.Enrollee;
 import bio.terra.pearl.core.model.study.StudyEnvironment;
-import bio.terra.pearl.core.service.exception.PermissionDeniedException;
 import bio.terra.pearl.core.service.kit.KitRequestService;
 import bio.terra.pearl.core.service.study.StudyEnvironmentService;
 import bio.terra.pearl.core.service.study.StudyService;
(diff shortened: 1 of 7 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
```diff
@@ -72,9 +71,6 @@ public Collection<KitRequest> getKitRequests(AdminUser adminUser, String enrolle
 
   public void refreshKitStatuses(
       AdminUser adminUser, String portalShortcode, String studyShortcode) {
-    if (!adminUser.isSuperuser()) {
```
breilly2, line 75: Batch status fetching is intentionally done during non-busy hours for DSM to avoid the chance of overwhelming it. I think we can rely on us as superusers to understand that and weigh the risk of consequences, but I don't expect study staff to know, nor should they care. If you want to open this function up to study staff, please discuss with Pegah and the DSM team.

Instead, the idea was to expose (probably in the individual enrollee kit view) the date/time that the current known status was fetched and allow study staff to refresh the status for that one kit (for which there is already a service function, but not yet an API endpoint).

breilly2 (commented):
I like the concept and where you added it in the UI. The only issue is the question of whether or not we can/should offer it to non-superusers.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
breilly2: As a compromise, can we add the UI enhancement while keeping it superuser-only for now until we can have a conversation about it with the DSM team?

MatthewBemis (approved):
Code looks good. Approving now, but I agree with Brian about checking with the DSM team first before opening up to all admins.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
devonbush: good idea

devonbush:
I've hidden it for non-superusers in the UI which should avoid accidental spam-refreshes, while also giving us the ability to direct them to the swagger api if it turns out refreshes are useful to them

PR #562, opened 2023-09-21
JN-557 - robust handling of calculated values

devonbush:
#### DESCRIPTION
In implementing the improved OurHealth height/weight questions, we're using calculatedValues in regular surveys for the first time (previously, we only used them in consent and pre-enrollment forms).  This requires updating autosave to capture computed values (consents and pre-enrolls do not use autosave).  This also requires updating survey parsing to include calculated values as "questions", so that they can properly be included in the data export

1. repopulate ourhealth
2. log into the participant view as consented@test.com
3. fill out the height and/or weight questions in the basics survey (you don't have to fill out anything else)
4. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHSENT/surveys/oh_oh_basicInfo` 
5. confirm the entered and derived values appear
image

6. go to the export preview `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/export/dataBrowser`
7. confirm the entered and derived values appear correctly

### DIFF
--- a/populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/surveys/basic.json
+++ b/populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/surveys/basic.json
@@ -108,7 +108,7 @@
             "type": "text",
             "name": "oh_oh_basic_rawWeight",
             "visibleIf": "{oh_oh_basic_weightUnit} notempty",
-            "title": "Weight in {weightUnit}",
+            "title": "Weight in {oh_oh_basic_weightUnit}",
             "isRequired": true,
             "validators": [
               {
--- a/populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/surveys/basic.json
+++ b/populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/surveys/basic.json
@@ -48,12 +48,14 @@
                 "type": "text",
                 "title": "First name",
                 "isRequired": true
-              }, {
+              },
+              {
                 "name": "oh_oh_basic_lastName",
                 "type": "text",
                 "title": "Last name",
                 "isRequired": true
-              }, {
+              },
+              {
                 "name": "oh_oh_basic_middleInitial",
                 "type": "text",
                 "title": "Middle initial",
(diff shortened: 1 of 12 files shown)


breilly2:
@devonbush Is it safe to assume that the size of the JSON diff is due to incorporating edits made by the OurHealth team in prod? It looks like there are some minor changes beyond just height/weight and lots of formatting changes (like removing some "select all that apply" text).

devonbush:
yes, the basic info survey was updated to be what's currently on demo sandbox, which has been synced with prod

devonbush:
the sonarcloud hotspot is a warning to make sure the regex -- declared as a static final constant, is not vulnerable to DDoS attacks.  So I'm going to say, no, it is not vulnerable...

breilly2:
Testing note: At least by the time we're exporting data (if not earlier), I think we should blank out the height value(s) for the units not being used. Otherwise, conflicting information could be entered. The logic for reading height _should_ be "first check the unit then read the value(s) for that unit", but someone analyzing the data could instead write something like "if cm is not empty, use that, otherwise use ft/in" or something like that. Blanking out the values not being used (and in fact not being shown to the participant, so not expected to be submitted) would avoid that.

breilly2 (commented):
Code looks good for the current behavior. I do think that the behavior should change slightly as I mentioned in my testing note comment.

breilly2 (approved):

populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/surveys/basic.json
```diff
@@ -77,197 +81,114 @@
...
+                "value": "cm",
+                "text": "centimeters (cm)"
+              }
             ]
           },
+          {
+            "type": "text",
+            "name": "oh_oh_basic_rawHeightCm",
+            "visibleIf": "{oh_oh_basic_heightUnit} = 'cm'",
+            "title": "Height in {heightUnit}",
```
MatthewBemis:
```suggestion
            "title": "Height in {oh_oh_basic_heightUnit}",
```

populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/surveys/basic.json
```diff
@@ -77,197 +81,114 @@
...
+                "value": "kg",
+                "text": "kilograms (kg)"
+              }
+            ]
+          },
+          {
+            "type": "text",
+            "name": "oh_oh_basic_rawWeight",
+            "visibleIf": "{oh_oh_basic_weightUnit} notempty",
+            "title": "Weight in {weightUnit}",
```
MatthewBemis:
```suggestion
            "title": "Weight in {oh_oh_basic_weightUnit}",
```

MatthewBemis (approved):
Looks good. I suggested 2 changes to the survey json that should be made to fix this text substitution issue:
 
Screenshot 2023-09-25 at 9 18 13 AM

PR #563, opened 2023-09-21
[JN-563] Use correct B2C policy name for api-admin

MatthewBemis:
#### DESCRIPTION

The script was incorrectly using the participant policy name instead of the admin policy name.

Verify that the setup script still works


breilly2 (approved):
Can also test by rendering the environment, starting the admin UI, and signing in with B2C instead of the developer login. Should see the "Sign in with [Google|Microsoft]" buttons instead of the participant email/password form.

devonbush (approved):

PR #564, opened 2023-09-22
JN-590 - fixing consent form view for past versions

devonbush:
#### DESCRIPTION

Previously, If a participant completes a consent form, and then a new version of the consent form is released -- the participant's response no longer appears in the admin tool.  this fixes that behavior.  The fix is done in the same way that the survey code is fixed.  A later refactoring should change the architecture of the participant view -- right now it lays out the current framework of the studyEnvironment, and tries to map the enrollee's information to it, which can lead to omissions if the study environment has changed significantly from when the enrollee participated.  It should instead just list all the enrollee data, and then load the supporting forms as-needed

1. repopulate ourhealth
2. update the consent form by making a trivial change
3. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHSENT
4. confirm that the consent form response shows up as completed

### DIFF
--- a/ui-admin/src/study/participants/enrolleeView/EnrolleeView.test.tsx
+++ b/ui-admin/src/study/participants/enrolleeView/EnrolleeView.test.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 
 import EnrolleeView from './EnrolleeView'
 import { setupRouterTest } from 'test-utils/router-testing-utils'
-import { mockEnrollee, mockStudyEnvContext, taskForSurvey } from 'test-utils/mocking-utils'
+import { mockEnrollee, mockStudyEnvContext, taskForForm } from 'test-utils/mocking-utils'
 import { render, screen } from '@testing-library/react'
 
 
@@ -19,7 +19,6 @@ test('renders survey links for configured surveys', async () => {
   expect(surveyLink.querySelector('span')).toBeNull()
 })
 
-
 test('renders survey taken badges', async () => {
   const studyEnvContext = mockStudyEnvContext()
   const enrollee = mockEnrollee()
@@ -30,7 +29,8 @@ test('renders survey taken badges', async () => {
     complete: false,
     enrolleeId: enrollee.id
   })
-  enrollee.participantTasks.push(taskForSurvey(studyEnvContext.currentEnv.configuredSurveys[0].survey, enrollee.id))
+  enrollee.participantTasks
+    .push(taskForForm(studyEnvContext.currentEnv.configuredSurveys[0].survey, enrollee.id, false))
 
   const { RoutedComponent } = setupRouterTest(
     // eslint-disable-next-line @typescript-eslint/no-empty-function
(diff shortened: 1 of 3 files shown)


breilly2:
My first test was great in that I consented OHSENT before running the new code to reproduce the problem. I then ran the new code and saw the consent, which is great!

However, I then consented OHNEWB and noticed that I saw the old consent. Fair enough... OHNEWB registered when v1 of the consent was active. I then joined with a new user and, as expected saw v2 of the consent. The problem is, when I now look at the consents for OHSENT and OHNEWB, it's showing me v2 even though they consented to v1!

@devonbush I'd have to try reproducing this again to narrow down the exact conditions, but wanted to raise my preliminary findings in case you can think of an obvious cause and apply an easy fix.

Side note: the specific change that I made to the consent was the text of the agreement answer, so it happens to be very easy to notice in the admin UI. Maybe that's also a clue about the cause.

breilly2:
I did some more comprehensive testing. My test plan was a little more complicated than it needed to be given that this PR only touches admin UI (but I didn't notice that until I was in the middle of the test).

## Initial conditions

OHSENT has consented.
OHBASC has also, but has no answer for the oh_oh_consent_agree question. (!)
OHNEWB has not consented.

## Test plan

### Before applying code change

* Modify the consent form. Change some static text as well as question text and the “agree” answer text.
    * Expect OHSENT’s consent to disappear from the admin UI. (Same for OHNEWB)
    * OHNEWB should still be able to consent… but will the admin UI show it? My guess is that it will show nothing, assuming OHNEWB is presented with consent v1. If OHNEWB is presented with consent v2, the admin should show the v2 consent.
* Join with another account and leave unconsented. We’ll call this OHNEWB2 (though the ID will be something different in practice). Viewing the consent form should show v2.

### After applying code change

* View OHSENT’s consent. It should show v1.
* View OHNEWB’s consent, which should match what OHNEWB saw when consenting.
* Consent OHNEWB2. Admin UI should show v1 (or v2 if OHNEWB is presented with v2).
* View OHSENT’s consent. It should still show v1.
* View OHNEWB’s consent, which should match what OHNEWB saw when consenting.
* Join with another account. Should be presented with consent v2. Admin UI should show v2 consent.

## Test plan execution notes

### Pre code change

* When consenting OHNEWB, I don’t see the question text, “Do you agree to the above?”, so that part of the test won’t yield any data.
* OHNEWB was presented with consent v1 and was able to complete consent. Admin UI shows no consent, just like OHSENT (and OHBASC).
* OHNEWB2 sees consent v2

### Post code change

* OHSENT’s consent shows v2
* Same for (consented) OHNEWB (and OHBASC)
* OHNEWB2 also shows v2, but that’s expected

In all cases, the participant UI shows the correct version of the consent, which is good. Only the admin UI shows the wrong consent, but this is a problem given the print/download feature.

devonbush:
Yup, Brian, that's the expected (and you are correct that it is not desired) behavior.  Currently the admin tool always shows the latest version, and just tries to map the enrollee's data onto it. (this is also what it does for surveys, where that behavior makes a *little* more sense).  This is all based form the admin tool being architected initially as "load blob, show blob".  So we'll want a ticket to update that

devonbush:
In the short term, the desire is to keep ourHealth consents from disappearing when V2 of their consent rolls out (which only differs in trivial typo fixes)

breilly2:
When it comes to something as important as consents, I think having incorrect information is potentially more dangerous than having missing information. The latter leads to a manual investigation to locate the correct data. The former potentially promotes spreading the wrong information. I understand that, right now, we're just talking about a few typos in the OurHealth consent, but it still feels like shaky footing to me. The next time we would likely revisit this is when a participant comes back and says, "that's not the consent that I signed" which would be very bad.

@JeffKorte, what's your thought about this from a product perspective?

devonbush:
I think missing is worse, because it will cause panic on the part of the study staff "what do you mean there's no record of consent?!?!?!" and urgent support escalations.  Again, this is just a stopgap for a couple of weeks

JeffKorte:
If this is just a stop gab iteration in the current sprint that can then be followed up by a ticket to polish this behavior up in the next sprint (I'll write that ticket up now), then I'm OK with releasing this in the meantime.  I agree with the concept of "showing something it better than showing nothing", but also agree with Brian's point that we need to improve upon this quickly so as not to propagate incorrect information.

breilly2 (approved):

MatthewBemis (approved):

PR #565, opened 2023-09-25
JN-591: enabling site content publish

devonbush:
#### DESCRIPTION

This enables selection of sitecontent changes to be published 
image

1. On the ourhealth publishing view, copy your Sandbox env. to irb `https://localhost:3000/ourhealth/studies/ourheart/publishing`
2. Make a small change to the siteContent in the sandbox, and save it, creating a new version
3. go back to the publishing view again, and select copy from sandbox to IRB
4. select the siteContent change
5. hit "copy"
6. confirm the updated version appears

### DIFF
--- a/ui-admin/src/portal/publish/PortalEnvDiffProvider.tsx
+++ b/ui-admin/src/portal/publish/PortalEnvDiffProvider.tsx
@@ -1,4 +1,4 @@
-import { useNavigate, useParams } from 'react-router-dom'
+import { useParams } from 'react-router-dom'
 import React, { useEffect, useState } from 'react'
 import Api, { PortalEnvironmentChange } from 'api/api'
 import { Store } from 'react-notifications-component'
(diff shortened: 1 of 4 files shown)


ui-admin/src/portal/publish/PortalEnvDiffProvider.tsx
```diff
@@ -33,7 +32,8 @@ const PortalEnvDiffProvider = ({ portal, studyShortcode, updatePortal }: PortalE
       const envIndex = updatedPortal.portalEnvironments.findIndex(env => env.environmentName === result.environmentName)
       updatedPortal.portalEnvironments[envIndex] = result
       updatePortal(updatedPortal)
-      navigate(studyPublishingPath(portal.shortcode, studyShortcode))
+      // for now, do a hard reload to make sure all changes are propagated
+      window.location.pathname = studyPublishingPath(portal.shortcode, studyShortcode)
```
MatthewBemis: I know the focus is to unblock, but in the interest of not accumulating small TODOs, it would only be a few line change to do an `await portalContext.reloadPortal(portal.shortcode)` above and keep the `navigate` here.

MatthewBemis (approved):
one small comment but it's not blocking

breilly2 (approved):

PR #569, opened 2023-09-26
[JN-549] adding scroll-to-top for admin

devonbush:
#### DESCRIPTION

When navigating in the admin tool, users should be taken to the top of each new page.  this uses the same component as the participant frontend.  I did not pull it into a common component, because I'm not sure we'll always want scrolling to be the same in admin and participant UXes.  Also, I don't have a test for this -- testing scrolling seemed too fiddly to be worth the cost

1. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/export/dataBrowser`
2. scroll down a little bit
3. click "website" in the left menu
4. confirm you are taken to the top of the website page

### DIFF
--- a/ui-admin/src/App.tsx
+++ b/ui-admin/src/App.tsx
@@ -1,11 +1,11 @@
-import React, {useEffect} from 'react'
+import React, { useEffect } from 'react'
 import 'react-notifications-component/dist/theme.css'
 import 'styles/notifications.css'
 import 'survey-core/defaultV2.min.css'
 import './App.css'
 import './print.css'
 
-import {BrowserRouter, Outlet, Route, Routes, useLocation} from 'react-router-dom'
+import { BrowserRouter, Outlet, Route, Routes, useLocation } from 'react-router-dom'
 import { ReactNotifications } from 'react-notifications-component'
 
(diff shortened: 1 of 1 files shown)


ui-admin/src/App.tsx
```diff
@@ -26,6 +26,16 @@ import AdminSidebar from './navbar/AdminSidebar'
 import NavContextProvider from 'navbar/NavContextProvider'
 import PopulateRouteSelect from './populate/PopulateRouteSelect'
 
+/** auto-scroll-to-top on any navigation */
+const ScrollToTop = () => {
+  const location = useLocation()
+  useEffect(() => {
+    // @ts-expect-error TS thinks "instant" isn't a valid scroll behavior.
+    window.scrollTo({ top: 0, left: 0, behavior: 'instant' })
```
breilly2, line 34: FWIW, `smooth` appears to be Bootstrap's influence as expressed in its `_reset.scss` (applied to the root `html` element). If we care to avoid the `@ts-expect-error`, we could override that opinion with our own of `auto` which will restore instant scrolling. Then we wouldn't have to specify `behavior` here at all.

breilly2 (approved):

MatthewBemis (approved):

PR #570, opened 2023-09-27
[JN-592] Allow leading whitespace and case insensitivity in multiple-combobox questions

MatthewBemis:
#### DESCRIPTION

Previously, users might accidentally type a space when searching for their medication and this would prevent any results from being returned. This changes the search behavior and trims the leading whitespace when trying to match search results. This also treats searches as case-insensitive.

I tried to write a test for this but I couldn't get the react testing utilities to work. I think it has to do with how the virtualization for the MultipleCombobox was being mocked? I can dig into it more if reviewers want.

https://github.com/broadinstitute/juniper/assets/7257391/db308fcd-5b33-4ea3-9999-7234a2808565

Try adding some leading whitespace to a medications search
Try searching for a medication in all caps


devonbush (approved):
Code looks good.  Unless it's inordinately difficult, please add at least one frontend test for the multipleCombobox component--it looks like we don't have any right now.

MatthewBemis:
I added a very basic test to make sure that the combobox at least renders. I opted for this basic test after I struggled to implement something more complex (like testing the functionality in the PR). I think the issue is getting it to work with the mocked virtualization.

denniscunningham (approved):

PR #573, opened 2023-09-29
[JN-488] "actionable" participant notes

devonbush:
#### DESCRIPTION

This adds basic infrastructure for assigning notes to an admin user, and then being able to resolve them.  This is not intended to be a fully-polished workflow, just the first iteration of functionality that will be iterated a lot as we see how people use Juniper and/or what new customers we get. 

image

This includes a new piece of core code -- the 'DataAuditedService' -- the goal is to have a base CRUD class for all our audited models that makes it clearer to the developer "you need to supply audit info to perform crud operations".  If we agree on this pattern, subsequent PRs can extend it out to our other audited models.  (reminder that our database ables generally fall into two categories:  "config" like surveys which are immutable, and "data" which is mutable and audited through DataChangeRecords.

TO REVIEW (recommended review steps since this is a big change)
1.  look at the schema update in `2023_08_25_admin_task.yaml` to get a sense for what tasks are and what we can hook them to
2. look at jsalk.json to see how we populate the tasks attached to notes for developer convenience
3. Click through the instructions below to get a feel for the workings
4. review the rest of the code
5. Look at the DataAuditedService and think about how we may/may not apply it to our other audited entities (answers, profiles, etc...)

1. redeploy ApiAdminApp
6. repopulate OurHealth  (`./scripts/populate_portal.sh ourhealth`)
7. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHSALK`
8. observe that two of the notes have assignees
9. click one of them to pull up the task edit modal
10. Set the status to "COMPLETE" and add a disposition note.
11. click "save"
12. observe the note list gets updated
image


ui-admin/src/study/participants/enrolleeView/ParticipantNoteView.tsx
```diff
@@ -1,37 +1,71 @@
...
-      <span className="text-muted ms-3">{instantToDefaultString(note.createdAt)}</span>
+      </div>
+      <div className="text-muted ms-3">{instantToDefaultString(note.createdAt)}</div>
+      { matchedTask && <button className="badge ms-auto btn-secondary"
+        style={{ backgroundColor: '#eee' }}
+        onClick={() => setShowTaskEdit(!showTaskEdit)}>
+        { matchedTask.status === 'NEW' && <span>
+            Assigned: {users.find(user => user.id === matchedTask.assignedAdminUserId)?.username}
+          <FontAwesomeIcon icon={faCaretDown} className="fa-sm ms-2"/>
+        </span> }
```
MatthewBemis: Unassigned tasks show up as an empty `Assigned: `.

`const matchedAssignedUser = users.find(user => user.id === matchedTask?.assignedAdminUserId)`

would allow you to:

```suggestion
        { matchedTask.status === 'NEW' && <span>
          { matchedAssignedUser ?
            `Assigned: ${matchedAssignedUser.username}` : 'Unassigned' }
          <FontAwesomeIcon icon={faCaretDown} className="fa-sm ms-2"/>
        </span> }
```

ui-admin/src/study/participants/enrolleeView/ParticipantNotesView.tsx
```diff
@@ -18,37 +20,56 @@ type ParticipantNotesViewProps = {
...
+    <h3 className="h4">Notes</h3>
+    <button className="btn btn-secondary" onClick={() => setShowAdd(!showAdd)}>
+      <FontAwesomeIcon icon={faPlus}/> Add
+    </button>
     {showAdd && <div className="pb-3">
       <textarea rows={5} cols={80} value={newNoteText} onChange={e => setNewNoteText(e.target.value)}/>
       <div>
+        <label htmlFor={userSelectId}>Assign to:</label>
+        <AdminUserSelect selectedUser={newNoteAssignee} setSelectedUser={setNewNoteAssignee} users={users}
+          readOnly={false} id={userSelectId}/>
```
MatthewBemis, line 61: Testing note: if you create a new task as Unassigned, it becomes assignable.

core/src/main/java/bio/terra/pearl/core/service/workflow/AdminTaskService.java
```diff
@@ -0,0 +1,85 @@
+package bio.terra.pearl.core.service.workflow;
+
+import bio.terra.pearl.core.dao.workflow.AdminTaskDao;
+import bio.terra.pearl.core.model.participant.Enrollee;
+import bio.terra.pearl.core.model.participant.ParticipantNote;
+import bio.terra.pearl.core.model.workflow.AdminTask;
+import bio.terra.pearl.core.model.workflow.DataAuditInfo;
+import bio.terra.pearl.core.model.workflow.ParticipantTask;
+import bio.terra.pearl.core.service.CascadeProperty;
+import bio.terra.pearl.core.service.CrudService;
```
MatthewBemis:
```suggestion
```

Unused imports

MatthewBemis (approved):
Testing was good, just a few small notes which aren't blockers given this isn't intended to be completely polished. But I still wanted to make note of them so they weren't lost.

Code generally looks good. The DataAuditedService stuff seems easy enough to reason about.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/admin/AdminTaskController.java
```diff
@@ -0,0 +1,71 @@
...
+        adminTaskExtService.getByStudyEnvironment(
+            portalShortcode, studyShortcode, environmentName, includedRelations, user);
+    return ResponseEntity.ok(tasks);
+  }
+
+  @Override
+  public ResponseEntity<Object> getByEnrollee(
+      String portalShortcode, String studyShortcode, String envName, String enrolleeShortcode) {
+    AdminUser user = authUtilService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
```
denniscunningham, line 54: Is this needed? Do we need to validate the env if it is not used?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/AuthUtilService.java
```diff
@@ -104,4 +104,20 @@ public Enrollee authAdminUserToEnrollee(AdminUser user, String enrolleeShortcode
     }
     return enrolleeService.findOneByShortcode(enrolleeShortcode).get();
   }
+
+  /**
+   * returns the enrollee if the user is authorized to access/modify it, throws an error otherwise
+   */
+  public Enrollee authEnrollee(AdminUser user, String enrolleeShortcode) {
```
denniscunningham: Is this different from `authAdminUserToEnrollee`?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/admin/AdminTaskExtService.java
```diff
@@ -0,0 +1,71 @@
...
+    this.studyEnvironmentService = studyEnvironmentService;
+  }
+
+  public AdminTaskService.AdminTaskListDto getByStudyEnvironment(
+      String portalShortcode,
+      String studyShortcode,
+      EnvironmentName environmentName,
+      List<String> includedRelations,
+      AdminUser user) {
+    PortalStudy portalStudy =
```
denniscunningham: Is this assignment necessary?
Noticing that the return value is rarely used, is `verifyUserForStudy` a better method name.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/ParticipantNoteExtService.java
```diff
@@ -4,27 +4,40 @@
...
+import bio.terra.pearl.core.model.workflow.TaskStatus;
 import bio.terra.pearl.core.service.kit.KitRequestService;
 import bio.terra.pearl.core.service.participant.ParticipantNoteService;
+import bio.terra.pearl.core.service.workflow.AdminTaskService;
+import java.util.UUID;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
 
 @Service
 public class ParticipantNoteExtService {
```
denniscunningham, line 18: Not specific to this PR but I'd like to better understand the relationship between `ParticipantNoteExtService` and `ParticipantNoteService`, and similar usages, like `AdminUserExtService. I see that the ext service adds auth and sometimes a bit more, but I'm not sure I completely understand the boundaries.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/admin/AdminTaskExtService.java
```diff
@@ -0,0 +1,71 @@
...
+import bio.terra.pearl.core.model.workflow.AdminTask;
+import bio.terra.pearl.core.model.workflow.DataAuditInfo;
+import bio.terra.pearl.core.service.study.StudyEnvironmentService;
+import bio.terra.pearl.core.service.workflow.AdminTaskService;
+import java.util.List;
+import java.util.UUID;
+import org.springframework.stereotype.Service;
+
+@Service
+public class AdminTaskExtService {
```
denniscunningham, line 17: There is not a way to delete just one task?
Or do tasks just hang around as COMPLETED until the ptp or study is deleted?
If so, I wonder if there is a valid oops user journey, where I created a task and just want to get rid of it.

core/src/main/resources/db/changelog/changesets/2023_08_25_admin_task.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - changeSet:
+      id: "adminTask"
+      author: dbush
+      changes:
+        - createTable:
+            tableName: admin_task
+            columns:
+              - column: { name: id, type: uuid, defaultValueComputed: gen_random_uuid(), constraints: { nullable: false, primaryKey: true } }
+              - column: { name: created_at, type: datetime, constraints: { nullable: false } }
+              - column: { name: last_updated_at, type: datetime, constraints: { nullable: false } }
```
denniscunningham, line 11: Seems like `completed_at` could be computed from `last_updated_at` and `status`.
Are there cases where a completed task can be updated? If so, that feels confusing.

core/src/main/resources/db/changelog/changesets/2023_08_25_admin_task.yaml
```diff
@@ -0,0 +1,43 @@
...
+              - column: { name: completed_at, type: datetime }
+              - column: { name: status, type: text, constraints: { nullable: false } }
+              - column: { name: disposition_note, type: text }
+              - column: { name: description, type: text }
+              - column: { name: study_environment_id, type: uuid, constraints:
+                { nullable: false, foreignKeyName: fk_admin_task_study_env, references: study_environment(id) }
+              }
+              - column: { name: participant_note_id, type: uuid, constraints:
```
denniscunningham, line 19: So we are committing that task will only go with notes?
Or since this is not listed as nullable (I could be reading the constraints incorrectly), that we would join on other tables for other task associations?

core/src/main/java/bio/terra/pearl/core/model/workflow/AdminTask.java
```diff
@@ -0,0 +1,35 @@
...
+public class AdminTask extends BaseEntity {
+    private Instant completedAt;
+    @Builder.Default
+    private TaskStatus status = TaskStatus.NEW;
+    private UUID studyEnvironmentId;
+    private UUID enrolleeId;
+    private UUID participantNoteId;
+    private UUID creatingAdminUserId;
+    private UUID assignedAdminUserId;
+    private String description;
```
denniscunningham, line 33: I'm trying to correlate these fields with the UI...
I see a "Note" field in the update modal... but apparently this has nothing to do with the underlying note the task is assigned to?
In the UI tasks do not seem to have title or `description` separate from the note, so wondering what `description` is for.

ui-admin/src/api/api.tsx
```diff
@@ -613,16 +635,24 @@ export default {
...
+    const response = await fetch(url, this.getGetInit())
+    return await this.processJsonResponse(response)
+  },
+
   async createParticipantNote(portalShortcode: string,
     studyShortcode: string,
     envName: string,
     enrolleeShortcode: string,
-    noteText: string): Promise<ParticipantNote> {
+    note: { text: string, assignedAdminUserId?: string }): Promise<ParticipantNote> {
```
denniscunningham, line 658: There does not appear to be a way to assign a task to an existing note?

denniscunningham (commented):
I've been reviewing this for a few hours, so it's probably a good time to get answers to these questions then finish the review later...

You can see a bit of this in my comments, but it looks like task are heavily wired into notes, which may make it difficult to associate task with other items (although given my Juniper inexperience I am not a great resource for suggesting where else tasks would be useful). 

It also looks like tasks are associated with aspects of enrollment (this is noticeable int the ptp detail page -> Task list), which apparently comes from the origin of the underlying note. But in some of those aspects (sorry, I assume there is a better term) there is not a way to provide a note (that I saw). 

A couple other more minor things...
Having a `note` field on a task that is associated with a note, seems confusing.
It would be helpful to delete a task separately from a note. Common cases are: changed my mind, put the task on the wrong note. etc.

core/src/main/resources/db/changelog/changesets/2023_08_25_admin_task.yaml
devonbush: yup, you're right -- we do want to be able to attach tasks to lots of different things.  `nullable: true` is the default, so this column will be nullable since we don't explicitly say `nullable: false` like we do for some of  the above columns

ui-admin/src/api/api.tsx
devonbush: correct, not yet

core/src/main/java/bio/terra/pearl/core/model/workflow/AdminTask.java
devonbush: I figured `description` might be useful for tasks that are attached to things that aren't notes (e.g. kit requests).  You're right it's not used yet

core/src/main/resources/db/changelog/changesets/2023_08_25_admin_task.yaml
devonbush: We could--although I've gotten burned on that a lot in the past though.  What if someone updates the disposition note (maybe fixing a typo) after completion, or makes some other change?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/admin/AdminTaskExtService.java
devonbush: currently, yes, tasks just hang around as "completed".  The idea is that it's kind of like Jira -- you don't delete tasks, you just mark them as done and they (hopefully) fade from view.  Obviously, the UI on this isn't polished yet, but that's the idea.  I think you make a good case for the accidental "oops, misclick, undo that" task deletion desire, although in audited environments like ours, it's sometimes cleaner to just mark "oopses" as "complete" or "rejected" rather than having to manage a separate delete workflow.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/ParticipantNoteExtService.java
devonbush: yeah, https://broadworkbench.atlassian.net/wiki/spaces/PEARL/pages/2841509953/Example+code+-+Mailing+list+administration walks through the layer logic -- you're right that the boundaries are blurry.  Probably worth a discussion in a week or so once you've seen a bit more code to see if we can identify a better pattern.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/AuthUtilService.java
devonbush: lol, no, and it's also not used.  I must have been thinking of doing something, changed my mind, and forgot to cleanup.  good catch

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/admin/AdminTaskController.java
devonbush: good question.   My thought It's probably good to validate it, since it can catch typos in enrollee shortcodes (on the off chance someone is using the API directly), but agree it's not strictly necessary to validate.  (And I think you've already identified a couple places in our codebase where we don't validate it)

devonbush:
Thanks for your great comments Dennis and Matt.  I'll be sure bring up the flow and functional issues  you raised in demo tomorrow to the team

api-admin/src/main/java/bio/terra/pearl/api/admin/service/admin/AdminTaskExtService.java
devonbush: good catch, yup, assignment isn't needed. 

  I'm not sure I can parse a difference between verifyUserForStudy and authUserToStudy as far as implied return types.  is it the difference between "for" and "to"?

ui-admin/src/api/api.tsx
denniscunningham: Just curious why not? Was it deemed unimportant for initial use? Seems like a common use case.

core/src/main/java/bio/terra/pearl/core/model/workflow/AdminTask.java
denniscunningham: I brought this up in a separate comment, but I'm wondering if the "Note" field on a task, might be confusing, especially since it seems like there is not a way to edit notes (but I could have missed that)?
You can imagine without note editing, the note field in an associated task becomes a way to update a note, but I assume the "Task" note field is really designed to supplement the tasks, like "Can you complete this by Friday".
Not sure if this is a terminology thing, a design thing, or me not understanding. :-)

core/src/main/java/bio/terra/pearl/core/service/DataAuditedService.java
```diff
@@ -0,0 +1,164 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * base class for entities that should have changes tracked via DataChangeRecords.
+ * This currently allows 'auditInfo' to be null--a later PR should remove that, once we upgrade all the calling services
+ * to also support passing auditInfo to create/deletes
+ * */
+public abstract class DataAuditedService<M extends BaseEntity, D extends BaseMutableJdbiDao<M>> {
```
denniscunningham, line 24: Is the idea that this class will be implicitly tested by higher level tests, or is there not a current way to deal with an actual DB for automated tests (thus testing at this level would not be very effective)?

core/src/main/java/bio/terra/pearl/core/service/workflow/DataChangeRecordService.java
```diff
@@ -14,7 +16,7 @@ public class DataChangeRecordService extends ImmutableEntityService<DataChangeRe
     private EnrolleeService enrolleeService;
 
 
-    public DataChangeRecordService(DataChangeRecordDao dao, EnrolleeService enrolleeService) {
+    public DataChangeRecordService(DataChangeRecordDao dao, @Lazy EnrolleeService enrolleeService) {
         super(dao);
```
denniscunningham, line 20: Just curious about the change to @lazy

denniscunningham (approved):
I don't want to get in the way getting this PR completed, so I just added a few more questions. Nice work, and thank you for your patience. 

Since I am currently working on deleting users, what should happen when an assignee is deleted? I assume there are other places where admin users are associated with Juniper items? (I can probably check that by looking at DB foreign keys.)

core/src/main/java/bio/terra/pearl/core/model/workflow/AdminTask.java
devonbush: I'm not sure which "note" field you mean -- tasks can be linked to ParticipantNotes?  But not all tasks are linked to ParticipantNotes, and so there's a need for multiple ways to specify text

core/src/main/java/bio/terra/pearl/core/service/DataAuditedService.java
devonbush: Yes, the idea is that this is tested by child classes.  All our automated tests are run against an actual postgres db (that's in the same docker container as the application postgres db--see `core/src/test/resources/application.yaml` for the config).  Once we get a sense of how many services will inherit from this, it might be worth testing this independently with some model subclass, but for now the implicit coverage by the one-and-only subclass should be good enough

core/src/main/java/bio/terra/pearl/core/service/workflow/DataChangeRecordService.java
devonbush: Was needed to resolve a circular dependency.  IIRC it was EnrolleeService -> DataChangeRecordService -> AdminTaskService -> EnrolleeService.  Which are all dependencies that are needed, so we just have to specify the order they're loaded

ui-admin/src/api/api.tsx
devonbush: yup, you're right that's a common case and I'm sure we'll support it soon enough -- a lot of leaving it out was driven by my desire to cut off this PR before it got even more unwieldy.  :)

PR #574, opened 2023-10-02
[JN-564] Fix: Switching to an uninitialized study env breaks the survey view

MatthewBemis:
#### DESCRIPTION

Previously if you viewed a survey in the admin tool and switched to an uninitialized study env, the screen would go blank and the user would be forced to refresh. This adds a check earlier in the component to check if the study env is initialized and display a message if not.

1) Repopulate ourhealth: `./scripts/populate_portal.sh ourhealth`
2) Load a survey in the sandbox env: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms/surveys/oh_oh_basicInfo?readOnly=false
3) Use the study environment dropdown to switch to IRB or Live.
4) You should see `Study environment not initialized` rather than a blank screen or React error

### DIFF
--- a/ui-admin/src/study/surveys/SurveyView.tsx
+++ b/ui-admin/src/study/surveys/SurveyView.tsx
@@ -89,8 +89,13 @@ export const useSurveyParams = () => {
 
 /** routable component for survey editing */
 function SurveyView({ studyEnvContext }: {studyEnvContext: StudyEnvContextT}) {
-  const { isReadOnly, version, stableId } = useSurveyParams()
   const { currentEnv, portal } = studyEnvContext
+
+  if (!currentEnv.studyEnvironmentConfig.initialized) {
+    return <span>Study environment not initialized</span>
+  }
+
+  const { isReadOnly, version, stableId } = useSurveyParams()
   const applyReadOnly = isReadOnly || currentEnv.environmentName !== 'sandbox'
   const envSurvey = currentEnv.configuredSurveys
     .find(s => s.survey.stableId === stableId)?.survey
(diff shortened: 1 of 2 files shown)


ui-admin/src/study/surveys/SurveyView.tsx
```diff
@@ -89,8 +89,13 @@ export const useSurveyParams = () => {
...
 /** routable component for survey editing */
 function SurveyView({ studyEnvContext }: {studyEnvContext: StudyEnvContextT}) {
-  const { isReadOnly, version, stableId } = useSurveyParams()
   const { currentEnv, portal } = studyEnvContext
+
+  if (!currentEnv.studyEnvironmentConfig.initialized) {
+    return <span>Study environment not initialized</span>
+  }
+
+  const { isReadOnly, version, stableId } = useSurveyParams()
```
devonbush, line 98: in general, you shouldn't return before a hook (we're somewhat sloppy with this in the code)  See https://stackoverflow.com/questions/54938236/can-you-early-return-with-react-hooks.  

We've been super-sloppy with this in the codebase (you can see we already violated that rule in this component below your edit).  The correct thing here is to make replace the SurveyView component with a useRoutedSurvey hook that contains all the logic that SurveyView has now and returns {isLoading, readOnly, survey}.  
Then have RawSurveyView call the new useRoutedSurvey hook.  Then RawSurveyView will own all the display logic (it will take over rendering the LoadingSpinner).  And then you can rename RawSurveyView to SurveyView.

devonbush: We have a ton of places in our codebase where we have loader/router "components" that really should be refactored into pure hooks.  So as you are updating the codebase, be on the lookout for refactors like this

devonbush (changes requested):

PR #575, opened 2023-10-02
JN-575 adding last login column to participant list

devonbush:
#### DESCRIPTION

This adds a "last login" column to the participant list.  Technically, this will show the last time they have logged into *any* Juniper study, not just the last time they logged into the given study.  But tracking per-study logins is a ticket for another day.

image

This accomplishes this by adding another join to the base search query.  This is fine since enrollee->participant_user is a many-to-one so it shouldn't unduly impact performance.  But eventually we'll want to shift the search architecture from a single large query returning a complex object to a few small queries returning collections--likely when we start adding answers to search results.   

1. redeploy ApiAdminApp
2. repopualte ourHealth 
3. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants
4. confirm you see the login column and populated values


devonbush (commented):
drive by fix of a missing annotation

MatthewBemis (approved):
Looks good. Not critical, but should random lastLogin times be added to the bulk populate endpoint as well?

denniscunningham:
I'm sure I could figure this out with some study, but to save some time on this review: It might help me better understand `ParticipantUser` if you could explain why there are many `Enrollee` to one `ParticipantUser`.

devonbush:
ParticipantUser is a one-to-one mapping to a B2C record.  A B2C login could correspond to multiple enrollees if a given user joins mulitple studies (e.g. a HeartHive user joins two different HeartHive studies -- each study would create a separate enrollment record).  What's currently fuzzy is how this plays into PortalParticipantUser. Orginally, Juniper was designed as single-login across all Portals.  And so someone would log into Juniper, and then pick which portal(s) to join.  Then later we decided we also needed to support whitelabelling, which meant separate B2C tenants for the whitelabeled customers so they'd be isolated.  It's still not clear yet whether everything will be witelabelled (and so PortalParticipantUser is redundant since each ParticipantUser (b2c account) corresponds to only one portal.

devonbush:
good idea on updating the bulk populate too -- I also added a trivial shell script to interface with the bulk populate for convenience

core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeSearchDao.java
```diff
@@ -96,7 +101,12 @@ protected String generateSearchQueryString(List<SqlSearchableFacet> facets) {
     selects.add(0, baseSelectString);
     String selectQuery = selects.stream().collect(Collectors.joining(","));
 
-    String baseFromQuery = " from enrollee left join profile on profile.id = enrollee.profile_id left join kit_request on enrollee.id = kit_request.enrollee_id";
+    String baseFromQuery = """
```
denniscunningham, line 104: Curious why this is not a class level static string.
Is this join sequence unlikely to be reused or amended as new tables are added?

core/src/main/java/bio/terra/pearl/core/service/participant/ParticipantUserService.java
```diff
@@ -14,7 +15,7 @@
 import java.util.UUID;
 
 @Service
-public class ParticipantUserService extends ImmutableEntityService<ParticipantUser, ParticipantUserDao> {
```
denniscunningham, line 17: Is this a drive-by change?
Not sure I understand how this fits in here.

denniscunningham (approved):
Looks good.
A couple of questions.

core/src/main/java/bio/terra/pearl/core/service/participant/ParticipantUserService.java
devonbush: Good question -- I needed to be able to update the lastLogin value of an existing user using this service, which means it needs the `update` method from CrudService.  Previously, updates to lastLogin were only done through CurrentUserService, and so the fact that this inherited from the wrong class (ParticipantUser objects are definitely not immutable) wasn't noticed

core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeSearchDao.java
devonbush: Yeah, I could see this going either way -- no plans to reuse it so I thought the code was more readable to have it in the flow.  I know Pepper used to be relatively strict on "all strings declared at top of files", but I have not adopted that convention for Juniper

PR #576, opened 2023-10-04
[JN-496] Save SendGrid activity logs

MatthewBemis:
#### DESCRIPTION

SendGrid provides a maximum of 30 days of activity history. History includes information about deliverability of emails, as well as user interactions with emails (number of opens and clicks). This PR brings all of that event history into our database by adding a new background process that queries for new activity every 30 minutes. The primary reason for doing this is so we can keep track of history beyond 30 days, which will provide a long-term audit trail and make it easier to debug email issues as they come up. It's also a bit easier to query our Postgres database than deal with the Sendgrid API/UI directly.

In the future we can think about surfacing some of this information in the admin UI. It might be nice to show an "opened" column in the participant email history table so study admins can see if a user has read the email and follow up accordingly.

**PREREQ BEFORE MERGING/RELEASING**: Upgrade production SendGrid account to have 30-day activity history (dev is already upgraded)

Test backfilling:

* Start admin API server
* Wait ~1 minute for the scheduled process to kick off. You should see a log message such as: `Querying for SendGrid activity between 2023-09-04T00:34:58.257Z and 2023-10-04T00:34:58.257Z`
* You should eventually see a message like `Sendgrid activity query returned 504 events`
* If this process is running for the first time, it will backfill 30 days worth of events. This is the maximum amount made available to us by SendGrid. Otherwise, it will query for events that have occurred since the most recently recorded event.
* Verify that the events look good compared to what's in https://app.sendgrid.com/email_activity

Test paginated results:

* Delete everything from the `sendgrid_event` table: `delete from sendgrid_event where 1=1;`
* Reduce `PAGE_SIZE` down to 100 (don't go too low with this otherwise you'll get rate limited very quickly)
* Restart admin API server
* You should see a bunch of log messages such as `Sendgrid activity query returned the limit of 100 events. Querying for additional events between 2023-09-04T00:34:58.257Z and 2023-09-28T17:12:05.000Z`
* Verify that the events look good compared to what's in https://app.sendgrid.com/email_activity

Test new/updated activity:

* Modify the schedule + shedlock settings in ScheduledSendgridActivityService.java to query every minute or so
* Fire off a test email from the Admin UI and open an old email that was sent to you from the *dev or local* server.
* Wait about 5 minutes for the events to propagate to SendGrid
* Query the `sendgrid_event` table to see that the newly sent email is present in the DB, and the recently opened email event has also been recorded (you'll see the `opens_count` field has been incremented by 1)

In general, it's good to cross-reference events in the DB versus what's shown in https://app.sendgrid.com/email_activity


core/src/main/java/bio/terra/pearl/core/service/notification/email/SendgridClient.java
```diff
@@ -38,6 +44,29 @@ public void sendEmail(Mail mail) throws Exception {
     sg.api(request);
```
MatthewBemis, line 46: I looked into what it would take to tie these notifications to the actual events (opens, clicks, etc) that we also record in the new `sendgrid_event` table. It's possible, but there are some things worth noting related to how SendGrid handles Message IDs. See https://docs.sendgrid.com/glossary/what-is-x-message-id and https://docs.sendgrid.com/glossary/message-id

(The TL;DR of the two links is that the message ID returned by the sendEmail endpoint is not quite the same as the message IDs in the activity logs, but with some finesse you can correlate them)

It most likely wouldn't be a lot of work to deal with this, and it could provide some very cool functionality in the admin UI that would allow study admins to monitor if participants are opening their emails or not, but it felt like scope creep for this ticket.

core/src/main/resources/db/changelog/changesets/2023_10_03_sendgrid_event.yaml
```diff
@@ -0,0 +1,19 @@
...
+      id: "sendgrid_event"
+      author: mbemis
+      changes:
+        - createTable:
+            tableName: sendgrid_event
+            columns:
+              - column: { name: id, type: uuid, defaultValueComputed: gen_random_uuid(), constraints: { nullable: false, primaryKey: true } }
+              - column: { name: msg_id, type: text, constraints: { nullable: false, unique: true } }
+              - column: { name: subject, type: text }
+              - column: { name: to_email, type: text }
```
MatthewBemis, line 12: wondering if `to_email` should be a FK.

Additionally, in my experience so far, none of these nullable fields have ever been null in the API response, but I'm still hesitating to assume that will always be the case

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/ScheduledSendgridActivityService.java
```diff
@@ -0,0 +1,24 @@
+package bio.terra.pearl.api.admin.service.notifications;
+
+import java.util.concurrent.TimeUnit;
+import net.javacrumbs.shedlock.spring.annotation.SchedulerLock;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Service;
+
```
devonbush: this class could use a top-level comment about what it's about.  Also, it would be nice if the classname made clear this is just getting the activity logs -- it's not an activity source itself.  Maybe `ScheduledSendgridEventFetcher` or somesuch

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/SendgridActivityService.java
```diff
@@ -0,0 +1,70 @@
...
+import bio.terra.pearl.core.service.notification.email.SendgridClient;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.List;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+
+@Service
```
devonbush: top level comment on this class would be helpful too
Should this be called "SendgridEventService" since it's corresponding to the SendgridEvent model?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/SendgridActivityService.java
```diff
@@ -0,0 +1,70 @@
...
+    Instant mostRecentEventDate =
+        mostRecentEvent
+            .map(SendgridEvent::getLastEventTime)
+            .orElse(Instant.now().minus(30, ChronoUnit.DAYS));
+
+    List<SendgridEvent> events = getAllRecentSendgridEvents(mostRecentEventDate, Instant.now());
+    // Over the course of gathering the results from Sendgrid's paginated API, it's possible that
+    // we've accumulated duplicate events at the fringes of the pages. The database will handle
+    // this fine, but this could result in the following log message indicating that we've inserted
+    // more events than we actually have.
```
devonbush: great comments in this method, thanks!

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/SendgridActivityService.java
```diff
@@ -0,0 +1,70 @@
...
+import java.time.temporal.ChronoUnit;
+import java.util.List;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+
+@Service
+public class SendgridActivityService {
+  private static final Logger logger = LoggerFactory.getLogger(SendgridActivityService.class);
```
devonbush: Dennis just clued me in that we can use a lombok @Slf4j annotation and that will automatically add an appropriate logger for the class.  I'm very excited about that because I always copy-paste these loggerFactory statements and then forget to change the classname :P

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/SendgridActivityService.java
```diff
@@ -0,0 +1,70 @@
...
+            .orElse(Instant.now().minus(30, ChronoUnit.DAYS));
+
+    List<SendgridEvent> events = getAllRecentSendgridEvents(mostRecentEventDate, Instant.now());
+    // Over the course of gathering the results from Sendgrid's paginated API, it's possible that
+    // we've accumulated duplicate events at the fringes of the pages. The database will handle
+    // this fine, but this could result in the following log message indicating that we've inserted
+    // more events than we actually have.
+    logger.info("Sendgrid activity query returned {} events", events.size());
+
+    sendgridEventDao.bulkUpsert(events);
```
devonbush: this should be pulled out into a separate "bulkUpsert" method in the service class.  We want to make sure that updates to the database are wrapped in `@Transactional`, but we don't want to wrap this entire method, since we don't want to keep a transaction open while we're doing external API calls.

devonbush:
This is good stuff -- really appreciate the attention to detail on the comments and log messages.  Worked great when testing locally.  A couple of comments on naming, and one relatively important adding of `@Transactional` annotation, but otherwise looks good.

devonbush (approved):

core/src/main/resources/db/changelog/changesets/2023_10_03_sendgrid_event.yaml
```diff
@@ -0,0 +1,19 @@
+databaseChangeLog:
+  - changeSet:
+      id: "sendgrid_event"
+      author: mbemis
+      changes:
+        - createTable:
+            tableName: sendgrid_event
+            columns:
+              - column: { name: id, type: uuid, defaultValueComputed: gen_random_uuid(), constraints: { nullable: false, primaryKey: true } }
+              - column: { name: msg_id, type: text, constraints: { nullable: false, unique: true } }
```
denniscunningham, line 10: Is `msg_id` a Sendgrid ID? Is it important to keep? (Do we need to go back to the sendgrid service during the period they retain their info?)

denniscunningham: I now see why this is important to deal to help us handle updates, since we need to page and deal with date ranges, but I am still curious if `msg_id` is potentially useful in other ways.

denniscunningham (approved):
Nice!
I left a couple questions.

core/src/main/resources/db/changelog/changesets/2023_10_03_sendgrid_event.yaml
MatthewBemis: It'll be potentially useful for tying these activity events with the actual notification events that triggered them from within Juniper. SendGrid returns a partial message ID when you initially send the email, so if we also save that partial ID (not done in this PR), we can then attempt to correlate that with the `msg_id` in this new activity table. This comment has a little bit more context about this: https://github.com/broadinstitute/juniper/pull/576#discussion_r1345033213

I also imagine it's useful to have the `msg_id` if we ever need to follow up with SendGrid support about an issue with a specific email for some reason, especially if it falls outside of our 30 day activity window. They'll only give us 30 days of activity, but I would be very surprised if they didn't hold onto this information for much longer.

PR #577, opened 2023-10-04
JN-606 Adding integration UX dashboard

devonbush:
#### DESCRIPTION

This adds a new section "Integrations" to superuser functions in the admin UI.  The goal is to support dashboards that show config, debug and test information for external integrations.  It adds a Kit Integration dashboard which for now is just a "test" button.  To support the "Test" button, the status sync features have been updated to be environment specific so that the "Test" button can be pressed without causing any side effects in live data.  (Eventually we might want a dedicated test endpoint, but the goal was good-enough-for-now).  

This also cleans up the stub service to add logging and use valid Pepper kit statuses

Later PR will update the dashboard with current config information, so the user can see, e.g., which DSM instance is being pointed to.

1. Redeploy ApiAdminApp
2. go to `https://localhost:3000/integrations/kits`
3. click the "Test" button
4. confirm a success message appears
5. <Optional> change StubPepperDsmClient.java  `fetchKitStatusByStudy` to throw an exception, and retry.  Confirm error is indicated in UI

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/kit/KitController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/kit/KitController.java
@@ -92,9 +92,11 @@ public class KitController implements KitApi {
   }
 
   @Override
-  public ResponseEntity<Void> refreshKitStatuses(String portalShortcode, String studyShortcode) {
+  public ResponseEntity<Void> refreshKitStatuses(
+      String portalShortcode, String studyShortcode, String envName) {
     AdminUser adminUser = authUtilService.requireAdminUser(request);
-    kitExtService.refreshKitStatuses(adminUser, portalShortcode, studyShortcode);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+    kitExtService.refreshKitStatuses(adminUser, portalShortcode, studyShortcode, environmentName);
     return ResponseEntity.noContent().build();
   }
 }
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
@@ -70,9 +70,12 @@ public class KitExtService {
   }
 
   public void refreshKitStatuses(
-      AdminUser adminUser, String portalShortcode, String studyShortcode) {
+      AdminUser adminUser,
+      String portalShortcode,
+      String studyShortcode,
+      EnvironmentName environmentName) {
     var portalStudy = authUtilService.authUserToStudy(adminUser, portalShortcode, studyShortcode);
     var study = studyService.find(portalStudy.getStudyId()).get();
-    kitRequestService.syncKitStatusesForStudy(study);
+    kitRequestService.syncKitStatusesForStudy(study, environmentName);
   }
 }
--- a/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
@@ -2,6 +2,7 @@ package bio.terra.pearl.core.service.kit;
 
 import bio.terra.pearl.core.dao.kit.KitRequestDao;
 import bio.terra.pearl.core.dao.kit.KitTypeDao;
+import bio.terra.pearl.core.model.EnvironmentName;
 import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.kit.KitRequest;
 import bio.terra.pearl.core.model.kit.KitRequestStatus;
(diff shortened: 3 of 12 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/controller/kit/KitController.java
```diff
@@ -92,9 +92,11 @@ public ResponseEntity<Object> requestKits(
   }
 
   @Override
-  public ResponseEntity<Void> refreshKitStatuses(String portalShortcode, String studyShortcode) {
+  public ResponseEntity<Void> refreshKitStatuses(
+      String portalShortcode, String studyShortcode, String envName) {
     AdminUser adminUser = authUtilService.requireAdminUser(request);
-    kitExtService.refreshKitStatuses(adminUser, portalShortcode, studyShortcode);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
```
denniscunningham, line 98: Just curious, where in the call stack are exceptions processed (to build a response)?

core/src/main/java/bio/terra/pearl/core/service/kit/StubPepperDSMClient.java
```diff
@@ -6,6 +6,8 @@
 import bio.terra.pearl.core.model.participant.Enrollee;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
denniscunningham, line 11: I think you automate this with a @Slf4j annotation on the class.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
```diff
@@ -70,9 +70,12 @@ public Collection<KitRequest> getKitRequests(AdminUser adminUser, String enrolle
   }
 
   public void refreshKitStatuses(
-      AdminUser adminUser, String portalShortcode, String studyShortcode) {
```
denniscunningham, line 73: I notice `RequestKits` (above) has an environmentName parameter that is not used. Is that good for now?

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -179,19 +182,22 @@ public void syncKitStatusesForStudy(Study study) {
                 Collectors.toMap(PepperKitStatus::getJuniperKitId, Function.identity(),
                         (kit1, kit2) -> !kit1.getCurrentStatus().equals("Deactivated") ? kit1 : kit2));
 
-        var studyEnvironments = studyEnvironmentService.findByStudy(study.getId());
-        for (StudyEnvironment studyEnvironment : studyEnvironments) {
-            var incompleteKits = findIncompleteKits(studyEnvironment.getId());
+        studyEnvironmentService.findByStudy(study.getShortcode(), environmentName).ifPresent(
+                studyEnvironment -> {
+                    var incompleteKits = findIncompleteKits(studyEnvironment.getId());
```
denniscunningham, line 187: Apparently we are good with using `var` rather than explicit types?

core/src/main/java/bio/terra/pearl/core/service/kit/StubPepperDSMClient.java
```diff
@@ -40,19 +44,21 @@ public String sendKitRequest(String studyShortcode, Enrollee enrollee, KitReques
 
     @Override
     public PepperKitStatus fetchKitStatus(UUID kitRequestId) {
+        logger.info("STUB fethcing kit status");
         return PepperKitStatus.builder()
                 .juniperKitId(kitRequestId.toString())
-                .currentStatus("SHIPPED")
+                .currentStatus("SENT")
```
denniscunningham, line 51: Should SENT be defined in a constant somewhere (for confidently changing the as we do here)?

denniscunningham:
By "causing any side effects in live data", do you mean performance impact?

denniscunningham (approved):
Added some comments and questions, but nothing that will hold up this PR.

MatthewBemis (approved):

devonbush:
I mean it won't update any live data -- no statuses will be unchanged.  And theoretically the performance impacts should be minimal because I wouldn't expect more than ~20 participants in the sandbox.  But it's more about data than performance

core/src/main/java/bio/terra/pearl/core/service/kit/StubPepperDSMClient.java
devonbush: did not know about that annotation -- nice!

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/kit/KitController.java
devonbush: that's a great question -- right now we have a very basic `GlobalExceptionHandler` class that use magic Spring annotations to package up exceptions into JSON responses.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
devonbush: yeah, that's good for now -- the convention we've adopted is that API urls always include an environment if applicable, even if it might be redundant (above the environment name is redundant to the enrollee shortcode) to make the access logs easier to filter

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
devonbush: I don't have a strong opinion on var -- I typically don't use `var` it unless the type is both particularly gruesome to type out and obvious.  Brian was more of a fan of it in regular use.  What's your take?

core/src/main/java/bio/terra/pearl/core/service/kit/StubPepperDSMClient.java
devonbush: good thought.  Here the relevant enum is PepperKitStatus, and this test is testing that we can use both upper and lowercase versions of the value, so we're not using the enum itself

PR #578, opened 2023-10-04
[JN-574] Participant table client side pagination

MatthewBemis:
#### DESCRIPTION

This adds client side pagination to the participant table. Based on Erin's mockups in [JN-574](https://broadworkbench.atlassian.net/browse/JN-574)

Screenshot 2023-10-04 at 1 42 13 PM

* Start up local admin API + UI
* Run the bulk populate script to add 1000 new users (see the new helper script [here](https://github.com/broadinstitute/juniper/blob/development/scripts/populate_enrollee_bulk.sh))
* Play around with pagination and filtering to make sure it works as expected and performs well (I tested with about 4000 participants and it was fine)
* Test out that your pageSize preference is remembered after refreshing and that it's updated when you change to a different pageSize. Also verify that this preference persists across study environments within the same study.

[JN-574]: https://broadworkbench.atlassian.net/browse/JN-574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui-admin/src/study/participants/participantList/ParticipantList.tsx
+++ b/ui-admin/src/study/participants/participantList/ParticipantList.tsx
@@ -135,7 +135,7 @@ function ParticipantList({ studyEnvContext }: {studyEnvContext: StudyEnvContextT
     },
     initialState: {
       pagination: {
-        pageSize: Number(localStorage.getItem(preferredNumRowsKey) || '10')
+        pageSize: parseInt(localStorage.getItem(preferredNumRowsKey) || '10')
       }
     },
     onColumnVisibilityChange: setColumnVisibility,
--- a/ui-admin/src/study/participants/participantList/ParticipantList.tsx
+++ b/ui-admin/src/study/participants/participantList/ParticipantList.tsx
@@ -6,7 +6,7 @@ import { StudyEnvContextT } from '../../StudyEnvironmentRouter'
 import {
   ColumnDef,
   getCoreRowModel,
-  getFilteredRowModel,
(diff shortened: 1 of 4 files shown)


ui-admin/src/study/participants/participantList/ParticipantList.tsx
```diff
@@ -131,12 +133,18 @@ function ParticipantList({ studyEnvContext }: {studyEnvContext: StudyEnvContextT
       rowSelection,
       columnVisibility
     },
+    initialState: {
+      pagination: {
+        pageSize: parseInt(localStorage.getItem(preferredNumRowsKey) || '10')
```
devonbush: lets pull this computation into a separate line
```
const pageSize = parseInt(localStorage.getItem(preferredNumRowsKey) || '10')
```

devonbush (approved):
Originally I was thinking that these params would be url-persisted.  I think you're right that the pageSize probably belongs in localStorage for user convenience.  I would lean towards also making it available as a url param so that links can be shared with higher fidelity.  And then the table uses the url param if specified, and localStorage if not, and the default size if neither exists.  Worth talking about at demo tomorrow

denniscunningham (approved):

MatthewBemis:
Added URL params in 7d1a6cc657858898d8b50974cf903dae4c8048c8 (and subsequently cleaned up in a7c44fe5)

PR #579, opened 2023-10-05
[JN-606] kit configuration UI

devonbush:
#### DESCRIPTION

Adds a screen for showing the config values for the kit configuration.  We can add more values to the config endpoint later as we flesh out the integrations page and other places where seeing environment variables will be useful. 

This also refactors some sub menu styling to give a cleaner, more consistent look to the page.  

image

1. redeploy ApiAdminApp
2. go to `https://localhost:3000/integrations/kits` 
3. observe the config values on the screen


api-admin/src/main/java/bio/terra/pearl/api/admin/service/ConfigExtService.java
```diff
@@ -39,4 +46,33 @@ private Map<String, String> buildConfigMap() {
...
+      throw new PermissionDeniedException("You do not have permission to view this config");
+    }
+    var configMap =
+        Map.of(
+            "pepperDsmConfig",
+            Map.of(
+                "useLiveDsm",
+                pepperDSMConfig.useLiveDsm(),
+                "secret",
+                pepperDSMConfig.getSecret().substring(0, 4)
```
denniscunningham: Nit, but it will likely improve readability... does it make sense to assign `pepperDSMConfig.getSecret()` before assembling `configMap`.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/ConfigExtService.java
```diff
@@ -39,4 +46,33 @@ private Map<String, String> buildConfigMap() {
...
+                    + "..."
+                    + pepperDSMConfig
+                        .getSecret()
+                        .substring(
+                            pepperDSMConfig.getSecret().length() - 3,
+                            pepperDSMConfig.getSecret().length()),
```
MatthewBemis: this assumes that this particular secret will never be rotated for a shorter secret, which could lead to this either revealing the entire secret, or revealing enough of it where it becomes easily guessable. which is probably fine to assume for now, but might not always be the case.

not a blocker and I know you just want it in order to debug live DSM right now, but since it's security related I wanted to at least mention the concern, especially since it might be more applicable to other config maps that we add to this endpoint in the future.

MatthewBemis (approved):
Just one thought re: secrets but 👍

api-admin/src/main/java/bio/terra/pearl/api/admin/service/ConfigExtService.java
denniscunningham: This substringing is failing when I run the code, presumably because there is no secret?

api-admin/src/test/java/bio/terra/pearl/api/admin/service/ConfigExtServiceTests.java
```diff
@@ -28,8 +32,20 @@ public void testConfigMap() {
...
+  public void testInternalConfigRequiresSuperuser() {
+    AdminUser user = AdminUser.builder().superuser(false).build();
+    ConfigExtService configExtService =
+        new ConfigExtService(b2CConfiguration, applicationRoutingPaths, pepperDSMConfig);
+    Assertions.assertThrows(
+        PermissionDeniedException.class,
+        () -> {
+          configExtService.getInternalConfigMap(user);
+        });
+  }
```
denniscunningham, line 53: Is it straightforward to provide `PepperDSMConfig` values to test the actual assembly? (Sorry, I have not used MockBean, so I do not know what that provides.)

core/src/main/java/bio/terra/pearl/core/service/kit/LivePepperDSMClient.java
```diff
@@ -9,7 +9,10 @@
 import lombok.Getter;
 import lombok.Setter;
 import lombok.experimental.Accessors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
denniscunningham: Nit, this can be replaced with @Slf4j.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/ConfigExtService.java
devonbush: yeah, hopefully we never have <10 character secrets, but I like the idea of being super safe

ui-admin/src/api/api.tsx
```diff
@@ -913,6 +917,12 @@ export default {
     return await this.processJsonResponse(response)
   },
 
+  async fetchInternalConfig(): Promise<InternalConfig> {
+    const url = `${API_ROOT}/internal/v1/config`
```
denniscunningham, line 921: How will this work when we add additional, non-kit config to the UI?
Does it make sense to make this kit specific now?

ui-admin/src/integration/KitIntegrationDashboard.tsx
```diff
@@ -20,9 +23,26 @@ export default function KitIntegrationDashboard() {
...
+    setConfig(response)
+  })
+
   return <div>
-    <div>
-            Test kit status sync &nbsp;
+    <h2 className="h3">Kits</h2>
+    <div className="mt-4">
+      <h3 className="h5">Config</h3>
+      {!isLoading && <dl>
```
denniscunningham, line 35: Sorry, dumb question, but dsm config is constant across environments now and in the future?

denniscunningham (changes requested):
It looks like most things are setup to return general Juniper config, across studies, etc., but the UI seems to be kit specific config.
IOW, there is (maybe not?) Juniper-wide config, there are integrations (currently only kits), but the interface to the backend implies there is just one blob of config. How do things work with the next integration, or if Juniper wants to show config that is not for integrations?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/ConfigExtService.java
devonbush: this is fixed by extracting this whole thing into a helper method

api-admin/src/main/java/bio/terra/pearl/api/admin/service/ConfigExtService.java
devonbush: fixed to handle blank strings

api-admin/src/test/java/bio/terra/pearl/api/admin/service/ConfigExtServiceTests.java
devonbush: yeah, we could, I was running fast and loose with this because I want to get it working, and I also always forget how MockBeans work :)

ui-admin/src/api/api.tsx
devonbush: the idea is there is a single config endpoint -- I don't want the API endpoint to dictate how we slice the UI, and some config values may be applicable to multiple integrations.  So this endpoint will be unchanged, we'll just add ux pages (and update the configMap on the backend with the values we want to expose)

ui-admin/src/integration/KitIntegrationDashboard.tsx
devonbush: dsm config is configured via Spring Environment, which is picked up from application.yaml files, which in turn typically delegate to shell environment variables.  So the values are different in every environment.  

the names of the values are hardcoded.  I could see this instead implemented as a UI that just dumps the received config object on the screen, but I think it's helpful for the UI to indicate what config values are expected and be able to order them.

denniscunningham: Thx for that explanation.
I was thinking of sandbox, irb, live, so I probably used the wrong term?

denniscunningham (approved):

ui-admin/src/integration/KitIntegrationDashboard.tsx
devonbush: ah, yup, that was the right term -- I just misunderstood.  Yes, the idea is that sandbox/irb/live will interface with DSM in the same way, so that we can test "real" integrations with sandbox data (e.g. we can ship a kit to someone not enrolled in the study proper).  This might evolve as our integrations become more complex, and we want to fully isolate sandboxes from some external integrations

devonbush: I could see a useful feature where "useLiveDsm" is togglable on/off (maybe even defaulting off) for sandbox/irb environments so it's harder to accidentally ship a kit to a fake participant

PR #580, opened 2023-10-06
[JN-609] adding sort order to export

devonbush:
#### DESCRIPTION

When smoke testing, it's nice to have a predictable order for exports.  This uses most-recent first so that the most recent participant will appear in the export preview which will help study staff confirm the export of the most recent participant.  Along the way, this cleans up some unused params, and adds a test.  

Export was a very-quickly-thrown-together feature that will need some work in the future (most notably `loadAllEnrolleesForExport` loads enrollees individually, which won't scale for long).

Also notable, this PR introduces a DAO getByX() method that takes sorting parameters.  I'm not thrilled with this, since at some level it breaks DAO encapsulation -- column names and SQL sort directions are a DAO concern, and so they shouldn't be specified by services.  But this seems better than adding a million different permutations of getByXSortedByY methods.  I suspect the long term solution is a composable framework, possibly based on JOOQ, that lets DAOs retain responsibility for query formulation, without requiring a separate method for every query.  In the meantime, we can see how this feels

1. repopulate OurHealth `./scripts/populate_portal ourhealth`
2. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/export/dataBrowser`
3. confirm the enrollees appear most-recent-created first (order will be 
OHBASC | OHSALK | OHSENT | OHNEWB | OHKITS )

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
@@ -18,13 +18,13 @@ import bio.terra.pearl.core.service.survey.SurveyResponseService;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.OutputStream;
 import java.util.*;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 
 @Service
+@Slf4j
 public class EnrolleeExportService {
-    private static final Logger logger = LoggerFactory.getLogger(EnrolleeExportService.class);
     private final ProfileService profileService;
     private final AnswerDao answerDao;
     private final SurveyQuestionDefinitionDao surveyQuestionDefinitionDao;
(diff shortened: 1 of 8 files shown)


MatthewBemis (approved):

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -53,17 +53,21 @@ public EnrolleeExportService(ProfileService profileService,
         this.objectMapper = objectMapper;
     }
 
-    public void export(ExportOptions exportOptions, UUID portalId, UUID studyEnvironmentId, OutputStream os) throws Exception {
-        List<ModuleExportInfo> moduleExportInfos = generateModuleInfos(exportOptions, portalId, studyEnvironmentId);
-        var enrolleeMaps = generateExportMaps(portalId, studyEnvironmentId,
+    /**
+     * exports the specified number of enrollees from the given environment
+     * The enrollees will be returned most-recent first
```
denniscunningham: Most recently enrolled? Or most recently updated? Or something else?

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -90,16 +94,20 @@ public Map<String, String> generateExportMap(EnrolleeExportData exportData,
...
         List<ModuleExportInfo> moduleInfo = new ArrayList<>();
         moduleInfo.add(new EnrolleeFormatter().getModuleExportInfo(exportOptions));
         moduleInfo.add(new ProfileFormatter().getModuleExportInfo(exportOptions));
-        moduleInfo.addAll(generateSurveyModules(exportOptions, portalId, studyEnvironmentId));
+        moduleInfo.addAll(generateSurveyModules(exportOptions, studyEnvironmentId));
         return moduleInfo;
     }
 
-    protected List<ModuleExportInfo> generateSurveyModules(ExportOptions exportOptions, UUID portalId, UUID studyEnvironmentId) throws Exception {
+    protected List<ModuleExportInfo> generateSurveyModules(ExportOptions exportOptions, UUID studyEnvironmentId) throws Exception {
```
denniscunningham, line 109: Does this throw a particular exception (`Exception` is pretty broad)? Any reason that cannot be caught and converted in this method?

denniscunningham (approved):
Looks good!
Left a couple questions.

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
devonbush: good question, updated comment to clarify it's most-recently created

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
devonbush: IIRC it's catching some JSONProcessing exceptions that can happen when trying to parse/stringify some things.  I think you're right that they should just be converted to runtime exceptions in the source methods, since we're not doing any handling of them.  Eventually, we'll want the export to generate a nice list of "here are the participants/modules that couldn't be exported and the exceptions for each".  If it's ok with you, I'll delay cleaning up this exception handling until then

denniscunningham: Sure.
I am probably hyper-sensitive to exception handling since it was quite a mess in Pepper. Skipping a lot of detail, the main problem was mapping exceptions to response codes, which I assert we want to get right. In Pepper we added a couple exception classes that developers could be confident in how they mapped to status codes (effectively 400 and 500), since it was rife with RuntimeExceptions and uncaught (and thus unconverted) exceptions. RuntimeException is almost never the right exception to throw.

PR #581, opened 2023-10-07
[JN-597] admin version display

devonbush:
#### DESCRIPTION

Adds version taken information to the admin survey response display

image

This required some updates to the populate infrastructure to make sure we could test the eventuality where someone's response spans multiple versions.  

1. repopulate demo portal `./scripts/populate_portal.sh demo`
2. go to `https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants/HDVERS/surveys/hd_hd_cardioHx`, confirm "version 1" is displayed as above
3. go to `https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants/HDVERS/surveys/hd_hd_socialHealth`, confirm "versions 1, 2" is displayed for the versions


core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -55,6 +55,8 @@ public Optional<SurveyResponse> findOneWithAnswers(UUID responseId) {
         return dao.findOneWithAnswers(responseId);
     }
 
+
```
denniscunningham: Is this intentional?

core/src/main/java/bio/terra/pearl/core/dao/survey/AnswerDao.java
```diff
@@ -49,6 +49,10 @@ public List<Answer> findAll(UUID surveyResponseId, List<String> questionStableId
 
     }
 
+    public List<Answer> findAll(UUID enrolleeId, String surveyStableId) {
```
denniscunningham: Wondering if there is not a different future method that might be more appropriately named `findAll`. IOW, might there by a `findAll` that would return all answers for a given survey, or for all surveys? 
I am not familiar with out naming standards for DAOs, but I see `findByEnrollee` just below.

core/src/main/java/bio/terra/pearl/core/service/survey/AnswerService.java
```diff
@@ -26,6 +26,10 @@ public List<Answer> findAll(UUID surveyResponseId, List<String> questionStableId
         return dao.findAll(surveyResponseId, questionStableIds);
     }
 
+    public List<Answer> findAll(UUID enrolleeID, String surveyStableId) {
```
denniscunningham: See my question in `AnswerDao`.

ui-admin/src/study/participants/survey/EnrolleeSurveyView.tsx
```diff
@@ -41,13 +42,20 @@ export function RawEnrolleeSurveyView({ enrollee, configSurvey, responses }:
     return <div>Most recent response has no data yet </div>
   }
 
+  const answerVersions = _uniq(lastResponse.answers.map(ans => ans.surveyVersion))
+  const versionString = `version${answerVersions.length > 1 ? 's' : ''} ${answerVersions.join(', ')}`
```
MatthewBemis: terra-ui uses [pluralize](https://www.npmjs.com/package/pluralize) for strings like this. It's obviously trivial to implement it each time, but I'm wondering if we should bring that package in since this is a very common thing to need to do and it's a nice utility to have

MatthewBemis (approved):

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -231,7 +238,11 @@ private Answer createAnswer(Answer answer, SurveyResponse response,
         answer.setCreatingParticipantUserId(ppUser.getParticipantUserId());
         answer.setSurveyResponseId(response.getId());
         answer.setSurveyStableId(survey.getStableId());
-        answer.setSurveyVersion(survey.getVersion());
+        if (answer.getSurveyVersion() == 0) {
```
denniscunningham, line 239: I realize this bit of code is trivial, but does it make sense to encapsulate it in a method, just to make it clear that the same logic is needed in each case?

denniscunningham (commented):
Looks good.
A couple minor questions.

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
devonbush: nope, thanks!

core/src/main/java/bio/terra/pearl/core/dao/survey/AnswerDao.java
devonbush: good question.  As you can tell, I've waffled between calling them all `findAll` and letting the arguments distinguish them, and calling them `findBy...`  While I think there are some cases where findBy... can get extreme "findByXAndYAndZ" I think you're right that calling out the args in the method name is probably cleaner

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
devonbush: I'm not sure if it will be the same logic in the future, which is why I didn't DRY this.  Essentially, how we treat updates and how we treat fresh answers might diverge when it comes to how we assign the version.

denniscunningham (approved):

PR #582, opened 2023-10-07
[JN-598] deactivating studyEnvSurvey configs

devonbush:
#### DESCRIPTION

In order to robustly handle multi-version export, we first need to enhance how we track which survey versions have been associated with a given studyEnvironment.  (And this is a good idea anyway).  So now instead of just deleting StudyEnvironmentSurveys, we deactivate them.  This PR has no functional changes to the application -- from a user perspective, everything appears the same as before.  

1. repopulate ourhealth (do this step *before* restarting the app, the goal is to simulate pre-existing data)
2. redeploy ApiAdminApp
3. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms, confirm surveys appear as usual
4. go to https://localhost:3000/ourhealth/studies/ourheart/publishing, copy the sandbox to irb
5. go back to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms, make a change to a survey, then save it
6. go back to https://localhost:3000/ourhealth/studies/ourheart/publishing, copy the sandbox to irb again.  Note the survey you changed should appear as selectable, select it and publish.
7. go to https://localhost:3000/ourhealth/studies/ourheart/env/irb/forms confirm that the updated survey version appears correctly

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
@@ -295,6 +295,23 @@ public abstract class BaseJdbiDao<T extends BaseEntity> {
         );
     }
 
+    protected List<T> findAllByTwoPropertiesSorted(String column1Name, Object column1Value,
+                                                   String column2Name, Object column2Value,
+                                                   String sortProperty, String sortDir) {
+        return jdbi.withHandle(handle ->
+                handle.createQuery("""
+                                select * from %s 
+                                where %s = :column1Value
+                                and %s = :column2Value
+                                order by %s %s
+                                """.formatted(tableName, column1Name, column2Name, sortProperty, sortDir))
+                        .bind("column1Value", column1Value)
+                        .bind("column2Value", column2Value)
+                        .mapTo(clazz)
+                        .list()
+        );
+    }
+
     /** defaults to matching on id if provided. */
     public Optional<T> findOneMatch(T matchObj) {
         if (matchObj.getId() != null) {
--- a/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
@@ -59,7 +59,7 @@ public class StudyEnvironmentSurveyDao extends BaseMutableJdbiDao<StudyEnvironme
     }
 
     /** finds by a surveyId and studyEnvironment, limited to active surveys */
-    public Optional<StudyEnvironmentSurvey> findBySurvey(UUID studyEnvId, UUID surveyId) {
+    public Optional<StudyEnvironmentSurvey> findActiveBySurvey(UUID studyEnvId, UUID surveyId) {
         return jdbi.withHandle(handle ->
                 handle.createQuery("""
                                 select * from %s
@@ -74,7 +74,7 @@ public class StudyEnvironmentSurveyDao extends BaseMutableJdbiDao<StudyEnvironme
         );
     }
 
-    public List<StudyEnvironmentSurvey> findBySurvey(UUID studyEnvId, String surveyStableId) {
+    public List<StudyEnvironmentSurvey> findActiveBySurvey(UUID studyEnvId, String surveyStableId) {
         return jdbi.withHandle(handle ->
                 handle.createQuery("""
                                 select %s from %s a
--- a/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
@@ -29,8 +29,13 @@ public class StudyEnvironmentSurveyDao extends BaseMutableJdbiDao<StudyEnvironme
     }
 
     public List<StudyEnvironmentSurvey> findAllByStudyEnvironmentId(UUID studyEnvId) {
-        return findAllByPropertySorted("study_environment_id", studyEnvId,
-                "survey_order", "ASC");
+        return findAllByStudyEnvironmentId(studyEnvId, true);
+    }
+
+    public List<StudyEnvironmentSurvey> findAllByStudyEnvironmentId(UUID studyEnvId, boolean active) {
+        return findAllByTwoPropertiesSorted("study_environment_id", studyEnvId,
+                            "active", active,
+                             "survey_order", "ASC");
     }
 
     /** gets all the study environment surveys and attaches the relevant survey objects in a batch */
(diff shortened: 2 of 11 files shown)


core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
```diff
@@ -29,8 +29,13 @@ public void deleteByStudyEnvironmentId(UUID studyEnvId) {
     }
 
     public List<StudyEnvironmentSurvey> findAllByStudyEnvironmentId(UUID studyEnvId) {
-        return findAllByPropertySorted("study_environment_id", studyEnvId,
-                "survey_order", "ASC");
+        return findAllByStudyEnvironmentId(studyEnvId, true);
+    }
+
+    public List<StudyEnvironmentSurvey> findAllByStudyEnvironmentId(UUID studyEnvId, boolean active) {
```
denniscunningham, line 35: Did we lose the ability to actually get "all" (active and not active), and if so, does that matter?

core/src/main/java/bio/terra/pearl/core/service/study/StudyEnvironmentSurveyService.java
```diff
@@ -22,6 +24,13 @@ public Optional<StudyEnvironmentSurvey> findBySurvey(UUID studyEnvId, UUID surve
         return dao.findBySurvey(studyEnvId, surveyId);
     }
 
+    @Transactional
+    public StudyEnvironmentSurvey deactivate(UUID id) {
```
denniscunningham, line 27: Does it make sense to override `delete` just to ensure no one accidentally calls it (or at least has to be deliberate if that's what is needed)?

core/src/main/resources/db/changelog/changesets/2023_10_06_survey_config_active.yaml
```diff
@@ -0,0 +1,9 @@
+databaseChangeLog:
+  - changeSet:
+      id: "survey_config_active"
+      author: dbush
+      changes:
+        - addColumn:
+            tableName: study_environment_survey
+            columns:
+              - column: { name: active, type: boolean, defaultValueBoolean: false, constraints: { nullable: false } }
```
denniscunningham: So false by default here, but true by default in StudyEnvironmentSurvey? Does that mean all existing `study_environment_survey` records are not active, but newly created ones will be active?

core/src/test/java/bio/terra/pearl/core/service/publishing/StudyPublishingServiceTests.java
```diff
@@ -76,16 +109,20 @@ public void testApplyChangesConsents() throws Exception {
...
-        studyPublishingService.applyChanges(loadedLiveEnv, changes, null);
+        diffAndApplyChanges(study.getShortcode(), EnvironmentName.irb, EnvironmentName.live);
 
         assertThat(studyEnvironmentConsentService.findByConsentForm(liveEnv.getId(), form.getId()).isPresent(), equalTo(true));
 
         form = consentFormService.find(form.getId()).get();
         assertThat(form.getPublishedVersion(), equalTo(1));
     }
 
+    private void diffAndApplyChanges(String studyShortcode, EnvironmentName src, EnvironmentName dest) throws Exception {
```
denniscunningham, line 120: Small thing, but Nice.

core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
```diff
@@ -53,18 +58,31 @@ public void deleteBySurveyId(UUID surveyId) {
         deleteByProperty("survey_id", surveyId);
     }
 
-    /** finds by a surveyId and studyEnvironment */
+    /** finds by a surveyId and studyEnvironment, limited to active surveys */
     public Optional<StudyEnvironmentSurvey> findBySurvey(UUID studyEnvId, UUID surveyId) {
```
denniscunningham: Might be helpful to change the method name to indicate that only active surveys are returned?

core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
devonbush: right on.  This is addressed in https://github.com/broadinstitute/juniper/pull/583, where the argument is upgraded to Boolean

core/src/main/resources/db/changelog/changesets/2023_10_06_survey_config_active.yaml
devonbush: Yup, this is fixed in https://github.com/broadinstitute/juniper/pull/583

core/src/main/java/bio/terra/pearl/core/service/study/StudyEnvironmentSurveyService.java
devonbush: good question.  We do need 'delete' for populating, and I wouldn't be surprised if we also will need it in the near future for some sort of "cleanup" operation that study staff will want to do:  "I created 10 versions of this survey while I was futzing around, please destroy all but the most recent, none of which ever got used, because I don't want them hanging around in the UI".  That said, I think your concern is a good one, in that the service doesn't delineate that the delete should be an unusual operation.

core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
```diff
@@ -295,6 +295,23 @@ protected List<T> findAllByPropertySorted(String columnName, Object columnValue,
         );
     }
 
+    protected List<T> findAllByTwoPropertiesSorted(String column1Name, Object column1Value,
```
MatthewBemis, line 298: With the recent addition of all of these methods like `findXByNProperties[Sorted|Unsorted](...)` etc, should we instead add a method like: `findAllByProperties(Map<String, Object> columns, Optional<SortOptions> sortOptions)`?

We could get rid of `findAllByPropertySorted`, `findAllByTwoPropertiesSorted`, `findAllByProperty`, and `findAllByTwoProperties`, plus avoid the inevitable 3 or 4 property variants, for the small cost of a slightly more complex DAO method that generates the correct SQL.

denniscunningham (approved):

core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
devonbush: yes, you're absolutely right that DAO methods are starting to get out of hand.  I'd propose we make a ticket for "DAO composability" and address it wholesale. Does https://broadworkbench.atlassian.net/browse/JN-632 address your concern?

MatthewBemis: Follow-up ticket works for me 👍

MatthewBemis (approved):

PR #583, opened 2023-10-10
[JN-598] export old versions

devonbush:
#### DESCRIPTION
This upgrades the export pipeline to support export of survey questions from prior versions of surveys.  It builds on 
https://github.com/broadinstitute/juniper/pull/581, https://github.com/broadinstitute/juniper/pull/580, and https://github.com/broadinstitute/juniper/pull/582, so review those first.  

The export pipeline is a four-stage process
 (1) generate "module info"--collecting all the study config information into a data structure that tells us the fields we need to export
 (2) fetch all the Enrollees we need to export, along with all their data
 (3) parse all the enrollee data into String->String map of columnName->rendered value
 (4) generate an appropriate file form the map.

This updates step (1) to pull inactive surveys into the module info so it can be used for export, and updates (3) to use past versions of questions to do, e.g. choice lookup.  This does not yet handle updating the data dictionary to list out the past versions.

1. Redeploy ApiAdminApp
2. repopulate demo  `./scripts/populate_portal.sh demo`
3. go to 
4. Scroll down to the socialHealth questions
5. Confirm you see "Disagre" (with typo) for the `hd_hd_socialHealth.neighborhoodSharesValues` question, and that `hd_hd_socialHealth.neighborhoodIsWalkable` appears as blank for all participants except one
image

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
@@ -29,8 +29,13 @@ public class StudyEnvironmentSurveyDao extends BaseMutableJdbiDao<StudyEnvironme
     }
 
     public List<StudyEnvironmentSurvey> findAllByStudyEnvironmentId(UUID studyEnvId) {
-        return findAllByPropertySorted("study_environment_id", studyEnvId,
-                "survey_order", "ASC");
+        return findAllByStudyEnvironmentId(studyEnvId, true);
+    }
+
+    public List<StudyEnvironmentSurvey> findAllByStudyEnvironmentId(UUID studyEnvId, boolean active) {
+        return findAllByTwoPropertiesSorted("study_environment_id", studyEnvId,
+                            "active", active,
+                             "survey_order", "ASC");
     }
 
     /** gets all the study environment surveys and attaches the relevant survey objects in a batch */
@@ -53,18 +58,31 @@ public class StudyEnvironmentSurveyDao extends BaseMutableJdbiDao<StudyEnvironme
         deleteByProperty("survey_id", surveyId);
     }
 
-    /** finds by a surveyId and studyEnvironment */
+    /** finds by a surveyId and studyEnvironment, limited to active surveys */
     public Optional<StudyEnvironmentSurvey> findBySurvey(UUID studyEnvId, UUID surveyId) {
-        return findByTwoProperties("study_environment_id",studyEnvId,
-                "survey_id", surveyId);
+        return jdbi.withHandle(handle ->
+                handle.createQuery("""
+                                select * from %s
+                                    where study_environment_id = :studyEnvId
+                                    and survey_id = :surveyId
+                                    and a.active = true;
+                                """.formatted(prefixedGetQueryColumns("a"), tableName))
+                        .bind("surveyId", surveyId)
+                        .bind("studyEnvId", studyEnvId)
+                        .mapTo(clazz)
+                        .findOne()
+        );
     }
 
     public List<StudyEnvironmentSurvey> findBySurvey(UUID studyEnvId, String surveyStableId) {
         return jdbi.withHandle(handle ->
-                handle.createQuery("select " + prefixedGetQueryColumns("a") + " from " + tableName +
-                                " a join survey on survey.id = a.survey_id" +
-                                " where survey.stable_id = :stableId " +
-                                " and a.study_environment_id = :studyEnvId;")
+                handle.createQuery("""
+                                select %s from %s a
+                                    join survey on survey.id = a.survey_id
+                                    where survey.stable_id = :stableId
+                                    and a.study_environment_id = :studyEnvId
+                                    and a.active = true;
+                                """.formatted(prefixedGetQueryColumns("a"), tableName))
                         .bind("stableId", surveyStableId)
                         .bind("studyEnvId", studyEnvId)
                         .mapTo(clazz)
--- a/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
@@ -46,6 +46,12 @@ public class StudyEnvironmentSurveyDao extends BaseMutableJdbiDao<StudyEnvironme
         return findAllByStudyEnvIdWithSurvey(studyEnvId, true);
     }
 
+    public List<StudyEnvironmentSurvey> findAllByStudyEnvironmentId(UUID studyEnvId, boolean active) {
+        return findAllByTwoPropertiesSorted("study_environment_id", studyEnvId,
+                            "active", active,
+                             "survey_order", "ASC");
+    }
+
     /** gets all the study environment surveys and attaches the relevant survey objects in a batch */
     public List<StudyEnvironmentSurvey> findAllByStudyEnvIdWithSurvey(UUID studyEnvId, Boolean active) {
         List<StudyEnvironmentSurvey> studyEnvSurvs = findAllByStudyEnvironmentId(studyEnvId, active);
@@ -67,7 +73,7 @@ public class StudyEnvironmentSurveyDao extends BaseMutableJdbiDao<StudyEnvironme
     }
 
     /** finds by a surveyId and studyEnvironment, limited to active surveys */
-    public Optional<StudyEnvironmentSurvey> findBySurvey(UUID studyEnvId, UUID surveyId) {
+    public Optional<StudyEnvironmentSurvey> findActiveBySurvey(UUID studyEnvId, UUID surveyId) {
         return jdbi.withHandle(handle ->
                 handle.createQuery("""
                                 select * from %s
@@ -82,7 +88,7 @@ public class StudyEnvironmentSurveyDao extends BaseMutableJdbiDao<StudyEnvironme
         );
     }
 
-    public List<StudyEnvironmentSurvey> findBySurvey(UUID studyEnvId, String surveyStableId) {
+    public List<StudyEnvironmentSurvey> findActiveBySurvey(UUID studyEnvId, String surveyStableId) {
         return jdbi.withHandle(handle ->
                 handle.createQuery("""
                                 select %s from %s a
--- a/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
@@ -32,15 +32,23 @@ public class StudyEnvironmentSurveyDao extends BaseMutableJdbiDao<StudyEnvironme
         return findAllByStudyEnvironmentId(studyEnvId, true);
     }
 
-    public List<StudyEnvironmentSurvey> findAllByStudyEnvironmentId(UUID studyEnvId, boolean active) {
-        return findAllByTwoPropertiesSorted("study_environment_id", studyEnvId,
-                            "active", active,
-                             "survey_order", "ASC");
+    public List<StudyEnvironmentSurvey> findAllByStudyEnvironmentId(UUID studyEnvId, Boolean active) {
+        if (active != null) {
+            return findAllByTwoPropertiesSorted("study_environment_id", studyEnvId,
+                    "active", active,
+                    "survey_order", "ASC");
+        } else {
+            return findAllByPropertySorted("study_environment_id", studyEnvId, "survey_order", "ASC");
+        }
+    }
+
+    public List<StudyEnvironmentSurvey> findAllByStudyEnvIdWithSurvey(UUID studyEnvId) {
+        return findAllByStudyEnvIdWithSurvey(studyEnvId, true);
     }
 
     /** gets all the study environment surveys and attaches the relevant survey objects in a batch */
-    public List<StudyEnvironmentSurvey> findAllByStudyEnvIdWithSurvey(UUID studyEnvId) {
-        List<StudyEnvironmentSurvey> studyEnvSurvs = findAllByStudyEnvironmentId(studyEnvId);
+    public List<StudyEnvironmentSurvey> findAllByStudyEnvIdWithSurvey(UUID studyEnvId, Boolean active) {
+        List<StudyEnvironmentSurvey> studyEnvSurvs = findAllByStudyEnvironmentId(studyEnvId, active);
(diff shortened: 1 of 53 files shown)


devonbush:
@denniscunningham  @MatthewBemis  merge conflicts now resolved so this is ready for review

devonbush:
There was a test error the first time I ran this about a database deadlock detected when trying to truncate some tables as part of a test setup unrelated to this PR.  Have either of you seen that before?  We only have a couple of test setups that do truncation, so if this error becomes frequent, we can just get rid of that and do the table clears using our regular DAO methods

populate/src/main/java/bio/terra/pearl/populate/service/EnrolleePopulator.java
```diff
@@ -141,8 +141,8 @@ private void populateResponse(Enrollee enrollee, SurveyResponsePopDto responsePo
                     .findTaskForActivity(ppUser.getId(), enrollee.getStudyEnvironmentId(), survey.getStableId()).get();
             if (responsePopDto.getSurveyVersion() != task.getTargetAssignedVersion()) {
                 /**
-                 * in simulateSubmission mode, tasks will be automatically created with the curren versions of the survey
-                 * to enable mocking of submissions of prior versions, if the version specified in the enrollee's seed file
+                 * in simulateSubmission mode, tasks will be automatically created with the curren versions of the survey.
+                 * To enable mocking of submissions of prior versions, if the version specified in the enrollee's seed file
```
denniscunningham, line 145: Nit, but the prior version seems to make more sense.

core/src/test/java/bio/terra/pearl/core/service/export/EnrolleeExportServiceTests.java
```diff
@@ -3,20 +3,27 @@
...
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
 
 import java.util.List;
 import java.util.Map;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.*;
```
denniscunningham, line 25: Do we have a policy on when `*` is acceptable for imports?

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -55,6 +55,8 @@ public Optional<SurveyResponse> findOneWithAnswers(UUID responseId) {
         return dao.findOneWithAnswers(responseId);
     }
 
+
```
denniscunningham: Is this intentional?

core/src/main/java/bio/terra/pearl/core/service/study/StudyEnvironmentSurveyService.java
```diff
@@ -1,6 +1,7 @@
 package bio.terra.pearl.core.service.study;
 
 import bio.terra.pearl.core.dao.study.StudyEnvironmentSurveyDao;
+import bio.terra.pearl.core.model.notification.NotificationConfig;
```
denniscunningham: It is not obvious why this is now necessary.

core/src/main/java/bio/terra/pearl/core/service/export/formatters/SurveyFormatter.java
```diff
@@ -87,20 +86,28 @@ public String getColumnSubHeader(ModuleExportInfo moduleExportInfo, ItemExportIn
...
         itemExportInfos.add(ExportFormatUtils.getItemInfoForBeanProp(moduleName, "lastUpdatedAt", SurveyResponse.class));
         itemExportInfos.add(ExportFormatUtils.getItemInfoForBeanProp(moduleName, "complete", SurveyResponse.class));
-        List<SurveyQuestionDefinition> sortedDefs = new ArrayList<>(questionDefs);
-        sortedDefs.sort(Comparator.comparing(SurveyQuestionDefinition::getExportOrder));
-        for (SurveyQuestionDefinition questionDef : sortedDefs) {
-            itemExportInfos.add(getItemExportInfo(exportOptions, moduleName, questionDef));
+
+        // group all the questions that share a stableId (i.e. different versions of the same question), and then sort them by
+        // the export order of the most recent version
+        Collection<List<SurveyQuestionDefinition>> questionDefsByStableId = questionDefs.stream().collect(groupingBy(
```
denniscunningham, line 98: So this is an unordered collection of all survey questions, which are represented as an ordered list of N question versions, where N varies per question?

core/src/main/java/bio/terra/pearl/core/service/export/formatters/SurveyFormatter.java
```diff
@@ -114,6 +121,17 @@ protected String stripStudyPrefixes(String stableId) {
         return stableId.substring(stableId.lastIndexOf('_') + 1);
     }
 
+    public ItemExportInfo getItemExportInfo(ExportOptions exportOptions, String moduleName, List<SurveyQuestionDefinition> questionDefs)
+            throws JsonProcessingException {
```
denniscunningham, line 130: What is the HTTP response code associated with `JsonProcessingException`?

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -107,15 +111,20 @@ public List<ModuleExportInfo> generateModuleInfos(ExportOptions exportOptions, U
...
+        // get all surveys that have ever been attached, including inactive
+        List<StudyEnvironmentSurvey> configuredSurveys = studyEnvironmentSurveyService.findAllByStudyEnvIdWithSurvey(studyEnvironmentId, null);
+        configuredSurveys.sort(Comparator.comparing(StudyEnvironmentSurvey::getSurveyOrder));
         SurveyFormatter surveyFormatter = new SurveyFormatter(objectMapper);
         List<ModuleExportInfo> moduleExportInfos = new ArrayList<>();
-        for (StudyEnvironmentSurvey configuredSurvey : latestConfiguredSurveys) {
-            var surveyQuestionDefinitions = surveyQuestionDefinitionDao
-                    .findAllBySurveyId(configuredSurvey.getSurvey().getId());
-            moduleExportInfos.add(surveyFormatter.getModuleExportInfo(exportOptions, configuredSurvey.getSurvey(), surveyQuestionDefinitions));
+        Map<String, List<StudyEnvironmentSurvey>> surveysByStableId = configuredSurveys.stream().collect(
```
denniscunningham: Sorry, trying to understand what this is.. 
The stable ID identifies a survey, which may have multiple versions? And those versions are represented by `StudyEnvironmentSurvey`?

core/src/main/java/bio/terra/pearl/core/service/export/formatters/SurveyFormatter.java
```diff
@@ -114,6 +121,17 @@ protected String stripStudyPrefixes(String stableId) {
         return stableId.substring(stableId.lastIndexOf('_') + 1);
     }
 
+    public ItemExportInfo getItemExportInfo(ExportOptions exportOptions, String moduleName, List<SurveyQuestionDefinition> questionDefs)
```
denniscunningham: Might be helpful to have a JavaDoc comment, or some in-line comments explaining what's going on here. 
I am new to this code, but I am unclear on whether the `List<SurveyQuestionDefinition>` represents a list of versions for a specific question, or a list of questions in a survey.
Likewise, what a module represents is not clear from the context in this file, so someone new to this code has to switch context to refresh on that.
In short there are a handful of entities in play here, and it would be helpful to have better names for those entities and/or some hints in comments.

denniscunningham (commented):
For the more intricate parts of export survey construction, I could use some help to make the review more effective. See my comments in `SurveyFormatter`.

core/src/test/java/bio/terra/pearl/core/service/export/EnrolleeExportServiceTests.java
devonbush: I just follow Intellij's automatic rules, which I think collapse to * when there are 3 or more.

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
devonbush: nope, thanks!

core/src/main/java/bio/terra/pearl/core/service/study/StudyEnvironmentSurveyService.java
devonbush: gah, intellij leftover.  I need to setup better automatic unused import clearing

core/src/main/java/bio/terra/pearl/core/service/export/formatters/SurveyFormatter.java
devonbush: yes, that's right

core/src/main/java/bio/terra/pearl/core/service/export/formatters/SurveyFormatter.java
devonbush: That Exception would mean there's something in a previously saved survey definition that contains an error.  The root line that throws this is line 139: `objectMapper.readValue(questionDef.getChoices(), new TypeReference<List<QuestionChoice>>(){})`  So if somehow some bad choices previously snuck past our on-save validation, that would error here.  I think 500 is the correct error because it's an error not related to the client request for export.  That's also why I thought just catching that error on line 139 and rethrowing it as a runtime exception might be appropriate.  Essentially "this study has corrupt survey definitions, we don't know what to do"

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
devonbush: Close!  A stableId identifies a survey, which might have multiple versions.  Some of those versions may have been attached to a StudyEnvironment at one time, via a StudyEnvironmentSurvey.

devonbush:
@denniscunningham  I took a first stab at adding some comments and cleaning up some variable names.  Let's chat sometime tomorrow to get on the same page

core/src/main/java/bio/terra/pearl/core/service/export/formatters/SurveyFormatter.java
denniscunningham: I'm probably hypersensitive to exception handling since we had to spend a lot of effort to get that cleaned up in Pepper, and we are still not done. It really helps to get it right from the get go.

Catching and rethrowing here is provides an opportunity to say exactly what you said above ("this study has corrupt survey definitions..."), unless that is easily deduced as is. 

I'm not a fan of throwing RuntimeException, since developers may throw that in a variety of situations. How about creating two exception classes like `JuniperInternalError` and `BadRequestException`? Those allow developers to unambiguously choose the upstream handling (in most cases that means choosing the HTTP response code). Presumably there is a way for us to map these exceptions to HTTP response handling.

This also provides an opportunity to log and strip information from internal exceptions, since we may not want to share that with end users.

core/src/test/java/bio/terra/pearl/core/service/export/EnrolleeExportServiceTests.java
denniscunningham: Ah, okay. 
I've worked a few places where is was strictly forbidden.

core/src/main/java/bio/terra/pearl/core/service/export/instance/ItemExportInfo.java
```diff
@@ -29,4 +33,10 @@ public class ItemExportInfo {
     private String questionText;
     @Builder.Default
     private List<QuestionChoice> choices = null;
+    /**
```
denniscunningham, line 36: The `ItemExportInfo` description says it is used for a survey question and "field of a data model". Just curious, what is the latter? 
The fields in this class seem biased toward question content. What is the advantage of sharing its use with something else? Do they share similar export processing?

denniscunningham (approved):
Looks good.
Thank you for adding comments to the code. Those are a nice readability improvement.
I left a few new questions/comments.

core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
```diff
@@ -32,15 +32,29 @@ public List<StudyEnvironmentSurvey> findAllByStudyEnvironmentId(UUID studyEnvId)
         return findAllByStudyEnvironmentId(studyEnvId, true);
     }
 
+    public List<StudyEnvironmentSurvey> findAllByStudyEnvironmentId(UUID studyEnvId, Boolean active) {
```
MatthewBemis, line 35: Maybe this is my Scala bias showing, but I'm a fan of using `Optional` for these cases where we're knowingly supporting `null` parameters, since it hints in the signature that the empty/null case is expected and handled.

core/src/main/java/bio/terra/pearl/core/service/export/formatters/SurveyFormatter.java
MatthewBemis: +1 for improving exception handling in general. I'm definitely guilty of throwing RuntimeExceptions in places where I should've been more specific. It might be worth creating a tech debt/developer's choice ticket to spend a bit of time auditing our current exception handling and improving it, while the size of the app is still manageable.

MatthewBemis (approved):
Looks good, one non-blocking comment about using `Optional` instead of `null`

core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
devonbush: I like that idea generally.  Here, to me at least, the use of Boolean instead of boolean is what communicates nulls are expected

core/src/main/java/bio/terra/pearl/core/service/export/instance/ItemExportInfo.java
devonbush: "field of a data model" would be something like "dateOfBirth" from a Profile object, or "complete" from SurveyResponse.
  I think you're right that this class could probably be usefully refactored into two classes that share an interface.

devonbush:
For `*` imports, I have no strong opinion on the matter (except that I want my IDE to do it for me whichever we decide).  If you lean one way, let's do it.

devonbush:
Thanks for the great comments Matt and Dennis.  Going to go ahead and merge this because we've got bigger fish to fry this week getting ready for HeartHive, but will definitely implement those next time we revisit export (which will likely be in about 2 months...)

PR #585, opened 2023-10-11
[JN-576] Download participant list

MatthewBemis:
#### DESCRIPTION

This adds the ability to download the participant list as a .csv file. The resulting .csv file will be exactly what is visible in the table when the user clicks the download button; we will respect filters and column selections. I also made a few other related UX tweaks to better match Erin's mockups for the buttons in the upper right corner of the table.

Screenshot 2023-10-11 at 10 31 09 AM

Screenshot 2023-10-11 at 10 49 22 AM

* Launch admin API and UI
* Verify that the download button is disabled if there aren't any rows visible
* Verify that you can download the csv when there are visible rows. Apply some filters and hide some columns, and verify that only the visible columns and rows are in the csv
* Verify that there aren't any formatting issues in the file (making sure that everything is escaped when necessary, specifically for dates/times)

### DIFF
--- a/ui-admin/src/study/participants/participantList/ParticipantList.tsx
+++ b/ui-admin/src/study/participants/participantList/ParticipantList.tsx
@@ -14,9 +14,9 @@ import {
   useReactTable,
   VisibilityState
 } from '@tanstack/react-table'
-import { basicTableLayout, ColumnVisibilityControl, IndeterminateCheckbox } from 'util/tableUtils'
+import { basicTableLayout, ColumnVisibilityControl, DownloadControl, IndeterminateCheckbox } from 'util/tableUtils'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faCheck } from '@fortawesome/free-solid-svg-icons'
+import { faCheck, faEnvelope } from '@fortawesome/free-solid-svg-icons'
 import AdHocEmailModal from '../AdHocEmailModal'
 import {
   ALL_FACETS,
@@ -25,7 +25,6 @@ import {
   facetValuesToString,
   KEYWORD_FACET
 } from 'api/enrolleeSearch'
-import { Button } from 'components/forms/Button'
 import { instantToDefaultString } from 'util/timeUtils'
 import { useLoadingEffect } from 'api/api-utils'
 import { FacetView, getUpdatedFacetValues } from './facets/EnrolleeSearchFacets'
@@ -76,12 +75,18 @@ function ParticipantList({ studyEnvContext }: {studyEnvContext: StudyEnvContextT
     id: 'createdAt',
     accessorKey: 'enrollee.createdAt',
     enableColumnFilter: false,
+    meta: {
+      columnType: 'instant'
+    },
     cell: info => instantToDefaultString(info.getValue() as unknown as number)
   }, {
     id: 'lastLogin',
     header: 'Last login',
     accessorKey: 'participantUser.lastLogin',
     enableColumnFilter: false,
+    meta: {
+      columnType: 'instant'
+    },
     cell: info => instantToDefaultString(info.getValue() as unknown as number)
   }, {
     id: 'familyName',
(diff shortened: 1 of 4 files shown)


ui-admin/src/study/participants/participantList/ParticipantList.tsx
```diff
@@ -187,22 +192,28 @@ function ParticipantList({ studyEnvContext }: {studyEnvContext: StudyEnvContextT
       <div className="col-12">
         <LoadingSpinner isLoading={isLoading}>
```
MatthewBemis, line 194: I suggest reviewing the rest of this file with whitespace changes hidden to get a better idea of what actually changed: https://github.com/broadinstitute/juniper/pull/585/files?w=1

ui-admin/src/study/participants/participantList/ParticipantList.tsx
```diff
@@ -187,22 +192,28 @@ function ParticipantList({ studyEnvContext }: {studyEnvContext: StudyEnvContextT
...
+                  {numSelected} of{' '}
+                  {/* eslint-disable-next-line max-len */}
+                  {table.getPreFilteredRowModel().rows.length} selected ({table.getFilteredRowModel().rows.length} shown)
+                </span>
+                { showEmailModal && <AdHocEmailModal enrolleeShortcodes={enrolleesSelected}
+                  studyEnvContext={studyEnvContext}
+                  onDismiss={() => setShowEmailModal(false)}/> }
+              </div>
+              <div className="d-flex">
+                <button
```
devonbush: Is there a reason you switched from our "Button" default button component to this?  This won't support tooltips when it's disabled, which are nice to give help to users wondering why they can't click somewhere?

ui-admin/src/util/tableUtils.test.tsx
```diff
@@ -9,11 +9,15 @@ import {
   getSortedRowModel,
   useReactTable
 } from '@tanstack/react-table'
-import { basicTableLayout, checkboxColumnCell, ColumnVisibilityControl } from './tableUtils'
+import { basicTableLayout, checkboxColumnCell, ColumnVisibilityControl, DownloadControl } from './tableUtils'
 import userEvent from '@testing-library/user-event'
 
-/** simple table with filters and a show/hide column control */
-const TestTableComponent = ({ initialValue }: {initialValue: ColumnFiltersState}) => {
+const DEFAULT_INITIAL_DATA = [{ consented: true, name: 'Fred' }, { consented: false, name: 'James' }]
```
devonbush: nit: default to me implies it would be empty [].

```suggestion
const SAMPLE_INITIAL_DATA = [{ consented: true, name: 'Fred' }, { consented: false, name: 'James' }]
```

ui-admin/src/util/tableUtils.tsx
```diff
@@ -266,6 +269,73 @@ export function ColumnVisibilityControl<T>({ table }: {table: Table<T>}) {
   </div>
 }
 
+/**
+ * Converts a cell value to an escaped string for csv export
+ */
```
devonbush, line 276: nice, I like how this is more robust than the basic approach from MailingListView.tsx

ui-admin/src/util/tableUtils.tsx
```diff
@@ -266,6 +269,73 @@ export function ColumnVisibilityControl<T>({ table }: {table: Table<T>}) {
...
+      <Modal.Header closeButton>
+        <Modal.Title>
+                Download
+        </Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        <div className="border-b border-black">
+                This will download <strong>{table.getFilteredRowModel().rows.length}</strong> rows
+                to <code>{fileName}.csv</code>. Your current row and column filters will be applied to the
+                downloaded data.
```
devonbush: possible wordsmithing for more conciseness? 
```suggestion
                Download <strong>{table.getFilteredRowModel().rows.length}</strong> rows
                to <code>{fileName}.csv</code>. The current data filters and shown columns will be applied.
```

devonbush (changes requested):
Works well, and love the updated styling.  Some nits and questions below, and two other things:
1. if it's not too hard, we should automatically filter out the "selected" column from the download -- I can't think of any reason why someone would want that, and it would be nice for them not to have to remove it manually
2. Can you go ahead and update MailingListView.tsx to use this new component?

ui-admin/src/study/participants/participantList/ParticipantList.tsx
MatthewBemis: Just an oversight. I'll switch that.

ui-admin/src/util/tableUtils.test.tsx
MatthewBemis: Done in 10e6f428

ui-admin/src/study/participants/participantList/ParticipantList.tsx
MatthewBemis: This is done in a693bd01. I also switched `ColumnVisibilityControl` and `DownloadControl` to use `Button`

denniscunningham (approved):

MatthewBemis:
@devonbush I've updated the DownloadControl to automatically filter out the `select` column. I took a look at MailingList, and the download flow is a little bit different because it only downloads the selected rows. It wouldn't be a lot of effort to support both cases, but I'd rather check with Erin about how to merge the two different modes in the UI- maybe clicking the `Download` button shows a dropdown or the modal has a radio button that let's the user select between "Download All" and "Download Selected".

Unless you're suggesting that I port the MailingList download to the "Download All" use-case, is it alright if I defer on refactoring that for the moment?

devonbush:
yup, that makes sense Matt, I forgot the functionality is different.  Thanks for checking it out.

devonbush (approved):

PR #589, opened 2023-10-12
[JN-614] create study modal

devonbush:
#### DESCRIPTION

Allows users to create an empty study inside an existing portal.

image

As a drive-by change, this switches the collection of StudyEnvironments in Study from Set to List.  The use of Set is a holdover from when this project was Hibernate based, as Sets are easier to work with in Hibernate.  But now our DAOs return lists, so we should go with that so we don't have to worry about equality operators when we least expect it.

1. redeploy ApiAdminApp
2. repopulate OurHealth and hearthive
3. go to https://localhost:3000/
4. click "add study"
5. fill in name and shortcode
6. hit "Create"
7. confirm study appears in list


core/src/main/java/bio/terra/pearl/core/model/study/Study.java
```diff
@@ -8,7 +8,9 @@
 import lombok.Setter;
 import lombok.experimental.SuperBuilder;
 
+import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
```
denniscunningham, line 12: I assume HashSet and Set are no longer needed?
I also assume whatever checkstyle rules we have don't check for unused imports?

ui-admin/src/navbar/NavContextProvider.tsx
```diff
@@ -29,23 +32,22 @@ export const useNavContext = () => useContext(NavContext)
 export default function NavContextProvider({ children }: { children: React.ReactNode}) {
   const [breadCrumbs, setBreadCrumbs] = useState<React.ReactNode[]>([])
   const [portalList, setPortalList] = useState<Portal[]>([])
-  const [isLoading, setIsLoading] = useState(true)
   const navigate = useNavigate()
 
+  const { reload, isLoading } = useLoadingEffect(async () => {
+    const result = await Api.getPortals()
+    setPortalList(result)
+    /** if there's only one portal & study, and the user is on the homepage, direct to that study's participant page */
```
denniscunningham, line 40: Dumb question, but how would the user add a new study when these conditions are met? IOW, this only works when there are multiple portals, so how do I add a study when I only have one portal?
I assume I'm just missing some context here.

ui-admin/src/study/CreateNewStudyModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+        </label>
+        <input type="text" size={20} id="studyName" className="form-control" value={studyName}
+          onChange={e => setStudyName(e.target.value)}/>
+        <label className="form-label mt-3" htmlFor="studyShortcode">
+                    Study shortcode
+        </label>
+        <InfoPopup content={`Unique identifier that will be used in urls and data exports to 
+                    indicate this study.  Must be all lowercase with no spaces`}/>
+        <input type="text" size={20} id="studyShortcode" className="form-control" value={studyShortcode}
+          onChange={e => setStudyShortcode(e.target.value.trim().toLowerCase())}/>
```
denniscunningham, line 56: It looks like we don't actually do shortcode validation here? 
Does the backend do that?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/study/StudyController.java
```diff
@@ -31,4 +36,12 @@ public ResponseEntity<Object> getKitTypes(String portalShortcode, String studySh
     var kitTypes = studyExtService.getKitTypes(adminUser, portalShortcode, studyShortcode);
     return ResponseEntity.ok(kitTypes);
   }
+
+  @Override
+  public ResponseEntity<Object> create(String portalShortcode, Object body) {
+    AdminUser operator = requestService.requireAdminUser(request);
+    var studyDto = objectMapper.convertValue(body, StudyExtService.StudyCreationDto.class);
```
denniscunningham, line 55: It feels like this is a bad request if the conversion fails, but I'm guessing the throw from here will result in a 500 status code?
In any case, does it make sense to add a try-catch here to at least acknowledge that this will fail with bad input?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/StudyExtService.java
```diff
@@ -1,24 +1,88 @@
...
+
+  @Transactional
+  public Study create(String portalShortcode, StudyCreationDto study, AdminUser operator) {
+    if (!operator.isSuperuser()) {
+      throw new PermissionDeniedException("You do not have permission to create studies");
+    }
+    Portal portal = authUtilService.authUserToPortal(operator, portalShortcode);
+    List<StudyEnvironment> studyEnvironments =
+        Arrays.stream(EnvironmentName.values())
+            .map(envName -> makeEmptyEnvironment(envName, envName == EnvironmentName.sandbox))
```
denniscunningham: Why are we doing this? Maybe a brief comment?

api-admin/src/test/java/bio/terra/pearl/api/admin/BaseSpringBootTest.java
```diff
@@ -1,8 +1,14 @@
...
+import org.junit.jupiter.api.TestInfo;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.context.ActiveProfiles;
 
 @SpringBootTest
 @ActiveProfiles({"test", "human-readable-logging"})
-public abstract class BaseSpringBootTest {}
+public abstract class BaseSpringBootTest {
+  /** convenience for getting the simple method name of a test from its TestInfo arg */
+  protected String getTestName(TestInfo testInfo) {
```
denniscunningham, line 11: Nice!

api-admin/src/main/java/bio/terra/pearl/api/admin/service/StudyExtService.java
```diff
@@ -1,24 +1,88 @@
...
+            .toList();
+
+    Study newStudy =
+        Study.builder()
+            .shortcode(study.getShortcode())
+            .name(study.getName())
+            .studyEnvironments(studyEnvironments)
+            .build();
+    newStudy = studyService.create(newStudy);
```
denniscunningham: If you attempt to create a new study with the same shortcode as an existing study, a SQL error finds its way to the UI. (I noticed a similar error when create admin user with an in-use email address.) Perhaps we can do some pre-checking or exception handling to clean that up?
Also, it looks like you cannot reuse the same study shortcode across portals, but that may be an unlikely occurrence.

ui-admin/src/navbar/NavContextProvider.tsx
devonbush: yup, they couldn't. But that's okay, because this is limited to superusers for now, who will always have >= two portals in prod

ui-admin/src/study/CreateNewStudyModal.tsx
devonbush: nope, this is at-your-own-risk, which is one of the many reasons this is superuser-only

api-admin/src/main/java/bio/terra/pearl/api/admin/service/StudyExtService.java
devonbush: yes, study shortcodes are intended to be globally unique, since the same study can be added to multiple portals.  Agree on the error checking

devonbush:
Thanks for the reviews Dennis.  I think I'll mothball this PR for a bit -- we've got bigger fish to fry with content creation.

ui-admin/src/study/CreateNewStudyModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+export default function CreateNewStudyModal({ onDismiss }: {onDismiss: () => void}) {
+  const { portalList, reload } = useNavContext()
+  const [studyName, setStudyName] = useState('')
+  const [studyShortcode, setStudyShortcode] = useState('')
+  const { onChange, options, selectedItem: selectedPortal, selectedOption, selectInputId } =
+        useReactSingleSelect(portalList, (portal: Portal) => ({ label: portal.name, value: portal }), portalList[0])
+
+  const [isLoading, setIsLoading] = useState(false)
+  const createStudy = async () => {
+    if (!selectedPortal) { return }
```
MatthewBemis, line 25: Since this is just immediately returning if there isn't a selected portal, is this check better suited for the `disabled` clause of the create `Button`?

ui-admin/src/HomePage.test.tsx
```diff
@@ -25,9 +25,9 @@ describe('HomePage', () => {
       setBreadCrumbs: jest.fn(),
       portalList,
       setPortalList: jest.fn()
-    }
+    };
 
-    ;(useNavContext as jest.Mock).mockReturnValue(mockContextValue)
```
MatthewBemis: nit: as you know I don't have too many opinions about code style, but this is one that I _do_ have an opinion on. 🙂  I learned this convention from Nick, and I personally like keeping semi-colons at the beginning of the line when they're required. Since we don't normally use them, it's a good signal to the reviewer that using the semi-colon here is actually deliberate. And since it's this specific line of code that necessitates the use of the semi-colon in the first place, I think it makes sense to leave it here. Not a hill I'll die on, though.

MatthewBemis (approved):
Looks good. Just a couple of comments.

ui-admin/src/study/CreateNewStudyModal.tsx
devonbush: it is, but typescript can't guarantee that's the only place this function will be called from, so we have to do the check here for typescript

ui-admin/src/HomePage.test.tsx
devonbush: works for me -- that makes a lot of sense.  I had just assumed it was an artifact of a merge conflict rearrangement

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/study/StudyController.java
devonbush: ah, looks like `convertValue` throws IllegalArgumentException, which we'll convert into a 400 via AbstractGlobalExceptionHandler. So I think this is ok

api-admin/src/main/java/bio/terra/pearl/api/admin/service/StudyExtService.java
denniscunningham: Is the error checking something you plan to do for this PR? (Sorry if I missed something.)

devonbush: Yeah, I added an @ExceptionHandler({PSQLException.class}) in the controller, which checks for the specific constraint key to determine whether to throw 400 or 500

denniscunningham (approved):

PR #590, opened 2023-10-13
[JN-633] upgrading mock autosave infrastructure

devonbush:
#### DESCRIPTION

Previously, our tests of survey autosaving relied on specifying a specific timeout interval.  This made the tests flaky as timeout intervals, especially in a CI environment, can be very flaky.  We now mock the entire autosave hook, which lets us manually specify when to simulate the autosave interval elapsing, and is much more robust.  This makes no changes to the behavior of the product.

1. run `npm -w ui-participant run test` to confirm tests pass

### DIFF
--- a/ui-participant/src/hub/survey/SurveyView.test.tsx
+++ b/ui-participant/src/hub/survey/SurveyView.test.tsx
@@ -10,13 +10,23 @@ import { render, screen } from '@testing-library/react'
 import { PagedSurveyView, SurveyFooter } from './SurveyView'
 import { usePortalEnv } from 'providers/PortalProvider'
 import { useUser } from 'providers/UserProvider'
-import { Survey } from '@juniper/ui-core'
+import { Survey, useAutosaveEffect } from '@juniper/ui-core'
 import Api from 'api/api'
 import { mockEnrollee, mockHubResponse } from 'test-utils/test-participant-factory'
 import userEvent from '@testing-library/user-event'
 import { setupRouterTest } from 'test-utils/router-testing-utils'
 
 jest.mock('providers/PortalProvider', () => ({ usePortalEnv: jest.fn() }))
+
+// mock the useAutosaveEffect, but leave other core functions intact
+jest.mock('@juniper/ui-core', () => {
+  const original = jest.requireActual('@juniper/ui-core')
+  return {
+    ...original,
+    useAutosaveEffect: jest.fn()
+  }
+})
+
 beforeEach(() => {
   (usePortalEnv as jest.Mock).mockReturnValue({
     portal: { name: 'demo' },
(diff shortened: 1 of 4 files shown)


ui-participant/src/hub/survey/SurveyView.test.tsx
```diff
@@ -78,14 +88,15 @@ describe('Renders a survey', () => {
...
+    const { submitSpy, triggerAutosave } = setupSurveyTest(generateThreePageSurvey())
 
     await userEvent.click(screen.getByText('Green'))
     await userEvent.click(screen.getByText('Next'))
     expect(screen.getByText('You are on page2')).toBeInTheDocument()
     await userEvent.type(screen.getByText('text input'), 'my Text')
     await userEvent.click(screen.getByText('Next'))
-    await new Promise(r => setTimeout(r, 600))
+    triggerAutosave()
+    triggerAutosave()
```
denniscunningham, line 99: Why twice?

ui-participant/src/hub/survey/SurveyView.test.tsx
```diff
@@ -230,16 +244,25 @@ describe('Renders a survey', () => {
...
     survey
   }
   const { RoutedComponent } = setupRouterTest(
     <PagedSurveyView enrollee={mockEnrollee()} form={configuredSurvey}
-      studyShortcode={'study'} taskId={'guid34'} autoSaveInterval={200}/>)
+      studyShortcode={'study'} taskId={'guid34'}/>)
   render(RoutedComponent)
   expect(screen.getByText('You are on page1')).toBeInTheDocument()
-  return { submitSpy, RoutedComponent }
+  return { submitSpy, RoutedComponent, triggerAutosave }
```
denniscunningham, line 265: Sorry for my JS ignorance, but this is returning three things, but callers are expecting two things? (It looks like this used to return 2 things and the callers were expecting 1 thing.)

ui-participant/src/hub/survey/SurveyView.test.tsx
```diff
@@ -230,16 +244,25 @@ describe('Renders a survey', () => {
 const setupSurveyTest = (survey: Survey) => {
   const submitSpy = jest.spyOn(Api, 'updateSurveyResponse')
     .mockImplementation(() => Promise.resolve(mockHubResponse()))
+  const autosaveManager = {
+    trigger: (): void => { throw 'no autosave registered' }
+  };
+
+  (useAutosaveEffect as jest.Mock).mockImplementation(saveFn => {
+    autosaveManager.trigger = () => { saveFn() }
```
denniscunningham, line 252: Trying to connect the mocking dots... the `saveFn` here is a no-op function assigned at line 22?

denniscunningham (approved):
Looks good.
A couple questions.

ui-participant/src/hub/survey/SurveyView.test.tsx
devonbush: good question, will add a comment.  This is testing that we don't spam the save API if nothing has changed.

ui-participant/src/hub/survey/SurveyView.test.tsx
devonbush: this is returning an object that used to have two properties, but now has three.  Callers are free to use any/all/none of those properties, and they are referenced by name so adding new properties doesn't impact existing callers.

ui-participant/src/hub/survey/SurveyView.test.tsx
devonbush: the saveFn is what will be passed into the useAutosaveEffect by downstream consumers.  So, e.g. SurveyView will pass in the function that calls the save response API.

MatthewBemis (approved):
Nice!

PR #592, opened 2023-10-16
[JN-542] Support adding new sections to pages

MatthewBemis:
#### DESCRIPTION

* Adds ability to create new HtmlSections
* Disables section type dropdown for previously-saved questions
* Full freetex (i.e. structural) editing for the raw JSON is not yet supported (in the same way that full freetext editing for existing sections is not yet supported). There are future tickets to deal with this but putting guard rails in place for JSON validation felt out of scope given the timelines.

Video demonstrating adding new sections:

https://github.com/broadinstitute/juniper/assets/7257391/b3d89b3b-c5c7-43ae-b13b-5c4df460c2f0

* Start up Admin UI
* Copy/paste some JSON from another section
* Cycle through the section types and confirm the right-side preview works as expected
* Make some nonstructural changes to the JSON and confirm the preview updates
* Confirm you can save the changes and the new sections persist

### DIFF
--- a/ui-admin/src/portal/siteContent/HtmlPageEditView.test.tsx
+++ b/ui-admin/src/portal/siteContent/HtmlPageEditView.test.tsx
@@ -1,0 +1,37 @@
+import { render, screen } from '@testing-library/react'
+import React from 'react'
+import { setupRouterTest } from 'test-utils/router-testing-utils'
+import { mockHtmlPage } from 'test-utils/mock-site-content'
+import HtmlPageEditView from './HtmlPageEditView'
+import userEvent from '@testing-library/user-event'
+
+test('readOnly disables insert new section button', async () => {
+  const mockPage = mockHtmlPage()
+  const { RoutedComponent } = setupRouterTest(
+    <HtmlPageEditView htmlPage={mockPage} readOnly={true} updatePage={jest.fn()}/>)
+  render(RoutedComponent)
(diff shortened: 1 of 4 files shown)


devonbush (approved):
nice, worked well when I tested locally!

denniscunningham:
@MatthewBemis Perhaps something is wrong on my end, but I am only able to edit the new panel by selecting the placeholder text,`{}`, and pasting text. I am not able to type in the new panel, but I am able to navigate. The Website feature is new to me, so I probably do not understand how it works. (I notice somewhat similar behavior in existing panels: I am able to edit existing text but not change the structure (remove or add nesting brackets, etc.).

MatthewBemis:
@denniscunningham yeah, the website editor is still lacking some polish in terms of editing UX. The issue you're seeing is because we don't yet allow changing the structure of the section JSON. Because the JSON is re-validated after every character change, if you start typing a new bracket or quote to alter the structure, it will immediately be seen as invalid JSON and the change is rejected.

This is of course terrible UX, but my understanding is that we just need this to be able to get HeartHive off the ground, so being limited to pasting full sections into the textbox rather than doing all of the editing directly in the textbox is _good enough_ for now. The validation in the survey JSON editor is a good model for what I want this to look like and I don't think it's a heavy lift- it's just not something I wanted to add to the scope of this PR.

denniscunningham (approved):

PR #593, opened 2023-10-16
[JN-634] handling kit bad address errors in the UX

devonbush:
#### DESCRIPTION

This upgrades the error handling of our Pepper API connections, and fixes some stub and test methods that were out of date. Notably:

1. we now have separate Exception classes for "Pepper returned an error" vs. "Pepper returned success, but we don't understand what it said".   
2. We now use ControllerAdvice to handle mapping the PepperApiExceptions to a suitable response.  

The biggest struggle was wrestling with how our openapi source generation clashed with how Spring wants us to handle exceptions.  Despite an hour of research and trying, I was unable to get the generators to add "throws Exception" to any controller method (docs say you should just be able to add "unhandledException" as a config property -- this did not work). This meant that PepperApiException had to be a RuntimeException, otherwise the controller method would have had to catch it and rethrow it as a runtime exception or build an response from scratch, which seemed cumbersome and/or defeating the purpose of configurable exception handling.  I'd love to find a better solution, but this is better than what we had before, and we need to move this along.

Another area that could use improvement in the future is differentiating our response code based on whether Pepper gave us a 400 or 500 error.

1.  redploy ApiAdminApp
2. repopulate ourhealth
3. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHBADR`
4. click "request kit" and request a kit
5. confirm a message appears about the address being invalid
image


core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -57,43 +60,40 @@ public KitRequestService(KitRequestDao dao,
...
      */
-    @Transactional
-    public KitRequest requestKit(AdminUser adminUser, String studyShortcode, Enrollee enrollee, String kitTypeName) {
+    public KitRequest requestKit(AdminUser adminUser, String studyShortcode, Enrollee enrollee, String kitTypeName)
+    throws PepperApiException {
         // create and save kit request
-        Profile profile = profileService.loadWithMailingAddress(enrollee.getProfileId())
-                .orElseThrow(() -> new RuntimeException("Missing profile for enrollee: " + enrollee.getShortcode()));
+        if (enrollee.getProfileId() == null) {
+            throw new IllegalArgumentException("No profile for enrollee: " + enrollee.getShortcode());
```
denniscunningham, line 74: Not for this PR, but this becomes a (potentially) interesting error/exception handling case. Presuming this will be caught (at the latest) as part of response processing (there are other cases, like cron jobs and message handling, but let's put those aside for now), we will want to be consistent on response handling. IOW, is this a 400 or 500? 
If we move to more specific Juniper specific exception classes, we do not lose any expressiveness (the message conveys what is needed), but we are able to specify whether this is an internal error or bad user input. 

In this particular case it is unclear. Should the admin have not asked for a kit for an enrollee without a profile, or is this more like an assert: enrollees should have profiles at this point.

Does that make sense?

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -57,43 +60,40 @@ public KitRequestService(KitRequestDao dao,
...
         try {
-            var untypedStatuses = PepperKitStatusResponse.extractUntypedKitStatuses(response, objectMapper);
-            // Pepper response from requesting it kit must have one and only one kit status
-            var pepperStatusJson = objectMapper.writeValueAsString(untypedStatuses.get(0));
+            PepperKitStatus dsmKitStatus = pepperDSMClient.sendKitRequest(studyShortcode, enrollee, kitRequest, pepperKitAddress);
+            // write out the PepperKitStatus as a string for storage
+            var pepperStatusJson = objectMapper.writeValueAsString(dsmKitStatus);
             kitRequest.setDsmStatus(pepperStatusJson);
-            saveKitStatus(kitRequest, pepperStatusJson, kitRequest.getCreatedAt());
+        } catch (PepperParseException e) {
```
denniscunningham, line 86: Does it make sense to catch `PepperApiException` here if there is something sensible we can do with it? Seems like we'd want to at least handle 400 and 500 responses differently, and further up the stack is probably not the place to for that.
By re-throwing different exceptions here we can also convert the message to something that makes sense to the user,  and (implicitly) dictate the response handling.

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/LivePepperDSMClient.java
```diff
@@ -43,25 +44,31 @@ public LivePepperDSMClient(PepperDSMConfig pepperDSMConfig,
...
 
     @Override
-    public String sendKitRequest(String studyShortcode, Enrollee enrollee, KitRequest kitRequest, PepperKitAddress address) {
+    public PepperKitStatus sendKitRequest(String studyShortcode, Enrollee enrollee, KitRequest kitRequest, PepperKitAddress address)
+    throws PepperApiException, PepperParseException {
         var request = buildAuthedPostRequest("shipKit", makeKitRequestBody(studyShortcode, enrollee, kitRequest, address));
-        return retrieveAndDeserializeResponse(request, String.class);
+        PepperKitStatusResponse response = retrieveAndDeserializeResponse(request, PepperKitStatusResponse.class);
+        if (response.getKits().length != 1) {
+            throw new PepperApiException("Expected a single result from shipKit by ID (%s), got %d".formatted(
```
denniscunningham: Is this really a `PepperApiException`, since it seems to be a fundamental contract misunderstanding? IOW, if this lands back with the user, because of how we handle `PepperApiException` further up the stack, then what are they supposed to do to fix this?

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/PepperApiException.java
```diff
@@ -0,0 +1,39 @@
+package bio.terra.pearl.core.service.kit.pepper;
+
+import lombok.Getter;
+import org.springframework.http.HttpStatus;
+
+/**
+ *  Class for 400-500 responses returned from Pepper.  This intentionally extends
```
denniscunningham, line 7: I think the error handling gets easier if we use this only at the lowest levels of the Pepper interface code (throw and catches), and convert the errors to something more generic that the rest of the stack can confidently process. Not wed to the concept, but I suggested having two uber exception classes like `JuniperBadRequest` and `JuniperInternalError` (or whatever naming works).  
This requires that the higher levels have no need to process a Pepper error any differently from the error from any service, but I suspect that is true for most/all service errors. (And in cases where it is not true, we might want to take a closer look at the cases.

denniscunningham (commented):
I am not done with the review, but you asked for early feedback...

Following up on our conversation this afternoon, I think this gets less complicated if we move to more generic exception classes geared towards response handling. (See my comments for more details.) This is easy to do gradually, so it would not be a major change to introduce that new technique in this PR.

It is entirely possible I missed an important detail, so happy to learn more.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
devonbush: Agree -- also because this check likely will get finer-scale -- sure they have a profile, but what if they have a profile with no address?  or an address with no street?

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
devonbush: Hmm... I'm not sure what we'd do with it.  Especially thinking that this is a service that will later on be invoked by automatic processes, I think we want to leave it up to the caller how to handle the various flavors of PepperApiException.

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/LivePepperDSMClient.java
devonbush: Yeah, I think you're right that this is a PepperParseException

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/PepperApiException.java
devonbush: Interesting -- maybe we have those two base classes and then PepperApiException and friends extend them?  Then we don't have to do conversion -- each layer of the stack can operate on whatever granularity it likes?

core/src/main/java/bio/terra/pearl/core/dao/DaoUtils.java
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * gets a uuid from the database.
+     * Useful for when you need an id but do not want to create an object just yet
+     * */
+    public UUID generateUUID() {
+        return jdbi.withHandle(handle -> handle.createQuery("select gen_random_uuid ();")
+                .mapTo(UUID.class)
+                .one()
+        );
```
MatthewBemis, line 23: What's the benefit of going to the database to get a UUID instead of using `java.util.UUID.randomUUID()`? Is it to ensure consistency in how/which version of UUID is generated?

MatthewBemis (approved):
One question, otherwise looks solid.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
denniscunningham: But what is a different flavor? Just in terms of encapsulation, why would code external to this service need to understand how to interpret responses from Pepper, particularly if they arrive in the same packaging, PepperApiException.

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/PepperApiException.java
denniscunningham: Sure, if that works for you. I suspect that additional shell is not necessary, but there is not a lot of downside to doing that way (unless the Spring exception handling stuff cannot handle the inheritance properly).

core/src/main/java/bio/terra/pearl/core/dao/DaoUtils.java
devonbush: yeah, I think UUID.randomUUID would work just as well, this is just being super careful.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
devonbush: Ah, maybe the code/comments aren't clear about the control flow -- we need to catch the PepperParseException because that means that the request to Pepper was successful.  And if the request was successful, we need to make sure we create the matching record in our database for bookkeeping purposes.    And likewise, we need to return a 20x status code since the operation overall was a success, so rethrowing something here just means we'd have to catch it again at the controller.  Telling an admin "your kit request was successful but DSM included some unexpected fields in its response" isn't particularly useful--I think logging as error so it gets picked up in alerting/triage is probably best because chances are the error is going to be very inside-baseball.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
denniscunningham: As things stand now this exception will whiz past the exception handling in `KitExtService.requestKits`, and I assume respond with a 500. Is that what we want in the multi-enrollee case? I can't tell if this is an assert or a viable case.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
denniscunningham: Makes sense. If we get a muddled response do we have enough info to make a useful record in the DB?
Do we have alerting setup for 500 responses and logging at levels >= error?

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -57,43 +60,40 @@ public KitRequestService(KitRequestDao dao,
...
+            // response was successful, but we got unexpected format back from pepper
+            // we want to log the error, but still continue on to saving the kit
+            log.error("Unable to parse kit response status from Pepper: kit id %s", kitRequest.getId());
```
denniscunningham, line 89: You probably want to include the exception in the error logging here and below?

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -159,21 +159,26 @@ public PepperKitStatus syncKitStatusFromPepper(UUID kitId) {
...
     public void syncAllKitStatusesFromPepper() {
         var studies = studyService.findAll();
         for (Study study : studies) {
             for (EnvironmentName environmentName : EnvironmentName.values()) {
-                syncKitStatusesForStudy(study, environmentName);
+                // if one sync fails, keep trying others in case the failure is just isolated unexpected data
+                try {
+                    syncKitStatusesForStudy(study, environmentName);
+                } catch (PepperParseException | PepperApiException e) {
+                    log.error("kit status sync failed for study %s".formatted(study.getShortcode()), e);
```
denniscunningham, line 174: This probably deserves a comment, similar to the above, since swallowing exceptions should be considered exceptional (ahem).

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -215,32 +220,34 @@ public void deleteByEnrolleeId(UUID enrolleeId, Set<CascadeProperty> cascade) {
...
+        return kitRequest;
+    }
+
+    protected String stringifyPepperAddress(PepperKitAddress kitAddress) {
+        try {
+            return objectMapper.writeValueAsString(kitAddress);
+        } catch (JsonProcessingException e) {
+            // There's no good reason for PepperKitAddress serialization to fail, so if it does, something very
+            // unexpected is happening.  Throw RuntimeException to ensure @Transactional annotations will rollback.
+            throw new RuntimeException(e);
```
denniscunningham, line 249: This is a great place for `JuniperInternalError`, when we invent it.

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/PepperApiException.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public PepperApiException(String message) {
+        super(message);
+    }
+
+    public PepperApiException(String message, Exception cause) {
+        super(message, cause);
+    }
+
+    public PepperApiException(String message, PepperErrorResponse errorResponse, HttpStatus httpStatus) {
```
denniscunningham, line 24: Are you using these methods now (where you include the HttpStatus)? I think it's worth discussion on alternative approaches.

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/PepperErrorResponse.java
```diff
@@ -0,0 +1,30 @@
...
+    @NotNull
+    private String juniperKitId;
+    private PepperErrorValue value;
+
+    @Getter @Setter
+    @SuperBuilder
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @ToString
+    public static class PepperErrorValue extends PepperResponse {
```
denniscunningham: Not sure I understand why this extends `PepperReponse`, since it appears only be used as a field in `PepperErrorResponse`.

denniscunningham (commented):
A few more questions. I realize we want to get this in soon, so I'm trying to not hold this up and move on to an exception handling discussion later. This is a clear improvement over the existing code, which I assume is our main objective.

I did not manually test a multi-enrollee kit request, so let me know if you would like me to do that.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
devonbush: It's an IllegalArgumentException, which AbstractGlobalExceptionHandler will turn into a BAD_REQUEST response.  You're right that we probably want to catch this in the Ext service for multi-requests.  Will add.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
devonbush: we don't yet have alerting -- just the weekly log review.  We'll likely want to up that to daily and add alerting once OurHealth traffic ramps up.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
devonbush: we're very leery of including exceptions and responses in logs--the directive from AppSec/Compliance is to avoid logging PII/PHI, and logging the exception/response here seems like a reasonable risk that the response from Pepper would have PII/PHI in it

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/PepperApiException.java
devonbush: yes, capturing the httpStatus that pepper returned seems very useful, in terms of differentiating failures due to auth, input, or other.  Happy to talk about other ways to pass that info along

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/PepperErrorResponse.java
devonbush: I can't remember why I did that either, removing

devonbush (commented):
thanks for the good comments, just pushed some cleanups

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
denniscunningham: Not sure I agree with that, but is the stack trace helpful here? If so, we could just add that part. (Perhaps not for this PR.)

api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
```diff
@@ -47,21 +47,19 @@ public KitRequestListResponse requestKits(
...
         KitRequest kitRequest = requestKit(adminUser, studyShortcode, enrolleeShortcode, kitType);
         response.kitRequests.add(kitRequest);
-      } catch (PepperApiException pepperApiException) {
-        // add the enrollee shortcode to the message for disambiguation
-        response.pepperApiExceptions.add(
-            new PepperApiException(
-                enrolleeShortcode + ": " + pepperApiException.getMessage(),
-                pepperApiException.getErrorResponse(),
-                pepperApiException.getHttpStatus()));
+      } catch (Exception e) {
```
denniscunningham, line 50: You could keep the `PepperApiException` processing and add `Exception` processing since that seems useful?

denniscunningham (approved):
Looks good.
Just a couple comments. Approving so you can decide if/when to address those.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
devonbush: good point.  we'll see how this shakes out once we have a better overall exception-handling strategy.

PR #595, opened 2023-10-18
[JN-543] image list page in admin tool

devonbush:
#### DESCRIPTION

In preparation for allowing image upload, this adds an image list table to the admin tool.  Shows the images in a sorted list, with the ability to click on the image thumbnails to show a larger preview.  
image

I added some clarifying comments around the pre-existing public single-image API endpoint, but did not make any functional changes to it.  It now routes through a SiteImageextService so that the controller does not need to import both SiteImageService and SiteImageExtService, which could lead to typos that compromise auth patterns.

this also has a drive-by fix of removing a TODO comment related to kits -- Pegah and I agreed that url casing change is not worth it to be done since everything is already working.  

1. Redeploy ApiAdminApp
2. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/siteContent`
3. click the new "Manage images" link
4. Confirm you see a table of images
5. click an image, and confirm a larger preview appears


core/src/main/java/bio/terra/pearl/core/dao/site/SiteImageDao.java
```diff
@@ -2,14 +2,23 @@
...
 import org.jdbi.v3.core.Jdbi;
 import org.springframework.stereotype.Component;
 
 @Component
 public class SiteImageDao extends BaseJdbiDao<SiteImage> {
+    private String fieldStringWithoutDataColumn;
     public SiteImageDao(Jdbi jdbi) {
+
         super(jdbi);
+        List<String> colsWithoutDataCol = getGetQueryColumns();
```
denniscunningham: A comment here may help. 
Nit: It definitely depends on the reader, but I find the var name `fieldStringWithoutDataColumn` more confusing than helpful.

core/src/test/java/bio/terra/pearl/core/service/site/SiteImageServiceTests.java
```diff
@@ -33,7 +41,7 @@ public void testGenerateSpecialCharShortcode() {
 
     @Test
     @Transactional
-    public void testCrud() {
+    public void testCrud(TestInfo testInfo) {
```
denniscunningham, line 45: Looks like you did not end up using `TestInfo`?

ui-admin/src/api/api.tsx
```diff
@@ -183,6 +183,13 @@ export type AdminTask = {
   dispositionNote?: string
 }
 
+export type SiteImage = {
```
denniscunningham: Nit: Maybe `SiteImageMetadata`?

core/src/main/java/bio/terra/pearl/core/service/site/SiteImageService.java
```diff
@@ -33,6 +33,10 @@ public Optional<SiteImage> findOneLatestVersion(String portalShortcode, String c
         return dao.findOneLatestVersion(portalShortcode, cleanFileName);
     }
 
+    public List<SiteImage> findByPortalWithoutDataColumn(String portalShortcode) {
```
denniscunningham: Does it make sense to not pretend `SiteImage` is complete and instead create a `SiteImageMetadata` class? (SiteImage can extend SiteImageMetadata.) That would make the method naming more straightforward, and we would not have to account for `SiteImage` instances that do not actually have an image?

denniscunningham (commented):
Dumb question, but what's the advantage of downloading images separately from their metadata? Is this to ensure that we only fetch the images for a page at a time, or to deal with how HTML deals with images, or something else?

devonbush:
Fetching the images separately means the page can be functional before the images all load.  And it means we don't have to send the images not on the current page.  Not a big deal for OurHealth whose total image directory is ~30MB, but that's because we helped them compress all their images and pick appropriate sizes.  When users are uploading their own images (likely pulled from splash sites on the internet) the image sizes could get massive.  

Also, individual images are very easily cacheable, but a big image bundle would only be cacheable if no single image had changed

core/src/test/java/bio/terra/pearl/core/service/site/SiteImageServiceTests.java
devonbush: yeah, I forgot the getTestName helper is stranded on the study creation modal branch

devonbush:
And yeah, I think you're right that a SiteImageMetadata DTO will make a lot of this cleaner

ui-admin/src/util/tableUtils.tsx
```diff
@@ -388,6 +389,22 @@ export function basicTableLayout<T>(table: Table<T>, config: BasicTableConfig =
...
+ */
+export const useRoutableTablePaging = (tableIdentifier: string) => {
+  const [searchParams] = useSearchParams()
+  const preferredNumRowsKey = `${tableIdentifier}.preferredNumRows`
+  const paginationState: PaginationState = {
+    pageIndex: parseInt(searchParams.get('pageIndex') || '0'),
+    pageSize: parseInt(searchParams.get('pageSize') || localStorage.getItem(preferredNumRowsKey) || '10')
+  }
+  return { paginationState, preferredNumRowsKey }
+}
```
MatthewBemis, line 406: oooo nice

MatthewBemis (approved):

core/src/testFixtures/java/bio/terra/pearl/core/factory/site/SiteImageFactory.java
```diff
@@ -11,19 +12,25 @@
 public class SiteImageFactory {
     @Autowired
     private PortalFactory portalFactory;
+    @Autowired
+    private SiteImageService siteImageService;
 
-    public SiteImage.SiteImageBuilder builder(String testName) {
+    public SiteImage.SiteImageBuilder<?, ?> builder(String testName) {
```
denniscunningham, line 18: Just curious, why was the generic wildcarding needed here?

core/src/main/java/bio/terra/pearl/core/model/site/SiteImageMetadata.java
```diff
@@ -0,0 +1,24 @@
...
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class SiteImageMetadata extends BaseEntity {
+    /**
+     * the cleanFileName essentially serves as a stableId -- we call it cleanFileName here to indicate its origin.
+     * and because it is not required to be globally unique like other stableIds -- it just has to be unique within a portal
+     */
+    private String cleanFileName;
```
denniscunningham, line 22: Ah, what was not clear until I read the frontend code is that effectively `cleanFileName`, `version` and `portalShortcode` are a key. That may be worth mentioning here since it makes the field use clearer.
If that's true, can `uploadFileName` vary across different versions?

denniscunningham (approved):
Looks good.
A couple of questions.

core/src/main/java/bio/terra/pearl/core/model/site/SiteImageMetadata.java
devonbush: Yeah, I'll add that comment.  TBH I'm not sure how the use case for uploadFileName is going to play out.  Originally, that was a bookkeeping column for populate methods to track where a file came from, and I thought it might also be useful to tell staff users what the file name uploaded was so they could find it on their machine.  Now I don't think we use it for populate anymore, and I think tracking it for staff users is probably just as likely to cause confusion.   I'll think more about that and might remove `uploadFileName` in the next PR.

core/src/testFixtures/java/bio/terra/pearl/core/factory/site/SiteImageFactory.java
devonbush: There's some deep lombok magic for SuperBuilder to enable it to chain the builders to the right types when there is inheritance.  I'm not 100% sure I understand it myself, but if you don't have this, it will return a SiteImageBuilder, but whose methods return SiteImageMetadataBuilders.  

So, concretely, without these wildcards, the method below it
```
public SiteImage.SiteImageBuilder<?, ?> builderWithDependencies(String testName) {
        Portal portal = portalFactory.buildPersisted(testName);
        return builder(testName)
                .portalShortcode(portal.getShortcode());
}
```
would fail to compile because the `.portalShortcode` method would return a SiteImageMetadataBuilder instead of a SiteImageBuilder

PR #597, opened 2023-10-18
[JN-645] study setting config

devonbush:
#### DESCRIPTION

This enables changing password protection and other study-environment level configurations.  This is what I had to do manually for OurHealth today to make their study public.  This also cleans up a bunch of the PortalSettings controls and brings them inline with our latest patterns.

image

One pattern I've been trying out in the last couple PRs is using "real" data in our ExtService auth tests instead of the "mock everything and just confirm a method is called" pattern we've been using.  The real data approach requires a bit more setup, and the tests will take a bit longer to execute, but I think it's worth it because then our ExtService tests serve as backup module/integration tests (and also encourage us to develop more streamlined factory patterns).  Curious about y'all's thoughts. 

1. redeploy ApiAdminApp 
2. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/settings
3. edit the study settings password
4. click "Save study config"
5. confirm the value is persisted.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/StudyEnvironmentController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/StudyEnvironmentController.java
@@ -3,7 +3,7 @@ package bio.terra.pearl.api.admin.controller;
 import bio.terra.pearl.api.admin.api.StudyEnvironmentApi;
 import bio.terra.pearl.api.admin.model.StudyEnvironmentDto;
 import bio.terra.pearl.api.admin.service.AuthUtilService;
-import bio.terra.pearl.api.admin.service.StudyEnvironmentExtService;
+import bio.terra.pearl.api.admin.service.study.StudyEnvironmentExtService;
 import bio.terra.pearl.core.model.EnvironmentName;
 import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.study.StudyEnvironment;
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/StudyEnvironmentController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/StudyEnvironmentController.java
@@ -7,6 +7,7 @@ import bio.terra.pearl.api.admin.service.StudyEnvironmentExtService;
 import bio.terra.pearl.core.model.EnvironmentName;
 import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.study.StudyEnvironment;
+import bio.terra.pearl.core.model.study.StudyEnvironmentConfig;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import javax.servlet.http.HttpServletRequest;
 import org.springframework.http.ResponseEntity;
@@ -35,10 +36,26 @@ public class StudyEnvironmentController implements StudyEnvironmentApi {
   public ResponseEntity<StudyEnvironmentDto> patch(
       String portalShortcode, String studyShortcode, String envName, StudyEnvironmentDto body) {
     AdminUser adminUser = requestService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
     StudyEnvironment envUpdate = objectMapper.convertValue(body, StudyEnvironment.class);
     StudyEnvironment savedEnv =
-        studyEnvExtService.update(adminUser, portalShortcode, studyShortcode, envUpdate);
+        studyEnvExtService.update(
+            adminUser, portalShortcode, studyShortcode, environmentName, envUpdate);
     return ResponseEntity.ok(objectMapper.convertValue(savedEnv, StudyEnvironmentDto.class));
+  }
+
+  /** updates the config object associated with the environment */
+  @Override
+  public ResponseEntity<Object> patchConfig(
+      String portalShortcode, String studyShortcode, String envName, Object body) {
+    AdminUser adminUser = requestService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+    StudyEnvironmentConfig configUpdate =
+        objectMapper.convertValue(body, StudyEnvironmentConfig.class);
+    StudyEnvironmentConfig savedConfig =
+        studyEnvExtService.updateConfig(
+            adminUser, portalShortcode, studyShortcode, environmentName, configUpdate);
+    return ResponseEntity.ok(savedConfig);
   }
 
   @Override
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtService.java
@@ -1,5 +1,6 @@
-package bio.terra.pearl.api.admin.service;
+package bio.terra.pearl.api.admin.service.study;
 
+import bio.terra.pearl.api.admin.service.AuthUtilService;
 import bio.terra.pearl.core.model.EnvironmentName;
 import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.study.StudyEnvironment;
--- a/api-admin/src/test/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtServiceTests.java
+++ b/api-admin/src/test/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtServiceTests.java
@@ -1,0 +1,62 @@
+package bio.terra.pearl.api.admin.service.study;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+import bio.terra.pearl.api.admin.BaseSpringBootTest;
+import bio.terra.pearl.core.factory.StudyEnvironmentFactory;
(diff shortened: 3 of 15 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtService.java
```diff
@@ -0,0 +1,79 @@
...
+      String studyShortcode,
+      EnvironmentName envName,
+      StudyEnvironmentConfig update) {
+    StudyEnvironment studyEnv = authToStudyEnv(operator, portalShortcode, studyShortcode, envName);
+    StudyEnvironmentConfig existing =
+        studyEnvConfigService.find(studyEnv.getStudyEnvironmentConfigId()).get();
+    // we don't allow directly setting the 'initialized' field -- that comes from the publishing
+    // flows
+    existing.setPasswordProtected(update.isPasswordProtected());
+    existing.setAcceptingEnrollment(update.isAcceptingEnrollment());
```
MatthewBemis, line 59: From testing, this field can be set, but it doesn't look like it's honored anywhere- I can still enroll even when this is set to false. Are we expecting this to work? More of a general question, it's certainly not a result of this PR.

ui-admin/src/study/StudySettings.tsx
```diff
@@ -1,39 +1,79 @@
...
+  const { user } = useUser()
+  const [isLoading, setIsLoading] = useState(false)
+  /** update a given field in the config */
+  const updateConfig = (propName: string, value: string | boolean) => {
+    setConfig(set(propName, value))
+  }
+
+  /** saves any changes to the server */
+  const save = async (e: React.MouseEvent) => {
+    e.preventDefault()
```
MatthewBemis: The `Button` component specifies `type="button"` which should prevent the form submit behavior without this line. (if that's what you were preventing here)

core/src/testFixtures/java/bio/terra/pearl/core/factory/StudyEnvironmentFactory.java
```diff
@@ -62,4 +69,26 @@ public StudyEnvironment buildPersisted(EnvironmentName envName, UUID studyId, St
...
+        Study study = studyFactory.buildPersisted(portalEnvironment.getPortalId(), testName);
+        StudyEnvironment studyEnvironment = buildPersisted(envName, study.getId(), testName);
+        Portal portal = portalService.find(portalEnvironment.getPortalId()).get();
+        return StudyEnvironmentBundle.builder()
+                .study(study)
+                .studyEnv(studyEnvironment)
+                .portal(portal)
+                .portalEnv(portalEnvironment)
+                .build();
+    }
```
MatthewBemis, line 83: Nice, seems like a quick way to get everything you need. Not a concern yet, but re: longer execution times, we should keep that in mind and periodically check in on execution time. One of the many nice things about this project is that it's pretty quick to iterate on tests (running tests locally, waiting for CI to run tests, etc). Easier to deal with and remediate any slowness when we stay ahead of those things.

MatthewBemis (approved):

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/StudyEnvironmentController.java
```diff
@@ -35,12 +36,28 @@ public StudyEnvironmentController(
...
   }
 
+  /** updates the config object associated with the environment */
+  @Override
+  public ResponseEntity<Object> patchConfig(
+      String portalShortcode, String studyShortcode, String envName, Object body) {
+    AdminUser adminUser = requestService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+    StudyEnvironmentConfig configUpdate =
+        objectMapper.convertValue(body, StudyEnvironmentConfig.class);
```
denniscunningham, line 54: If this throws I assume it will result in a 500 response, but it probably should be a 400?

core/src/testFixtures/java/bio/terra/pearl/core/factory/StudyEnvironmentFactory.java
devonbush: yeah, agree on fast tests being nice.  my main goal is to make sure we're keeping our tests independent.  As long as they're independent, and don't rely on, e.g. populating a big shared DB, we can make them fast in CI by parallelizing them.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtService.java
devonbush: wow, you're right.  I'm trying to remember what that field was supposed to do -- I'll make a ticket to either remove it or implement it from the backend, and I'll hide it for now

api-admin/src/test/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtServiceTests.java
```diff
@@ -0,0 +1,62 @@
...
+  @Autowired private StudyEnvironmentConfigService studyEnvironmentConfigService;
+
+  @Test
+  @Transactional
+  public void updateConfigAuthsToStudy() {
+    var studyEnvBundle =
+        studyEnvironmentFactory.buildBundle("updateConfigAuthsToStudy", EnvironmentName.irb);
+    AdminUser operator = adminUserFactory.buildPersisted("updateConfigAuthsToStudy", false);
+    Assertions.assertThrows(
+        NotFoundException.class,
```
denniscunningham, line 32: I'm not clear on what this is testing. What is not found in this case?

api-admin/src/test/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtServiceTests.java
```diff
@@ -0,0 +1,62 @@
...
+        studyEnvBundle.getPortal().getShortcode(),
+        studyEnvBundle.getStudy().getShortcode(),
+        EnvironmentName.irb,
+        StudyEnvironmentConfig.builder().password("test456").build());
+
+    var updatedConfig =
+        studyEnvironmentConfigService
+            .find(studyEnvBundle.getStudyEnv().getStudyEnvironmentConfigId())
+            .get();
+    assertThat(updatedConfig.getPassword(), equalTo("test456"));
```
denniscunningham, line 60: So with this pattern we do some functional testing via the `ExtService` and some via the `Service`. Is there a particular line between the two?

api-admin/src/test/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtServiceTests.java
```diff
@@ -0,0 +1,62 @@
...
+              EnvironmentName.irb,
+              new StudyEnvironmentConfig());
+        });
+  }
+
+  @Test
+  @Transactional
+  public void updateConfigAllowsSuperuser() {
+    var studyEnvBundle =
+        studyEnvironmentFactory.buildBundle("updateConfigAuthsToStudy", EnvironmentName.irb);
```
denniscunningham: Is using the same bundle seed/test name in both `updateConfigAuthsToStudy` and `updateConfigAllowsSuperuser` intentional?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/StudyEnvironmentController.java
devonbush: This will throw IllegalArgumentException, which will be converted to 400

api-admin/src/test/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtServiceTests.java
devonbush: The Portal.  authUserToPortal throws a 404 not found if the user is not authorized to access that portal.  Similar to how github returns a 404 if you try to access a private repo you don't have access to.

api-admin/src/test/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtServiceTests.java
devonbush: Great question.  ExtServices do Auth, and core services do not.  core services handle the very basic CRUD operations and DAO usage, while ExtServices do not access DAOs.  There's mushy middle in between.  So this test pattern is trying to make sure we cover that mushy middle wherever it is.

api-admin/src/test/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtServiceTests.java
devonbush: copypasta -- good catch.  fixed

api-admin/src/test/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtServiceTests.java
denniscunningham: A comment may help future readers, particularly with this new pattern where it is not obvious we are fishing for a auth failure.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/StudyEnvironmentController.java
denniscunningham: My bad, I was thinking a JSON exception but should have checked.

denniscunningham (approved):

PR #598, opened 2023-10-19
[JN-625] Add new site pages + bonus features

MatthewBemis:
#### DESCRIPTION

This adds a bunch of new functionality to the site content editor. The main thing is the ability to add new pages (which automatically adds them to the navbar). I was having fun with it so I added a few extra small features that were inspired by the functionality in the survey editor. I also spent some time making the styling a little more consistent with the survey editor and making it so components on the page erratically resize as content changed on the screen. Videos below.

Add a new page:

https://github.com/broadinstitute/juniper/assets/7257391/3608d15f-02da-4485-a8e9-6dd2b9e89b0b

Delete a page section:

https://github.com/broadinstitute/juniper/assets/7257391/f50cf7d6-bade-4529-be34-0b23f7baf47a

Move page sections:

https://github.com/broadinstitute/juniper/assets/7257391/79e83ab8-86e7-4b75-a13b-2bb64f7dcc50

* Start up admin + participant API servers and UIs
* Try adding a new page, confirm it shows up in the navbar after saving
* Confirm you can delete sections
* Confirm you can move sections around
* In general just try modifying as much state as possible in-between saves and confirm the final result after saving is as-expected

### DIFF
--- a/ui-admin/src/portal/siteContent/AddPageModal.tsx
+++ b/ui-admin/src/portal/siteContent/AddPageModal.tsx
@@ -1,0 +1,76 @@
+import React, { useState } from 'react'
+import Modal from 'react-bootstrap/Modal'
+import { HtmlPage, PortalEnvironment } from '@juniper/ui-core'
+import Api from 'api/api'
+import { useConfig } from 'providers/ConfigProvider'
+
+/** renders a modal that adds a new page to the site */
+const AddPageModal = ({ portalEnv, portalShortcode, insertNewPage, show, setShow }: {
+  portalEnv: PortalEnvironment, portalShortcode: string, insertNewPage: (page: HtmlPage) => void,
+  show: boolean, setShow:  React.Dispatch<React.SetStateAction<boolean>>
+}) => {
+  const zoneConfig = useConfig()
+
+  const [pageTitle, setPageTitle] = useState('')
+  const [pagePath, setPagePath] = useState('')
+
+  const addPage = async () => {
+    insertNewPage({
+      title: pageTitle,
+      path: pagePath,
+      sections: []
+    })
+    setShow(false)
+  }
+
+  const clearFields = () => {
+    setPageTitle('')
+    setPagePath('')
+  }
+
+  const portalUrl = Api.getParticipantLink(portalEnv.portalEnvironmentConfig, zoneConfig.participantUiHostname,
+    portalShortcode, portalEnv.environmentName)
+
+  return <Modal show={show}
+    onHide={() => {
+      setShow(false)
+      clearFields()
+    }}>
+    <Modal.Header closeButton>
+      <Modal.Title>Add New Page</Modal.Title>
+    </Modal.Header>
+    <Modal.Body>
+      <form onSubmit={e => e.preventDefault()}>
+        <label htmlFor="inputPageTitle">Page Title</label>
+        <input type="text" size={50} className="form-control mb-3" id="inputPageTitle" value={pageTitle}
+          onChange={event => {
+            setPageTitle(event.target.value)
+          }}/>
+        <label htmlFor="inputPagePath">Page Path</label>
+        <div className="input-group">
+          <div className="input-group-prepend">
+            <span className="input-group-text" id="pathPrefix">{portalUrl}/</span>
+          </div>
+          <input type="text" className="form-control" id="inputPagePath"
+            value={pagePath} aria-describedby="pathPrefix"
+            onChange={event => {
+              setPagePath(event.target.value)
+            }}/>
+        </div>
+      </form>
+    </Modal.Body>
+    <Modal.Footer>
+      <button
+        className="btn btn-primary"
+        disabled={!pageTitle || !pagePath}
+        onClick={addPage}
+      >Create</button>
+      <button className="btn btn-secondary" onClick={() => {
+        setShow(false)
+        clearFields()
+      }}>Cancel</button>
+    </Modal.Footer>
+  </Modal>
+}
+
+export default AddPageModal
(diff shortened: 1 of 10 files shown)


ui-admin/src/portal/siteContent/SiteContentEditor.tsx
```diff
@@ -2,15 +2,16 @@ import React, { useState } from 'react'
...
 import HtmlPageEditView from './HtmlPageEditView'
 import { HtmlPage, LocalSiteContent, ApiProvider, SiteContent, ApiContextT } from '@juniper/ui-core'
 import { Link } from 'react-router-dom'
 import SiteContentVersionSelector from './SiteContentVersionSelector'
 import { Button } from '../../components/forms/Button'
+import AddPageModal from './AddPageModal'
 
-type NavbarOption = {label: string, value: string | null}
-const landingPageOption = { label: 'Landing page', value: null }
+type NavbarOption = {label: string, value: string}
```
MatthewBemis, line 13: Previously the `value` field of `NavbarOption` was the ID of the `NavbarItem`. Landing page was the null/special case because it wasn't associated with a navbar item. I had to switch `value` to be the page title/text of the navbar item because newly added pages don't have an ID until they've been saved to the database. So the special-case for Landing page was no longer unique because I don't want to force a save just to make the page dropdown work correctly. My assumption is that navbar item text is unique _enough_ to not really be a concern, but maybe it'd be worth protecting against that?

ui-admin/src/portal/siteContent/SiteContentLoader.tsx
```diff
@@ -47,7 +47,7 @@ const SiteContentLoader = ({ portalEnvContext }: {portalEnvContext: PortalEnvCon
       Store.addNotification(failureNotification('save failed'))
       return
     }
-    await switchToVersion(newVersion.id, newVersion.stableId, newVersion.version, newVersion)
+    await switchToVersion(newVersion.id, newVersion.stableId, newVersion.version, undefined)
```
MatthewBemis: This fixes a previous bug that caused all pages other than the landing page to be blank after saving a new version. `switchToVersion` allows you to pass in previously-loaded content or force it to fetch the version from the server. Here we were using the response from `createNewSiteContentVersion` to avoid going back to the server to fetch the newly-saved version. Sadly that response only includes the htmlPage contents for the landing page, hence the blank pages for everything else. It's cheap enough to just fetch the version after saving it but I could have also updated the response for that endpoint.

ui-admin/src/portal/siteContent/AddPageModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const clearFields = () => {
+    setPageTitle('')
+    setPagePath('')
+  }
+
+  const portalUrl = Api.getParticipantLink(portalEnv.portalEnvironmentConfig, zoneConfig.participantUiHostname,
+    portalShortcode, portalEnv.environmentName)
+
+  return <Modal show={show}
```
MatthewBemis, line 34: There are probably other options that we'll want to surface in this modal in the future, like allowing the user to set the new page as the landing page or not add it to the navbar. But for now this should be enough?

ui-admin/src/portal/siteContent/SiteContentLoader.tsx
MatthewBemis: Bug fixed _for real_ courtesy of Devon 35cbfa074001955a980e9145b94bb0122b9fa43f

ui-admin/src/portal/siteContent/SiteContentEditor.tsx
devonbush: yeah, I can't imagine people wanting nav links with duplicate text

devonbush (approved):
Created and saved a bunch of stuff locally, seems to work and look great.  nice!

ui-admin/src/portal/siteContent/SiteContentEditor.tsx
```diff
@@ -53,20 +55,38 @@ const SiteContentEditor = (props: InitializedSiteContentViewProps) => {
     setWorkingContent(newWorkingContent)
   }
 
+  const insertNewPage = (page: HtmlPage) => {
+    if (!localContent) {
+      return
+    }
+    const newNavBarItem: NavbarItemInternal = {
```
denniscunningham, line 62: Just curious, what's the difference between NavBarItems of type `EXTERNAL`, `INTERNAL`, etc.?

ui-admin/src/portal/siteContent/HtmlPageEditView.tsx
```diff
@@ -47,11 +25,20 @@ const HtmlPageView = ({ htmlPage, updatePage, readOnly }: HtmlPageViewProps) =>
   }
 
   return <div>
+    <div className="col-md-12 my-2" style={{ backgroundColor: '#eee' }}>
+      <Button variant="secondary"
+        aria-label={'Insert a blank section'}
+        tooltip={'Insert a blank section'}
+        disabled={readOnly}
+        onClick={() => insertNewSection(0, { id: '', sectionType: 'HERO_WITH_IMAGE' })}>
```
denniscunningham, line 33: What role does the JS template code associated with the `sectionType` play? Just trying to understand how this works since the code section is blank when you insert the section.

denniscunningham (approved):
Very nice.
Just a couple of questions that have no bearing on the review per se.

ui-admin/src/portal/siteContent/SiteContentEditor.tsx
devonbush: Check out NavbarItemType.java for the enum with comments

ui-admin/src/portal/siteContent/HtmlPageEditView.tsx
MatthewBemis: Good question, this doesn't play with those template configs yet. But that would be a good next step to make the newly-added sections more useful, so you could have a pre-populated template in the section editor instead of having to copy-paste.

PR #600, opened 2023-10-19
[JN-543] upload image function

devonbush:
DESCRIPTION
This adds an "Add image" and "update" buttons on the image manager which allow a user to upload files.
image
The "update" buttons will upload a new version of the existing image. To see the update in the website, the user will then have to change the image version in the siteContent

TO TEST: (simple manual steps for confirming core behavior -- used for pre-release checks)
redeploy ApiAdminApp
go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/images
click "add Image"
choose an image file to upload
click upload
confirm the updated image can be found in the image list

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/SiteImageController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/SiteImageController.java
@@ -5,7 +5,7 @@ import bio.terra.pearl.api.admin.service.AuthUtilService;
 import bio.terra.pearl.api.admin.service.siteContent.SiteImageExtService;
 import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.site.SiteImage;
-import com.fasterxml.jackson.databind.ObjectMapper;
+
 import java.io.IOException;
 import java.net.URLConnection;
 import java.util.Optional;
@@ -22,17 +22,14 @@ public class SiteImageController implements SiteImageApi {
   private AuthUtilService authUtilService;
   private HttpServletRequest request;
   private SiteImageExtService siteImageExtService;
-  private ObjectMapper objectMapper;
 
   public SiteImageController(
       AuthUtilService authUtilService,
       HttpServletRequest request,
-      SiteImageExtService siteImageExtService,
-      ObjectMapper objectMapper) {
+      SiteImageExtService siteImageExtService) {
     this.authUtilService = authUtilService;
     this.request = request;
     this.siteImageExtService = siteImageExtService;
-    this.objectMapper = objectMapper;
   }
 
   /**
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/SiteImageController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/SiteImageController.java
@@ -5,7 +5,6 @@ import bio.terra.pearl.api.admin.service.AuthUtilService;
 import bio.terra.pearl.api.admin.service.siteContent.SiteImageExtService;
 import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.site.SiteImage;
-
 import java.io.IOException;
 import java.net.URLConnection;
 import java.util.Optional;
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/SiteImageController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/SiteImageController.java
@@ -5,6 +5,8 @@ import bio.terra.pearl.api.admin.service.AuthUtilService;
 import bio.terra.pearl.api.admin.service.siteContent.SiteImageExtService;
 import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.site.SiteImage;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.IOException;
 import java.net.URLConnection;
 import java.util.Optional;
 import javax.servlet.http.HttpServletRequest;
@@ -13,20 +15,24 @@ import org.springframework.core.io.Resource;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
+import org.springframework.web.multipart.MultipartFile;
 
 @Controller
 public class SiteImageController implements SiteImageApi {
   private AuthUtilService authUtilService;
   private HttpServletRequest request;
   private SiteImageExtService siteImageExtService;
+  private ObjectMapper objectMapper;
 
   public SiteImageController(
       AuthUtilService authUtilService,
       HttpServletRequest request,
-      SiteImageExtService siteImageExtService) {
+      SiteImageExtService siteImageExtService,
+      ObjectMapper objectMapper) {
     this.authUtilService = authUtilService;
     this.request = request;
     this.siteImageExtService = siteImageExtService;
+    this.objectMapper = objectMapper;
   }
 
   /**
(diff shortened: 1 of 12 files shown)


ui-admin/src/portal/images/SiteImageUploadModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+    doApiLoad(async () => {
+      await Api.uploadPortalImage(portalContext.portal.shortcode, fileName, version, file)
+      Store.addNotification(successNotification('image saved'))
+      onSubmit()
+    }, { setIsLoading })
+  }
+
+  return <Modal show={true} onHide={onDismiss}>
+    <Modal.Header closeButton>
+      <Modal.Title>{existingImage ? 'Update' : 'Upload'} image</Modal.Title>
```
denniscunningham, line 56: This is not working for me. I see "Update image" after choosing "Add image". Perhaps some lingering `existingImage`?

ui-admin/src/portal/images/SiteImageUploadModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+                    File:
+          <div>
+            { FileChooser }
+            <span className="text-muted fst-italic ms-2">{file?.name}</span>
+          </div>
+        </div>
+        <label className="mt-3 mb-2 form-label">
+                    Name:
+          <input type="text" className="form-control" readOnly={!!existingImage}
+            onChange={e => setFileName(cleanFileName(e.target.value))} value={fileName}/>
```
denniscunningham, line 73: Perhaps because `existingImage` is not set correctly (see above), the "Name" field is not updated when I add a new image.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/siteContent/SiteImageExtService.java
```diff
@@ -34,4 +34,20 @@ public List<SiteImageMetadata> list(String portalShortcode, AdminUser operator)
...
+      String portalShortcode, String uploadFileName, byte[] imageData, AdminUser operator) {
+    authUtilService.authUserToPortal(operator, portalShortcode);
+    String cleanFileName = SiteImageService.cleanFileName(uploadFileName);
+    int version = siteImageService.getNextVersion(cleanFileName, portalShortcode);
+    SiteImage image =
+        SiteImage.builder()
+            .portalShortcode(portalShortcode)
+            .version(version)
+            .data(imageData)
+            .uploadFileName(uploadFileName)
```
denniscunningham, line 48: Feels a bit cleaner if we populate the `cleanFileName` here?
Also, why do we store the `uploadFileName"?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/SiteImageController.java
```diff
@@ -13,20 +15,24 @@
...
   public SiteImageController(
       AuthUtilService authUtilService,
       HttpServletRequest request,
-      SiteImageExtService siteImageExtService) {
+      SiteImageExtService siteImageExtService,
+      ObjectMapper objectMapper) {
     this.authUtilService = authUtilService;
     this.request = request;
     this.siteImageExtService = siteImageExtService;
+    this.objectMapper = objectMapper;
```
denniscunningham: Is this used somewhere?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/SiteImageController.java
```diff
@@ -48,6 +54,19 @@ public ResponseEntity<Object> list(String portalShortcode) {
     return ResponseEntity.ok(siteImageExtService.list(portalShortcode, operator));
   }
 
+  @Override
+  public ResponseEntity<Object> upload(
```
denniscunningham, line 54: image

Can we check for file type?
Is it possible to catch this error and present something better? I realize this is in the bowels of the Web server.

denniscunningham (commented):
A nit: The `update` button feels somewhat unfinished. Not sure if it is as simple as the uppercasing `update`, but there is row jitter when hovering since the button text is bolded.
Thinking ahead to deleting images, is there an icon that works for update?

ui-admin/src/portal/images/SiteImageUploadModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+    // only autofill the name field if this is a new image
+    if (!existingImage) {
+      setFileName(cleanFileName(newFile.name))
+    }
+  }
+
+  const { file, FileChooser } = useFileUploadButton(handleFileChange)
+  const uploadImage = () => {
+    if (!file) { return }
+    const version = existingImage?.version ? existingImage.version + 1 : 1
```
MatthewBemis, line 37:
```suggestion
  const uploadImage = () => {
    if (!file) { return }
    
    if (!file.type.match(/image\/(gif|ico|jpeg|jpg|png|svg|webp)/)) {
      Store.addNotification(failureNotification('File must be an image.'))
      return
    }

    if (file.size > 10 * 1024 * 1024) {
      Store.addNotification(failureNotification('File size exceeds the 10MB limit.'))
      return
    }
    
    const version = existingImage?.version ? existingImage.version + 1 : 1
```

Quick pass at adding some frontend validation for file type and file size

ui-admin/src/portal/images/SiteImageUploadModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+    if (hiddenFileInput.current) {
+      hiddenFileInput.current.click()
+    }
+  }
+
+  return {
+    FileChooser: <span>
+      <Button variant="primary" outline={true} onClick={handleClick}>
+                    Choose file
+      </Button>
```
MatthewBemis, line 115: TOL for future improvements, I really like when websites use something like https://react-dropzone.js.org/

MatthewBemis (approved):

MatthewBemis:
@devonbush Whoops sorry, I did not mean to approve this as part of my initial comments- so I re-requested a review. :) I still want to take another pass...

ui-admin/src/portal/images/SiteImageUploadModal.tsx
devonbush: yup, great catch -- wasn't ever clearing out the existingImage -- fixed

ui-admin/src/portal/images/SiteImageUploadModal.tsx
devonbush: fixed

api-admin/src/main/java/bio/terra/pearl/api/admin/service/siteContent/SiteImageExtService.java
devonbush: ah, yes, you asked about that in the last PR.  Because I'm a data hoarder?  :).  I'll make a ticket to remove it

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/SiteImageController.java
devonbush: deleted

ui-admin/src/portal/images/SiteImageUploadModal.tsx
devonbush: nice, I'm going to go ahead and put this logic in the component itself so you can't submit

devonbush:
Ah, yeah good thought on the secondary button styling.  I'll make a ticket for that -- that's a sitewide style, and we do want to update it, but it will require regression testing beyond the scope of this ticket

devonbush:
Thanks Matt for the good validation logic suggestions--here's what it looks like now:
image

MatthewBemis (approved):
Looks good, had one additional thought:

This is existing code, but `SiteImageService.java` only validates file type using file extension, it doesn't use something more reliable like MIME type (you can currently upload a .zip file if you modify the extension). This doesn't break the UI when rendering the image preview, but should we make that a little more robust on the backend? Just thinking through extra layers of protection since we're accepting arbitrary file uploads.

core/src/main/java/bio/terra/pearl/core/service/site/SiteImageService.java
```diff
@@ -40,6 +40,7 @@ public List<SiteImageMetadata> findMetadataByPortal(String portalShortcode) {
 
     @Override
     public SiteImage create(SiteImage image) {
+
```
denniscunningham, line 43: It feels like now when clean name is now provided we can remove the part where we reprocess the clean file name?

denniscunningham (approved):
Looks good.
One question.

core/src/main/java/bio/terra/pearl/core/service/site/SiteImageService.java
devonbush: there are other places that call this function (e.g. the populate code), so I'd just as soon keep the reprocessing in for safety.

devonbush:
Interesting point on MIME type -- I hadn't thought of MIME types as more robust than extensions, since both are user-supplied.  But you're probably right that double-checking mime type would add additional checks against someone who accidentally named/saved a file with a bad suffix.  Can think about adding next time we touch these files

PR #603, opened 2023-10-23
JN-646 Analysis-friendly export

devonbush:
#### DESCRIPTION

Adds support for the analysis-friendly export of multi-select questions, with a separate column for each choice.  See https://broad-juniper.zendesk.com/hc/en-us/articles/18259824756123-Participant-List-Export-details for a description with examples.  

I'm sure I'm late to the game, but I finally set up Github copilot and used it while I was coding this.  Overall it was 'meh' but there were two places where it wrote entire reasonably complex and correct lines for me after I just typed 3-4 characters.  Bonus points to any reviewer who can identify those lines :)

1. redeploy ApiAdminApp
2. go to `localhost:3000/ourhealth/studies/ourheart/env/live/export/dataBrowser`
3. select 'Download'
4. choose 'analysis-friendly' format
5. downlaod
6. confirm the data exports and multiselect questions are rendered as columns.  (`oh_oh_cardioHx_diagnosedHeartConditions` should show 1s for JOSALK)

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/export/formatters/SurveyFormatter.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/export/formatters/SurveyFormatter.java
@@ -10,14 +10,23 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.*;
 import java.util.stream.Collectors;
+
+import lombok.extern.slf4j.Slf4j;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static java.util.stream.Collectors.groupingBy;
 
+/**
+ * See https://broad-juniper.zendesk.com/hc/en-us/articles/18259824756123-Participant-List-Export-details
+ * for information on the export format of survey questions
+ */
+@Slf4j
 public class SurveyFormatter implements ExportFormatter {
-    private static final Logger logger = LoggerFactory.getLogger(SurveyFormatter.class);
     public static String OTHER_DESCRIPTION_KEY_SUFFIX = "_description";
+    public static String OTHER_DESCRIPTION_HEADER = "other description";
+    public static String SPLIT_OPTION_SELECTED_VALUE = "1";
+    public static String SPLIT_OPTION_UNSELECTED_VALUE = "0";
     private ObjectMapper objectMapper;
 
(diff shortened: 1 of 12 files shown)


core/src/main/java/bio/terra/pearl/core/service/export/formatters/SurveyFormatter.java
```diff
@@ -57,32 +67,43 @@ public Map<String, String> toStringMap(EnrolleeExportData exportData, ModuleExpo
 
     @Override
     public String getColumnKey(ModuleExportInfo moduleExportInfo, ItemExportInfo itemExportInfo, boolean isOtherDescription, QuestionChoice choice) {
-        if (isOtherDescription) {
+         if (isOtherDescription) {
```
denniscunningham: Typo

core/src/main/java/bio/terra/pearl/core/service/export/formatters/SurveyFormatter.java
```diff
@@ -200,37 +226,64 @@ public String valueAsString(Answer answer, List<QuestionChoice> choices, boolean
...
             return answer.getNumberValue().toString();
         } else if (answer.getObjectValue() != null) {
-            return formatObjectValue(answer.getStringValue(), choices, stableIdForOptions, answer);
+            return formatObjectValue(answer, choices, stableIdForOptions);
         }
         return "";
     }
 
+    /** adds an entry to the valueMap for each selected option of a 'splitOptionsIntoColumns' question */
+    public void addSplitOptionSelectionsToMap(ItemExportInfo itemExportInfo,
```
denniscunningham: It seems like this class could be reworked to make a handful of methods private/protected and static.

denniscunningham (commented):
Not particular to the changes here, but perhaps now more important... In the Excel export numbers are formatted as labels, so if I user want to compute using those values, they first need to convert the cell content. The TSV version does not have this issue.

devonbush:
Good point on the values all being strings in Excel.   IIRC, that's also how we handle it in Pepper export, and no users have complained yet -- I suspect either because their formulas can use strings just fine, or they're immediately exporting to a stats tool that's going to treat those as categorical values anyway. We'll see if that causes any issues for Shriie.

devonbush:
Moved some methods to static/protected, agree that will make the class more legible and easier to test

denniscunningham (approved):

MatthewBemis (approved):

PR #604, opened 2023-10-23
JN-654:  adding link branding color

devonbush:
#### DESCRIPTION

OurHealth participant site links were showing as blue instead their theme color.  This fixes that.  Adding @nawatts  as a reviewer, because this fix seems too obvious, and so I'm wondering if there's some complexity about the theme application that I'm missing, and why this wasn't already a style.

BEFORE
image

NOW:
image

1.  Go to https://sandbox.ourhealth.localhost:3001/
2. confirm footer and other links show up as the theme color

### DIFF
--- a/ui-participant/src/index.scss
+++ b/ui-participant/src/index.scss
@@ -54,7 +54,7 @@ code {
 
 a {
   text-decoration: none;
-  color: var(--brand-color);
+  color: var(--brand-color, blue);
 }
 
 .App .navbar-nav {
--- a/ui-participant/src/index.scss
+++ b/ui-participant/src/index.scss
@@ -54,6 +54,7 @@ code {
 
 a {
   text-decoration: none;
+  color: var(--brand-color);
 }
 
 .App .navbar-nav {


denniscunningham (approved):
Looks fine, but I lack useful domain knowledge here.

MatthewBemis (approved):
Looks good, but I agree with you- I'm always suspicious of the fixes that seem too easy

ui-participant/src/index.scss
```diff
@@ -54,6 +54,7 @@ code {
 
 a {
   text-decoration: none;
+  color: var(--brand-color);
```
nawatts: It might be worth adding a fallback here or scoping this `.App a` as the brand color variables are only defined within the App container.

https://github.com/broadinstitute/juniper/blob/fdb7d4267740b90906ea9e611b8d9f2b18b67359/ui-participant/src/App.tsx#L105-L108

Though practically, there might not be any links outside of the App container.

devonbush: that makes sense, thanks!

devonbush: adding .App breaks some button styling, so I added a fallback

PR #606, opened 2023-10-26
[JN-629], [JN-630] Add consent form and preEnroll buttons

devonbush:
#### DESCRIPTION

This enables adding consent forms and pre-enrollment surveys to studies.  out of scope for this PR are removing/deactivating them.  This is almost entirely modeled after the survey infrastructure, and so this also updates some of the survey UX logic to match our latest modal patterns. 
image

image

1. redeploy ApiAdminApp
2. populate HeartHive `./scripts/populate_portal.sh hearthive`
3. go to https://localhost:3000/hearthive/studies/pacing/env/sandbox/forms
4. try adding some consent forms.  Although it's not critical and out-of-scope to test, the participant flow should work with multiple consents (and it did when I quickly tested it) -- the participant shouldn't be able to access surveys until all consents are complete.
5. try adding a pre-enrollment survey

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConsentFormController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConsentFormController.java
@@ -39,4 +39,16 @@ public class ConsentFormController implements ConsentFormApi {
         consentFormExtService.createNewVersion(portalShortcode, consentForm, adminUser);
     return ResponseEntity.ok(savedConsent);
   }
+
+  @Override
+  public ResponseEntity<Object> create(String portalShortcode, String stableId, Object body) {
+    AdminUser adminUser = authUtilService.requireAdminUser(request);
+    ConsentForm consentForm = objectMapper.convertValue(body, ConsentForm.class);
+    if (!stableId.equals(consentForm.getStableId())) {
+      throw new IllegalArgumentException("form parameters don't match");
+    }
+    ConsentForm savedConsent =
+        consentFormExtService.create(portalShortcode, consentForm, adminUser);
+    return ResponseEntity.ok(savedConsent);
+  }
 }
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
@@ -86,7 +86,7 @@ public class ConsentFormExtService {
     }
     if (!studyEnv.getId().equals(updatedObj.getStudyEnvironmentId())) {
       throw new IllegalArgumentException(
-              "Study environment id in request body does belong to this study");
+          "Study environment id in request body does belong to this study");
     }
     return studyEnv;
   }
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
@@ -6,9 +6,10 @@ import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.consent.ConsentForm;
 import bio.terra.pearl.core.model.consent.StudyEnvironmentConsent;
 import bio.terra.pearl.core.model.portal.Portal;
+import bio.terra.pearl.core.model.study.StudyEnvironment;
 import bio.terra.pearl.core.service.consent.ConsentFormService;
-import bio.terra.pearl.core.service.exception.PermissionDeniedException;
 import bio.terra.pearl.core.service.study.StudyEnvironmentConsentService;
+import bio.terra.pearl.core.service.study.StudyEnvironmentService;
 import org.springframework.beans.BeanUtils;
 import org.springframework.stereotype.Service;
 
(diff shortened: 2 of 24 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
```diff
@@ -17,38 +18,76 @@ public class ConsentFormExtService {
...
+   * made in.
+   */
+  protected StudyEnvironment authConfiguredConsentRequest(
+      String portalShortcode,
+      EnvironmentName envName,
+      String studyShortcode,
+      StudyEnvironmentConsent updatedObj,
+      AdminUser user) {
+    authUtilService.authUserToStudy(user, portalShortcode, studyShortcode);
+    StudyEnvironment studyEnv = studyEnvironmentService.findByStudy(studyShortcode, envName).get();
```
denniscunningham: Will `studyEnv` always be present? Did we already validate the `envName`? Maybe a comment explaining why, or an `orElseThrow"?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
```diff
@@ -17,38 +18,76 @@ public class ConsentFormExtService {
...
+    if (!EnvironmentName.sandbox.equals(envName)
+        || !EnvironmentName.sandbox.equals(studyEnv.getEnvironmentName())) {
```
denniscunningham: When is `envName` different from `studyEnv.getEnvironmentName()`?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
```diff
@@ -17,38 +18,76 @@ public class ConsentFormExtService {
...
+      throw new IllegalArgumentException(
+          "Updates can only be made directly to the sandbox environment".formatted(envName));
```
denniscunningham: Did you intend to add a string replacement specifier to this string?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
```diff
@@ -17,38 +18,76 @@ public class ConsentFormExtService {
...
+    }
+    if (!studyEnv.getId().equals(updatedObj.getStudyEnvironmentId())) {
```
denniscunningham, line 89: What does `updatedObj` represent at this point? It feels like `updated` might not have much meaning in this context. Perhaps a java doc comment for this parameter would help demystify?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
```diff
@@ -173,4 +164,29 @@ public Survey authSurveyToPortal(Portal portal, String stableId, int version) {
     }
     return survey;
   }
+
+  /**
+   * confirms the user has access to the study and that the configured survey belongs to that study,
+   * and that it's in the sandbox environment. Returns the study environment for which the change is
+   * being made in.
+   */
+  protected StudyEnvironment authConfiguredSurveyRequest(
```
denniscunningham, line 173: How does this differ from `authConfiguredConsentRequest`?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConsentFormController.java
```diff
@@ -39,4 +39,16 @@ public ResponseEntity<Object> newVersion(String portalShortcode, String stableId
         consentFormExtService.createNewVersion(portalShortcode, consentForm, adminUser);
     return ResponseEntity.ok(savedConsent);
   }
+
+  @Override
+  public ResponseEntity<Object> create(String portalShortcode, String stableId, Object body) {
+    AdminUser adminUser = authUtilService.requireAdminUser(request);
+    ConsentForm consentForm = objectMapper.convertValue(body, ConsentForm.class);
+    if (!stableId.equals(consentForm.getStableId())) {
```
denniscunningham, line 47: When would this error occur? Is there anything the user can do about it?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
```diff
@@ -17,38 +18,76 @@ public class ConsentFormExtService {
...
     return studyEnvironmentConsentService.update(existing);
   }
+
+  public StudyEnvironmentConsent createConfiguredConsent(
+      String portalShortcode,
+      EnvironmentName envName,
+      String studyShortcode,
+      StudyEnvironmentConsent newObj,
+      AdminUser user) {
+    authConfiguredConsentRequest(portalShortcode, envName, studyShortcode, newObj, user);
```
denniscunningham, line 65: What does the term `configured` mean here and in related classes?

ui-admin/src/study/consents/CreateConsentModal.tsx
```diff
@@ -0,0 +1,96 @@
...
+    studyEnvContext: StudyEnvContextT
+    onDismiss: () => void }) {
+  const [isLoading, setIsLoading] = useState(false)
+  const { formName, formStableId, clearFields, nameInput, stableIdInput } = useFormCreationNameFields()
+
+  const portalContext = useContext(PortalContext) as PortalContextT
+  const navigate = useNavigate()
+
+
+  const attachConsentToEnv = async (createdConsent: ConsentForm) => {
```
denniscunningham, line 26: Wondering why these two operations are not combined into one on the backend?
It feels like even if there were instances where a consent form would be created and attached to different entities, the backend would handle that, perhaps through parameters that indicate what is desired?

denniscunningham (changes requested):
Not completely done with the review, but it would probably help my understanding to get this pass out of the way.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
devonbush: There's an invariant assumed throughout our application that a study will always have the sandbox, irb, and live environments.  so the validation of envName happens when it's converted to an EnvironmentName enum -- if it doesn't match one of those, the controller will throw IllegalArgument

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
devonbush: ah, good catch, they can't be different -- I think this is a leftover check from when the environment was fetched by id

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
devonbush: it's for surveys :).  I could see putting this into a common method somewhere, although our code generally tries to avoid sharing code between Survey and ConsentForm handling methods, despite their frequent similarities, so that it's very clear when ConsentForms are being manipulated, as that will likely come with extra legal/audit/workflow concerns over time

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
devonbush: "ConfiguredConsent" and "ConfiguredSurvey" are used in various places in the code as a more readable shorthand for "SutdyEnvironmentConsent" and "StudyEnvironmentSurvey", since they apply studyEnv-specific configuration to a survey/consent.  If I had to do it all over again, StudyEnvironmentConsent would probably just be called "ConfiguredConsent"

ui-admin/src/study/consents/CreateConsentModal.tsx
devonbush: that was how this API was originally structured: the backend took a "targetStudyEnvs" property.  But we want the form save to succeed  if at all possible, since that's saving user work.  And so it led to the need to pass complex  error status/message states for the backend to be able to communicate "the form save succeeded but the study environment wasn't updated".  It wasn't a big deal (all these errors are rare), and I think both approaches are totally valid, but IMO the current approach was slightly more readable on both the front and backend.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
denniscunningham: This code had the same errors as `authConfiguredConsentRequest` if that helps motivate combining them. :-)

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConsentFormController.java
devonbush: we would never expect (unless there's a bug) from someone using our UI.  It would only happen if someone is accessing the API directly, in which case they'd fix it by correcting the invalid parameter

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ConsentFormExtService.java
devonbush: Created https://broadworkbench.atlassian.net/browse/JN-677 for the rename

ui-admin/src/api/api.tsx
```diff
@@ -559,6 +570,19 @@ export default {
     return await this.processJsonResponse(response)
   },
 
+  async createConfiguredConsent(portalShortcode: string, studyShortcode: string, environmentName: string,
+    configuredConsent: StudyEnvironmentConsent): Promise<StudyEnvironmentSurvey> {
```
MatthewBemis: should this return a `StudyEnvironmentConsent`?

ui-admin/src/study/StudyContent.tsx
```diff
@@ -40,11 +45,36 @@ function StudyContent({ studyEnvContext }: {studyEnvContext: StudyEnvContextT})
...
-            </li></ul>}
+            { preEnrollSurvey && <ul className="list-unstyled">
+              <li className="d-flex align-items-center">
+                <Link to={`preEnroll/${preEnrollSurvey.stableId}?readOnly=${isReadOnlyEnv}`}>
+                  {preEnrollSurvey.name} <span className="detail">v{preEnrollSurvey.version}</span>
+                </Link>
+
+                { !isReadOnlyEnv && <div className="nav-item dropdown ms-1">
+                  <IconButton icon={faEllipsisH}  data-bs-toggle="dropdown"
+                    aria-expanded="false" aria-label="configure survey menu"/>
```
MatthewBemis:
```suggestion
                    aria-expanded="false" aria-label="configure pre-enroll menu"/>
```

ui-admin/src/study/StudyContent.tsx
```diff
@@ -109,30 +145,29 @@ function StudyContent({ studyEnvContext }: {studyEnvContext: StudyEnvContextT})
...
+              </li> }
             </ul>
           </div>
-
         </li>
       </ul> }
-      { <CreateSurveyModal studyEnvContext={studyEnvContext}
-        isReadOnlyEnv={isReadOnlyEnv}
-        show={showCreateSurveyModal}
-        setShow={setShowCreateSurveyModal}/> }
```
MatthewBemis, line 127: nice, thanks for updating these to be consistent

ui-admin/src/study/surveys/CreateSurveyModal.tsx
```diff
@@ -116,3 +87,37 @@ const CreateSurveyModal = ({ studyEnvContext, isReadOnlyEnv, show, setShow }: {s
 }
 
 export default CreateSurveyModal
+
+/** hook for a form that sets a name and stableId */
+export const useFormCreationNameFields = () => {
```
MatthewBemis, line 92: sweet! should this live in its own file?

MatthewBemis (approved):
Just a few small things, nothing blocking. Worked well when tested!

ui-admin/src/study/surveys/CreateSurveyModal.tsx
devonbush: I don't have a strong opinion about it.  do you have a leaning?

MatthewBemis: I mainly pointed it out because I've seen some hooks that have their own files, but others that don't. I lean towards having it in a separate file since it's sort of arbitrary that it lives in this particular Modal. Might also make it easier to know that these re-usable hooks exist

denniscunningham (approved):
I don't have additional feedback, so approved whenever you feel you have addressed everything you want to.

ui-admin/src/study/surveys/CreateSurveyModal.tsx
devonbush: cool, I'll keep it as-is in this PR, and then separate it out in the upcoming PR that uses it fro pre-reg surveys

PR #607, opened 2023-10-27
[JN-630] add prereg survey

devonbush:
#### DESCRIPTION

This adds a "Pre-registration" button on the Website editor to enable creation/editing of a pre-registration survey for the portal as a whole (as distinct from a pre-enrollment survey which only gates a particular study).  See our [new docs](https://broad-juniper.zendesk.com/hc/en-us/articles/19922803461915-How-do-I-create-a-Pre-registration-survey-) about it.  

image

This also adds a new zendeskUtils file so we can centrally manage our zendesk stuff.  Most importantly, this should make it easier for engineers to identify when they're working on an existing page with ZenDesk docs, so they'll know to keep the docs current

This needs tests, but I wanted to get this up now so you'd have plenty of time to review and we could get this out in the Monday release.  There are some very complex participant workflows around pre-registration surveys, and how they intersect with pre-enrollment, but those are outside the scope of this PR.

1. redeploy ApiAdminApp
2. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/siteContent
3. click the "Pre-registration" button
4. add a survey
5. confirm you can edit/save the survey
6. (out of scope of this PR) go to sandbox.ourhealth.localhost:3001.
7. Click "Join"
8. fill out the pre-enrollment survey (it's expected that you get the study survey first since the OurHealth 'join' link is to a specific study)
9. confirm you see your updated pre-registration survey afterwards

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
@@ -59,7 +59,7 @@ public class PortalExtService {
       AdminUser user) {
     if (!user.isSuperuser()) {
       throw new PermissionDeniedException(
-              "You do not have permissions to update portal configurations");
+          "You do not have permissions to update portal configurations");
     }
     authUtilService.authUserToPortal(user, portalShortcode);
     PortalEnvironment portalEnv = portalEnvironmentService.findOne(portalShortcode, envName).get();
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
@@ -12,7 +12,6 @@ import bio.terra.pearl.core.service.portal.PortalEnvironmentConfigService;
 import bio.terra.pearl.core.service.portal.PortalEnvironmentService;
 import bio.terra.pearl.core.service.portal.PortalService;
 import java.util.List;
-import java.util.Objects;
 import java.util.UUID;
 import org.springframework.beans.BeanUtils;
 import org.springframework.stereotype.Service;
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
@@ -24,9 +24,6 @@ public class PortalExtService {
   private PortalEnvironmentConfigService portalEnvironmentConfigService;
   private PortalAdminUserService portalAdminUserService;
   private AuthUtilService authUtilService;
-  private static final List<String> SUPERUSER_CONFIG_IGNORE_FIELDS = List.of();
-  private static final List<String> USER_CONFIG_IGNORE_FIELDS =
-      List.of("id", "createdAt", "participantHostname", "emailSourceAddress");
 
   public PortalExtService(
       PortalService portalService,
@@ -60,6 +57,10 @@ public class PortalExtService {
       EnvironmentName envName,
       PortalEnvironmentConfig newConfig,
       AdminUser user) {
+    if (!user.isSuperuser()) {
+      throw new PermissionDeniedException(
+              "You do not have permissions to update portal configurations");
+    }
     authUtilService.authUserToPortal(user, portalShortcode);
     PortalEnvironment portalEnv = portalEnvironmentService.findOne(portalShortcode, envName).get();
     PortalEnvironmentConfig config =
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
@@ -7,10 +7,12 @@ import bio.terra.pearl.core.model.portal.Portal;
 import bio.terra.pearl.core.model.portal.PortalEnvironment;
 import bio.terra.pearl.core.model.portal.PortalEnvironmentConfig;
 import bio.terra.pearl.core.service.admin.PortalAdminUserService;
+import bio.terra.pearl.core.service.exception.PermissionDeniedException;
 import bio.terra.pearl.core.service.portal.PortalEnvironmentConfigService;
 import bio.terra.pearl.core.service.portal.PortalEnvironmentService;
 import bio.terra.pearl.core.service.portal.PortalService;
 import java.util.List;
+import java.util.Objects;
 import java.util.UUID;
 import org.springframework.beans.BeanUtils;
 import org.springframework.stereotype.Service;
@@ -22,6 +24,9 @@ public class PortalExtService {
   private PortalEnvironmentConfigService portalEnvironmentConfigService;
   private PortalAdminUserService portalAdminUserService;
   private AuthUtilService authUtilService;
+  private static final List<String> SUPERUSER_CONFIG_IGNORE_FIELDS = List.of();
+  private static final List<String> USER_CONFIG_IGNORE_FIELDS =
+      List.of("id", "createdAt", "participantHostname", "emailSourceAddress");
 
   public PortalExtService(
       PortalService portalService,
@@ -55,10 +60,17 @@ public class PortalExtService {
       EnvironmentName envName,
       PortalEnvironmentConfig newConfig,
       AdminUser user) {
-    Portal portal = authUtilService.authUserToPortal(user, portalShortcode);
+    authUtilService.authUserToPortal(user, portalShortcode);
     PortalEnvironment portalEnv = portalEnvironmentService.findOne(portalShortcode, envName).get();
     PortalEnvironmentConfig config =
         portalEnvironmentConfigService.find(portalEnv.getPortalEnvironmentConfigId()).get();
+    if (!user.isSuperuser()) {
+      if (!Objects.equals(config.getParticipantHostname(), newConfig.getParticipantHostname())
+          || !Objects.equals(config.getEmailSourceAddress(), newConfig.getEmailSourceAddress())) {
+        throw new PermissionDeniedException(
+            "ParticipantHostname and emailSourceAddress can only be updated by superusers");
+      }
+    }
     BeanUtils.copyProperties(newConfig, config, "id", "createdAt");
     config = portalEnvironmentConfigService.update(config);
     return config;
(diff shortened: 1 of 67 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
```diff
@@ -56,10 +55,6 @@ public PortalEnvironmentConfig updateConfig(
       EnvironmentName envName,
       PortalEnvironmentConfig newConfig,
       AdminUser user) {
-    if (!user.isSuperuser()) {
-      throw new PermissionDeniedException(
-          "You do not have permissions to update portal configurations");
-    }
```
devonbush, line 62: this restriction was added before out of an abundance of caution.  But now we need to allow this so that HeartHive can make their site live/password protected on the day of their conference without our intervention.

ui-admin/src/study/surveys/PreRegView.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+    <SurveyEditorView
+      studyEnvContext={studyEnvContext}
+      currentForm={currentSurvey}
+      readOnly={readOnly}
+      onCancel={() => navigate(studyEnvSiteContentPath(portal.shortcode,
+        studyEnvContext.study.shortcode, portalEnv.environmentName))}
+      onSave={createNewVersion}
+    />
+  </div>
```
MatthewBemis:
```suggestion
  return <>
    <div className="d-flex align-items-center">
      <h2 className="h3 ms-2">Pre-registration survey </h2>
      <InfoPopup placement="bottom" content={<span>
        Pre-registration surveys are shown to participants before they create a user account on the portal.
        <ZendeskLink doc={DocsKey.PREREG_SURVEYS} className="ms-4">More documentation.</ZendeskLink>
      </span>}/>

    </div>

    <SurveyEditorView
      studyEnvContext={studyEnvContext}
      currentForm={currentSurvey}
      readOnly={readOnly}
      onCancel={() => navigate(studyEnvSiteContentPath(portal.shortcode,
        studyEnvContext.study.shortcode, portalEnv.environmentName))}
      onSave={createNewVersion}
    />
  </>
```

`SurveyEditorView` doesn't render on the screen for me unless I replace the outer `div` with fragment tags. Do you see the same behavior?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
```diff
@@ -60,28 +61,42 @@ public PortalEnvironmentConfig updateConfig(
...
     PortalEnvironment portalEnv = portalEnvironmentService.findOne(portalShortcode, envName).get();
     PortalEnvironmentConfig config =
         portalEnvironmentConfigService.find(portalEnv.getPortalEnvironmentConfigId()).get();
+    if (!user.isSuperuser()) {
+      if (!Objects.equals(config.getParticipantHostname(), newConfig.getParticipantHostname())
+          || !Objects.equals(config.getEmailSourceAddress(), newConfig.getEmailSourceAddress())) {
+        throw new PermissionDeniedException(
+            "ParticipantHostname and emailSourceAddress can only be updated by superusers");
+      }
+    }
```
MatthewBemis: This stuff seems unrelated to pre-reg, but I assume it's something you want in the upcoming release?

ui-admin/src/portal/CreatePreRegSurveyModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+    <Modal.Header closeButton>
+      <Modal.Title>Create Pre-registration Survey</Modal.Title>
+      <div className="ms-4">
+        {portalEnvContext.portal.name}: {portalEnvContext.portalEnv.environmentName}
+      </div>
+    </Modal.Header>
+    <Modal.Body>
+      <p>This survey will be shown to all participants wishing to create a user account on the portal.
+            This form is associated with the portal as a whole, not any particular study.
+      <ZendeskLink doc={DocsKey.PREREG_SURVEYS}>More info.</ZendeskLink>
```
MatthewBemis:
```suggestion
      <ZendeskLink className="px-1" doc={DocsKey.PREREG_SURVEYS}>More info.</ZendeskLink>
```
nit: spacing

MatthewBemis (commented):
I'm curious if you're also seeing the same rendering issue that I saw. That's really the only blocking thing, otherwise I don't think I'll have any additional comments before approving

ui-admin/src/study/surveys/PreRegView.tsx
devonbush: yup, great catch. fixed

api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
devonbush: ah yup, this belongs in a different PR -- removing

MatthewBemis (approved):

api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
```diff
@@ -60,7 +60,7 @@ public PortalEnvironmentConfig updateConfig(
       throw new PermissionDeniedException(
           "You do not have permissions to update portal configurations");
     }
-    Portal portal = authUtilService.authUserToPortal(user, portalShortcode);
+    authUtilService.authUserToPortal(user, portalShortcode);
```
denniscunningham, line 63: Not new to this PR, but it seems like if the two fetches below are invariants (and therefore we feel confident about the `get()`), it would be self-documenting and cleaner to have an `orElseThrow` instead of `get`. 
Unless the code just prior explicitly checks for presence, it is hard for a reader to understand expectations.

denniscunningham (changes requested):
It would be helpful to have a way to delete the pre-registration survey. (Sorry if I missed how to do this.)

Will users understand the difference between pre-registration and pre-enrollment? It's not clear that portal vs study is evident from the admin UX (but I could be missing something). Also, one of the confirmations when creating a pre-registration survey says "Environment updated"... Do users have any idea what an environment is? More broadly, is the idea to get this feature in now and do the UX later?

It seems odd that pre-registration is not listed on the "Forms and Surveys" page, despite the difference between portal and study. I know you asked for feedback last week, but I lacked the context to have an opinion.

I assume the current presentation on the participant side is temporary? (That is, when/where the pre-registration is presented to the ptp.)

api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
devonbush: Interesting -- to me a naked `get()` is what communicates that "if it fails, a deep invariant has been violated". And to me the code is more readable than adding a bunch of "orElseThrow(SomeInvariantViolatedException)". That said, I think your take is more correct as the codebase scales in usage and team size since it always identifies which invariant.  The question for us is that given our team size and turnover, and since those invariant code exceptions will rarely get exercised/tested, is adding them worth the maintenance costs (since those paths would have to be tested as otherwise they'd be a prime source of copypasta error.)

devonbush:
Responding to @denniscunningham 
```It would be helpful to have a way to delete the pre-registration survey. (Sorry if I missed how to do this.)```
Yup, just wasn't in the scope of this PR

```Will users understand the difference between pre-registration and pre-enrollment? It's not clear that portal vs study is evident from the admin UX (but I could be missing something). Also, one of the confirmations when creating a pre-registration survey says "Environment updated"... Do users have any idea what an environment is? More broadly, is the idea to get this feature in now and do the UX later?```
Probably not -- the UX links to documentation, but our documentation is still very much in progress.  I think we're going to need more than a few actual customers using this to get feedback on how to structure the documents and terms in a way that people understand.  In the meantime, my goal for the product is "make it as safe and easy as possible for users to preview their changes, so they don't have to understand from reading, they can just try stuff out and see if it does what they want". 

```It seems odd that pre-registration is not listed on the "Forms and Surveys" page, despite the difference between portal and study. I know you asked for feedback last week, but I lacked the context to have an opinion.```
Yup, this could go either way (and will probably end up on both)-- consensus from Jeff and Erin last week was to start with it on the SiteContent page, so we're starting here


```I assume the current presentation on the participant side is temporary? (That is, when/where the pre-registration is presented to the ptp.)```
Possibly?  at this point the participant UX of Preregistration is best thought of as "a prototype that HeartHive nodded approvingly at ~1 year ago, but may or may not reflect actual needs"

File: api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java (hunk 5 above)
Comment (denniscunningham) [reply api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java:63]: A naked `get()` to me is ambiguous: careless coding or an invariant?
I don't think I have a strong opinion about how to indicate an invariant (old school asserts did that), but I feel pretty strongly that a well understood way of expressing that is important.

**Comment by denniscunningham** [comment]
I seems like we do not have a good UX for distinguishing portal from study, but I feel like we will need one.
I'm not clear if you feel that we *do* have that UX, or if it is just not important at this time?

I don't see how users will understand the difference between pre-enroll and pre-registration without that, but that could just be a deficit in my understanding.

**Comment by devonbush** [comment]
If we're waiting to merge this until we have UX/documentation that gives us confidence that  users will understand the difference between portal and study, I suspect we'll merge this sometime next year.  Which isn't to say that it's not important to try, but given that the Study/Portal distinction is something that often gets confused *internally*, I think it's a tough road.  Can we treat https://broad-juniper.zendesk.com/hc/en-us/articles/19922803461915-How-do-I-create-a-Pre-registration-survey- as a small step forward in the process (it indicates that the portal is the website, and is distinct from a study), and continue iterating from there once this has made contact with a user or two?

File: api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java (hunk 5 above)
Comment (devonbush) [reply api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java:63]: Reading https://stackoverflow.com/questions/49158806/optional-get-versus-overloaded-optional-orelsethrow has confirmed my suspicion that there are a large number of folks like me who prefer `get()` 
```
in Java 10 we proposed to deprecate Optional.get(), but the public response to that was ... let's say cold.
```
but that your view is seen as the correct one.  So I'll give it a shot.  Created some new exception classes and switched to orElseThrow

devonbush:
oops, sorry Matt, didn't mean to re-request review from you

api-admin/src/main/java/bio/terra/pearl/api/admin/service/portal/PortalExtService.java
denniscunningham: First, thx for your thoughtful reply.
FWIW, although I prefer to construct an exception in `orElseThrow`, calling it without an argument also implies an assertion.
I still have PTSD from optional use in Pepper, so I may be overly sensitive to misuse.

denniscunningham:
@devonbush I honestly did not realize that creating UX to distinguish portal from study was a big lift. Not to slight the Juniper UX design, but the opposite... I guess I've grown accustomed to the magic. It just seems like the feature lacks context and thus may be troublesome for users. But of course it is easy enough to move around since it seems like the portal and study concepts are well modeled on the backend. 
In any case, just so we are on the same page, I see part of my review role as raising points that might not be top of mind for others. I definitely benefit the new eyes on reviews of my stuff, since it's easy to lose sight of the forest after spending a lot of time in a small set of trees. But most of the time this is just a discussion, and given the feedback, the implementer makes the call of when the code is ready.

denniscunningham (approved):

PR #608, opened 2023-10-27
[JN-80] adding video embedding support

devonbush:
#### DESCRIPTION

Adds the ability to embed youtube or vimeo videos on our site by supplying a "videoLink" in place of an image cleanFileName/version.  Vimeo is where HeartHive's video is currently hosted, and YouTube seemed like the other obvious provider.  The internet suggested that cross-site iframes are generally safe, although to the extent we take responsibility for content, we'd want to advise customers about the possibility of those sites tracking users.  
image

1.  Go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/siteContent
2. In the second section, replace the "image" attribute with the content below (or something similar).  Confirm the video appears
```
"image": {
    "videoLink": "https://player.vimeo.com/video/306028272?autoplay=true&muted=true&loop=false&byline=false&portrait=false&title=false"
  }
  ```

### CODE REVIEW

**Review by MatthewBemis — APPROVED** [review]
Didn't test this one out, but the code looks legit 👍

File: ui-core/src/participant/landing/ConfiguredMedia.tsx [hunk 1]
```diff
@@ -0,0 +1,88 @@
...
+  cleanFileName: string
+  version: number
+  alt?: string
+  className?: string
+  style?: CSSProperties
+}
+
+export type MediaConfig = ImageConfig | VideoConfig
+
+// TODO: Add JSDoc
```
denniscunningham, line 25: Dumb question, but why do we often leave this for later?

ui-core/src/participant/landing/ConfiguredMedia.tsx
```diff
@@ -0,0 +1,88 @@
...
+export default function ConfiguredMedia(props: ConfiguredImageProps) {
+  const { media, className, style } = props
+  const { getImageUrl } = useApiContext()
+  if ((media as VideoConfig).videoLink) {
+    const videoLinkMedia = media as VideoConfig
+    const videoAllowed = isVideoLinkAllowed(videoLinkMedia.videoLink)
+    return <div style={{ ...style, ...media.style }}
+      className={classNames('configured-image', className, media.className)}>
+      {videoAllowed && <iframe src={videoLinkMedia.videoLink} frameBorder="0" allowFullScreen={true}
+        data-testid="media-iframe"></iframe> }
```
denniscunningham, line 65: I noticed that depending on where I placed the image on the page, the full screen button was visible but unreachable. Could just be an issue with interactive content that users will work around, or perhaps something that could be helped with a larger border?

denniscunningham (approved):
Looks good. A couple of comments.

ui-core/src/participant/landing/ConfiguredMedia.tsx
devonbush: we made a change to our linting policy this summer -- we didn't use to require comments for const-declared functions.  Then when we decided we wanted to require those docs, we automatically added TODOs to the 300+ existing functions so that existing code would pass linting.  This is one of those (git is telling you this is a new file, but I just pasted it from the previous file).

ui-core/src/participant/landing/ConfiguredMedia.tsx
devonbush: yeah, that's something to keep an eye on.  In the meantime, users can add a className of "p-{x}" to add padding as they need

ui-core/src/participant/landing/ConfiguredMedia.tsx
devonbush: Normally I try to clean up the TODOs as I touch those old files, I just missed this one

PR #609, opened 2023-10-28
[JN-681]. syncing all kit statuses

devonbush:
#### DESCRIPTION

Previously, we only synched active kits.  Now we sync all kits.  IIRC Brian was worried synching all kits would send too many DB requests if we synched every kit every time.  We can address that concern later if the sync becomes too slow by implementing a bulk update.   This also updates the stubClient to progress the status of kits, so that you can see what kit statuses look like as they move through the application. 

1.  redeploy ApiAdminApp
2. repopulate ourhealth
3. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/requested`
4. confirm you see two kits, one created, one with issues
5. click the "Refresh" button.  
6. confirm the deactivated kit (previously under "issues"), is now under "created"

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
@@ -189,11 +189,11 @@ public class KitRequestService extends CrudService<KitRequest, KitRequestDao> {
 
         studyEnvironmentService.findByStudy(study.getShortcode(), environmentName).ifPresent(
                 studyEnvironment -> {
-                    var incompleteKits = findIncompleteKits(studyEnvironment.getId());
+                    var kits = dao.findByStudyEnvironment(studyEnvironment.getId());
 
                     // The set of kits returned from DSM may be different from the set of incomplete kits in Juniper, but
                     // we want to update the records in Juniper so those are the ones we want to iterate here.
-                    for (KitRequest kit : incompleteKits) {
+                    for (KitRequest kit : kits) {
                         var pepperKitStatus = pepperKitStatusByKitId.get(kit.getId().toString());
                         if (pepperKitStatus != null) {
                             saveKitStatus(kit, pepperKitStatus, pepperStatusFetchedAt);
@@ -201,8 +201,10 @@ public class KitRequestService extends CrudService<KitRequest, KitRequestDao> {
                     }
                 }
         );
+    }
 
-
+    public List<KitRequest> findByStudyEnvironment(UUID studyEnvironmentId) {
+        return dao.findByStudyEnvironment(studyEnvironmentId);
     }
 
     public List<KitRequest> findIncompleteKits(UUID studyEnvironmentId) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -250,28 +252,6 @@ protected String stringifyPepperAddress(PepperKitAddress kitAddress) {
         }
     }
```
devonbush, line 254: this was unused

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/StubPepperDSMClient.java
```diff
@@ -72,12 +77,26 @@ public PepperKitStatus fetchKitStatus(UUID kitRequestId) {
...
-                    .currentStatus("SENT")
+                    .currentStatus(getNextStatus(kit))
                     .build();
             return status;
         }).toList();
     }
+
+    /** helper to get the next status for a kit */
+    private String getNextStatus(KitRequest kit) {
+        List<String> statusVals = Arrays.stream(PepperKitStatus.Status.values()).map(status -> status.currentStatus).toList();
```
denniscunningham: Nit: might be nice note on the enum that we are taking advantage of the order, in case new statuses get added.

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/StubPepperDSMClient.java
```diff
@@ -8,14 +8,19 @@
...
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.commons.lang3.RandomUtils;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Component;
 
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
```
denniscunningham, line 23: I'm confused by the two Implementations of `PepperDSMClient`.
It seems like we only use `StubPepperDSMClient` for testing? But the code is much different from `LivePepperDSMClient`.
There is also `PepperDSMClientProvider` which I have not figured out yet.

Is there doc somewhere? I could not find anything on Confluence.

denniscunningham (changes requested):
I need some information on the use of `PepperDSMClient` implementations to complete my review.

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/StubPepperDSMClient.java
devonbush: PepperDSMClientProvider is essentially just a switcher configuration class that switches which client gets used based on the value of USE_LIVE_DSM (which comes from the container environment variables or spring config file)).  If true, the live client gets used, otherwise, the stub client gets used.

And yes, StubDSMClient is used for local testing where spinning up or connecting to Dev DSM isn't worth it.  It is currently too stubby for my liking.  Ideally, it would be more fully-functioning, and maybe using an internal cache of it's own.  This is why you're seeing it gradually accumulate semi-realistic behavior like in this PR, and gradually mimic more of Pepper's responses more precisely. It has a ways to go before you could develop against it with confidence for non-trivial integration work

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/StubPepperDSMClient.java
devonbush: I updated this to manually specify the order in the Stub, I think that makes it more evident what's going on

MatthewBemis (approved):

denniscunningham (approved):
Thank you for the explanation!

PR #612, opened 2023-10-30
[JN-683] Section editor now allows typing invalid JSON

MatthewBemis:
#### DESCRIPTION

I'm not 100% thrilled with how this turned out, but it was the least invasive way of allowing for invalid JSON to be typed during section editing. Ideally we'll have a UI slapped on top of this textbox so hopefully this won't live forever. I tried a couple of other approaches that followed a cleaner hierarchy of which component should own the validation, but they all ended up being pretty gnarly when it came down to actual implementation.

The implementation of this final approach disables a lot of functionality until a user has corrected their invalid JSON (add page, insert section, move section, etc). Because newly-inserted sections don't yet have IDs and section indexes can easily change, it was pretty difficult to accurately keep track of which sections were actually invalid after moving them around or changing pages. Disabling that functionality adds enough guardrails where it becomes much easier to manage.

https://github.com/broadinstitute/juniper/assets/7257391/4c555c52-97fa-40ea-ae58-30771308d6ab

* Launch admin UI + API
* Edit a section and type in some invalid JSON. Ensure that there is no possible way to save the page with that invalid section
* Ensure that section previews are still working
* Make sure that no functionality is inadvertently disabled even when JSON is valid

### DIFF
--- a/ui-admin/src/portal/siteContent/HtmlPageEditView.tsx
+++ b/ui-admin/src/portal/siteContent/HtmlPageEditView.tsx
@@ -9,11 +9,13 @@ import { sectionTemplates } from './sectionTemplates'
 type HtmlPageViewProps = {
   htmlPage: HtmlPage
   readOnly: boolean
+  siteInvalid: boolean
+  setSiteInvalid: (invalid: boolean) => void
   updatePage: (page: HtmlPage) => void
 }
 
 /** Enables editing of a given page, showing the config and a preview for each section */
-const HtmlPageView = ({ htmlPage, updatePage, readOnly }: HtmlPageViewProps) => {
+const HtmlPageView = ({ htmlPage, readOnly, siteInvalid, setSiteInvalid, updatePage }: HtmlPageViewProps) => {
   const DEFAULT_SECTION_TYPE = {
     id: '',
     sectionType: 'HERO_WITH_IMAGE' as SectionType,
@@ -36,7 +38,7 @@ const HtmlPageView = ({ htmlPage, updatePage, readOnly }: HtmlPageViewProps) => 
       <Button variant="secondary"
         aria-label={'Insert a blank section'}
         tooltip={'Insert a blank section'}
-        disabled={readOnly}
+        disabled={readOnly || siteInvalid}
         onClick={() => insertNewSection(sectionIndex, DEFAULT_SECTION_TYPE)}>
         <FontAwesomeIcon icon={faPlus}/> Insert section
       </Button>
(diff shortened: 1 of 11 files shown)


ui-admin/src/portal/siteContent/HtmlPageEditView.tsx
```diff
@@ -9,11 +9,13 @@ import { sectionTemplates } from './sectionTemplates'
...
   htmlPage: HtmlPage
   readOnly: boolean
+  siteInvalid: boolean
+  setSiteInvalid: (invalid: boolean) => void
   updatePage: (page: HtmlPage) => void
 }
 
 /** Enables editing of a given page, showing the config and a preview for each section */
-const HtmlPageView = ({ htmlPage, updatePage, readOnly }: HtmlPageViewProps) => {
+const HtmlPageView = ({ htmlPage, readOnly, siteInvalid, setSiteInvalid, updatePage }: HtmlPageViewProps) => {
```
devonbush: hmm.. maybe   `hasInvalidContentFormat` or `hasInvalidSection` as a better name for `siteInvalid`?

ui-admin/src/portal/siteContent/HtmlSectionEditor.tsx
```diff
@@ -27,26 +28,30 @@ const HtmlSectionEditor = ({
...
-      JSON.parse(updatedSection.sectionConfig ?? '{}')
-    } catch (e) {
-      // for now, we just don't allow changing the object structure itself -- just plain text edits
-      return
-    }
+  const [editorValue, setEditorValue] = useState(() =>
+    JSON.stringify(JSON.parse(section?.sectionConfig ?? '{}'), null, 2))
 
+  useEffect(() => {
+    setEditorValue(JSON.stringify(JSON.parse(section?.sectionConfig ?? '{}'), null, 2))
```
devonbush, line 52: I inadvertently started testing an earlier version of this PR where this section had a _setEditorValue and then a useCallback.  I'm glad you took the extra time to clean this up!

devonbush (approved):
seemed great with local testing, this is a big step forward for usability

denniscunningham (approved):
Looks good.
One situation to consider... If a user messes up their content they may end up stuck. Perhaps they cannot find the error in content they've been working on for a while. Perhaps they selected something large and inadvertently deleted it. If they cannot fix the  problem, their only choice is to cancel and lose their work. Undo would be helpful, but does not cover all cases. (I realize they can also lose their work by navigating away from the page.)
I'm not sure that's a big enough concern to reconsider this PR, but wanted to be sure we had those cases in mind when evaluating this.

MatthewBemis:
@denniscunningham that's a good point, I don't think it would be too hard to implement a button to reset the section to the last valid entry. The survey builder has similar issues, we could take a similar approach for both.

And re: "they can also lose their work by navigating away from the page", another improvement I'd like to make is to add a warning when the user is about to navigate away from the page with unsaved content. The survey editor auto-saves to browser local storage every few seconds, but I'm not sold on taking that route again.

PR #613, opened 2023-10-30
[JN-624] Add support for multiple B2C tenants

denniscunningham:
#### DESCRIPTION
This PR combines with Juniper deployment changes (https://github.com/broadinstitute/terraform-ap-deployments/pull/1240) to enable a map of tenant/portal to B2C parameters, enabling separate participant app B2C configuration for each Juniper portal. (See [Juniper B2C tenant parameter storage](https://docs.google.com/document/d/1fbvsiWlwjOTkifCVFTrfZAog7yrHz3hOxy-xdp521ak/edit) for a detailed discussion.) 

It replaces the existing B2C configuration technique, which used a set of environment variables matching the B2C parameters for one portal, with a file containing a per-portal map of B2C parameters.  In deployed environments a Kubernetes ConfigMap is manifest as a file mounted in the Juniper participant container. In the local environment the configuration is in a resource file. The latter is temporary until we find a better home for the configuration.

I considered a handful of local configuration choices before deciding on this implementation, so I am happy to discuss alternatives and tradeoffs. Ideally we decide on a permanent solution for local B2C configuration soon, and evaluate this one on its short term merits and faults.

Note: There are a small set of changes that I will put in a PR for `terraform-ap-deployments`, so this PR will not be merged until that work is complete. Also, I plan to carefully deploy this to the dev/demo environment, since there is not a practical way to test the deployed behavior of this new technique.

Run `bash ./local-dev/render_environment_vars.sh ApiParticipantApp <YOUR_EMAIL>` to regenerate a new set of environment variables.
After starting the participant app, go to sandbox.ourhealth.localhost:3001.
Choose "Login" and confirm that you see the OurHealth login dialog.
image

Choose "Join", complete the pre-enroll survey, get and enter the verification code, and confirm that you see the OurHealth "Create Account" dialog.
image


api-participant/build.gradle
```diff
@@ -3,7 +3,7 @@ plugins {
     id 'de.undercouch.download'
     id 'com.google.cloud.tools.jib'
     id 'com.srcclr.gradle'
-
+    id "io.freefair.lombok" version "6.5.1"
```
denniscunningham, line 6: I noticed this is causing a conflict (here and in admin app), which appears to impact testing parallelism, so I can address that in a separate PR.

api-participant/build.gradle
```diff
@@ -37,11 +37,13 @@ dependencies {
             'io.jsonwebtoken:jjwt-jackson:0.11.5' // or 'io.jsonwebtoken:jjwt-gson:0.11.5' for gson
 
     testImplementation 'org.junit.jupiter:junit-jupiter-api'
+    testImplementation 'org.junit.jupiter:junit-jupiter-params'
     testImplementation('org.springframework.boot:spring-boot-starter-test:2.7.10') {
         // Fixes warning about multiple occurrences of JSONObject on the classpath
         exclude group: 'com.vaadin.external.google', module: 'android-json'
     }
     testImplementation 'org.mockito:mockito-inline'
+    testImplementation 'commons-io:commons-io:2.13.0'
```
denniscunningham, line 45: For temp file support.

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
```diff
@@ -0,0 +1,99 @@
...
+  public B2CConfigurationService() {
+    b2cConfiguration = new B2CConfiguration();
+  }
+
+  /**
+   * Get the B2C configuration for a portal. Returns an empty map if portal config for portal is not
+   * specified
+   */
+  public Map<String, String> getB2CForPortal(String portalShortcode) {
+    // see if already constructed
```
devonbush:
```suggestion
    /** if it's already constructed use that to save the expense of recomputing.  Otherwise, try rereading the config in case new values have been propagated */
```
Is the reason we're not just reading the config into the string maps all at once at startup to better pick up dynamic changes?

api-participant/src/main/resources/application.yml
```diff
@@ -15,11 +15,7 @@ env:
   swagger:
     # if true, the swagger UI page will be made available at swagger-ui.html -- should be false for production
     enabled: ${SWAGGER_ENABLED:false}
-  b2c:
-    tenantName: ${B2C_TENANT_NAME:missing_tenant_name}
```
devonbush, line 19: nice, I like how this file gets shorter

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.stereotype.Service;
+import org.yaml.snakeyaml.LoaderOptions;
+import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.Constructor;
+
```
devonbush, line 19:
```suggestion
/** this is implemented as a service, rather than static config so that it may eventually support dynamic updates to the B2C configs */
```

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
```diff
@@ -0,0 +1,99 @@
...
+   */
+  public void initB2CConfig(String b2cConfigFile) {
+    if (StringUtils.isBlank(b2cConfigFile)) {
+      log.error("b2c-config-file property is not set");
+      return;
+    }
+
+    Yaml yaml = new Yaml(new Constructor(B2CConfiguration.class, new LoaderOptions()));
+
+    File file = new File(b2cConfigFile);
```
devonbush, line 54:
```suggestion
    File file = new File(b2cConfigFile);
    /** an absolute file path will be used in deployed environments as the file will be placed by terraform.  In local dev, we read the config from source */
```

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
```diff
@@ -0,0 +1,99 @@
...
+      try (InputStream str = new FileInputStream(file)) {
+        b2cConfiguration = yaml.load(str);
+      } catch (Exception e) {
+        log.error("Error loading b2c config file: {}", b2cConfigFile, e);
+      }
+    } else {
+      // relative path, load from classpath
+      try (InputStream str =
+          Thread.currentThread().getContextClassLoader().getResourceAsStream(b2cConfigFile)) {
+        b2cConfiguration = yaml.load(str);
```
devonbush: FWIW the pattern we've used elsewhere to read from the classpath is
```
ClassPathResource cpr = new ClassPathResource(relativePath);
cpr.getInputStream();
  ```
in `FilePopulateService`. Any advantages/disadvantages to pulling it from the currentThread?

File: api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java [hunk 8]
```diff
@@ -0,0 +1,99 @@
...
+    Map<String, String> config =
+        Map.of(
+            "b2cTenantName",
+            StringUtils.defaultIfEmpty(b2cConfig.getTenantName(), ""),
+            "b2cClientId",
+            StringUtils.defaultIfEmpty(b2cConfig.getClientId(), ""),
+            "b2cPolicyName",
+            StringUtils.defaultIfEmpty(b2cConfig.getPolicyName(), ""),
+            "b2cChangePasswordPolicyName",
+            StringUtils.defaultIfEmpty(b2cConfig.getChangePasswordPolicyName(), ""));
```
devonbush, line 105: this might be more trouble than it's worth, but you could enforce the convention of the `b2c` prefix and make the code a bit more typo-proof with something like:
```suggestion
    Map<String, String> config = new HashMap<>();
    List.of("TenantName", "ClientId", ...).stream().forEach(propName -> {
       config.put(B2C_PREFIX + propName, StringUtils.defaultIfEmpty(PropertyUtils.getProperty(b2cConfig, propName)));
       });
```

devonbush (approved):
Looks good, and worked when I tested locally.  A few suggestions for comments (half because I think the comment would be helpful and half to confirm my own understanding), and some code suggestions, but nothing blocking

devonbush:
btw, I assumed this was draft just because it was waiting on the corresponding terraform PR, and so wasn't mergable yet.  hope I didn't jump the gun on review

denniscunningham:
@devonbush Thx for the early look. I actually had to do some restructuring due to Spring application context madness during test. If anything, it made the implementation more in line with what we do elsewhere. I wandered off to enable some stuff, then realized I could come back to something more conventional.

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
denniscunningham: It mostly a service because we have to read a file, which is not standard config AFAIK. It also moves some logic out of the controller that should probably be an a service. Also, if B2C config gets more interesting in the future we will want a service (for instance, a better way to read local B2C config).
But I might be missing your point?

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
denniscunningham: Lazy initialization, for no particular reason. Can change if it makes things clearer.
Not sure what you mean about re-reading the config. What would tigger that?

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
denniscunningham: With this implementation we can also support a local file at an absolute path, if that becomes our longer term plan.

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
denniscunningham: Sure. I can make that change if it makes things clearer. I was not aware of PropertyUtils.

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
devonbush: Yeah, I expected the map to be read all at once, and so when I saw this I assumed it was re-reading the config file every time in case it had been updated. (now I see that's not the case).  I might lean towards reading it all at once because it will pick up any config parsing errors immediately, as opposed to waiting until the first request comes in.  But I could see it either way

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
devonbush: yeah, Spring reading files for configuration is standard, and I think you're right that since this isn't just reading Spring-standard config files from spring-standard places it deserves a service

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
devonbush: Yeah, to me, PropertyUtils falls under the "use sparingly" list of tools, since it throws just about any kind of type safety out the window.  So this was definitely a "this might make things worse, but it might be worth a try" sort of suggestion

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
denniscunningham: Just curious, for "Spring reading files for configuration is standard" do you mean properties files or something else? One option I explored (and coded) for this was reading this as standard config via application.yml, but that config could have been pulled into that file from elsewhere. 
Is that what you mean, or something else?

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
denniscunningham: That makes sense. Will change.

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
denniscunningham: I can leave this as is or implement any improvement you suggest. It feels like boilerplate, and thus can be improved, but we will probably not add much to this list of properties in the future.

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
denniscunningham: Not that I am aware of. Will change.

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
devonbush: yup, that's what I meant--just reading properties from application.yaml would be better fit for a Configuration bean, but you're doing more here, because you have to support different ways of getting the file in

api-participant/src/main/java/bio/terra/pearl/api/participant/config/B2CConfigurationService.java
devonbush: no strong opinion

devonbush (approved):
looks good, thanks for the explanations

MatthewBemis (approved):

PR #614, opened 2023-10-31
[JN-675] Site footer self service

MatthewBemis:
#### DESCRIPTION

Adds the ability to add, remove, & edit site footers from any page. And while you can edit it from any page, it's still global (although you could implement a different site footer per-page if you really wanted to)

https://github.com/broadinstitute/juniper/assets/7257391/7e3adce3-9e42-4323-a38d-5ec089d428b3

* Launch admin tool
* Edit and save the footer section from any page: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/siteContent
* Delete the footer section and save, confirm it disappears
* Try adding a footer back to the site
* Make sure participant site actually works too


core/src/main/java/bio/terra/pearl/core/service/site/LocalizedSiteContentService.java
```diff
@@ -52,6 +52,8 @@ public LocalizedSiteContent create(LocalizedSiteContent localSite) {
             dao.setLandingPageId(savedSite.getId(), landingPage.getId());
         }
         savedSite.setLandingPage(landingPage);
+        localSite.getFooterSection().setId(savedSite.getFooterSectionId());
+        savedSite.setFooterSection(localSite.getFooterSection());
```
MatthewBemis, line 56: These were not attached to the SiteContent response when a new version was created (so the site footer would disappear from the editor after you saved)

ui-admin/src/portal/siteContent/HtmlPageEditView.tsx
```diff
@@ -33,6 +37,51 @@ const HtmlPageView = ({ htmlPage, readOnly, siteInvalid, setSiteInvalid, updateP
     updatePage(htmlPage)
   }
 
+  const removeSection = (sectionIndex: number) => () => {
```
MatthewBemis, line 40: I should have put these in the page editor to begin with. Sections don't need to care about their index or the overall structure of the page that they belong to. This also made it a lot cleaner to deal with footers which are sections that don't live within an `HtmlPage`.

core/src/main/java/bio/terra/pearl/core/service/site/LocalizedSiteContentService.java
```diff
@@ -52,6 +52,7 @@ public LocalizedSiteContent create(LocalizedSiteContent localSite) {
             dao.setLandingPageId(savedSite.getId(), landingPage.getId());
         }
         savedSite.setLandingPage(landingPage);
+        savedSite.setFooterSection(localSite.getFooterSection());
```
MatthewBemis, line 56: This was not attached to the SiteContent response when a new version was created (so the site footer would disappear from the editor after you saved)

ui-admin/src/portal/siteContent/HtmlPageEditView.tsx
```diff
@@ -11,11 +11,15 @@ type HtmlPageViewProps = {
   readOnly: boolean
   siteInvalid: boolean
   setSiteInvalid: (invalid: boolean) => void
+  footerSection: HtmlSection | undefined
```
devonbush:
```suggestion
  footerSection?: HtmlSection
```

ui-admin/src/portal/siteContent/HtmlPageEditView.tsx
```diff
@@ -11,11 +11,15 @@ type HtmlPageViewProps = {
...
+  updateFooter: (section: HtmlSection | undefined) => void
```
devonbush:
```suggestion
  updateFooter: (section?: HtmlSection) => void
```

devonbush: I think this is what Nick said is the preferred style

ui-admin/src/portal/siteContent/HtmlPageEditView.tsx
```diff
@@ -49,16 +98,41 @@ const HtmlPageView = ({ htmlPage, readOnly, siteInvalid, setSiteInvalid, updateP
...
+            moveSection={moveSection(index)} removeSection={removeSection(index)} allowTypeChange={section.id === ''}
+            siteInvalid={siteInvalid} section={section} readOnly={readOnly}/>
         </div>
         <div className="col-md-8">
           <HtmlSectionView section={section}/>
         </div>
         { renderAddSectionButton(index + 1) }
       </div>
     })}
+    { footerSection && <div key={`${footerSection.id}`} className="row g-0">
```
devonbush:
```suggestion
    { footerSection && <div className="row g-0">
```
I don't think we need a key on this div since it's not being rendered as part of an array

core/src/main/java/bio/terra/pearl/core/service/site/LocalizedSiteContentService.java
devonbush: I'm still seeing this behavior -- after I save, the footer disappears

ui-admin/src/portal/siteContent/HtmlPageEditView.tsx
```diff
@@ -49,16 +98,41 @@ const HtmlPageView = ({ htmlPage, readOnly, siteInvalid, setSiteInvalid, updateP
     { renderAddSectionButton(0) }
```
devonbush, line 98: we could really use some extra padding at the bottom of the page to make the footer more accessible.  I'd suggest adding a `pb-4` or `pb-5` to the containing div on line 48

devonbush (changes requested):
Code looks good, and appreciate the cleanup of the moveSection code, and extracting Footer into its own component.  A couple of nice-to-have suggestions.  The blocker for me is that I'm still seeing the footer-disappears-after-saving bug you mention.  the save succeeds with my changes, but the footer disappears from the UI.  So I have to reload the page to see it

core/src/main/java/bio/terra/pearl/core/service/site/LocalizedSiteContentService.java
devonbush: For good hygiene, I would add `localSite.setFooterSection(footer);` below line 37 above.  that way the attached object will have all the correct ids from the database, in case we ever decide to use them

devonbush (approved):
previous behavior was me forgetting to restart apiAdminApp.  works great now

denniscunningham (approved):
I'm seeing the footer text change to fuchsia-red on the ptp page. I'm pretty sure it was blue when first loaded, but on reload it stays fuchsia. 

image

MatthewBemis:
> I'm seeing the footer text change to fuchsia-red on the ptp page. I'm pretty sure it was blue when first loaded, but on reload it stays fuchsia.
> 
> image

@denniscunningham that's a good catch, it's related to this: https://github.com/broadinstitute/juniper/pull/604 I don't think the admin UI is able to get the correct branding CSS. I'll add a ticket to look into that, would be nice to get that working

PR #615, opened 2023-11-01
[JN-665] Adding asset fingerprint fallback to participant site

devonbush:
#### DESCRIPTION

We had a problem where rolling deploys was causing the site to appear unavailable as requests for js assets 404'ed due to the request being served by a pod with a different version of the fingerprinted asset. Investigation into devOps-side ways of solving this problem (blue/green deployments, sticky load balancer pods, static asset hosting, etc..) suggested they would all involve more time and moving parts than we are comfortable spending right now, and likely introduce more risk than we want given our recent and upcoming launches.  See https://broadinstitute.slack.com/archives/CADM7MZ35/p1698759809580049 for some discussion

 So as a temporary band-aid, this updates our application servers to fallback to serving the js/css assets they have, even if the fingerprint doesn't match.  

This bandaid is reasonably safe for us now, though, since main.js is typically the only asset that changes between versions -- the majority of our css is inlined, and our JS chunks are for things like the privacy policy that are rarely used/updated. And the fingerprints are still included in index.html, so the fingerprints still do their job of preventing unwanted browser caching.  And it's certainly better than the alternative of the site always being down for 5-10mins during deploys.

My current thinking is that this change should get us at least  through the holidays, and then in Jan/Feb we can look at more robust devOps solutions. 

Also, I only included this band-aid on the participant side, since participant site availability is way more important than admin site availability.  

1.  Clean out any existing built assets: from the repo root, run `./gradlew api-participant:clean`
2. Run the task to build and place all the assets for the participant server:  from the repo root , run `./gradlew api-participant:jibDockerBuild`
3. Restart your ApiParticipantApp (either intelliJ or `./gradlew api-participant:bootRun`)
4. in your browser, go to http://localhost:8081/static/js/main.11111111.js
5. confirm you get Javascript served to you, despite that asset name not existing in the built assets included in `api-participant/build/resources/main/static/static/js`

### DIFF
--- a/api-participant/build.gradle
+++ b/api-participant/build.gradle
@@ -47,15 +47,42 @@ dependencies {
 }
 
 task copyWebApp(type: Copy) {
+    dependsOn(rootProject.bundleParticipantUI)
     from "$rootDir/ui-participant/build"
     into "$rootDir/api-participant/build/resources/main/static"
 }
 
-// for now, only jib depends on copyWebApp, so that a npm rebuild/install will not be triggered for
-// development redeploys.  this means to deploy locally with the static assets in place you'll
-// need to run the copyWebApp task yourself
-copyWebApp.dependsOn(rootProject.bundleParticipantUI)
-jibDockerBuild.dependsOn('copyWebApp')
+// creates copies of the fingerprinted js files without the asset fingerprint.
+task createUnfingerprintedJs(type: Copy) {
+    dependsOn('copyWebApp')
+    dependsOn('processResources')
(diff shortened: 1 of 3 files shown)


denniscunningham:
@devonbush You have a tantalizing "See " with no link in your first paragraph of the PR description.

denniscunningham (approved):

devonbush:
sorry, meant to link the slack thread, but then got distracted looking at Slack.  Now added

MatthewBemis (approved):

PR #619, opened 2023-11-03
[JN-693] Allow study admins to mark surveys as required or optional

MatthewBemis:
#### DESCRIPTION

This allows study admins to set surveys as optional or required during creation or from the study content view.

https://github.com/broadinstitute/juniper/assets/7257391/1a8482f9-e0c9-464f-88fb-96723e6c6969

* Go to the form view: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms
* Toggle an existing survey as required/optional
* Verify that you can create a new required survey

### DIFF
--- a/ui-admin/src/study/StudyContent.tsx
+++ b/ui-admin/src/study/StudyContent.tsx
@@ -130,7 +130,7 @@ function StudyContent({ studyEnvContext }: {studyEnvContext: StudyEnvContextT}) 
                         <Link to={`surveys/${survey.stableId}?readOnly=${isReadOnlyEnv}`}>
                           {survey.name}
                           <span className="mx-1 detail">v{survey.version}</span>
-                          <span>{surveyConfig.required && <span className="detail">(required)</span>}</span>
+                          {surveyConfig.required && <span className="detail">(required)</span>}
                         </Link>
                       </div>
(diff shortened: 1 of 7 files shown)


ui-admin/src/study/StudyContent.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useContext, useState } from 'react'
```
MatthewBemis, line 1: You'll absolutely want to review this file with whitespace hidden: https://github.com/broadinstitute/juniper/pull/619/files?w=1

denniscunningham (approved):
Looks good.
Noticed a re-render of the full page when toggling required/optional on ui-admin/src/study/StudyContent.tsx. Perhaps unavoidable since the text for the changed item is updated, but surprising.

devonbush:
Yeah, the "reloadPortal" call reloads the portal 'god object' which is going to force a rerender of pretty much everything.  It's outside the scope of this PR, but as we're working in the admin tool, we need to be slowly reducing the dependence on that 'god object' and having each subpage load its own content.  That will allow for smarter refresh behavior.

devonbush (approved):
looks good, worked locally

PR #622, opened 2023-11-07
[JN-657] Fix error handling for login of participants not in portal

denniscunningham:
#### DESCRIPTION
This PR addresses a couple of issues:

1. Improves error handling and reporting when a participant with a valid token does not have an account with a given portal.
2. Starts the process of accurately mapping exceptions to the appropriate response code.

(The ticket mentioned here (JN-657) is about the latter. I can open a separate ticket for the former, or associate one of the B2C tickets with that work.)

The scenario for (1) is hopefully rare with real participants in prod, since a participants need a valid B2C account to get to Juniper login processing unless something has gone wrong with the study joining process. However, I stumbled on it since dev/demo and local environments share B2C directories (two DBs, one directory). In any case, the backend was not returning 403 for users who did not have portal accounts, and the frontend lacked proper error handling for any Juniper login failure. 

For (2), the GlobalExceptionHandler for admin and participant relied on a Terra library for mapping exceptions to response status, and since we want more precise control over that mapping, we no longer rely on that library. Also, we want to use error level logging carefully since it will be an alert trigger. Thus internal errors and exceptions should have an associated error log statement so the team can address them ASAP. Bad requests, etc., should not log at error level.

(Note that the code in each GlobalExceptionHandler is currently identical. It may diverge in the future, so I did not bother experimenting with @ExceptionHandler behavior in a shared library. But I can consolidate any code that makes sense to share, which is at least `buildErrorReport`. I did not want to create a new shared package for that small piece of code without getting feedback from the team.)

1. Ensure OurHealth is populated.
2. Join OurHealth on dev/demo as a new user, completing the process with a new account with a password.
3. In your local environment navigate to OurHealth participant portal (https://sandbox.ourhealth.localhost:3001).
4. Login using the account created in step 2.
5. You should see an error modal, and 403 response code. You will be returned to the login dialog.


api-participant/src/main/java/bio/terra/pearl/api/participant/service/CurrentUnauthedUserService.java
```diff
@@ -24,21 +27,22 @@ public CurrentUnauthedUserService(
...
-  public Optional<CurrentUserService.UserWithEnrollees> unauthedLogin(
+  public CurrentUserService.UserWithEnrollees unauthedLogin(
       String username, String portalShortcode, EnvironmentName environmentName) {
     Optional<ParticipantUser> userOpt = participantUserDao.findOne(username, environmentName);
-    if (userOpt.isPresent()) {
-      ParticipantUser user = userOpt.get();
-      user = updateUnauthedUserToken(user);
-      return Optional.of(currentUserService.loadFromUser(user, portalShortcode));
+    if (userOpt.isEmpty()) {
+      log.info("User {} not found for environment {}", username, environmentName);
```
denniscunningham: This logging is helpful since we otherwise do not record this info (and the detail is not exposed in the front end), but note that we are exposing the participant username. Just want to be sure we are deliberate about our PII boundaries, since there will be other cases where info like username will be important for troubleshooting.

devonbush: Yeah, our current marching orders from AppSec are to not log usernames, especially participant usernames.  You're right though, that we don't have any other identifiers at this point.  Maybe we obfuscate the email by just logging first letter + domain?  So e.g. dbush@broadinstitute.org would be `d...@broadinstitute.org`?

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/registration/RegistrationController.java
```diff
@@ -42,11 +41,13 @@ public RegistrationController(
   public ResponseEntity<Object> register(
       String portalShortcode, String envName, UUID preRegResponseId, RegistrationInfo body) {
     var environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+    // TODO: seems like user should be authorized (tokenLogin) before registering -DC
```
devonbush: thanks for calling this out.  there are some registration flows where the registration process is what creates the user, so we can't require the user (See lines 101-107 of RegistrationService).  What we should do though, is move the `requireToken` check below to the first line of this method, and then use the username from the token for registration, rather than the username from the body

api-participant/src/main/java/bio/terra/pearl/api/participant/service/CurrentUserService.java
```diff
@@ -38,46 +42,46 @@ public CurrentUserService(
...
-  public Optional<UserWithEnrollees> tokenLogin(
+  public UserWithEnrollees tokenLogin(
       String token, String portalShortcode, EnvironmentName environmentName) {
-    Optional<UserWithEnrollees> userOpt = loadByToken(token, portalShortcode, environmentName);
-    userOpt.ifPresent(
-        userAndEnrolles -> {
-          userAndEnrolles.user.setLastLogin(Instant.now());
-          participantUserDao.update(userAndEnrolles.user);
-        });
+    UserWithEnrollees userOpt = loadByToken(token, portalShortcode, environmentName);
```
devonbush:
```suggestion
    UserWithEnrollees user = loadByToken(token, portalShortcode, environmentName);
```
No need for the Opt suffix since now this is the object proper

api-participant/src/main/java/bio/terra/pearl/api/participant/service/CurrentUserService.java
```diff
@@ -38,46 +42,46 @@ public CurrentUserService(
...
+  public UserWithEnrollees refresh(
       String token, String portalShortcode, EnvironmentName environmentName) {
     return loadByToken(token, portalShortcode, environmentName);
   }
 
-  protected Optional<UserWithEnrollees> loadByToken(
+  protected UserWithEnrollees loadByToken(
       String token, String portalShortcode, EnvironmentName environmentName) {
-    var decodedJWT = JWT.decode(token);
+    DecodedJWT decodedJWT = JWT.decode(token);
```
devonbush, line 65: thanks for removing those `var`s

api-participant/src/main/java/bio/terra/pearl/api/participant/service/CurrentUserService.java
```diff
@@ -38,46 +42,46 @@ public CurrentUserService(
...
     var email = decodedJWT.getClaim("email").asString();
-    var userOpt = participantUserDao.findOne(email, environmentName);
-    if (userOpt.isPresent()) {
-      var user = userOpt.get();
-      return Optional.of(loadFromUser(user, portalShortcode));
+    Optional<ParticipantUser> userOpt = participantUserDao.findOne(email, environmentName);
+    if (userOpt.isEmpty()) {
+      log.info("User {} not found for environment {}", email, environmentName);
+      throw new UnauthorizedException("User not found for environment " + environmentName);
```
devonbush, line 70: here's another place where we need to avoid logging the full username

api-participant/src/main/java/bio/terra/pearl/api/participant/service/CurrentUserService.java
```diff
@@ -38,46 +42,46 @@ public CurrentUserService(
...
-    // this get() will fail if the user has not registered for the given portal, which is what we
-    // want
-    PortalParticipantUser portalParticipantUser =
-        portalParticipantUserService.findOne(user.getId(), portalShortcode).get();
-    user.getPortalParticipantUsers().add(portalParticipantUser);
-    List<Enrollee> enrollees = enrolleeService.findByPortalParticipantUser(portalParticipantUser);
+    Optional<PortalParticipantUser> portalParticipantUser =
+        portalParticipantUserService.findOne(user.getId(), portalShortcode);
+    if (portalParticipantUser.isEmpty()) {
+      log.info("User {} not found for portal {}", user.getUsername(), portalShortcode);
```
devonbush: same, let's obfuscate the username.

api-participant/src/main/java/bio/terra/pearl/api/participant/service/CurrentUserService.java
```diff
@@ -38,46 +42,46 @@ public CurrentUserService(
...
+    participantUserDao.update(userOpt.user);
     return userOpt;
   }
 
   @Transactional
-  public Optional<UserWithEnrollees> refresh(
+  public UserWithEnrollees refresh(
       String token, String portalShortcode, EnvironmentName environmentName) {
     return loadByToken(token, portalShortcode, environmentName);
   }
```
devonbush, line 62: Nice, I like this refactor to return the object instead of an Optional and handle exception throwing here -- makes the other code more readable

devonbush (changes requested):
Nice, thanks for digging in on this error.  I think you need to either remove/obfuscate the username logging.  In addition to being (as far as I know) the current compliance practice dictated by AppSec, I like keeping logs free of PII because then that gives us more freedom to, e.g. send log errors to Slack, or easily dump them into log processors that might not be HIPAA compliant

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/registration/RegistrationController.java
denniscunningham: Sure, I can do that, but it seems like this code, along with RegistrationService.register could use some reorganization? 
Not investigating too deeply, it feels like new user vs existing user could be sorted out at this level to make the service code cleaner. But my original concern here is the fate of the registration data if `tokenLogin` fails. Does it get stranded?
Also, we are using email from the request body, but not verifying it against the email in the token claim?

api-participant/src/main/java/bio/terra/pearl/api/participant/service/CurrentUnauthedUserService.java
denniscunningham: Thx for clarifying our position on logging PII. Also, I appreciate your workaround suggestion.

I chose to just remove the PII in the couple cases where there was no fallback, and substituted the user ID in the case where it is available. 

I don't think obfuscated username buys us much in these cases, since we have no Juniper data to compare it to (like a list of users in a DB). For cases like this, rare but interesting, we might be better served by logging it to a table, which we can manage with time-to-live records. But for that we should probably wait until we see a specific pattern we are interested in.

Does that make sense?

devonbush: works for me.  the obfuscated username could help since this is the case where they've successfully logged in, and so we'll have record in the B2C access log in Azure.  So then we could just match on timestamp and confirm with the obfuscated username.  But I'm not sure that buys us much (since we can already do the timestamp match), so I agree to keep it simple for now and just not log it

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/registration/RegistrationController.java
devonbush: Yeah, I think we'll learn more from HeartHive about how this is going to flow, and might want to restructure, but that's beyond the scope of this PR.  

Yes, there are some corner cases where registration data gets stranded -- since they complete the registration data before they sign in, with certain pernicious patterns of browser refresh/internet connectivity problems, we can lose the user->registration connection.  We attempt to mitigate this with local storage and browser retries, but there's only so much we can do since they completed the form prior to creating an account.  And it's not worth too much worrying, since the registration data is usually just a couple of checks of "yes" on a prequal survey

Yes, it looks like this is using email from request body without validating against the token claim, which we should fix.  (either in this or in a separate ticket)

devonbush (approved):
Looks good--thanks for cleaning the PII out of the logs.  One thing I just realized, you're just logging the environment name right now--you should log the portal shortcode too for easier troubleshooting.

api-participant/src/main/java/bio/terra/pearl/api/participant/service/CurrentUnauthedUserService.java
denniscunningham: Ah, great point about the B2C access log.
Hopefully we can match by timestamp for now, and embellish as needed.

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/registration/RegistrationController.java
denniscunningham: I'd like to leave the email validation for a separate ticket where we clean up this code, if that's okay with you. 
By separating the two cases the logic will be cleaner, since as things stand now we do not have the email from the token until after registration.

denniscunningham:
Sure, I can add the portal shortcode to the two logging instances where it is missing, but I purposely did not include it since it seem irrelevant and perhaps confusing: In each case we are logging that the user could not be found, regardless of portal.
What do you think?

devonbush:
I think it's useful to know which portal they were trying to sign up/into.  that could help us troubleshoot if, for example, there's something wonky about the preregistration survey or b2c configuration for a particular customer

core/src/main/java/bio/terra/pearl/core/service/study/StudyService.java
```diff
@@ -45,13 +46,21 @@ public List<Study> findByPortalId(UUID portalId) {
...
+        try {
+            Study newStudy = dao.create(study);
+            study.getStudyEnvironments().forEach(studyEnv -> {
+                studyEnv.setStudyId(newStudy.getId());
+                StudyEnvironment newEnv = studyEnvironmentService.create(studyEnv);
+                newStudy.getStudyEnvironments().add(newEnv);
+            });
+            return newStudy;
+        } catch (Exception e) {
+            if (e.getCause() instanceof PSQLException && e.getMessage().contains("study_shortcode_key")) {
```
denniscunningham, line 58: Moved this here from StudyController to bring error handling closer to source of error.

devonbush (approved):

MatthewBemis (approved):

PR #625, opened 2023-11-09
[JN-706] notification editing

devonbush:
#### DESCRIPTION

This enables very rough-and-ready editing of notification emails from the admin tool.    While I was initially worried about security for this feature, my online research indicates emails are a less dangerous attack vector in many ways than website content, since email clients are typically more robustly sandboxed than websites.  So this, combined with the fact that we still control publishing, gives me confidence that we can expose this feature to our current limited user base without thorough safeguards around, e.g. script tag inclusion.   

image

Along the way, this upgrades the notification list page to match our latest api fetching practices, so it's no longer reliant on a portal 'god object'.
This also includes some quality-of-life improvements for the test email sender, automatically inserting the admin user's name into the "to" field, and defaulting to the support@juniper.terra.bio address if an email address hasn't been configured yet for the portal.  

Not yet supported:
  deletion/deactivation of configs
  library of default templates
  sensible guards like checks on whether or not duplicate configs exist for a single event
  user-useful error handling
  sending test emails for templates that aren't yet saved

1. repopulate hearthive ./scripts/populate_portal.sh hearthive
2. redploy apiAdminApp
3. go to https://localhost:3000/hearthive/studies/cmyop/env/sandbox/notificationContent
4. Click "Add"
5.  enter "consent" for the name, and choose "consent" for the event type
6.  Confirm the config is created, and you are taken to the screen shown above, where you can edit and save the email text


api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
```diff
@@ -39,16 +40,43 @@ public List<NotificationConfig> findForStudy(
...
-        authUtilService.authUserToStudy(user, portalShortcode, studyShortcode);
+    authUtilService.authUserToPortal(user, portalShortcode);
+    authUtilService.authUserToStudy(user, portalShortcode, studyShortcode);
     StudyEnvironment studyEnvironment =
-        studyEnvironmentService.findByStudy(studyShortcode, environmentName).get();
-    var configs = notificationConfigService.findByStudyEnvironmentId(studyEnvironment.getId());
+        studyEnvironmentService
+            .findByStudy(studyShortcode, environmentName)
+            .orElseThrow(StudyEnvironmentMissing::new);
+    var configs =
```
denniscunningham: Nit: would be nice to have a type here.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
```diff
@@ -70,19 +98,60 @@ public NotificationConfig replace(
...
+    authNotificationConfig(existing, portalEnvironment, studyEnvironment);
+    NotificationConfig newConfig = create(update, studyEnvironment, portalEnvironment);
+    // after creating the new config, deactivate the old config
     existing.setActive(false);
     notificationConfigService.update(existing);
     return newConfig;
   }
+
+  /**
+   * deactivates the notification config with configId, and adds a new config as specified in the
```
denniscunningham: This is an update or a create?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
```diff
@@ -39,16 +40,43 @@ public List<NotificationConfig> findForStudy(
...
+        studyEnvironmentService
+            .findByStudy(studyShortcode, environmentName)
+            .orElseThrow(StudyEnvironmentMissing::new);
+    var configs =
+        notificationConfigService.findByStudyEnvironmentId(studyEnvironment.getId(), true);
     notificationConfigService.attachTemplates(configs);
     return configs;
   }
 
+  public Optional<NotificationConfig> find(
```
denniscunningham, line 56: Maybe a brief comment since 'find' is ambiguous? Also, since a UUID is unique, it would be good to know what the other params are for.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
```diff
@@ -70,19 +98,60 @@ public NotificationConfig replace(
...
+        portalEnvironmentService.findOne(portalShortcode, environmentName).get();
+    authUtilService.authUserToStudy(user, portalShortcode, studyShortcode);
+    StudyEnvironment studyEnvironment =
+        studyEnvironmentService.findByStudy(studyShortcode, environmentName).get();
+
+    return create(newConfig, studyEnvironment, portalEnvironment);
+  }
+
+  /** confirms the given config is associated with the given study and portal environments */
+  private void authNotificationConfig(
```
denniscunningham: Nit: Why 'auth'? Does 'verify' work better?

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -688,7 +688,39 @@ paths:
...
+      tags: [ notificationConfig ]
+      operationId: create
+      parameters:
+        - { name: portalShortcode, in: path, required: true, schema: { type: string } }
+        - { name: studyShortcode, in: path, required: true, schema: { type: string } }
+        - { name: envName, in: path, required: true, schema: { type: string } }
+      requestBody:
+        required: true
+        content: { application/json: { schema: { type: object } } }
+      responses:
```
denniscunningham, line 702: I realize this is the practice, but for most of these API calls (and here specifically) we typically return 4xx class errors as well. I'm guessing only Swagger uses this info? Should we leave that to a general cleanup task?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
```diff
@@ -70,19 +98,60 @@ public NotificationConfig replace(
...
+   */
+  @Transactional
+  public NotificationConfig create(
+      String portalShortcode,
+      String studyShortcode,
+      EnvironmentName environmentName,
+      NotificationConfig newConfig,
+      AdminUser user) {
+    authUtilService.authUserToPortal(user, portalShortcode);
+    PortalEnvironment portalEnvironment =
```
denniscunningham, line 123: Sorry, I still struggle with the naked `get()` pattern. 
I assume if any of `portalShortCode`, `studyShortcode` or `environmentName` are invalid, we want to respond with bad request, but this code does not make that obvious.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
```diff
@@ -70,19 +98,60 @@ public NotificationConfig replace(
...
+      StudyEnvironment studyEnvironment,
+      PortalEnvironment portalEnvironment) {
+    newConfig.cleanForCopying();
+    newConfig.setStudyEnvironmentId(studyEnvironment.getId());
+    newConfig.setPortalEnvironmentId(portalEnvironment.getId());
+    if (newConfig.getEmailTemplate() != null && newConfig.getEmailTemplateId() == null) {
+      // this is a new email template, set the portal appropriately
+      newConfig.getEmailTemplate().setPortalId(portalEnvironment.getPortalId());
+    }
+    NotificationConfig savedConfig = notificationConfigService.create(newConfig);
```
denniscunningham, line 158: Nit: Why the assignment?

ui-admin/src/study/notifications/CreateNotificationConfigModal.tsx
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  return <Modal show={true} className="modal-lg" onHide={onDismiss}>
+    <Modal.Header closeButton>
+      <Modal.Title>Create Notification</Modal.Title>
+    </Modal.Header>
+    <Modal.Body>
+      <form className="bg-white p-3 my-2">
+        <label className="form-label" htmlFor="templateName">
+                    Notification name (for internal reference)
```
denniscunningham: I am not sure what "for internal reference" means. Maybe some help text?

ui-admin/src/study/notifications/EmailTemplateEditor.tsx
```diff
@@ -0,0 +1,82 @@
...
+        })
+      })
+    })
+  }
+
+  const templateVersionString = `v${emailTemplate.version}`
+  return <div className="mt-3">
+    <div className="d-flex align-items-baseline">
+      <h3 className="h6">{emailTemplate.name}</h3>
+      <div className="ms-2 text-muted fst-italic">
```
denniscunningham, line 42: I found this template reference confusing. 
Where do I find the template? Am I editing the template now?

core/src/main/java/bio/terra/pearl/core/service/study/exception/StudyEnvironmentMissing.java
```diff
@@ -0,0 +1,5 @@
+package bio.terra.pearl.core.service.study.exception;
+
+/** throw if the invariant that all studies must have the sandbox/irb/live environments is violated */
+public class StudyEnvironmentMissing extends RuntimeException {
```
denniscunningham, line 4: So you are expecting this to resolve to an internal error response?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
```diff
@@ -39,16 +40,43 @@ public List<NotificationConfig> findForStudy(
...
+        studyEnvironmentService
+            .findByStudy(studyShortcode, environmentName)
+            .orElseThrow(StudyEnvironmentMissing::new);
+    Optional<NotificationConfig> configOpt = notificationConfigService.find(configId);
+    configOpt.ifPresent(
+        config -> {
+          authNotificationConfig(config, portalEnvironment, studyEnvironment);
+          notificationConfigService.attachTemplates(List.of(config));
+        });
+    return configOpt;
```
denniscunningham, line 78: Feels like there is a lot more going on here than with most `ext` services. Perhaps a test?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/notifications/NotificationConfigController.java
```diff
@@ -41,15 +42,38 @@ public ResponseEntity<Object> findByStudy(
...
 
+  @Override
+  public ResponseEntity<Object> get(
+      String portalShortcode, String studyShortcode, String envName, UUID configId) {
+    AdminUser operator = authUtilService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+    Optional<NotificationConfig> configOpt =
+        notificationConfigExtService.find(
+            operator, portalShortcode, studyShortcode, environmentName, configId);
+    return ResponseEntity.of(configOpt.map(opt -> opt));
```
denniscunningham, line 53: Not a big fan of this pattern since it bypasses any standardized process (typically logging) we might want to do.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
```diff
@@ -39,16 +40,43 @@ public List<NotificationConfig> findForStudy(
...
-    PortalStudy portalStudy =
-        authUtilService.authUserToStudy(user, portalShortcode, studyShortcode);
+    authUtilService.authUserToPortal(user, portalShortcode);
+    authUtilService.authUserToStudy(user, portalShortcode, studyShortcode);
     StudyEnvironment studyEnvironment =
-        studyEnvironmentService.findByStudy(studyShortcode, environmentName).get();
-    var configs = notificationConfigService.findByStudyEnvironmentId(studyEnvironment.getId());
+        studyEnvironmentService
+            .findByStudy(studyShortcode, environmentName)
+            .orElseThrow(StudyEnvironmentMissing::new);
```
denniscunningham, line 48: Nice to get rid of the naked `get()`!

denniscunningham (changes requested):
Left some comments, but from a more meta POV, I am unclear about the user journey/experience via the UI. 
image

What is the difference between an email and a notification? If all notifications are email, perhaps we can get rid of the term "notification"?
It might be less confusing if this page just allowed me to configure automated emails sent to participants. (And perhaps that is all it does now?)
In that case events might be best listed as "Study enrollment *complete*".
"Task reminder" might be confusing since we have admin tasks. Are these "Participant task reminders"?
Does it make sense to use a unit like hours to increase usability? (We could always add a choice of time unit later.)
What is "Ad-hoc"?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
devonbush: copypasta comment, good catch, fixed

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
devonbush: I called it auth since I was thinking of it in terms of this preventing unauthorized access of a config outside your study/portal.  But I think you're right that this method isn't limited to be used in auth contexts, so verify works

api-admin/src/main/resources/api/openapi.yml
devonbush: yeah, this is only for Swagger AFAIK.  Agree that belongs in a general cleanup task.  Probably whenever we decide we want to start supporting our swagger API as a first-class citizen, as opposed to a "internal development-only convenience utility" we'll want to address this.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
devonbush: Thanks for catching this -- I am unfortunately extremely accustomed to the pattern of naked gets, so it'll probably take a few more PRs before I stop typing them out of habit

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
devonbush: I know some style guides frown on this, for me these last-line assignments are an extremely useful/convenient place to put a debug breakpoint -- that way I can quickly inspect what was already returned.  So that's what you're seeing here.

ui-admin/src/study/notifications/CreateNotificationConfigModal.tsx
devonbush: just means that participants will never see it, I'll add a comment

core/src/main/java/bio/terra/pearl/core/service/study/exception/StudyEnvironmentMissing.java
devonbush: Yes--this means something went horribly wrong with the study creation (whenever that happened) and so the application is in an illegal state

ui-admin/src/study/notifications/EmailTemplateEditor.tsx
devonbush: Yes, you are editing the template now.  I think I see your confusion though "template" is kind of an overloaded word.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/notifications/NotificationConfigController.java
devonbush: Is the preferred pattern that we confirm optional existence in the ExtService?

devonbush:
` I am unclear about the user journey/experience via the UI.`
The user journey is "Devon or Whitney needs to update the HeartHive email templates this week to match some text that Rachel sent us on Friday, and doing it via the database directly is hard given how long the template strings are."  Of course, it would be great if this also laid a foundation for self-service, but that's secondary to this PR.

`What is the difference between an email and a notification? If all notifications are email, perhaps we can get rid of the term "notification"?`
Yup, notifications may also include text messages, but we're not supporting those yet, so can simplify to Emails

`It might be less confusing if this page just allowed me to configure automated emails sent to participants. (And perhaps that is all it does now?)`
Yup, that's what this page does

`In that case events might be best listed as "Study enrollment complete".`
To me the 'complete' is redundant. the event is "a study enrollment", and adding 'complete' might imply they've also completed all study activities.  I could see that going either way 

`"Task reminder" might be confusing since we have admin tasks. Are these "Participant task reminders"?`
yeah, I had thought it was obvious this was configuring participant reminders, but I suspect that's my own bias.  Will clarify

`Does it make sense to use a unit like hours to increase usability? (We could always add a choice of time unit later.)`
Yup, hours or days is probably preferable.  

`What is "Ad-hoc"?`
This could use some docs.  that's the template that gets used when you send a custom email. E.g. you select a bunch of participants in the participant list, and say "send email".  It lets you include a custom message, but still have study logo, header and footer

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/notifications/NotificationConfigController.java
denniscunningham: My thoughts are that a `find` method returning nothing is fine, and optional is not a terrible way to express that.
The `.of` shorthand is handy, although readers have to remember or lookup what it does (returns a 404). But ideally we throw exactly what we want here and allow the global exception handling deal with it in a standard way (including logging something that supports a Juniper monitoring plan).
What do you think?

In this case, is this really a bad request, or not found? If we don't think the UI will ever care about the difference, then perhaps that does not matter much.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/notifications/NotificationConfigExtService.java
denniscunningham: You are correct.  Sonar, among others, flag this pattern, so it stands out.
But your reason for use makes sense to me.

core/src/main/java/bio/terra/pearl/core/service/study/exception/StudyEnvironmentMissing.java
denniscunningham: As I mentioned in a separate review, extending `InternalErrorException` makes this clearn.

ui-admin/src/study/notifications/EmailTemplateEditor.tsx
denniscunningham: When the template name appears on the page it is unclear what it is referring to. I'd be surprised if users understood it, particularly if the template name does not appear anywhere else (does it?). IOW, do users need to know this, and if so where? 
Is there a way to make that clearer?

denniscunningham:
@devonbush Thx for your explanation on the UI.
Any chance I can talk you into titling the page "Participant emails", and changing the use of "notification" to "email"?
We can always change this when we offer text message notifications, but it seems like there are enough differences in text notifications that we might have a different approach anyway.

devonbush:
It's getting late and so I don't trust my typing to convert the 3 fields to hours from minutes without errors.  So if it's ok with you, I'll take an IOU and we can convert those time units to something more user-friendly the next time we touch this page

ui-admin/src/study/notifications/EmailTemplateEditor.tsx
```diff
@@ -0,0 +1,82 @@
...
+                ({emailTemplate.stableId} {templateVersionString})
+      </div>
+    </div>
+    <div>
+      <label className="form-label">Subject
+        <input className="form-control" type="text" size={100} value={emailTemplate.subject}
+          onChange={e => updateEmailTemplate({
+            ...emailTemplate,
+            subject: e.target.value
+          })}/>
```
MatthewBemis, line 52: Testing note: editing the subject, and then editing the body in the Designer tab will cause the subject to be reset to the original value (you'll need to click somewhere else on the page to trigger the event listener)

ui-admin/src/study/notifications/EmailTemplateEditor.tsx
```diff
@@ -0,0 +1,82 @@
...
+    <div>
+      <Tabs
+        activeKey={activeTab ?? undefined}
+        className="mb-1"
+        mountOnEnter
+        unmountOnExit
+        onSelect={setActiveTab}
+      >
+        <Tab eventKey="designer" title="Designer">
+          <EmailEditor
```
MatthewBemis, line 64: Testing note: I get `GET https://editor.unlayer.com/1.5.33/tinymce/skins/content/default/content.css net::ERR_ABORTED 403 (Forbidden)` when the editor first loads. 

I also get an error in the console when I try to collapse the toolbar:

`editor.js:2 Uncaught TypeError: Cannot read properties of undefined (reading 'style')
    at n (editor.js:2:709842)
    at Object.onAction (editor.js:2:711331)
    at editor.js:2:5489116
    at editor.js:2:5297770
    at editor.js:2:5489077
    at editor.js:2:5196794
    at V (editor.js:2:5163973)
    at run (editor.js:2:5196771)
    at editor.js:2:5197545
    at editor.js:2:5161687`

MatthewBemis (commented):
I'm not finished with my review but figured I'd post some testing notes since it's late. Nothing major.

ui-admin/src/study/notifications/EmailTemplateEditor.tsx
devonbush: oooh, that's weird, investigating now

ui-admin/src/study/notifications/EmailTemplateEditor.tsx
devonbush: hmm... I noticed that a lot of stuff in the editor was  wonky -- e.g. I spent 20mins trying to futz with options to make the preview read-only, or hide the toolbars, and nothing worked.  I wonder if we'll need to switch to a different, better-supported editor when we revisit this page

MatthewBemis: Yeah, and overall it works fine for now, so definitely not blocking.

ui-admin/src/study/notifications/NotificationConfigView.tsx
```diff
@@ -119,28 +89,66 @@ export default function NotificationConfigView({ studyEnvContext }: {studyEnvCon
...
-        <textarea rows={20} cols={100} value={config.emailTemplate.body} readOnly={true}/>
-      </div>}
+      { hasTemplate && <EmailTemplateEditor emailTemplate={workingConfig.emailTemplate}
+        portalShortcode={portal.shortcode}
+        updateEmailTemplate={updatedTemplate => setWorkingConfig({
+          ...workingConfig,
+          emailTemplate: {
+            ...updatedTemplate,
+            id: undefined,
+            version: config!.emailTemplate.version + 1
```
MatthewBemis:
```suggestion
            version: config ? config.emailTemplate.version + 1 : 1
```

MatthewBemis (approved):
One additional nit. Overall looks good and the editor works well enough for what we're trying to achieve right now.

denniscunningham (approved):

PR #629, opened 2023-11-13
[JN-719] Delete pages

MatthewBemis:
#### DESCRIPTION

Not much to say about this one. You can delete pages now 🥳 You can't delete a landing page, but that doesn't seem like a desirable option anyway.

The modal in the video is not current and has been updated to look like this:
Screenshot 2023-11-13 at 3 02 10 PM

https://github.com/broadinstitute/juniper/assets/7257391/da19941a-98f3-40fc-8b2a-cfacd54d9876

* Launch admin API + UI
* Populate ourhealth
* Go to the site content editor: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/siteContent
* Try deleting pages (and note that you cannot delete a landing page)
* Save and confirm that the pages are deleted

### DIFF
--- a/ui-admin/src/portal/siteContent/DeletePageModal.test.tsx
+++ b/ui-admin/src/portal/siteContent/DeletePageModal.test.tsx
@@ -6,41 +6,6 @@ import React from 'react'
 import { mockHtmlPage } from 'test-utils/mock-site-content'
 
 describe('DeletePageModal', () => {
-  test('disables Delete button when confirm string is not typed in', async () => {
-    //Arrange
-    const { RoutedComponent } = setupRouterTest(<DeletePageModal
-      onDismiss={jest.fn()}
-      deletePage={jest.fn()}
-      renderedPage={mockHtmlPage()}
-    />)
-
-    render(RoutedComponent)
-
-    //Assert
-    const deleteButton = screen.getByText('Delete')
-    expect(deleteButton).toBeDisabled()
-  })
-
-  test('enables Delete button when confirm string is typed in', async () => {
-    //Arrange
-    const { RoutedComponent } = setupRouterTest(<DeletePageModal
-      onDismiss={jest.fn()}
-      deletePage={jest.fn()}
-      renderedPage={mockHtmlPage()}
-    />)
-
-    render(RoutedComponent)
-
-    //Act
-    const confirmDeleteInput = screen.getByLabelText('Confirm by typing "delete example page" below.')
-    const deleteButton = screen.getByText('Delete')
-
-    await userEvent.type(confirmDeleteInput, 'delete example page')
-
-    //Assert
-    expect(deleteButton).toBeEnabled()
-  })
-
   test('Delete button calls deletePage with the page', async () => {
     //Arrange
     const mockDeletePageFn = jest.fn()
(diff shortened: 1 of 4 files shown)


devonbush:
looks good, and works well when I tested it locally.  IMHO the requirement to type the name is unnecessary given that the change won't take effect unless/until they hit "Save" (and since they can go back to prior versions as you note).  And it might be worth calling that out in the confirmation dialog that they need to save before the deletion will take effect.  the confirmation text initially made me think the delete would be persisted immediately.

devonbush (approved):

MatthewBemis:
@devonbush ~~that's a good point. I went back and forth with it... Initially I felt like the UX without it wasn't great because you'd hit the delete button and sometimes it was hard to tell that anything happened. But I realize that both approaches incur  UX-debt. Maybe the one that involves committing less code is the lesser evil...~~

~~In general, I think we need to adopt better visual patterns for drawing attention after certain types of events occur (re-ordering survey pages/questions is another example)~~

MatthewBemis:
Oh wait I completely misread what you were saying Devon, clearly I still have a bad case of vacation brain.

Removing the confirmation field but keeping the dialog is totally reasonable and a good improvement. Will do that...

denniscunningham (commented):
Although the delete page modal says I have to save my changes before deleting, I was able to delete before saving. 
Did I misunderstand, or does it make sense to remove that stipulation, or actually require that for deletion?

MatthewBemis:
@denniscunningham the dialog text might need some wordsmithing. But what I was trying to get at is that page deletions aren't persisted to the database until you hit "Save" in the top right. Every change is essentially just a local draft until you hit Save.

denniscunningham:
@MatthewBemis Ah, I did not get that. 
Is there a way we can make this smoother? Is there a way to automatically save, since that intent seems pretty clear once a user has completed the dialog.

MatthewBemis:
@denniscunningham it's possible, but it introduces some inconsistency into how certain editor actions behave. We can assume intent on behalf of the user for each action, and I think your suggestion about deleting pages is accurate, but I have a preference for treating everything as a draft until "Save". I think it's an easier UX to understand because saves aren't happening piecemeal. It's also cleaner from a versioning standpoint because we aren't splitting multiple changes across multiple versions, even if they're part of one single update.

I'll ultimately defer to the reviewers for this, but I'd propose leaving the save behavior as-is and removing the text that talks about hitting the Save button in order to persist the changes. It was originally added because the first version of the PR (see video) had a confirmation input that implied that the modal _did_ save immediately, so the clarification was needed. But that's gone now.

denniscunningham:
@MatthewBemis Perhaps we can chat with @devonbush about this? Whatever the group decides is fine by me.
My POV as a user deleting the page is a meta action. As a user, I am past committing incremental changes, I'm done with the page, get rid of it. It's odd to have to save a delete unless, it is an immediate "do you really want to do that" action.
But I could be misunderstanding, so we can probably talk this through quickly f2f.

devonbush:
I agree that the correct solution isn't obvious.  Let's merge something ASAP (Matt, take your pick--I think all are viable options)--that will make sure this is up in time for team testing.  And then we can get feedback from Whitney, who is currently the most experienced site content editor, about what she'd expect and/or prefer.

MatthewBemis:
I removed the note at the bottom of the modal, so the final version I'm proposing looks like this:

Screenshot 2023-11-14 at 8 35 11 AM

denniscunningham (approved):

PR #630, opened 2023-11-13
JN-718: enabling study staff withdrawal

devonbush:
#### DESCRIPTION

OurHealth needs to be able to withdraw some soft-launch participants.  This removes the superuser limitation on withdrawing participants.  This also cleans up the component to match latest patterns

1. redeploy apiAdminApp
2. log in to admin tool using development mode as `staff@ourhealth.org`
3. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHNEWB/withdrawal
4. type in the confirmation, and click withdraw
5. confirm the participant is withdrawn


devonbush:
CI failiure is the StudyExtServiceTests > testStudyCreation(TestInfo) test, which is known to be flaky (fixed in another branch)

MatthewBemis (approved):

denniscunningham (approved):
Looks good.
Kinda surprised that the withdrawal confirmation is first name and last name. It feels like we use email as a primary identifier.
Are there cases (now or in the future) where we do not have first and last name?

devonbush:
Yup, that's already come up -- the screen makes you type "withdraw undefined undefined". :). The thinking was that since this screen already shows the participant email address, the first/last name gave extra confirmation information.  I could see that going either way, though

PR #634, opened 2023-11-15
JN-695 dashboard message tweaks

devonbush:
#### DESCRIPTION

Removes "study" from dashboard messages.   `RedirectFromOauth` handles when  you've joined a study and created your account in the same process (e.g. you signed up for the portal and joined the main study all at once).  `StudyEnrollRouter` handles when you join a study but you are already signed in.  There might be some refactoring to do to consolidate some logic/messages in the future, but I'm leaving it separate for now since I suspect how we will handle those cases will diverge as we learn more about workflows for multi-study portals.

1. repopulate hearthive
2. go to https://sandbox.hearthive.localhost:3001/ and click "join"
3. join, confirm you see the message below on your dashboard
image


MatthewBemis (approved):

ui-participant/src/login/RedirectFromOAuth.tsx
```diff
@@ -74,8 +74,8 @@ export const RedirectFromOAuth = () => {
               })
               const hubUpdate: HubUpdate = {
                 message: {
-                  title: 'Welcome to the study.',
-                  detail: 'Please read and sign the consent form below to complete registration.',
+                  title: `Welcome to ${portalStudy.study.name}`,
```
denniscunningham, line 77: Just curious since the JS is not strong, how do things work if `portalStudy` is null?

denniscunningham (approved):

ui-participant/src/login/RedirectFromOAuth.tsx
devonbush: if portal study (or study) were null here, an exception would be thrown and the user would probably get stuck with an eternal loading spinner.  Fortunately this is where typescript helps us confirm that that won't happen -- the `if` on line 70 checks for null.  If that `if` weren't there, (or if study was a nullable property on portalStudy), then typescript would refuse to allow this line

PR #635, opened 2023-11-15
JN-706 Notification config editing cleanups

devonbush:
#### DESCRIPTION
This cleans up a bunch of stuff from https://github.com/broadinstitute/juniper/pull/625

1. Reminder durations are in hours
2. Spacing between form fields are cleaner
3. NotFound exceptions are now explicitly handled
4. Our default exception mapper now includes the error class for better UX branching (not needed here, but will be useful for kit requests)
5. fixes bug where making changes to the email template would reset other form fields

1. redeploy ApiAdminApp
2. go to https://localhost:3000/hearthive/studies/hh_registry/env/sandbox/notificationContent
3. create a notification config, confirm it saves and fields are all editable
4. go to https://localhost:3000/hearthive/studies/hh_registry/env/sandbox/notificationContent/configs/125cb80f-ca12-422f-8105-d2bd2734e215
5. confirm you get a reasonable 404 error message in the UI as below
image

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
@@ -1,10 +1,8 @@
 package bio.terra.pearl.api.admin.controller;
 
-import bio.terra.common.exception.BadRequestException;
-import bio.terra.common.exception.InternalServerErrorException;
-import bio.terra.common.exception.UnauthorizedException;
-import bio.terra.common.exception.ValidationException;
+import bio.terra.common.exception.*;
 import bio.terra.pearl.api.admin.model.ErrorReport;
+import bio.terra.pearl.core.service.exception.PermissionDeniedException;
 import com.auth0.jwt.exceptions.JWTDecodeException;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.HttpStatus;
@@ -38,11 +36,18 @@ public class GlobalExceptionHandler {
     return buildErrorReport(ex, HttpStatus.UNAUTHORIZED);
   }
 
-  @ExceptionHandler({
-    UnauthorizedException.class,
-  })
+  @ExceptionHandler({UnauthorizedException.class, PermissionDeniedException.class})
   public static ResponseEntity<ErrorReport> authorizationExceptionHandler(Exception ex) {
     return buildErrorReport(ex, HttpStatus.FORBIDDEN);
+  }
+
+  @ExceptionHandler({
+    NotFoundException.class,
+    javax.ws.rs.NotFoundException.class,
+    bio.terra.pearl.core.service.exception.NotFoundException.class
+  })
+  public static ResponseEntity<ErrorReport> notFoundExceptionHandler(Exception ex) {
+    return buildErrorReport(ex, HttpStatus.NOT_FOUND);
   }
 
   // catchall - internal server error
@@ -61,6 +66,10 @@ public class GlobalExceptionHandler {
     log.info("Global exception handler: " + causes, ex);
 
     return new ResponseEntity<>(
-        new ErrorReport().message(ex.getMessage()).statusCode(statusCode.value()), statusCode);
+        new ErrorReport()
+            .errorClass(ex.getClass().getName())
+            .message(ex.getMessage())
+            .statusCode(statusCode.value()),
+        statusCode);
   }
 }
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/notifications/NotificationConfigController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/notifications/NotificationConfigController.java
@@ -53,7 +53,7 @@ public class NotificationConfigController implements NotificationConfigApi {
             operator, portalShortcode, studyShortcode, environmentName, configId);
     return ResponseEntity.ok(
         configOpt.orElseThrow(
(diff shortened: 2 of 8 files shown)


denniscunningham (changes requested):
Can you merge with development branch since there may be conflicts in GlobalExceptionHandler?

devonbush:
@denniscunningham  Thanks for the good comments during demo about template/document/version and the structure of this page.  I'm marinating on those, and will definitely be grounds for improving the next iteration of this page UX, but I want to keep this PR  just focused on the bugfixes and easy UX polish so we can move on to the other stuff in the sprint

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/notifications/NotificationConfigController.java
```diff
@@ -50,7 +51,9 @@ public ResponseEntity<Object> get(
     Optional<NotificationConfig> configOpt =
         notificationConfigExtService.find(
             operator, portalShortcode, studyShortcode, environmentName, configId);
-    return ResponseEntity.of(configOpt.map(opt -> opt));
+    return ResponseEntity.ok(
+        configOpt.orElseThrow(
+            () -> new NotFoundException("a config with that id does not exist in this study")));
```
denniscunningham: Nit: The first letter in this sentence should be capitalized to remain consistent with other Juniper messages.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
```diff
@@ -50,13 +48,20 @@ public ResponseEntity<ErrorReport> authenticationExceptionHandler(Exception ex)
...
-  })
+  @ExceptionHandler({UnauthorizedException.class, PermissionDeniedException.class})
   public ResponseEntity<ErrorReport> authorizationExceptionHandler(Exception ex) {
     return buildErrorReport(ex, HttpStatus.FORBIDDEN, request);
   }
 
+  @ExceptionHandler({
+    NotFoundException.class,
+    javax.ws.rs.NotFoundException.class,
+    bio.terra.pearl.core.service.exception.NotFoundException.class
```
denniscunningham, line 59: That's a lot of NotFound variations!
I guess it's too late at this point but is there a reason Juniper invented it's own (and based other exception classes on that)?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
```diff
@@ -80,6 +85,10 @@ protected static ResponseEntity<ErrorReport> buildErrorReport(
         ex);
 
     return new ResponseEntity<>(
-        new ErrorReport().message(ex.getMessage()).statusCode(statusCode.value()), statusCode);
+        new ErrorReport()
```
denniscunningham, line 88: Are you comfortable having the admin and participant versions of GlobalExceptionHandler diverge?

denniscunningham (changes requested):
Look good. The exception handling improvements are nice!
Sorry to hold up on this, but the capitalization fix feels important, esp after seeing the error message in the UI.
Also, a question about GlobalExceptionHandler.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
devonbush: Yes, I agree.  I spent a good 5 minutes trying to figure out why this method wasn't being invoked before realizing that I had added the wrong NotFoundException classes.  It's related to our conversation from yesterday -- the other two exception classes are web-oriented and so are in packages not visible to `core`

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
devonbush: Yes, I would expect them to.  Also, since changes to Participant UI require about 10x more review and quality checks than admin UX, I like keeping controllers that might impact participant UI as separate as possible from admin controllers so we can iterate quickly on admin UX without thinking "will this trigger IRB scrutiny by impacting participant UI?"

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
devonbush: that might be worth an infrastructure ticket to investigate whether we should just say "we all know core is probably only ever getting used only in a webapp, so it might as well import some web packages to avoid having an extra translation layer"

devonbush: https://broadworkbench.atlassian.net/browse/JN-737

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
devonbush: And to the concrete question of "should this particular  change be in the participant exception handler", I'd lean towards "not until we need it".  Since the participant UI is much more public, we should error on the side of not sending details to the UX

denniscunningham (approved):
Looks good.
Thanks for your thoughtful answers to my questions.

MatthewBemis (approved):

PR #636, opened 2023-11-16
[JN-657] Add request info to exception logging

denniscunningham:
#### DESCRIPTION
For all exceptions handled by the GlobalExceptionHandler, logs exception information with the associated request. This makes troubleshooting via logs and alerts a lot easier.

1. Run ApiAdminApp.
2. Goto https://localhost:3000/
3. Do an operation via the UI that will cause an exception. (I like creating a study with an existing shortcode.)
4. Observe the exception logging. It should include request information (method, URI).


devonbush (approved):
Looks great.  this will really come in handy on the participant side where we can't just reach out to customers and say "hey, were you trying to do X?"

MatthewBemis (approved):

PR #640, opened 2023-11-17
[JN-657] Consolidate log handling in GlobalExceptionHandler

denniscunningham:
#### DESCRIPTION
Consolidates logging into a single place in GlobalExceptionHandler, ensuring consistent logging and providing a central place for future logging changes. This matters since Juniper alerting depends on the severity and structure of exception logging. (Also, I should have gotten this right in a prior PR.)

1. Run GlobalExceptionHandlerTest.
2. Observe the exception logging.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
@@ -18,7 +18,7 @@ import org.springframework.web.servlet.NoHandlerFoundException;
 @RestControllerAdvice
 public class GlobalExceptionHandler {
 
-  private HttpServletRequest request;
+  private final HttpServletRequest request;
 
   public GlobalExceptionHandler(HttpServletRequest request) {
     this.request = request;
@@ -65,7 +65,6 @@ public class GlobalExceptionHandler {
   // catchall - internal server error
   @ExceptionHandler({InternalServerErrorException.class, Exception.class})
   public ResponseEntity<ErrorReport> internalErrorExceptionHandler(Exception ex) {
-    log.error("Exception caught by catch-all handler", ex);
     return buildErrorReport(ex, HttpStatus.INTERNAL_SERVER_ERROR, request);
   }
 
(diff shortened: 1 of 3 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
```diff
@@ -76,13 +75,21 @@ protected static ResponseEntity<ErrorReport> buildErrorReport(
...
-        request.getRequestURI(),
-        statusCode.value(),
-        ex);
+
+    String logString =
+        String.format(
+            "%s%nRequest: %s %s %s",
+            causes, request.getMethod(), request.getRequestURI(), statusCode.value());
+
+    switch (statusCode) {
```
devonbush: nit: I generally avoid `switch` except in extreme cases because I can never remember the fallthrough/break/default rules across languages.

this code might be cleaner if you either made some static Lists 
`public static final List ERROR_LOG_STATUSES = List.of(...)`
and then used `includes` to branch.  Alternatively, you could use the `statusCode.is5xxServerError()`for branching

devonbush (approved):
nice, agree this will be a big win for logging.  Some codesmithing suggestions, but none I have a strong opinion about.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
denniscunningham: @devonbush 
Just curious, why are you negative on `switch`? Is that broadly frowned on these days, or a personal preference? I did not see any info on a quick search.

To my eye the alternatives seem harder to read, and this switch makes it clear how to add log customizations for other response statuses, but I could be missing something.

denniscunningham: There is also this style:
```    
switch (statusCode) {
      case INTERNAL_SERVER_ERROR -> log.error(logString, ex);
      case BAD_REQUEST, UNAUTHORIZED, FORBIDDEN, NOT_FOUND -> log.info(logString, ex);
      default -> log.warn(logString, ex);
    }
```

devonbush: It was frowned on at the mid-sized company I worked for ~10 years ago, but otherwise just a personal preference.  I just have a mental block where no matter how many times I reread the `switch` syntax rules for Java/JS, I can't remember them offhand, and so a `switch` statement always slows me down and makes me worried.   But I can live with it :)

devonbush: No opinion on which of the two switch syntaxes is best -- whatever you think is best

denniscunningham: @devonbush 
Can you re-review? While we were discussing the switch statements I did some manual testing which led to an update.

devonbush (approved):

MatthewBemis (approved):

PR #641, opened 2023-11-18
JN-698: case insensitive login

devonbush:
#### DESCRIPTION

Upgrades admin logins to be case-insensitive.  Surprisingly difficult to find authoritative information on Microsoft acocunt settings, I suspect because there are about a billion different microsoft credential systems.  But I think we're relatively safe here because (1) The major MS account sources (Outlook, Windows) are case-insensitive and (2) MS admin accounts should generally be within a company domain.  So, e.g. while it might be theoretically possible for some obscure active directory mode to enable`someone@mgh.org` and `someONE@mgh.org` to be different accounts, that's a pretty narrow attack vector, since the attacker would still need admin access to the `mgh.org` OAuth provider to make that account. 

1. Redeploy apiAdminApp
2. login in development mode as both `dbush@broadinstitue.org`  and `DBush@broadinstitute.org`, and observe successes.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/CurrentUnauthedUserService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/CurrentUnauthedUserService.java
@@ -3,6 +3,7 @@ package bio.terra.pearl.api.admin.service;
 import bio.terra.pearl.core.dao.admin.AdminUserDao;
 import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.admin.AdminUserWithPermissions;
+import bio.terra.pearl.core.service.admin.AdminUserService;
 import com.auth0.jwt.JWT;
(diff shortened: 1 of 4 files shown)


MatthewBemis (approved):

api-admin/src/main/java/bio/terra/pearl/api/admin/service/CurrentUnauthedUserService.java
```diff
@@ -1,34 +1,32 @@
...
 import java.time.Instant;
 import java.util.Optional;
 import java.util.UUID;
 import org.springframework.stereotype.Service;
-import org.springframework.transaction.annotation.Transactional;
 
 @Service
 public class CurrentUnauthedUserService {
-  private AdminUserDao adminUserDao;
+  private AdminUserService adminUserService;
```
denniscunningham, line 15: Nice!

denniscunningham (approved):

PR #642, opened 2023-11-20
[JN-714] Configurable dashboard alerts

MatthewBemis:
#### DESCRIPTION

This adds configurable participant dashboard alerts. Right now these are only configurable for a portal environment, but soon they will also be configurable per study environment (which is why the DB table supports taking an optional study env ID). This is also one more example of how we'll need to clean up the admin UX for handling configuration for portals versus studies. This only adds direct support for editing the "No activities remain" alert, but you can see how we'll be able to easily add support for alerts that use other triggers.

https://github.com/broadinstitute/juniper/assets/7257391/425985e8-6eec-46c3-9f25-52418814f87e

* Start up admin API + participant API + admin UI + participant UI
* Log into the participant UI and complete all activities (maybe I should add a user that has all of their activities completed...)
* Observe the No activities remain message
* Go into the admin UI and update the message for that study environment
* Refresh the participant UI and observe the custom message displays


ui-admin/src/portal/publish/PortalEnvDiffView.tsx
```diff
@@ -132,6 +132,13 @@ export default function PortalEnvDiffView(
           { !changeSet.siteContentChange.changed && <VersionChangeView record={changeSet.siteContentChange}/> }
         </div>
       </div>
+      <div className="my-2">
+        <h2 className="h6">
+          Dashboard config</h2>
+        <div className="ms-4 ">
+          <span className="text-muted fst-italic">not yet implemented</span>
```
MatthewBemis: I still need to add this to the content publishing flow, but didn't want to hold up review for that.

core/src/main/resources/db/changelog/changesets/2023_11_17_participant_dashboard_alerts.yaml
```diff
@@ -0,0 +1,28 @@
+databaseChangeLog:
+  - changeSet:
+      id: participant_dashboard_alert
+      author: mbemis
+      changes:
+        - createTable:
+            tableName: participant_dashboard_alert
```
devonbush, line 7: maybe `participant_dashboard_message` instead of alert?  These aren't always alerts in the sense of demanding immediate attention

core/src/main/resources/db/changelog/changesets/2023_11_17_participant_dashboard_alerts.yaml
```diff
@@ -0,0 +1,28 @@
...
+            columns:
+              - column: { name: id, type: uuid, defaultValueComputed: gen_random_uuid(), constraints: {
+                nullable: false, primaryKey: true
+              } }
+              - column: { name: title, type: text, constraints: { nullable: false } }
```
devonbush, line 12: Do we think these will ever want formatting beyond just text/markdown?  Would they ever want to be an HtmlSection-like object instead?

ui-core/src/participant/dashboard/Alert.tsx
```diff
@@ -42,7 +79,7 @@ export const Alert = (props: AlertProps) => {
       {renderedIcon && <FontAwesomeIcon className="me-3 fa-xl" icon={renderedIcon} />}
       <div className="flex-grow-1">
```
devonbush, line 83: I think we should upgrade both the title and text to support markdown -- I wouldn't be surprised if a lot of the messages will contain links

devonbush (changes requested):
This looks really good -- I appreciate the effort to factoring out common message code, and I think you did a good job of scoping this PR so I could get a good sense of what's happening, but without going too deep. 
 I haven't finished the code review yet, but I have some high-level comments that I wanted to put in since I might not get to the full code review until next Tuesday.  

Overall, my first impression of the code is that it's quite solid.  The functional comments I put in the files are suggestions and not blocking.  the blocker for me is that this needs to be wired into the populate infrastructure so that when a developer populates HeartHive, they can see some sample messages configured.

core/src/main/resources/db/changelog/changesets/2023_11_17_participant_dashboard_alerts.yaml
devonbush: thinking more about it, I think you were right to categorize these as "alerts" since they are only contingently displayed

core/src/main/java/bio/terra/pearl/core/model/ParticipantDashboardAlert.java
```diff
@@ -0,0 +1,22 @@
...
+@SuperBuilder
+@NoArgsConstructor
+public class ParticipantDashboardAlert extends BaseEntity {
+    private UUID portalEnvironmentId;
+    private UUID studyEnvironmentId; //nullable. if this is set for an alert, it takes precedence over an equivalent portal alert in the UI
+
+    private String title;
+    private String detail;
+    private String trigger;
+    private String type;
```
devonbush: `type` and `trigger` feel like they should be enums, at least until we have custom event infrastructure.  In fact, the `trigger` might want to share values with `NotificationEventType` (perhaps refactoring/renaming it into a more general "EventType" enum)

core/src/main/java/bio/terra/pearl/core/model/ParticipantDashboardAlert.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.experimental.SuperBuilder;
+
+//TODO: should these be versioned as well?
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class ParticipantDashboardAlert extends BaseEntity {
```
devonbush: I suspect we'll eventually have a parent "DashboardConfig" entity, and these will be attached to that, rather than the StudyEnvironment directly, but that's a task for another day

devonbush (commented):
more general thoughts -- have another third of the code to go, hopefully will get to it tonight

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/dashboard/DashboardController.java
```diff
@@ -0,0 +1,74 @@
...
+import bio.terra.pearl.core.model.ParticipantDashboardAlert;
+import bio.terra.pearl.core.model.admin.AdminUser;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.List;
+import javax.servlet.http.HttpServletRequest;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+
+@Controller
+public class DashboardController implements DashboardApi {
```
devonbush, line 17: nice, I like you you made this a more general DashboardController rather than a DashboardAlertController

core/src/main/resources/db/changelog/changesets/2023_11_17_participant_dashboard_alerts.yaml
```diff
@@ -0,0 +1,28 @@
...
+              - column: { name: type, type: text, constraints: { nullable: false } }
+              - column: { name: portal_environment_id, type: uuid, constraints: {
+                nullable: false, foreignKeyName: fk_dashboard_alert_portal, references: portal_environment(id)
+              } }
+              - column: { name: study_environment_id, type: uuid, constraints: {
+                nullable: true, foreignKeyName: fk_dashboard_alert_study, references: study_environment(id)
+              } }
+              - column: { name: created_at, type: datetime, constraints: { nullable: false } }
+              - column: { name: last_updated_at, type: datetime, constraints: { nullable: false } }
+        - addUniqueConstraint:
```
devonbush, line 21: nice

api-admin/src/main/java/bio/terra/pearl/api/admin/service/dashboard/DashboardExtService.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    PortalEnvironment portalEnv =
+        portalEnvironmentService
+            .findOne(authedPortal.getShortcode(), envName)
+            .orElseThrow(PortalEnvironmentMissing::new);
+
+    ParticipantDashboardAlert alert =
+        portalDashboardConfigService.findByPortalEnvId(portalEnv.getId()).stream()
+            .filter(a -> a.getTrigger().equals(newAlert.getTrigger()))
```
devonbush: since trigger is the key identifier for the alert, let's put the trigger name in the URL and method signatures for the API.  that makes it clearer that the API call is "update the alert with this trigger name"

api-admin/src/main/java/bio/terra/pearl/api/admin/service/dashboard/DashboardExtService.java
```diff
@@ -0,0 +1,92 @@
...
+            .findFirst()
+            .orElseThrow(() -> new NotFoundException("The specified alert does not exist"));
+
+    BeanUtils.copyProperties(newAlert, alert);
```
devonbush: BLOCKER: copying all properties opens up security holes since you haven't validated the properties of the newAlert match.  You'll want to either exclude id, createdAt, updatedAt, studyEnvironmentId, and portalEnvironmentId from the copy, or (probably safer), only copy over the title and message properties

api-admin/src/main/java/bio/terra/pearl/api/admin/service/dashboard/DashboardExtService.java
```diff
@@ -0,0 +1,92 @@
...
+      throw new IllegalArgumentException(
+          "You can only create dashboard alerts in the sandbox environment");
+    }
+
+    PortalEnvironment portalEnv =
+        portalEnvironmentService
+            .findOne(authedPortal.getShortcode(), envName)
+            .orElseThrow(PortalEnvironmentMissing::new);
+
+    newAlert.setPortalEnvironmentId(portalEnv.getId());
```
devonbush, line 102: BLOCKER: We need to either validate the studyEnvironmentId or null it
```suggestion
    newAlert.setStudyEnvironmentId(null);
```

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -982,6 +982,52 @@ paths:
...
+        '200':
+          description: OK
+          content: { application/json: { schema: { type: object } } }
+        '500':
+          $ref: '#/components/responses/ServerError'
+    post:
+      summary: Creates a participant dashboard alert config for a portal environment
+      tags: [ dashboard ]
+      operationId: createPortalEnvAlert
+      parameters:
```
devonbush, line 1035: similar to update, since "trigger" is kind of the primary key for the alert, we should put "trigger" in the api url path.

devonbush (commented):
Thinking more about this -- I think there might be two different kinds of alerts -- state-based and event-based.  State-based would be "show this alert any time the participant is in a state where they have no activities left to do"  event-based would be "after we get the kit request back from a participant, show this".  I think the alerts you have are mix of the two -- NO_ACTIVITIES_REMAIN is state-based, it will continue to show as long as there are no activities to do, while STUDY_ALREADY_ENROLLED is event-based "you tried to enroll in something you already enrolled for".  So I think it's good that you didn't extend/reuse NotificationEventType, since these are kind of different.  But as this feature evolves, we'll have to think about whether we want to treat state and event-based messages differently

core/src/main/resources/db/changelog/changesets/2023_11_17_participant_dashboard_alerts.yaml
MatthewBemis: Is there a use-case you have in mind where markdown might not be sufficient? I'm a little worried about introducing yet another complex/structured object into the self-service workflow. They generally feel rigid to work with (obviously they're rigid for good reason), but they require a lot more UX thought and documentation to roll out.

I think we can go a long way with just markdown, but the door is always open to add some more structure to it in the future if users start asking for more complex alert structures.

MatthewBemis:
@devonbush I believe I've addressed all comments, so this is ready for re-review. I'll open a separate PR into this one for the content publishing flow.

MatthewBemis:
Actually, one comment remains. I still need to wire this up to the populate infrastructure.

ui-core/src/participant/dashboard/Alert.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+export type ParticipantDashboardAlert = {
+  id?: string
+  title: string
+  detail?: string
+  type: AlertLevel
+  trigger: AlertTrigger
+}
+
+export const alertDefaults: Record<AlertTrigger, ParticipantDashboardAlert> = {
```
devonbush, line 28: something to keep in the back of our minds as we tackle the full multi-study dashboard is moving as much default text to the database as we can, so that we can change it and internationalize it more easily

ui-participant/src/hub/HubPage.tsx
```diff
@@ -1,24 +1,43 @@
...
-  const { portal } = usePortalEnv()
+  const { portal, portalEnv } = usePortalEnv()
   const { enrollees } = useUser()
+  const [noActivitiesAlert, setNoActivitiesAlert] = useState<ParticipantDashboardAlert>()
+
+  useEffect(() => {
+    loadDashboardAlerts()
+  }, [])
+
+  const loadDashboardAlerts = async () => {
```
devonbush, line 26: I like what you did here of loading the dashboard config separately from the big study/portal env blob that gets loaded when they hit the landing page

ui-participant/src/hub/HubPage.tsx
```diff
@@ -1,24 +1,43 @@
...
 
 /** renders the logged-in hub page */
 export default function HubPage() {
-  const { portal } = usePortalEnv()
+  const { portal, portalEnv } = usePortalEnv()
   const { enrollees } = useUser()
+  const [noActivitiesAlert, setNoActivitiesAlert] = useState<ParticipantDashboardAlert>()
+
+  useEffect(() => {
+    loadDashboardAlerts()
```
devonbush, line 23: Added an eng infrasructure ticket https://broadworkbench.atlassian.net/browse/JN-750 to handle extracting this pattern into a utility function

ui-participant/src/hub/HubPage.tsx
```diff
@@ -1,24 +1,43 @@
...
+
+  const loadDashboardAlerts = async () => {
+    try {
+      const alerts = await Api.getPortalEnvDashboardAlerts(portal.shortcode, portalEnv.environmentName)
+      setNoActivitiesAlert({
+        ...alertDefaults['NO_ACTIVITIES_REMAIN'],
+        ...alerts.find(msg => msg.trigger === 'NO_ACTIVITIES_REMAIN')
+      })
+    } catch (error) {
+      console.error(error)
```
devonbush: BLOCKER: we should call either `defaultApiErrorHandle` or `logError` here.  (and JN-750 should handle putting those into a nicer pattern so it's easier to remember that)

devonbush (approved):
One blocker around error logging for the participant UI that I trust you to put in, otherwise I think this is a great start.  This is the first bit of truly new configuration that we've added since Juniper started, so I appreciate your resilience in tackling it.  As we discussed, the publishing can be in a follow-on PR.

ui-participant/src/hub/HubPage.tsx
MatthewBemis: Ah good call. It looks like `processJsonResponse` has `defaultApiErrorHandle` built into it, so I'll remove the `console.error` here.

core/src/main/resources/db/changelog/changesets/2023_11_17_participant_dashboard_alerts.yaml
denniscunningham: FWIW, I like 'message' more than alerts. For instance, there is no call to action for the specific case created for populate in this PR. But it hardly seems worth the change at this point since 'alert' is everywhere.

ui-participant/src/api/api.tsx
```diff
@@ -172,6 +173,12 @@ export default {
     return parsedResponse
   },
 
+  async getPortalEnvDashboardAlerts(portalShortcode: string, envName: string): Promise<ParticipantDashboardAlert[]> {
+    const url = `${API_ROOT}/public/portals/v1/${portalShortcode}/env/${envName}/dashboard/config/alerts`
```
denniscunningham, line 177: Here and in the new admin dashboard alerts, what value does the term 'config' provide?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/dashboard/DashboardController.java
```diff
@@ -0,0 +1,81 @@
...
+      DashboardExtService dashboardExtService,
+      ObjectMapper objectMapper) {
+    this.request = request;
+    this.authUtilService = authUtilService;
+    this.dashboardExtService = dashboardExtService;
+    this.objectMapper = objectMapper;
+  }
+
+  @Override
+  public ResponseEntity<Object> listPortalEnvAlerts(String portalShortcode, String envName) {
```
denniscunningham, line 36: Just trying to connect this with your PR comment... It sounds like dashboard configuration will eventually be per study rather than per portal? Or is the plan to allow configuration at each level?
Nit: I'm not sure the term 'env' adds value here, but I may misunderstand the feature. I thought this was configuration that got published to each environment, like surveys? In that case, I get that env is used to validate the usage, but it does not seem like a principal player in the behavior, since differences in env are only due to publishing differences -- you cannot set a different alert in live than in sandbox.

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -982,6 +982,55 @@ paths:
           content: { application/json: { schema: { type: object } } }
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/portals/v1/{portalShortcode}/env/{envName}/dashboard/config/alerts:
```
denniscunningham, line 1016: Is the term `config` useful here? IOW, is it likely there will be dashboard aspects other than `config` in the future?

core/src/main/java/bio/terra/pearl/core/model/dashboard/ParticipantDashboardAlert.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.experimental.SuperBuilder;
+
+//TODO: should these be versioned as well?
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class ParticipantDashboardAlert extends BaseEntity {
+    private UUID portalEnvironmentId;
+    private UUID studyEnvironmentId; //nullable. if this is set for an alert, it takes precedence over an equivalent portal alert in the UI
```
denniscunningham: I don't understand this comment (but it seems important). :-)

local-dev/render_environment_vars.sh
```diff
@@ -27,6 +27,7 @@ PARTICIPANT_API_ENV_VARS=(
   "REDIRECT_ALL_EMAILS_TO:static:$DEV_EMAIL"
   "B2C_CONFIG_FILE:static:b2c-config.yml"
   "SENDGRID_API_KEY:vault:vault read -field=api_key secret/dsp/ddp/d2p/dev/sendgrid"
+  "SWAGGER_ENABLED:static:true"
```
denniscunningham, line 30: How does this fit in here?

ui-admin/src/study/StudyEnvironmentRouter.tsx
```diff
@@ -174,6 +177,11 @@ export const studyEnvNotificationsPath = (portalShortcode: string, studyShortcod
   return `/${portalShortcode}/studies/${studyShortcode}/env/${envName}/notificationContent`
 }
 
+/** helper for path to configure study messages */
+export const studyEnvMessagesPath = (portalShortcode: string, studyShortcode: string, envName: string) => {
```
denniscunningham: Are messages the same as alerts?

ui-core/src/participant/dashboard/Alert.tsx
```diff
@@ -0,0 +1,89 @@
...
+  'NO_ACTIVITIES_REMAIN': {
+    title: 'All activities complete',
+    detail: 'You have completed all activities for this study. We will notify you ' +
```
denniscunningham, line 31: I'm confused about why this is here when are allowing admins to edit this message. Presumably we want to get this from the DB?

denniscunningham (commented):
Can we discuss this PR in person, perhaps after standup on Friday?
I think I'm missing some context, since this seems purposely incomplete. 

In terms of alert/message life cycle, will all the logic for when a message is shown live in the ptp frontend?

Also, there is an implication that dashboard messages will eventually have a study and portal context, but I think why we want to do that, how that would work for admins, and how we implement that are worth discussing, not only for this feature but for how we think about similar study/portal situations.
Does that make sense?

ui-admin/src/study/StudyEnvironmentRouter.tsx
MatthewBemis: In the first few commits in this PR I originally called these "messages" rather than "alerts". They're the same, but looks like this usage slipped through the cracks when renaming.

I'm indifferent about what to call these, but if there are strong feelings about going back to "messages" I'm happy to make that change now while the concept is still quarantined to this PR

local-dev/render_environment_vars.sh
MatthewBemis: We disable the Swagger UI for the Participant API by default because we don't want participants to have access to it. There's no reason to disable it for local development though.

It's not directly related to any of the other changes in this PR, but during development/testing of this PR I ran into a brief headache because I hadn't realized that we disabled it by default. I added it to the setup script so other devs in the future wouldn't run into the same confusion

core/src/main/java/bio/terra/pearl/core/model/dashboard/ParticipantDashboardAlert.java
MatthewBemis: This relates to some of the configuration headaches that this PR spawned.

This PR only adds support for editing alerts for portalEnvironments, but eventually admins will likely want to get more specific with their edits and edit alerts per studyEnvironment. Should that be represented as two separate tables and two separate types? i.e. do we want to have `portal_environment_dashboard_alert` and `study_environment_dashboard_alert` tables? I chose to represent them as the same object in the same table because I think it simplifies the code quite a bit, admittedly at the cost of muddying the concept a bit.

There will ultimately be 3 "layers" of alert-specificity: 1) Default, 2) PortalEnv, 3) StudyEnv.

The UI will always need to show the most specific alert available, so the query will first look for any alert that matches the given studyEnvironmentId. If that returns nothing, then it will go up a level and look for a matching portalEnvironmentId. And lastly it will default to the hardcoded values that live in the UI code (but will eventually live in the DB).

As I'm typing all of this I'm realizing it was probably premature to set the stage for handling studyEnv alerts. Would it be better if I removed it?

api-admin/src/main/resources/api/openapi.yml
MatthewBemis: I'm not sure. I have PTSD from a previous project where every decision about our REST API was a life-sentence. So I've developed a habit of over-preparing for these types of potential future situations.

Fortunately that's not the reality of this project, so I could buy into getting rid of it. It's just been a hard habit to kick.

ui-core/src/participant/dashboard/Alert.tsx
MatthewBemis: Yes that is the eventual goal, especially with multi-language support as Devon mentioned.

core/src/main/java/bio/terra/pearl/core/model/dashboard/ParticipantDashboardAlert.java
denniscunningham: Thx for the explanation.
I think `studyEnvironmentId` should probably not be included in this PR.
I can think of a couple ways study level dashboard messages could be modeled that would not require this field (separate tables, using join tables on a generic message table).

ui-admin/src/study/StudyEnvironmentRouter.tsx
denniscunningham: Thx. I do not think changing everything from 'alert' to 'message' it is worth the trouble since the value is debatable.
It would be nice to use a consistent term. Perhaps changing these to 'alert' is something for a future PR?

api-admin/src/main/resources/api/openapi.yml
denniscunningham: Gotcha.
From a practical point of view, our contract for this is easily changed, so I could see changing this now or waiting until later.

denniscunningham (changes requested):
Looks great.
Thx for you patience explaining things to me.
At this point the only thing left is your consideration of omitting `studyEnvironmentId` from `ParticipantDashboardAlert`.
Happy with whatever you decide, but wanted to call it out.

ui-admin/src/study/StudyEnvironmentRouter.tsx
MatthewBemis: Since the code was already in front of me, I changed these to "alert"

core/src/main/java/bio/terra/pearl/core/model/dashboard/ParticipantDashboardAlert.java
MatthewBemis: 👍 Makes sense, removed studyEnvironmentId in the latest commit

denniscunningham (approved):
Looks great.
Thank you for helping me review this PR!

core/src/main/java/bio/terra/pearl/core/model/dashboard/AlertType.java
```diff
@@ -0,0 +1,9 @@
+package bio.terra.pearl.core.model.dashboard;
+
+public enum AlertType {
+    primary,
```
denniscunningham: Just realized while reviewing PR 663 that these enum elements should be capitalized.

PR #650, opened 2023-11-25
JN-711 adding environment to emails

devonbush:
#### DESCRIPTION

This adds the environment name (if not live) and the deployment zone (if not prod) to emails so that they are easier to distinguish in your inbox.  In the process, this promotes "DEPLOYMENT_ZONE", which was already being supplied, to a top-level `env` variable in our Environment (previously, it was only consumed by TDR services).

 Note that we don't append environment name to emails to admins, since those are not (yet) environment specific.  Right now the only email we send to admins is a "Welcome to Juniper" email.

 This also fixes a bug where we were sending the "Welcome to Juniper" emails to study staff from the info@ourhealthstudy.org domain, rather than from Juniper.  

1. redeploy ApiAdminApp
2. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/notificationContent`
3. pick a configuration, and then select "send test email" and click send
4. Notice that the email now appears in your inbox with (sandbox) (local) suffixed to it


api-admin/src/main/resources/application.yml
```diff
@@ -19,7 +20,6 @@ env:
     storageAccountKey: ${TDR_EXPORT_STORAGE_ACCOUNT_KEY:}
     storageContainerName: ${TDR_EXPORT_STORAGE_CONTAINER_NAME:}
     billingProfileId: ${BILLING_PROFILE_ID:550212cf-45dc-478f-9fe6-9b3290e2a1fe}
-    deploymentZone: ${DEPLOYMENT_ZONE:}
```
MatthewBemis, line 22: This is also referenced by `isTdrConfigured()` in `ScheduledDataRepoExportService`. Since there's a default now it can probably just be removed.

MatthewBemis (approved):

core/src/main/java/bio/terra/pearl/core/service/notification/email/SendgridClient.java
```diff
@@ -71,16 +73,18 @@ public List<SendgridEvent> getEvents(Instant startDate, Instant endDate, int que
...
     Email to = new Email(toAddress);
 
-    String fromName = "Juniper";
-    if (contextInfo.portal() != null) {
-      // Set the 'from' name on the email to the portal name
-      from.setName(contextInfo.portal().getName());
+    if (fromName == null) {
+      fromName = "Juniper";
+    }
+    if (!deploymentZone.equalsIgnoreCase("prod")) {
```
denniscunningham, line 83: Not sure why both the environment and deployment zone are useful. It's a nit, but just curious since it looks kind of odd in the email header.

denniscunningham (approved):
Looks good. One question.
Not related to this PR, but a UX issue... When using the send test email modal, "Send email" does dismiss the dialog. Since the email sent confirmation was delayed for me, I assumed my button click had not registered, so mashed the button again.

devonbush:
Yeah, I want to keep the dialog up so it's easier to send multiple test emails (e.g. I send one to myself and one to Erin) or testing how it works with various permutations user properties.   I'll see if I can add a loading spinner to avoid mashing

core/src/main/java/bio/terra/pearl/core/service/notification/email/SendgridClient.java
devonbush: at least for me, I get most of my emails from sandboxes, and so I want to know if it's local/prod/demo so I can tune it in/out appropriately, or withdraw the participant

devonbush:
thanks @denniscunningham   updated the UX so that there's a loading spinner and the button is disabled while a send is happening

PR #651, opened 2023-11-26
JN-745: fixing study settings bug

devonbush:
#### DESCRIPTION

Previously, changing the environment on the Study Settings page could leave stale values populated in the form.  This is because we weren't clearing the form state when the environment or study changed, and also because sometimes we were passing `undefined` as the value to text fields, which results in it becoming an uncontrolled field.

This was my first time using the `rerender` function in react tests -- this let me easily test how the component behaves when rerendered with different props.

1. repopulate OurHealth. `./scripts/populate_portal.sh ourhealth`
2. go to `https://localhost:3000/ourhealth/studies/ourheart/env/live/settings`
3. note the "Email source address" (the bottom-most form field) is empty
4. switch the environment to 'sandbox' 
5. confirm the "Email source address" now shows "info@ourhealthstudy.org"

### DIFF
--- a/ui-admin/src/portal/PortalEnvConfigView.test.tsx
+++ b/ui-admin/src/portal/PortalEnvConfigView.test.tsx
@@ -22,6 +22,29 @@ test('renders a portal env. config', async () => {
     .toBe(envConfig.acceptingRegistration)
 })
 
+test('changes state when portal env changes', async () => {
+  const portalContext = mockPortalContext()
+  const portalEnv = portalContext.portal?.portalEnvironments[0] as PortalEnvironment
+  const envConfig = portalEnv.portalEnvironmentConfig
+  const { rerender } = render(<MockSuperuserProvider>
+    <PortalEnvConfigView portalContext={portalContext} portalEnv={portalEnv}/>
+  </MockSuperuserProvider>)
+
+  expect(screen.getByLabelText('password')).toHaveValue(envConfig.password)
+  const differentEnv = {
+    ...portalEnv,
+    portalEnvironmentConfig: {
+      ...portalEnv.portalEnvironmentConfig,
+      password: 'newPass3'
+    },
+    environmentName: 'irb'
+  }
+  rerender(<MockSuperuserProvider>
+    <PortalEnvConfigView portalContext={portalContext} portalEnv={differentEnv}/>
+  </MockSuperuserProvider>)
+  expect(screen.getByLabelText('password')).toHaveValue('newPass3')
+})
+
 test('updates a portal env. config', async () => {
   const portalContext = mockPortalContext()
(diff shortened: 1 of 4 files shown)


ui-admin/src/portal/PortalEnvConfigView.tsx
```diff
@@ -27,6 +28,9 @@ const PortalEnvConfigView = ({ portalContext, portalEnv }: PortalEnvConfigViewPr
   const updateConfig = (propName: string, value: string | boolean) => {
     setConfig(set(propName, value))
   }
+  useUpdateEffect(() => {
```
denniscunningham, line 31: Sorry for my inexperience, but trying to understand how the impact of this is differs, from the `config` initialization on line 23. I assume the initialization on line 23 sets the config for the current render, and this `setConfig` here applies to a the next render (which may happen right away)?
Also, how do the two dependencies relate to the operation within the clause?

denniscunningham (changes requested):
Just a couple of questions, not anything wrong per se, but the answers will help me provide a better review.

ui-admin/src/portal/PortalEnvConfigView.test.tsx
```diff
@@ -22,6 +22,29 @@ test('renders a portal env. config', async () => {
...
+  expect(screen.getByLabelText('password')).toHaveValue(envConfig.password)
+  const differentEnv = {
+    ...portalEnv,
+    portalEnvironmentConfig: {
+      ...portalEnv.portalEnvironmentConfig,
+      password: 'newPass3'
+    },
+    environmentName: 'irb'
+  }
+  rerender(<MockSuperuserProvider>
```
MatthewBemis, line 42: nice, TIL about `rerender`

MatthewBemis (approved):

ui-admin/src/portal/PortalEnvConfigView.tsx
devonbush: Yes, `useUpdateEffect` will only trigger when the dependencies specified are updated (different from `useEffect` which will trigger both after initial render and on dependency updates).  So, line 32 will only run if the portal shortcode or environment name changes.  If one of those changes, we can safely assume that the portalEnvironmentConfig has changed to reflect the updated environment.  

An alternate way you could structure this would be to have `portalEnv.portalEnvironmentConfig.id` be the dependency.  The reason I went with environmentName and shortcode instead was that a common pattern in our UX is "what should this component do when the user switches the environment/portal?" since those are the dropdowns provided.  I could see either way working

devonbush: I could also see trying to write an `usePortalEnvUpdateEffect` hook to make that pattern more explicit

denniscunningham (approved):
Thanks for the explanation. Very helpful.

PR #652, opened 2023-11-27
JN-736 Extract & upload portal configurations

devonbush:
#### DESCRIPTION

This enables downloading of much of a portal's configuration (the settings, surveys, site content, images, consents, etc...) as a single zip file.  This zip file has an internal structure that largely mimics the file structure of the populate seed files we already use.  This zip file can then be uploaded to populate the portal configuration.  This is currently conceived as a *developer-only* feature.   The main use case is to extract portal configurations from the production server, so they can be spun up on either a local dev instance or the demo server  (e.g. get HeartHive's current prod configuration, and populate it locally to test a new feature on to ensure compatibility).  

In addition to some minor ancillary cleanups,  now the same ObjectMapper is used across all our modules (previously, core, ApiAdmin, ApiParticipant each configured their ObjectMapper separately).  This was originally done to enable faster iteration on the ApiAdmin mapper.  But we haven't really needed to change that mapper config, so it makes sense to consolidate them all to ensure tests all run using the same mapper config, and the Spring beans are easier to manage.

Not included in this PR:
1. copying of dashboard alert configs
2. copying of kit type configuration

1. restart ApiAdminApp
4. repopulate OurHealth `./scripts/populate_portal.sh ourhealth`
5. go to `https://localhost:3000/populate/extractPortal`
6. type 'ourhealth' and hit enter
7. observe a zip file is downloaded.
8. extract the zip file, and confirm the contents are sensibly arranged
9. Go to `https://localhost:3000/populate/portal`
10. Choose the zip file that you just downloaded
11. select "Yes" for "overwrite"
12. Hit "populate"
13. Confirm the populate succeeds
14. You can now browse the ourhealth study and confirm most things appear just as they were before.


devonbush:
@connorlbark  I'll set up a time next year to walk you through this PR -- it's definitely a lot.

devonbush:
The Sonar hotspot is because I used a psuedorandom generator for the temp path file name.  This is fine since this data isn't privileged and knowing the location of the file wouldn't help any attacker (unless they already had full access to our server)

populate/src/main/java/bio/terra/pearl/populate/service/extract/README.md
```diff
@@ -0,0 +1,9 @@
+### Extraction Services
+
+Extraction is the inverse of populate.  While populate takes a pointer to a file, and then populates the 
+database with the configuration from the files, extraction takes portal configuration from the database
+and writes it to a zip file.  The folder structure and the file contents that result from the extraction
+should be as close as possible to the original files that were used to populate the database.
+
+Most importantly, the files produced by extraction should be easily human readable and editable.  this will facilitate using
+these files as a starting point for test cases.
```
connorlbark, line 9: Per zoom discussion, I think we should add some documentation for the steps a user needs to take to extract and populate a portal from the UI. Also, we should add documentation for what adding a new populator/extractor would look like for a new feature, as it will likely be relatively common.

api-admin/src/main/resources/application.yml
```diff
@@ -96,8 +96,8 @@ spring:
 
   servlet:
     multipart:
-      max-file-size: 10MB
-      max-request-size: 10MB
+      max-file-size: 50MB
+      max-request-size: 50MB
```
MatthewBemis, line 107: I can't think of a reason why users would need to upload 50MB images, but since this is global and the site media upload endpoint will now support 50MB too, should we update the text/validation in [SiteImageUploadModal.tsx](https://github.com/broadinstitute/juniper/blob/development/ui-admin/src/portal/images/SiteImageUploadModal.tsx)?

I could also see the argument for keeping the frontend logic capped at 10MB to discourage large image/document uploads and keep participant sites mobile-friendly.

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -1173,6 +1173,41 @@ paths:
                 type: object
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/internal/v1/populate/portal/upload:
```
MatthewBemis, line 1176: The addition of this second populate endpoint makes the original `POST /api/internal/v1/populate/portal` endpoint a little more mysterious. Could you also update the summary in the swagger spec for that endpoint to make its purpose more clear?

populate/src/main/java/bio/terra/pearl/populate/service/extract/StudyExtractor.java
```diff
@@ -0,0 +1,139 @@
...
+        List<NotificationConfig> notificationConfigs = notificationConfigService.findByStudyEnvironmentId(studyEnv.getId());
+        for (NotificationConfig config : notificationConfigs) {;
+            NotificationConfigPopDto configPopDto = new NotificationConfigPopDto();
+            BeanUtils.copyProperties(config, configPopDto, "id", "studyEnvironmentId", "portalEnvironmentId", "emailTemplateId");
+            String filename = "../../" + context.getFileNameForEntity(config.getEmailTemplateId());
+            configPopDto.setPopulateFileName(filename);
+            studyEnvPopDto.getNotificationConfigDtos().add(configPopDto);
+        }
+
+        // TODO extract dashboard configs
```
MatthewBemis, line 119: I'm curious, why were dashboard configs and kit configuration types left out of this PR?

MatthewBemis (approved):
Overall looks good. Worked great when tested!

Just a few comments, nothing blocking.

populate/src/main/java/bio/terra/pearl/populate/dto/site/SiteContentPopDto.java
```diff
@@ -2,14 +2,15 @@
...
-import java.util.Set;
+
+import java.util.ArrayList;
+import java.util.List;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
 
 @Getter @Setter @NoArgsConstructor
 public class SiteContentPopDto extends SiteContent implements FilePopulatable {
```
connorlbark, line 13: outside of the scope of this PR for sure, but I noticed there were a lot of fields (e.g. "sections" from the extended SiteContent) which are serialized but seem to have no purpose now with the DTO. Is there any way to flag them to not serialize? Or is it more work than it's worth?

connorlbark (approved):
Short of going line by line, it looks good and works great! I think there's a lot of potential with a feature like this.

api-admin/src/main/resources/application.yml
devonbush: good question -- yeah, let's leave the frontend cap for now

populate/src/main/java/bio/terra/pearl/populate/service/extract/StudyExtractor.java
devonbush: just didn't have time, and wanted to get this in so I could update the demo and development stuff with prod HeartHive configs.  I'll make a ticket for it

populate/src/main/java/bio/terra/pearl/populate/dto/site/SiteContentPopDto.java
devonbush: Great question.  Our models, like SiteContent, serve double (or even triple) duty right now.  they serve as 
1. the definition of what's stored in the DB
2. the definition of what's stored in the populate flat files
3. the definition of what's sent from the server to the client. 

In this case, the "sections" aren't used in populate serializing, but they are still very much used for sending from the server to the client. 

We do it this way because it was faster and simpler to get the codebase started that way, and at the beginning, there wasn't much difference between all the differeny uses.  As the codebase matures, we'll likely want to create explicit DTOs for each model for each of those purposes, instead of using bits and pieces for each.

PR #653, opened 2023-11-27
[JN-676] Spring Boot 3

MatthewBemis:
#### DESCRIPTION

This [DSP blog post](https://broadworkbench.atlassian.net/wiki/spaces/DSPBLOG/blog/2023/11/16/2906521657/The+Great+Spring+Boot+3+Yak+Shaving+Expedition) about upgrading to Spring Boot 3 was helpful.

I did run into a couple of wrinkles getting us upgraded:

Wrinkle #1 (resolved): 
The new version of TCL assumes that you do tracing with OpenTelemetry. Even though we don't use any of the tracing utilities from TCL, I still had to bring in a couple of related OpenTelemetry packages to get the app to successfully boot up. Tracing is really nice to have and eventually we'll want to think about doing it, but that time is not now because it requires some exploration into how this will work for Azure deployments. So far, DSP services have only ever sent traces to Google Cloud Trace.

Wrinkle #2 (unresolved):
Right now, TDR only publishes a javax client. We're one of the first consumers of the TDR client that's asking for an upgrade (the other consumer is WDS). It's not a light lift for them to migrate TDR, so we're stuck at the moment. 

We have 3 potential paths forward:
1. Hold the Spring Boot 3 upgrade until TDR migrates to Jakarta
2. Publish our own Jakarta TDR client and use that until it's officially published by the Jade team (see precedent [here](https://broadinstitute.slack.com/archives/CD4HBRFMG/p1701203814132849))
3. Disable TDR integration for now....


api-admin/build.gradle
```diff
@@ -14,18 +14,18 @@ apply from: 'publishing.gradle'
...
+    implementation 'bio.terra:terra-common-lib:0.1.2-SNAPSHOT'
     implementation project(':core')
     implementation project(':populate')
     implementation 'org.apache.commons:commons-dbcp2'
     implementation 'org.apache.commons:commons-text:1.10.0'
     implementation 'commons-beanutils:commons-beanutils:1.9.4'
-    implementation 'org.springframework.boot:spring-boot-starter:2.7.10'
-    implementation 'org.springframework.boot:spring-boot-starter-data-jdbc:2.7.10'
-    implementation 'org.springframework.boot:spring-boot-starter-web:2.7.10'
+    implementation 'org.springframework.boot:spring-boot-starter:3.1.2'
```
MatthewBemis, line 23: Upgrading to 3.1.2 right now because TCL doesn't work with 3.1.5 yet. TCL logging utilities use some deprecated code from Logback which was removed in the version of Logback that comes along with Spring Boot 3.1.5.

PR #654, opened 2023-11-27
[JN-651] Switch to Lombok slf4j annotation

MatthewBemis:
#### DESCRIPTION

For code consistency, this switches all of our loggers to use the @Slf4j Lombok annotation.

* Start up Admin API and Participant API
* Perform some actions that trigger API calls to each app
* Verify that application logging is as-expected


devonbush (approved):
Nice,  looking through this, I see this fixed at least 3 places where we were instantiating the logger with the wrong class

denniscunningham (approved):

PR #656, opened 2023-11-29
JN-746 fixing consent history display

devonbush:
#### DESCRIPTION

The consent form version selector was never wired up, and Rachel from HeartHive needs it to get back to a previous version of one of their consents.  So this adds the needed endpoints and updates the version selector to be able to load either consents or surveys.  It also updates the consent form editor to fetch the consent from the server, rather than relying on the portal 'god object'

image

This adds a new interface "PortalAttached" to enable some auth code sharing.  

There definitely is some room to more cleanly factor the UI code--I don't like the "isConsentForm" boolean.  I suspect the better approach will be to break the Api object into modules, and then the controls will just pass in the Api module for forms/surveys as needed, and that will enable a lot more code sharing between the consent and survey editors.  But that's a bit beyond the scope of this PR.  

1. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms/consentForms/oh_oh_consent?readOnly=false`
2. make a trivial change to the form, and save it
3. click the "history" button
4. confirm the past version is available, and can be previewed

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConsentFormController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConsentFormController.java
@@ -29,6 +29,20 @@ public class ConsentFormController implements ConsentFormApi {
   }
 
   @Override
+  public ResponseEntity<Object> get(String portalShortcode, String stableId, Integer version) {
+    AdminUser adminUser = authUtilService.requireAdminUser(request);
+    ConsentForm form = consentFormExtService.get(portalShortcode, stableId, version, adminUser);
+    return ResponseEntity.ok(form);
+  }
+
+  @Override
+  public ResponseEntity<Object> getAllVersions(String portalShortcode, String stableId) {
+    AdminUser adminUser = authUtilService.requireAdminUser(request);
+    return ResponseEntity.ok(
+        consentFormExtService.listVersions(portalShortcode, stableId, adminUser));
+  }
+
+  @Override
   public ResponseEntity<Object> newVersion(String portalShortcode, String stableId, Object body) {
     AdminUser adminUser = authUtilService.requireAdminUser(request);
     ConsentForm consentForm = objectMapper.convertValue(body, ConsentForm.class);
--- a/ui-admin/src/study/surveys/ConsentView.tsx
+++ b/ui-admin/src/study/surveys/ConsentView.tsx
@@ -11,8 +11,8 @@ import Api, {
 
 import { failureNotification, successNotification } from 'util/notifications'
 import SurveyEditorView from './SurveyEditorView'
-import {useLoadingEffect} from "../../api/api-utils";
-import LoadingSpinner from "../../util/LoadingSpinner";
+import { useLoadingEffect } from '../../api/api-utils'
+import LoadingSpinner from '../../util/LoadingSpinner'
 
 /** Handles logic for updating study environment surveys */
 function RawConsentView({ studyEnvContext, consent, readOnly = false }:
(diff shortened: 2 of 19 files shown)


denniscunningham (changes requested):
Looks like you have some JS/TS lint issues.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConsentFormController.java
```diff
@@ -28,6 +28,20 @@ public ConsentFormController(
...
 
+  @Override
+  public ResponseEntity<Object> get(String portalShortcode, String stableId, Integer version) {
+    AdminUser adminUser = authUtilService.requireAdminUser(request);
+    ConsentForm form = consentFormExtService.get(portalShortcode, stableId, version, adminUser);
+    return ResponseEntity.ok(form);
+  }
+
+  @Override
+  public ResponseEntity<Object> getAllVersions(String portalShortcode, String stableId) {
```
denniscunningham, line 39: It feels like the implementation is customized to a particular use case that is at odds with the restful interface. IOW, this `get` does not return the whole consent form object.
I'm not sure I have a great solution other than a customized endpoint (e.g. `/api/portals/v1/{portalShortcode}/consentForms/{stableId}/metadata`) but there is something uncomfortable about this implementation.
Does that make sense, or is this common restful practice that I somehow missed?

ui-admin/src/study/surveys/ConsentView.tsx
```diff
@@ -55,32 +57,65 @@ function RawConsentView({ studyEnvContext, consent, readOnly = false }:
...
+
+/** read consent-form-related url params */
+export const useSurveyParams = () => {
+  const params = useParams<ConsentParamsT>()
+  const [searchParams] = useSearchParams()
+  const isReadOnly = searchParams.get('readOnly') === 'true'
+  const version = params.version ? parseInt(params.version) : undefined
+
+  return { isReadOnly, version, stableId: params.consentStableId }
 }
```
MatthewBemis: It doesn't look like this new hook ended up being used anywhere.

Regardless, I think `ConsentView` also needs to account for read-only study envs before passing `isReadOnly` into `RawConsentView `

ui-admin/src/study/surveys/ConsentView.tsx
```diff
@@ -55,32 +57,65 @@ function RawConsentView({ studyEnvContext, consent, readOnly = false }:
...
 
 export type ConsentParamsT = StudyParams & {
-  consentStableId: string
+  consentStableId: string,
+  version: string
+}
+
+/** loads a consentForm */
+export const useLoadedConsentForm = (portalShortcode: string, stableId: string, version: number) => {
+  const [form, setForm] = useState<ConsentForm | undefined>()
```
denniscunningham: Just for my information... What does `useState <ConsentForm | undefined>()` mean, and how is it different from `useState <ConsentForm>({})`?

MatthewBemis (approved):
Left one comment, but pre-approving.

denniscunningham (approved):
A couple comments/questions.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConsentFormController.java
devonbush: yeah, that's a good point -- I think you're right that a custom endpoint is more correct

ui-admin/src/study/surveys/ConsentView.tsx
devonbush: it's the same -- will fix, thanks

PR #657, opened 2023-12-01
JN-720: Refactoring kit statuses

devonbush:
#### DESCRIPTION

I started investigating including kit status into the participant data export, and using it to send reminders, and immediately got lost in the multiple layers of "statuses". So this PR is an effort to simplify and clean up our status code.  This has no functional changes except the addition of the "deactivated" tab (which was much easier to add as a result of this cleanup) 

The major change is that our KitRequestStatus is now a superset of statuses we get from Pepper, with a one-to-one mapping of statuses.  This enables our application logic to use the KitRequestStatus directly, instead of having to dive into the pepper data object.  I expect we will want to start extracting more fields like this from the pepper object as we add more admin capability.

image

This also does some minor prep for supporting non-Pepper manual kit tracking, by renaming KitRequest.dsmStatus to 'externalRequest' -- both so it's clear that's the representation of the entire request, and so that could represent state from somewhere other than pepper.

For the migration script, correctness of the status mapping is actually not that important to test (I lightly tested it locally), since the statuses will all be resynced nightly anyway, so the migration just ensures the data will look reasonable in the immediate aftermath of the release.   

1. Restart ApiAdminApp
2. repopulate OurHealth
3. go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/requested/created`
4. confirm you see two kits, one 'created', one 'deactivated'
5. refresh the kit statuses a few times, and confirm the kit statuses cycle appropriately.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
@@ -81,10 +81,10 @@ public class KitRequestService extends CrudService<KitRequest, KitRequestDao> {
 
         // send kit request to DSM
         try {
-            PepperKitRequest dsmKitStatus = pepperDSMClient.sendKitRequest(studyShortcode, enrollee, kitRequest, pepperKitAddress);
+            PepperKit dsmKitStatus = pepperDSMClient.sendKitRequest(studyShortcode, enrollee, kitRequest, pepperKitAddress);
             // write out the PepperKitStatus as a string for storage
             String pepperRequestJson = objectMapper.writeValueAsString(dsmKitStatus);
-            kitRequest.setExternalRequest(pepperRequestJson);
+            kitRequest.setExternalKit(pepperRequestJson);
         } catch (PepperParseException e) {
             // response was successful, but we got unexpected format back from pepper
             // we want to log the error, but still continue on to saving the kit
@@ -145,12 +145,12 @@ public class KitRequestService extends CrudService<KitRequest, KitRequestDao> {
      * Do _NOT_ call this repeatedly for a collection of kits. Use a bulk operation instead to avoid overwhelming DSM.
      */
     @Transactional
-    public PepperKitRequest syncKitStatusFromPepper(UUID kitId) throws PepperParseException, PepperApiException {
+    public PepperKit syncKitStatusFromPepper(UUID kitId) throws PepperParseException, PepperApiException {
         KitRequest kitRequest = dao.find(kitId).orElseThrow(() -> new NotFoundException("Kit request not found"));
         var pepperKitStatus = pepperDSMClient.fetchKitStatus(kitId);
         try {
-            kitRequest.setExternalRequest(objectMapper.writeValueAsString(pepperKitStatus));
-            kitRequest.setExternalRequestFetchedAt(Instant.now());
+            kitRequest.setExternalKit(objectMapper.writeValueAsString(pepperKitStatus));
+            kitRequest.setExternalKitFetchedAt(Instant.now());
         } catch (JsonProcessingException e) {
             throw new RuntimeException("Could not parse JSON response from DSM", e);
         }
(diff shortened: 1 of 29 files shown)


core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -145,12 +145,12 @@ public Collection<KitRequest> getSampleKitsByStudyEnvironment(StudyEnvironment s
      * Do _NOT_ call this repeatedly for a collection of kits. Use a bulk operation instead to avoid overwhelming DSM.
      */
     @Transactional
-    public PepperKitStatus syncKitStatusFromPepper(UUID kitId) throws PepperParseException, PepperApiException {
-        var kitRequest = dao.find(kitId).orElseThrow(() -> new NotFoundException("Kit request not found"));
+    public PepperKitRequest syncKitStatusFromPepper(UUID kitId) throws PepperParseException, PepperApiException {
+        KitRequest kitRequest = dao.find(kitId).orElseThrow(() -> new NotFoundException("Kit request not found"));
         var pepperKitStatus = pepperDSMClient.fetchKitStatus(kitId);
```
denniscunningham, line 150: Thank you for fixing all these vars.

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/PepperKitRequest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.Set;
+
+/** holds kit request data as it is received from Pepper */
+@Getter @Setter
+@SuperBuilder @NoArgsConstructor
+@EqualsAndHashCode
+@ToString
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class PepperKitRequest {
```
denniscunningham: I think capturing that this is status or a response in the class name is important. The default reading of `PepperKitRequest` is as the payload for a Pepper request.

core/src/main/resources/db/changelog/changesets/2023_11_30_kit_status.yaml
```diff
@@ -0,0 +1,25 @@
...
+                WHEN cast(dsm_status AS json)->>'currentStatus' = 'sent' THEN 'SENT'
+                WHEN cast(dsm_status AS json)->>'currentStatus' = 'received' THEN 'RECEIVED'
+                WHEN cast(dsm_status AS json)->>'currentStatus' = 'error' THEN 'ERRORED'
+                WHEN cast(dsm_status AS json)->>'currentStatus' = 'deactivated' THEN 'DEACTIVATED'
+                WHEN cast(dsm_status AS json)->>'currentStatus' = NULL THEN NULL
+                ELSE 'UNKNOWN'
+                END);
+        - renameColumn:
+            oldColumnName: dsm_status
+            newColumnName: external_request
```
denniscunningham: As above, 'request' denotes something different than what is stored.

ui-admin/src/study/participants/KitStatusCell.tsx
```diff
@@ -12,19 +12,20 @@ export default function KitStatusCell(
   { kitRequest: KitRequest, infoPlacement?: Placement }
 ) {
   const infoTexts: Record<string, string> = {
-    'kit without label': 'Kit request has been received',
-    'queue': 'Shipping label has been printed',
-    'sent': 'Kit has been sent to the participant',
-    'received': 'Kit has been returned by the participant',
-    'error': 'There was a problem fulfilling this kit request',
-    'deactivated': 'Kit is deactivated'
+    'CREATED': 'Kit has not yet shipped',
```
denniscunningham: I think the original text is more accurate.
A workflow of request received -> preparing to ship -> sent... makes logical sense and is consistent with general on-line ordering practices most people are familiar with.

denniscunningham (changes requested):
Thanks for making these changes.
A couple comments about terminology. Seems like a nit but hard for me to mentally work around. Interested in your thoughts.

ui-admin/src/study/participants/KitStatusCell.tsx
devonbush: I wanted to avoid confusion of using 'received' for both 'received by GP' and 'received by participant'. So maybe `Kit request received, not yet shipped`?

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/PepperKitRequest.java
devonbush: I see your point -- I was trying to have the class name capture "this is the Pepper representation of a kit request", but you're right that since 'request' is overloaded, it's confusing.  Maybe 'PepperKitState' just to avoid overloading 'status'

devonbush: or just PepperKit?

denniscunningham: I had not thought of that, but I think `PepperKit` works.

ui-admin/src/study/participants/KitStatusCell.tsx
denniscunningham: Ah, I see your point.
Yup, that works or something like "... awaiting processing"

denniscunningham (approved):

MatthewBemis (approved):

PR #659, opened 2023-12-01
[JN-732] Add pdf support

MatthewBemis:
#### DESCRIPTION

This adds pdf as an allowed filetype for the site file uploader. It also makes a few user-facing changes, like renaming "images" to "media" and changing some styling to be more consistent with other pages. 

Right now I only updated the "media" language where it was user facing. If we decide that media is the right word to use, it would make sense to propagate that language throughout the rest of our code. I just didn't want to be too quick to make that change before reaching consensus.

Screenshot 2023-12-01 at 1 35 17 PM

Screenshot 2023-12-01 at 1 34 56 PM

* Launch admin API and UI
* Try uploading a PDF: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/media
* Verify that the regex still only supports image types + pdf

### DIFF
--- a/ui-admin/src/portal/images/SiteImageUploadModal.tsx
+++ b/ui-admin/src/portal/images/SiteImageUploadModal.tsx
@@ -10,9 +10,9 @@ import { LoadedPortalContextT } from '../PortalProvider'
 
 
 export const allowedImageTypes = ['gif', 'ico', 'jpeg', 'jpg', 'png', 'svg', 'webp']
-export const allowedApplicationTypes = ['pdf']
+export const allowedDocumentTypes = ['pdf']
 const FILE_TYPE_REGEX = new RegExp(
-  `^(?:image\\/(?:${allowedImageTypes.join('|')}))|(?:application\\/(?:${allowedApplicationTypes.join('|')}))$`
+  `^(?:image\\/(?:${allowedImageTypes.join('|')}))|(?:application\\/(?:${allowedDocumentTypes.join('|')}))$`
 )
 /** Renders a modal for an admin to submit a sample collection kit request. */
 export default function SiteImageUploadModal({
@@ -62,7 +62,7 @@ export default function SiteImageUploadModal({
     <Modal.Body>
       <form onSubmit={e => e.preventDefault()}>
         <div>
-          <p>Supported extensions are {[...allowedImageTypes, ...allowedApplicationTypes].join(', ')}.
+          <p>Supported extensions are {[...allowedImageTypes, ...allowedDocumentTypes].join(', ')}.
             Maximum size is 10MB</p>
           File:
           <div>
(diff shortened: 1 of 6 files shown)


ui-admin/src/study/StudyEnvironmentRouter.tsx
```diff
@@ -91,7 +91,7 @@ function StudyEnvironmentRouter({ study }: {study: Study}) {
       <Route path="participants/*" element={<ParticipantsRouter studyEnvContext={studyEnvContext}/>}/>
       <Route path="kits/*" element={<KitsRouter studyEnvContext={studyEnvContext}/>}/>
       <Route path="siteContent" element={<SiteContentLoader portalEnvContext={portalEnvContext}/>}/>
-      <Route path="images" element={<SiteImageList portalContext={portalContext} portalEnv={portalEnv}/>}/>
+      <Route path="media" element={<SiteImageList portalContext={portalContext} portalEnv={portalEnv}/>}/>
```
MatthewBemis, line 94: Normally I wouldn't be so quick to change a URL but this doesn't seem like a mission-critical path

denniscunningham (commented):
Look good.
Dumb question... where will PDFs be used? The Jira ticket was not clear on that.
(Might matter to my thoughts on where they should be found after uploading.)

MatthewBemis:
@denniscunningham I think they would likely be used in a number of places- referenced from survey questions, site nav links, etc. I believe [this Slack thread](https://broadinstitute.slack.com/archives/C042QNA6C3S/p1700070084523079) is what prompted this feature request.

ui-admin/src/portal/images/SiteImageUploadModal.tsx
```diff
@@ -9,7 +9,11 @@ import { Button } from 'components/forms/Button'
 import { LoadedPortalContextT } from '../PortalProvider'
 
 
-const FLE_TYPE_REGEX = /image\/(gif|ico|jpeg|jpg|png|svg|webp)/
+export const allowedImageTypes = ['gif', 'ico', 'jpeg', 'jpg', 'png', 'svg', 'webp']
+export const allowedApplicationTypes = ['pdf']
```
devonbush:
```suggestion
export const allowedDocumentTypes = ['pdf']
```

ui-admin/src/study/StudyEnvironmentRouter.tsx
devonbush: concur, I just did a quick check of user docs/smoke test checklists and this was not referenced

devonbush (approved):
Looks good, one minor variable name suggestion

denniscunningham:
@MatthewBemis Thx for the link. That explains the need.
How would users create a link to an uploaded PDF? Or is that for a future PR?

devonbush:
HeartHive currently uses an uploaded pdf (that I snuck in by changing the file extension) for their privacy policy (See the link to https://thehearthive.org/api/public/portals/v1/hearthive/env/live/siteImages/2/privacy_notice.pdf about 2/3 of the way down on https://thehearthive.org/).  They can construct links it the same way that images are linked from the participant ui.  We don't have a UI for the making the construction of those links easy yet, so it's something that Whitney would probably have to help with in the short term, and is outside the scope of this ticket

denniscunningham (approved):

PR #660, opened 2023-12-01
JN-725: editing the nav logo

devonbush:
#### DESCRIPTION
This enables editing of the nav logo and the primary brand color from the SiteContentEditor
image
Clicking the "Branding" button pulls up a dialog
image

This is bare-bones, no error-checking, mainly intended for internal use when setting up a study just so we don't have to go to the database.  I thought about a color picker for brand color, but I'm not sure it would actually be used -- AFAIK all our past customers already had a brand color (or Erin chose one), and so exact typing is strongly preferred to picking.  

To make testing easier, I added a link to the participant view from the SiteContentEditor.  If you press it with unsaved changes, an alert pops up letting you know that the preview will not show unsaved changes.

image

This SiteContent editor is starting to get very noisy, I suspect it doesn't have room for many more controls before it will need some consolidation/rearrangement

1. Go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/siteContent
2. click the "Branding" button
3. edit the primary brand color, updating it to, say, `#33bb44`
4. pick a different nav logo image
5. click 'ok'
6. save 
7. click the participant view link
8. confirm the new primary brand color and logo is reflected in the UI

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17851,6 +17851,15 @@
         }
       }
     },
+    "node_modules/react-colorful": {
+      "version": "5.6.1",
+      "resolved": "https://registry.npmjs.org/react-colorful/-/react-colorful-5.6.1.tgz",
+      "integrity": "sha512-1exovf0uGTGyq5mXQT0zgQ80uvj2PCwvF8zY1RN9/vbJVSjSo3fsB/4L3ObbF7u70NduSiK4xu4Y6q1MHoUGEw==",
+      "peerDependencies": {
+        "react": ">=16.8.0",
+        "react-dom": ">=16.8.0"
+      }
+    },
     "node_modules/react-dev-utils": {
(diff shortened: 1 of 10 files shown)


denniscunningham (changes requested):
The create button on the "Create Study" modal is a bit wonky.
It seems like I have to change the portal selection for it to take. (The default choice is not recognized, and reselecting it does not help.)

MatthewBemis (approved):
Seeing the same Create Study modal behavior as Dennis. Otherwise, the branding stuff itself looks good and worked well.

Agreed that the page is starting to get very crowded, but it's fine for the moment.

devonbush:
Great catch on the study creation bug -- I wasn't handling initial/current values correctly

denniscunningham (approved):

PR #663, opened 2023-12-04
[JN-714] Publish dashboard alerts

MatthewBemis:
#### DESCRIPTION

This adds the alert publishing functionality that was missing from #642. Note that this is a PR into #642, not into `development`.

https://github.com/broadinstitute/juniper/assets/7257391/b2d17b37-7527-4653-aa06-07250d35c507

* Launch admin API and UI
* Update the default `NO_ACTIVITIES_REMAIN` alert in sandbox
* View the publish diff view for sandbox vs IRB: https://localhost:3000/ourhealth/studies/ourheart/diff/sandbox/irb
* Confirm that you see your changes represented in the diff
* Select some of the changes and publish
* Confirm that they show up in the IRB environment

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/publishing/PortalDiffService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/publishing/PortalDiffService.java
@@ -90,14 +90,14 @@ public class PortalDiffService {
 
         List<ParticipantDashboardAlert> destAlerts = new ArrayList<>(destEnv.getParticipantDashboardAlerts());
         List<ParticipantDashboardAlert> sourceAlerts = new ArrayList<>(sourceEnv.getParticipantDashboardAlerts());
-        List<ParticipantDashboardAlertChange> participantDashboardAlertChanges = diffAlertLists(sourceAlerts, destAlerts);
+        List<ParticipantDashboardAlertChange> alertChangeLists = diffAlertLists(sourceAlerts, destAlerts);
 
         return new PortalEnvironmentChange(
                 siteContentRecord,
                 envConfigChanges,
                 preRegRecord,
                 notificationConfigChanges,
-                participantDashboardAlertChanges,
+                alertChangeLists,
                 studyEnvChanges
         );
     }
@@ -106,24 +106,23 @@ public class PortalDiffService {
             List<ParticipantDashboardAlert> sourceAlerts,
             List<ParticipantDashboardAlert> destAlerts) throws Exception {
         List<ParticipantDashboardAlert> unmatchedDestAlerts = new ArrayList<>(destAlerts);
-        List<ParticipantDashboardAlertChange> changedAlerts = new ArrayList<>();
+        List<ParticipantDashboardAlertChange> alertChangeLists = new ArrayList<>();
         for (ParticipantDashboardAlert sourceAlert : sourceAlerts) {
-            ParticipantDashboardAlert matchedAlert = unmatchedDestAlerts.stream().filter(
-                            destAlert -> sourceAlert.getTrigger().equals(destAlert.getTrigger()))
-                    .findAny().orElse(null);
+            ParticipantDashboardAlert matchedAlert = unmatchedDestAlerts.stream().filter(destAlert ->
+                            sourceAlert.getTrigger().equals(destAlert.getTrigger())).findAny().orElse(null);
             if (matchedAlert == null) {
                 List<ConfigChange> newAlert = ConfigChange.allChanges(sourceAlert, null, CONFIG_IGNORE_PROPS);
-                changedAlerts.add(new ParticipantDashboardAlertChange(sourceAlert.getTrigger(), newAlert));
+                alertChangeLists.add(new ParticipantDashboardAlertChange(sourceAlert.getTrigger(), newAlert));
             } else {
                 unmatchedDestAlerts.remove(matchedAlert);
-                List<ConfigChange> changedAlert = ConfigChange.allChanges(sourceAlert, matchedAlert, CONFIG_IGNORE_PROPS);
-                if(!changedAlert.isEmpty()) {
-                    changedAlerts.add(new ParticipantDashboardAlertChange(sourceAlert.getTrigger(), changedAlert));
+                List<ConfigChange> alertChanges = ConfigChange.allChanges(sourceAlert, matchedAlert, CONFIG_IGNORE_PROPS);
+                if(!alertChanges.isEmpty()) {
+                    alertChangeLists.add(new ParticipantDashboardAlertChange(sourceAlert.getTrigger(), alertChanges));
                 }
             }
         }
 
-        return changedAlerts;
+        return alertChangeLists;
     }
 
     protected PortalEnvironment loadPortalEnvForProcessing(String shortcode, EnvironmentName envName) {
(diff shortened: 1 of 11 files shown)


core/src/main/java/bio/terra/pearl/core/service/publishing/PortalPublishingService.java
```diff
@@ -142,4 +150,44 @@ protected void applyChangesToNotificationConfigs(PortalEnvironment destEnv, List
...
+                applyAlertChanges(alert, change.changes());
+                portalDashboardConfigService.update(alert);
+            }
+        }
+    }
+
+    protected void applyAlertChanges(ParticipantDashboardAlert alert, List<ConfigChange> changes) {
+        try {
+            for (ConfigChange alertChange : changes) {
+                //The 'type' property is an enum, so we need to handle it differently than the other properties
```
devonbush: Ah, great catch.  I just realized this impacts notification configs too -- if you attempt to change the event type, it will error.  

Instead of hardcoding this, let's make a new utility method "setPropertyEnumSafe" that uses reflection to detect whether the target property is an enum, and if so, convert the string value to an enum before invoking PropertyUtils.setProperty.  

I believe you can get the target class by using `PropertyUtils.getPropertyDescriptor(alert.getClass(), alertChange.propertyName).getPropertyType()`

Then you can upgrade `PublishingUtils.applyChangesToVersionedConfig` to use that method

devonbush (changes requested):
Great stuff,  requested change will fix related bug with respect to NotificationConfigs

core/src/main/java/bio/terra/pearl/core/service/publishing/PortalDiffService.java
```diff
@@ -83,15 +88,43 @@ public PortalEnvironmentChange diffPortalEnvs(PortalEnvironment sourceEnv, Porta
...
                 preRegRecord,
                 notificationConfigChanges,
+                alertChangeLists,
                 studyEnvChanges
         );
     }
 
+    protected List<ParticipantDashboardAlertChange> diffAlertLists(
+            List<ParticipantDashboardAlert> sourceAlerts,
+            List<ParticipantDashboardAlert> destAlerts) throws Exception {
```
denniscunningham: What exception does this throw and where is it handled?

core/src/main/java/bio/terra/pearl/core/service/publishing/PortalDiffService.java
```diff
@@ -83,15 +88,43 @@ public PortalEnvironmentChange diffPortalEnvs(PortalEnvironment sourceEnv, Porta
...
+        List<ParticipantDashboardAlert> unmatchedDestAlerts = new ArrayList<>(destAlerts);
```
denniscunningham, line 110: Might be more efficient to represent `unmatchedDestAlerts` as a map of trigger to alert.

core/src/main/java/bio/terra/pearl/core/service/publishing/PortalPublishingService.java
```diff
@@ -142,4 +150,39 @@ protected void applyChangesToNotificationConfigs(PortalEnvironment destEnv, List
         }
     }
 
+    protected void applyChangesToParticipantDashboardAlerts(PortalEnvironment destEnv, List<ParticipantDashboardAlertChange> changes) throws Exception {
```
denniscunningham: What exception does this throw and where is it handled?

core/src/main/java/bio/terra/pearl/core/service/publishing/PortalPublishingService.java
```diff
@@ -142,4 +150,39 @@ protected void applyChangesToNotificationConfigs(PortalEnvironment destEnv, List
...
+        for(ParticipantDashboardAlertChange change : changes) {
+            Optional<ParticipantDashboardAlert> destAlert = portalDashboardConfigService.findByPortalEnvIdAndTrigger(destEnv.getId(), change.trigger());
+            if(destAlert.isEmpty()) {
+                // The alert doesn't exist in the dest env yet, so default all the required fields before
+                // applying the changes from the change list
+                ParticipantDashboardAlert newAlert = ParticipantDashboardAlert.builder()
```
denniscunningham, line 159: Nit: It might be helpful to pull the default setup into it's own method, to emphasize this is the default, and to consolidate the code on lines 168-173.

core/src/main/java/bio/terra/pearl/core/service/publishing/PublishingUtils.java
```diff
@@ -57,4 +57,12 @@ public static <T extends BaseEntity & Versioned, D extends BaseVersionedJdbiDao<
             service.assignPublishedVersion(entity.getId());
         }
     }
+
+    public static void setPropertyEnumSafe(Object object, String propertyName, Object newValue) throws Exception {
+        if(object.getClass().getDeclaredField(propertyName).getType().isEnum()) {
```
denniscunningham, line 62: Nit: Totally a style thing, but this might read better with an assignment in the `if` clause and a single call to `setProperty`.

denniscunningham (commented):
Looks good!
A couple comments and questions.

core/src/main/java/bio/terra/pearl/core/service/publishing/PublishingUtils.java
```diff
@@ -57,4 +57,12 @@ public static <T extends BaseEntity & Versioned, D extends BaseVersionedJdbiDao<
...
+            PropertyUtils.setProperty(object, propertyName, Enum.valueOf((Class<Enum>) PropertyUtils.getPropertyType(object, propertyName), newValue.toString()));
```
devonbush, line 63: I would pull this out into a helper method `public static <T> stringToEnum(String value, Class<T extends Enum> class)`

devonbush (approved):
looks good, thanks for adding that generalized enum setter.  Just one suggestion

core/src/main/java/bio/terra/pearl/core/service/publishing/PortalDiffService.java
denniscunningham: Just wanted to followup on this since I realize from another PR that my comments about exceptions may not be clear. That seems likely here, so my apologies for not being more specific.

This is my POV, so feel free to share your thoughts with me if they differ, but throwing `Exception` feels odd here. Library methods will often list the exceptions they throw since it is a way to communicate specific error conditions to callers, and if that's your aim here then I would suggest listing the specific exceptions you want callers to be aware of. You also must list any exceptions that do not inherit from RuntimeException -- that is, checked exceptions -- but again it is best to list the specific exceptions.
Many of our methods throw exceptions and do not list them,  since they do not expect any specific handling and are happy with the way those exceptions are handled in the GlobalExceptionHandler. If that's the case here, you can just omit the throws clause.

Does that make sense?

MatthewBemis: I appreciate the clarification, that does help me understand more. A lot of the methods in this class and related classes throw a broad `Exception`, so that's why I'm having to throw `Exception` here. My intent was to just let it bubble up and let the GlobalExceptionHandler deal with it -- there's not any special handling that I want to do here.

This part of the codebase should probably be combed over to make the exception handling more clear. In the meantime, I've updated the exceptions in this particular code path to (I hope) be a bit more clear (see 32923897f1988c77479b892fac9e02b83fb6f05e)

MatthewBemis:
For the sake of an audit trail, I misclicked and merged this PR into #642 (the 2 approval requirement is only for PRs into `development`). It's not straightforward to revive a merged PR (would have to revert the merge and open a brand new PR, which is of little value at this point imo), so I've addressed the remaining comments from here in the base PR: https://github.com/broadinstitute/juniper/pull/642

Followed up with Dennis over Slack about whether or not his original approval still stands and he said 👍

PR #664, opened 2023-12-05
JN-720 including kits in data export

devonbush:
#### DESCRIPTION

This adds KitRequests to the modules that we can export.  This meant expanding the capability of the export formatters to support iterated modules -- i.e. one enrollee can have export columns of `sample_kit.status` `sample_kit.2.status`, etc...  The convention is that the oldest kit is first, and is indicated without a numeral.  I personally think this first-one-doesn't-have-a-number convention is not particularly friendly, but it came out of multiple conversations with Pepper customers about what was easiest to work with, so I'm not inclined to change it without countervailing customer reports.

Because I honestly got lost a few times trying to start this PR, I took the opportunity to refactor the export code to make it (IMHO) cleaner and clearer.  I'm still not entirely happy with this refactoring--in particular I think the fact that a SurveyFormatter can have both PropertyItemFormatters and AnswerItemFormatters isn't clear, and processing that combined collection of AnswerItem and PropertyItem formatters isn't as clean as I would like.  Suggestions for improvement most welcome.

TO REVIEW: (sorry the PR got so large -- most of it is just rearranging)
1. start with core/src/main/java/bio/terra/pearl/core/service/export/README.md
2. then read through the EnrolleeFormatter, BeanModuleFormatter, and ModuleFormatter code to get a sense of the simplest formatter
3. read through KitRequestFormatter and KitTypeItemFormatter to see how the new pattern applies to the new kit request export code
4. skim the rest -- it's just the old stuff rearranged

1.  go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/export/dataBrowser`
5. scroll down a bit, to confirm that Jonas Salk has two sample kits, starting with the row `sample_kit.status`


core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -35,64 +32,68 @@ public class EnrolleeExportService {
...
-        BaseExporter exporter = getExporter(exportOptions.fileFormat(), moduleExportInfos, enrolleeMaps);
-        exporter.export(os);
+    public void export(ExportOptions exportOptions, UUID studyEnvironmentId, OutputStream os) {
+        List<ModuleFormatter> moduleFormatters = generateModuleInfos(exportOptions, studyEnvironmentId);
+        var enrolleeMaps = generateExportMaps(studyEnvironmentId, moduleFormatters, exportOptions.limit());
+        BaseExporter exporter = getExporter(exportOptions.fileFormat(), moduleFormatters, enrolleeMaps);
+        try {
+            exporter.export(os);
+        } catch (Exception e) {
+            throw new RuntimeException("Error writing export to output stream", e);
```
denniscunningham: Is this just here to add context to the thrown exception? RuntimeException seems like a bad choice since it is unclear how this will be handled upstream. 
Is there a way to achieve your aims without losing the intent of the caught exceptions?

core/src/main/java/bio/terra/pearl/core/service/export/DictionaryExportService.java
```diff
@@ -20,9 +19,9 @@ public DictionaryExportService(EnrolleeExportService enrolleeExportService) {
     public void exportDictionary(ExportOptions exportOptions, UUID portalId,
                                  UUID studyEnvironmentId,
                                  OutputStream os) throws Exception {
-        List<ModuleExportInfo> moduleInfos = enrolleeExportService.generateModuleInfos(exportOptions, studyEnvironmentId);
+        List<ModuleFormatter> moduleFormatters = enrolleeExportService.generateModuleInfos(exportOptions, studyEnvironmentId);
```
denniscunningham, line 22: Feels like `generateModuleInfos` should also be renamed?

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -102,21 +103,21 @@ public Map<String, String> generateExportMap(EnrolleeExportData exportData,
      * gets information about the modules, which will determine the columns needed for the export
      * e.g. the columns needed to represent the survey questions.
      */
-    public List<ModuleExportInfo> generateModuleInfos(ExportOptions exportOptions, UUID studyEnvironmentId) throws Exception {
-        List<ModuleExportInfo> moduleInfo = new ArrayList<>();
-        moduleInfo.add(new EnrolleeFormatter().getModuleExportInfo(exportOptions));
-        moduleInfo.add(new ProfileFormatter().getModuleExportInfo(exportOptions));
-        moduleInfo.addAll(generateSurveyModules(exportOptions, studyEnvironmentId));
-        return moduleInfo;
+    public List<ModuleFormatter> generateModuleInfos(ExportOptions exportOptions, UUID studyEnvironmentId)  {
```
denniscunningham, line 104: See my comment above about naming.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/EnrolleeExportExtService.java
```diff
@@ -40,11 +40,8 @@ public void export(
     authUtilService.authUserToStudy(user, portalShortcode, studyShortcode);
     StudyEnvironment studyEnv =
         studyEnvironmentService.findByStudy(studyShortcode, environmentName).get();
-    try {
-      enrolleeExportService.export(options, studyEnv.getId(), os);
-    } catch (Exception e) {
-      throw new RuntimeException(e);
-    }
+
```
denniscunningham, line 43: Nice change.

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -151,18 +152,19 @@ protected EnrolleeExportData loadEnrolleeData(Enrollee enrollee) {
                 profileService.loadWithMailingAddress(enrollee.getProfileId()).get(),
                 answerDao.findByEnrolleeId(enrollee.getId()),
                 participantTaskService.findByEnrolleeId(enrollee.getId()),
-                surveyResponseService.findByEnrolleeId(enrollee.getId())
+                surveyResponseService.findByEnrolleeId(enrollee.getId()),
```
denniscunningham, line 153: Not new, but this feels like a lot of DB calls to construct the export?

denniscunningham (changes requested):
Looks great. 
Thx for the refactoring. I recall reviewing code in this area a few months ago and struggling with the structure.
I spent a couple hours reviewing, but there is a lot here so I could have missed something. 
My comments are mostly nits, but I'd like you to consider my question about exception handling.

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
devonbush: yup, right now to export the ~500 participants in OurHealth takes 4-5 seconds in prod (producing a 2.6MB file). I haven't profiled it, but I'd guess the biggest latency is the ~2500 sequential DB calls that result from the code above.  So before OurHealth gets too much bigger we'll want to profile this operation and identify how to batch/group the queries better

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
devonbush: good catch--was just leftover sloppy error handling from before.  Fixed this by catching the errors (all IOExceptions) closer to the source

core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/KitRequestFormatter.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class KitRequestFormatter extends ModuleFormatter<KitRequest, PropertyItemFormatter<KitRequest>> {
+    private static final String KIT_REQUEST_MODULE_NAME = "sample_kit";
+    private static final List<String> KIT_REQUEST_INCLUDED_PROPERTIES = List.of("status", "sentToAddress", "externalKit");
+
+    private static final String KIT_TYPE_COLUMN_NAME = "kitType";
```
MatthewBemis, line 20: this is unused, was it supposed to be used in `KitRequestTypeFormatter`?

MatthewBemis (approved):
Looks good and worked well when tested.

I didn't see anything in the code that would make me think that it would break TDR export, but tested it anyway and it works great with the changes 👍

core/src/main/java/bio/terra/pearl/core/service/export/JsonExporter.java
```diff
@@ -18,15 +18,19 @@ public JsonExporter(List<ModuleFormatter> moduleFormatters, List<Map<String, Str
...
         List<String> subHeaderRowValues = getSubHeaderRow();
 
         JsonExport jsonExport = new JsonExport(columnKeys, headerRowValues, subHeaderRowValues, enrolleeMaps);
-        printWriter.println(objectMapper.writeValueAsString(jsonExport));
-        printWriter.flush();
+        try {
+            printWriter.println(objectMapper.writeValueAsString(jsonExport));
+            printWriter.flush();
+        } catch (IOException e) {
+            throw new RuntimeException("Error writing json to stream", e);
```
denniscunningham: Here, and in ExcelExporter and DataDictionaryExcelExporter can you change the thrown exception to something more specific than RuntimeException, which seems ambiguous from an error response perspective? 
I assume we want to return a 500, or perhaps a 503.

Also, does it make sense to indicate in the message that retrying the export may work?

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
denniscunningham: Thx for the info. Agree that current perf of 4-5s for this operation is acceptable.

denniscunningham (changes requested):

core/src/main/java/bio/terra/pearl/core/service/export/JsonExporter.java
devonbush: It's anyone's guess whether a retry will work, could be due to server conditions, or could be due to some really unexpected data properties.  I don't necessary want to encourage a retry, for example, the reason the writing failed was the server ran out of RAM after loading a large study.

core/src/main/java/bio/terra/pearl/core/service/exception/IORuntimeException.java
```diff
@@ -0,0 +1,4 @@
+package bio.terra.pearl.core.service.exception;
```
denniscunningham: Guessing this file is not needed?

denniscunningham (changes requested):

denniscunningham (approved):
Tada!

PR #665, opened 2023-12-05
[JN-749] Implement notification of kit sent events

denniscunningham:
#### DESCRIPTION
This PR implements a notification event when a kit status is changed to 'sent', so admins can configure an email message to be sent to participants when a kit is shipped. Most of the infrastructure for this was already in place -- as it would be for any new events -- so this PR mostly generates the event from KitRequestService.

The testing instructions below suggest using the kit 'Refresh' button. This uses the local simulated kit service to simultaneously advance the status of all requested kits. Since this is cumbersome Devon suggested providing a dropdown on kit list rows to change the status for each kit separately. I will do that in a separate PR, but if that's a blocker for testing this PR I can put that PR up first and incorporate that change into this PR.

image

1. Populate OurHealth.
2. Goto https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/notificationContent and setup an email notification for kit sent event.
3. Goto https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/requested/created.
4. Repeatedly mash the "Refresh" button until a participant status moves from 'Queued' to 'Sent'.
5. Check the logs. You should see an entry from EventService like "Kit status event for enrollee..." coinciding with the time the participant status changed to 'Sent'.
6. Closely following that message you should see an entry from EnrolleeEmailService like "Email sent...".
7. Check your inbox for a kit sent email notification.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/participant/PortalParticipantUserDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/participant/PortalParticipantUserDao.java
@@ -69,6 +69,9 @@ public class PortalParticipantUserDao extends BaseJdbiDao<PortalParticipantUser>
         return findAllByProperty("profile_id", profileId);
     }
 
+    public Optional<PortalParticipantUser> findByParticipantUserIdAndProfileId(UUID userId, UUID profileId) {
+        return findByTwoProperties("participant_user_id", userId, "profile_id", profileId);
+    }
 
 
     /**
--- a/core/src/main/java/bio/terra/pearl/core/dao/participant/PortalParticipantUserDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/participant/PortalParticipantUserDao.java
@@ -65,14 +65,9 @@ public class PortalParticipantUserDao extends BaseJdbiDao<PortalParticipantUser>
         return findAllByProperty("portal_environment_id", portalEnvId);
     }
 
-    public List<PortalParticipantUser> findByProfileId(UUID profileId) {
-        return findAllByProperty("profile_id", profileId);
+    public Optional<PortalParticipantUser> findByProfileId(UUID profileId) {
+        return findByProperty("profile_id", profileId);
     }
-
-    public Optional<PortalParticipantUser> findByParticipantUserIdAndProfileId(UUID userId, UUID profileId) {
-        return findByTwoProperties("participant_user_id", userId, "profile_id", profileId);
-    }
-
 
     /**
      * loads the user along with their profile
--- a/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
@@ -22,6 +22,7 @@ import bio.terra.pearl.core.service.participant.EnrolleeService;
 import bio.terra.pearl.core.service.participant.ProfileService;
 import bio.terra.pearl.core.service.study.StudyEnvironmentService;
 import bio.terra.pearl.core.service.study.StudyService;
+import bio.terra.pearl.core.service.workflow.EventService;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.extern.slf4j.Slf4j;
@@ -41,11 +42,10 @@ public class KitRequestService extends CrudService<KitRequest, KitRequestDao> {
     // possible statuses are.
     private static final Set<String> PEPPER_COMPLETED_STATUSES = Set.of("PROCESSED", "CANCELLED");
     private static final Set<String> PEPPER_FAILED_STATUSES = Set.of("CONTAMINATED");
-
     private final DaoUtils daoUtils;
 
     public KitRequestService(KitRequestDao dao,
-                             StudyEnvironmentKitTypeService studyEnvironmentKitTypeService, @Lazy EnrolleeService enrolleeService,
+                             EventService eventService, StudyEnvironmentKitTypeService studyEnvironmentKitTypeService, @Lazy EnrolleeService enrolleeService,
                              KitTypeDao kitTypeDao,
                              PepperDSMClient pepperDSMClient,
                              ProfileService profileService,
@@ -54,6 +54,7 @@ public class KitRequestService extends CrudService<KitRequest, KitRequestDao> {
                              ObjectMapper objectMapper,
                              DaoUtils daoUtils) {
         super(dao);
+        this.eventService = eventService;
         this.studyEnvironmentKitTypeService = studyEnvironmentKitTypeService;
         this.enrolleeService = enrolleeService;
         this.kitTypeDao = kitTypeDao;
@@ -275,6 +276,10 @@ public class KitRequestService extends CrudService<KitRequest, KitRequestDao> {
         try {
             kit.setExternalKit(objectMapper.writeValueAsString(pepperKit));
             kit.setExternalKitFetchedAt(pepperStatusFetchedAt);
+            KitRequestStatus newStatus = PepperKitStatus.mapToKitRequestStatus(pepperKit.getCurrentStatus());
+            if (newStatus != kit.getStatus()) {
+                eventService.publishKitStatusEvent()
+            }
             kit.setStatus(PepperKitStatus.mapToKitRequestStatus(pepperKit.getCurrentStatus()));
             dao.update(kit);
         } catch (JsonProcessingException e) {
@@ -292,5 +297,6 @@ public class KitRequestService extends CrudService<KitRequest, KitRequestDao> {
     private final StudyService studyService;
     private final StudyEnvironmentService studyEnvironmentService;
     private final ObjectMapper objectMapper;
+    private final EventService eventService;
     private StudyEnvironmentKitTypeService studyEnvironmentKitTypeService;
 }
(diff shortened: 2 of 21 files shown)


denniscunningham:
@devonbush Thanks for your help and your code. Regarding the latter, the branch you provided as a starter for this was based on another branch which had not been merged. I attempted a couple rebase strategies but ended up just cherry picking commits, which effectively erased your authorship. If you have suggestions I will try a different approach that makes the branch lineage clearer.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -202,19 +208,19 @@ public void syncKitStatusesForStudyEnv(Study study, EnvironmentName environmentN
     private void syncKitStatusesForStudyEnv(String studyShortcode, EnvironmentName envName, Collection<PepperKit> pepperKits) throws PepperParseException, PepperApiException {
         UUID studyEnvId = studyEnvironmentService.findByStudy(studyShortcode, envName)
                 .orElseThrow(() -> new NotFoundException("No matching study")).getId();
-        var pepperStatusFetchedAt = Instant.now();
-        var pepperKitStatusByKitId = pepperKits.stream().collect(
+        Instant pepperStatusFetchedAt = Instant.now();
+        Map<String, PepperKit> pepperKitByKitId = pepperKits.stream().collect(
```
devonbush, line 212: thanks for updating these -- I just found this https://stackoverflow.com/questions/57870236/intellij-search-and-replace-var-declarations-with-actual-types
I think I'm going to add an infrastructure ticket to do a global replace of all of these

devonbush: https://broadworkbench.atlassian.net/browse/JN-766

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -202,19 +208,19 @@ public void syncKitStatusesForStudyEnv(Study study, EnvironmentName environmentN
...
         studyEnvironmentService.find(studyEnvId).ifPresent(
                 studyEnvironment -> {
-                    var kits = dao.findByStudyEnvironment(studyEnvironment.getId());
+                    List<KitRequest> kits = dao.findByStudyEnvironment(studyEnvironment.getId());
 
                     // The set of kits returned from DSM may be different from the set of incomplete kits in Juniper, but
                     // we want to update the records in Juniper so those are the ones we want to iterate here.
                     for (KitRequest kit : kits) {
-                        var pepperKitStatus = pepperKitStatusByKitId.get(kit.getId().toString());
+                        PepperKit pepperKitStatus = pepperKitByKitId.get(kit.getId().toString());
```
devonbush, line 223: good method name change

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -272,18 +278,40 @@ protected String stringifyPepperAddress(PepperKitAddress kitAddress) {
...
+            log.error("Unable to serialize status JSON for kit %s: %s".formatted(kit.getId(), pepperKit.toString()), e);
+        }
+        try {
+            notifyKitStatus(kit, priorStatus);
+        } catch (Exception e) {
+            log.error("Error publishing kit status event for KitRequest enrollee %s".formatted(kit.getEnrolleeId()), e);
         }
     }
 
+    protected void notifyKitStatus(KitRequest kitRequest, KitRequestStatus priorStatus) {
```
devonbush: I'm not quite sure on the parsing of this name, but maybe it's a convention.  `notifyIfKitStatusChanged` would read more naturally to me.  And since this is only notifying on changes to sent, we might want to call that out explicitly `notifyIfKitStatusChangedToSent` (or at least in a comment)

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -272,18 +278,40 @@ protected String stringifyPepperAddress(PepperKitAddress kitAddress) {
...
+            return;
+        }
+        if (!KitRequestStatus.SENT.equals(kitRequest.getStatus())) {
+            return;
+        }
+
+        Enrollee enrollee = enrolleeService.find(kitRequest.getEnrolleeId()).orElseThrow(() ->
+                new IllegalStateException("Invalid enrollee for KitRequest %s: enrollee ID=%s".formatted(kitRequest.getId(), kitRequest.getEnrolleeId())));
+        List<PortalParticipantUser> ppUsers = portalParticipantUserService.findByParticipantUserId(enrollee.getParticipantUserId());
+        if (ppUsers.size() != 1) {
```
devonbush: this isn't an illegal state -- this is what we would expect if the same person (same email address) joined two different portals.  If this happens, we just need to identify which ppUser is the one that was used in creating this enrollee.  This can be done by find the ppUser with  `ppUser.profileId == enrollee.profileId`. Now if *that* returns multiple matches, we have an illegalState

devonbush: Incidentally, the ppUser<->enrollee relationship is one of the most confusing, because we didn't (and still don't) have a great sense of the arity of that relationship and how multi-study-multi-portal-mega-hub will impact it.  So I would be a big fan of taking the logic for navigating from an enrollee to a PortalParticipantUser and putting it in a `findPortalParticipantUser(Enrollee enrollee)` function in EnrolleeService so other folks can use it and it will be easier to update it as "hub" concepts clarify.

devonbush (changes requested):
Looks good, and no worries on the commit history.  Just one blocker related to handling the case of multiple PortalParticipantUsers, and a couple suggestions

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
denniscunningham: Yup, thx.
I forgot to highlight this in the PR (too much time since I wrote it), since I was unsure and wanted feedback.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
denniscunningham: I prefer `notifyIfKitStatusChanged` since this will likely handle other cases in the future.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
devonbush: Doing some digging, there's currently no uniqueness constraint on ppUser->profileId.  Probably because when these tables were created we were thinking that profile-sharing across a "hub" of portals might be a thing (or I just neglected it).  Would you mind writing a migration to add that constraint?  Something like:

```       
 - addUniqueConstraint:
                    tableName: portal_participant_user
                    constraintName: uc_pp_user_profile_id
                    columnNames: profile_id
            ```

File: core/src/main/java/bio/terra/pearl/core/service/participant/PortalParticipantUserService.java [hunk 5]
```diff
@@ -73,6 +74,13 @@ public List<PortalParticipantUser> findbyProfileId(UUID profileId) {
         return dao.findByProfileId(profileId);
     }
 
+    public PortalParticipantUser findForEnrollee(Enrollee enrollee) {
+        Optional<PortalParticipantUser> ppUser =
```
denniscunningham: @devonbush I implemented this method here rather than in EnrolleeService since it fit more naturally here. Let me know if you prefer a different approach.

devonbush: yup, this is the right place for it, thanks!

core/src/main/java/bio/terra/pearl/core/dao/participant/PortalParticipantUserDao.java
```diff
@@ -69,6 +69,9 @@ public List<PortalParticipantUser> findByProfileId(UUID profileId) {
         return findAllByProperty("profile_id", profileId);
     }
 
+    public Optional<PortalParticipantUser> findByParticipantUserIdAndProfileId(UUID userId, UUID profileId) {
```
devonbush: do we need to add the userId to this query or is profileId enough?  I think it's redundant, especially since we've added the profileId uniqueness constraint to ppUser

devonbush (approved):
Looks good, thanks for adding the uniqueness constraint.  One question about whether we need both params in a query, but not blocking

core/src/main/java/bio/terra/pearl/core/dao/participant/PortalParticipantUserDao.java
denniscunningham: Ah, good catch.
Took me a while to get that 'profile_id' is the join key between these tables. Sorry about that.
Can you give the latest commit a quick look?

devonbush: heh, as you can tell from the lack of previously existing constraints, helper methods, etc.., it also took me until today to remember that profile_id is functioning as the join key.  IIRC correctly, I did it that way because most of the time in the app we want to go direct from enrollee -> profile (e.g. when showing the details page), and don't care about the pp_user.  But it definitely makes the code harder to read, so might be worth a rethink

devonbush (approved):

MatthewBemis (approved):
Nice, looks good!

PR #667, opened 2023-12-07
JN-764 survey editor link browsing

devonbush:
#### DESCRIPTION

In working on JN-764, I found it convenient to add browsing buttons to navigate without using the table of contents.  So this PR adds clickable page, panel, and question listing

image

In the above image, clicking the question/panel titles will bring up that panel/question.  the item will also appear as selected on the left.  One thing this PR does NOT do is auto-scroll the table of contents to the selected item, that's a task for another day. 

We now also put the selectedItemPath in the url now, so that reloading the page doesn't lose your place in the survey

1. Go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms/surveys/oh_oh_basicInfo
2. click around on the links that now appear in the right panel of the designer
3. reload the page, confirm your place isn't lost
4. navigate to the preview tab, then back, and confirm you don't lose your place

### DIFF
--- a/ui-admin/src/forms/designer/PageElementList.tsx
+++ b/ui-admin/src/forms/designer/PageElementList.tsx
@@ -12,11 +12,13 @@ type PageElementListProps = {
   readOnly: boolean
   value: FormContentPage['elements']
   onChange: (newValue: FormContentPage['elements']) => void
+    setSelectedElementPath: (path: string) => void
+    selectedElementPath: string
 }
 
 /** UI for re-ordering a list of form elements. */
 export const PageElementList = (props: PageElementListProps) => {
-  const { readOnly, value, onChange } = props
+  const { readOnly, value, onChange, selectedElementPath, setSelectedElementPath } = props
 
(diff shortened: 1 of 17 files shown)


ui-admin/src/forms/designer/PageElementList.tsx
```diff
@@ -12,11 +12,13 @@ type PageElementListProps = {
   readOnly: boolean
   value: FormContentPage['elements']
   onChange: (newValue: FormContentPage['elements']) => void
+    setSelectedElementPath: (path: string) => void
```
denniscunningham, line 15: The indentation inconsistency makes this clumsy to read.

denniscunningham (approved):
This is pretty sweet. TIL.

ui-admin/src/forms/designer/PageElementList.tsx
devonbush: ugh, agree, was definitely not intentional

MatthewBemis (approved):
Very nice, these are such great UX improvements!

PR #668, opened 2023-12-08
JN-764 adding html to question type select

devonbush:
#### DESCRIPTION

this enables adding "html" type questions via the standard question flow.  

1. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms/surveys/oh_oh_basicInfo?readOnly=false
2. click "Page 1" on the left menu
3. click "add question"
4. add a stableId, then select "html" as the type, and fill in some text
5. confirm the question is added successfully, and shows up in preview/json editor.

### DIFF
--- a/ui-core/src/types/forms.ts
+++ b/ui-core/src/types/forms.ts
@@ -95,9 +95,12 @@ export type HtmlElement = {
 
 type BaseQuestion = BaseElement & {
   name: string
-  title: string
   description?: string
   isRequired?: boolean
+}
+
+export type TitledQuestion = BaseQuestion & {
+    title: string
 }
 
 export type QuestionChoice = {
@@ -112,7 +115,7 @@ type WithOtherOption<T> = T & {
   otherErrorText?: string
 }
 
-export type CheckboxQuestion = WithOtherOption<BaseQuestion & {
+export type CheckboxQuestion = WithOtherOption<TitledQuestion & {
   type: 'checkbox'
   choices: QuestionChoice[]
   showNoneItem?: boolean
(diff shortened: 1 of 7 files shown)


denniscunningham (approved):
This looks fine. I'm just not sure how/where this fits in. The ticket did not provide any details.

devonbush:
Good question.  This was an ad-hoc request, one of the HeartHive study staff was struggling with the creation of this type of question using just JSON.  See line 9 in https://docs.google.com/spreadsheets/d/1elOSE0QpXnwmUcPmAIgIETdSyOCuLhjPeOpiqzrJOw8/edit#gid=0. I just updated the ticket with that ref too

ui-core/src/types/forms.ts
```diff
@@ -112,46 +115,52 @@ type WithOtherOption<T> = T & {
...
-export type MedicationsQuestion = BaseQuestion & {
+export type MedicationsQuestion = TitledQuestion & {
   type: 'medications'
 }
 
+
+export type HtmlQuestion = BaseQuestion & {
+  type: 'html',
+  html: string
+}
```
MatthewBemis, line 162: Nit: should this omit `isRequired` from `BaseQuestion`?

MatthewBemis (approved):
Looks good. Just had one optional suggestion

ui-core/src/types/forms.ts
devonbush: I think you're right, but I'm not sure.  If so, I'll clear it in an upcoming PR

PR #669, opened 2023-12-08
[JN-413] Add kit received event handling

denniscunningham:
#### DESCRIPTION
This PR implements a notification event when a kit status is changed to 'RECEIVED', so admins can configure an email message to be sent to participants when their kit is received by GP. This is need to implement JN-749.

Also included are updates to two README files, documenting things I learned while talking with Devon. Feel free to suggest edits if the information is incorrect, or the wording can be improved.

image

1. Populate OurHealth.
2. Goto https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/notificationContent and setup an email notification for kit received event.
3. Goto https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/requested/created.
4. Repeatedly mash the "Refresh" button until a participant status moves from 'Sent' to 'Returned'.
5. Check the logs. You should see an entry from EventService like "Kit status event for enrollee..." coinciding with the time the participant status changed to 'Sent'.
6. Closely following that message you should see an entry from EnrolleeEmailService like "Email sent...".
7. Check your inbox for a kit received email notification.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/README.md
+++ b/core/src/main/java/bio/terra/pearl/core/dao/README.md
@@ -1,3 +1,14 @@
+The DAO layer interacts with the database. DAO concrete classes derive from BaseJdbiDao, via BaseMutableJdbiDao or
+BaseVersionedJdbiDao, and each is associated with an entity class that acts as a data transfer object (DTO).
+Those entity classes are may include nested entity classes, which are also DTOs. Typically, the nested entity classes
+are not instantiated by the associated DAO class, so callers should assume those nested entity classes are empty unless 
+the DAO method explicitly documents otherwise. 
+
+Nested entity classes are instantiated at various points in the code, often for the purpose of populating a response
+object. A common pattern for a service class method that takes an entity class argument with nested entities is to 
+take the nested entities as separate arguments, rather than relying on the entity class to have the nested entities 
+attached.
+
 When creating a DAO test:
 1. the DAO test should test its own DAO, but not call any other DAOs.
 To the extent it needs to create/read other objects, it should do so via factories and services.
--- a/core/src/main/java/bio/terra/pearl/core/service/README.md
+++ b/core/src/main/java/bio/terra/pearl/core/service/README.md
@@ -22,3 +22,10 @@ they are covered by ingress logging, DB logging, and exception handling.
    that allows setting of non-dependent entities that should be deleted if orphaned.  this cascades method 
    should be passed along to other service-delete methods called.  An enu 'AllowedCascades' can be
    provided in the service to specify which non-dependent deletes are implemented.
+
+If circular dependencies are encountered when wiring service dependencies with Spring, first consider the service design. If that is
+sound, then use the @Lazy annotation. Ideally the @Lazy annotation should be applied to services that are higher up in 
+the dependency tree, not necessarily on the usage in the new service.
+
+When wiring services dependencies with Spring via constructor arguments, do not call the services in the constructor.
+This ensures order of instantiation is not an issue.
--- a/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
@@ -299,8 +299,8 @@ public class KitRequestService extends CrudService<KitRequest, KitRequestDao> {
         if (priorStatus == kitRequest.getStatus()) {
             return;
         }
-        // for now, only notify when the kit is sent
-        if (!KitRequestStatus.SENT.equals(kitRequest.getStatus())) {
+        // only notify when the kit is sent or received
+        if (!KitRequestStatus.SENT.equals(kitRequest.getStatus()) && !KitRequestStatus.RECEIVED.equals(kitRequest.getStatus())) {
             return;
         }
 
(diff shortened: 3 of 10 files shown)


core/src/main/java/bio/terra/pearl/core/service/kit/KitStatusEvent.java
```diff
@@ -21,6 +21,11 @@ public static KitStatusEvent newInstance(KitRequest kitRequest, KitRequestStatus
                 .kitRequest(kitRequest)
                 .priorStatus(priorStatus)
                     .build();
+        } else if (KitRequestStatus.RECEIVED.equals(kitRequest.getStatus())) {
+            return KitReceivedEvent.builder()
+                .kitRequest(kitRequest)
+                .priorStatus(priorStatus)
+                    .build();
```
denniscunningham: The boilerplate code here is now repeated three times. Offhand I could not think of an elegant alternative, but I am open to suggestions, or I can propose something if this just feels bad.

core/src/main/java/bio/terra/pearl/core/dao/README.md
```diff
@@ -1,3 +1,14 @@
...
+BaseVersionedJdbiDao, and each is associated with an entity class that acts as a data transfer object (DTO).
+Those entity classes are may include nested entity classes, which are also DTOs. Typically, the nested entity classes
+are not instantiated by the associated DAO class, so callers should assume those nested entity classes are empty unless 
+the DAO method explicitly documents otherwise. 
+
+Nested entity classes are instantiated at various points in the code, often for the purpose of populating a response
+object. A common pattern for a service class method that takes an entity class argument with nested entities is to 
+take the nested entities as separate arguments, rather than relying on the entity class to have the nested entities 
+attached.
+
```
devonbush, line 11: Thanks for adding, well said!

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -299,8 +299,8 @@ protected void notifyKitStatusChange(KitRequest kitRequest, KitRequestStatus pri
         if (priorStatus == kitRequest.getStatus()) {
             return;
         }
-        // for now, only notify when the kit is sent
-        if (!KitRequestStatus.SENT.equals(kitRequest.getStatus())) {
+        // only notify when the kit is sent or received
+        if (!KitRequestStatus.SENT.equals(kitRequest.getStatus()) && !KitRequestStatus.RECEIVED.equals(kitRequest.getStatus())) {
```
devonbush: I think a list operation is clearer and more extensible here (possibly pulling the list out into a "EVENTED_STATUSES" constant too)
```suggestion
        if (!List.of(KitRequestStatus.SENT, KitRequestStatus.RECEIVED).contains(kitRequest.getStatus)) {
```

core/src/main/java/bio/terra/pearl/core/service/kit/KitStatusEvent.java
devonbush: Thanks for flagging this.  If we're using builders, I think we want something like the below, but I'm not 100% sure it will work.  Alternatively, you could just have a `KitStatusEvent event`, call the regular constructors in the `if` statements, and then use the regular setters to set the shared properties
```
KitStatusEvent.KitStatusEventBuilder builder = KitStatusEvent.builder();
 if (KitRequestStatus.SENT.equals(kitRequest.getStatus())) {
      builder = KitSentEvent.builder();
 } else if (KitRequestStatus.SENT.equals(kitRequest.getStatus())) {
    builder = KitReceivedEvent.builder();
 }
 return builder.priorStatus(priorStatus)
                .kitRequest(kitRequest)
                .build();
```

core/src/main/java/bio/terra/pearl/core/service/README.md
```diff
@@ -22,3 +22,10 @@ they are covered by ingress logging, DB logging, and exception handling.
    that allows setting of non-dependent entities that should be deleted if orphaned.  this cascades method 
    should be passed along to other service-delete methods called.  An enu 'AllowedCascades' can be
    provided in the service to specify which non-dependent deletes are implemented.
+
+If circular dependencies are encountered when wiring service dependencies with Spring, first consider the service design. If that is
+sound, then use the @Lazy annotation. Ideally the @Lazy annotation should be applied to services that are higher up in 
+the dependency tree, not necessarily on the usage in the new service.
+
+When wiring services dependencies with Spring via constructor arguments, do not call the services in the constructor.
+This ensures order of instantiation is not an issue.
```
devonbush, line 31: nice!

devonbush (approved):
Looks good, I trust your judgement on the code cleanup in the newInstance method.  One thing (outside the scope of this ticket), that we should think about is how we represent and troubleshoot these events.  Naively, I think very soon we will want an `event` table that logs all events that are fired, so that we can look at an enrollee in the admin tool and see a listing of what events were fired relating to that enrollee.  that will help us troubleshoot cases involving, e.g., duplicate events or "why didn't that email get sent?" questions.  Just created https://broadworkbench.atlassian.net/browse/JN-769

denniscunningham:
@devonbush Regarding your comment about recording event history...  Yup, that makes sense, particularly in helping admins see a timeline of participant history. One thing to consider is the overlap of information provided, since much of the data in that timeline appears elsewhere in different forms. For instance, kit requests are reported on the ptp page, and on the kit page, and (soon) as tasks, and as notifications. This feature would add kit events reporting. Not necessarily a bad thing, but it does thicken the information stream.

MatthewBemis (approved):

PR #671, opened 2023-12-11
JN-714 enabling delete of environment alerts in populate contexts

devonbush:
#### DESCRIPTION

Speaking of "adding a new configurable entity has way too many undocumented steps",  this enables the environments with alerts to be deleted/recreated for populating.  The main thing we were missing was a test that the HeartHive environment could be deleted/recreated.  We have that test for OurHealth, but since Matt only added the configurable alerts to HeartHive, we didn't notice that those alerts weren't able to be repopulated in development.

1. run `./scripts/populate_portal.sh hearthive`
2. run `./scripts/populate_portal.sh hearthive` again, confirm it works


MatthewBemis (approved):
Ah, good catch. Thanks for fixing this.

denniscunningham (approved):

devonbush:
test failure is unrelated

devonbush:
test failures are unrelated

PR #670, opened 2023-12-11
JN-770: adding new outreach survey type

devonbush:
#### DESCRIPTION

This adds the framework (but by no means a complete/refined journey) for a study staff to create messages that appear in the participant dashboard.  This accomplishes that by adding a 'type' field to Surveys, and splitting Surveys into RESEARCH surveys (questionnaires to build a dataset) and OUTREACH (ads).   I'm pleased at how quickly this came together, I think modeling it as a survey type was a big win--it let us implement this quickly while also enhancing general survey capability.

image

We've added a new task status of "VIEWED" to indicate that someone has affirmatively interacted with a task, but not necessarily started it.  

If this survey type pattern proves successful, this could also lay the groundwork for simplifying our code by treating ConsentForms as just a survey type.

This obviously opens up a TON of new functionality and workflows, and many need refinement.  but I don't wan this PR to get too big.  So what I'd propose is that this review focuses mostly on the technical approach, and that we assume at least 4-5 subsequent PRs will be needed before we tell Rachel this is ready to use.
 
 To be done in later PRs:
1. implementing the "blurb" capability (and general participant UX polish).  It's added to the data model in this PR, but to surface it in the dashboard I want to do some refactoring of how task information is cached/fetched.
2. admin UX help for creating the Outreach surveys -- they probably should default to a survey with 1 html question, and could use more info popups and/or Zendesk docs.
3. Ability to only have the outreach survey given to a subset of enrollees
4. Ability to reject/dismiss the outreach survey
5. Ability to track whether the enrollee "said yes" to the outreach survey (obviously, we can't track if they actually joined the external study, but we could track whether they followed a link to it).
6. Ability to assign outreach tasks to existing participants (this is covered in JN-762)

#### TO TEST:  *(simple manual steps for confirming core behavior -- used for pre-release checks)*

1.  restart ApiAdminApp
2. repopulate hearthive `./scripts/populate_portal.sh hearthive`
3. go to https://localhost:3000/hearthive/studies/hh_registry/env/sandbox/forms
4. see the new "Outreach surveys" section
5. go to https://sandbox.hearthive.localhost:3001/
6. sign in as `gelion@test.com`
7. check out the "Foobar study" invitation
8. go to https://localhost:3000/hearthive/studies/hh_registry/env/sandbox/participants/HHGELI/tasks
9. confirm the outreach task now has a status of "VIEWED"

PR #672, opened 2023-12-12
[JN-770] Create new outreach modal

MatthewBemis:
#### DESCRIPTION

This adds to the outreach modal in #670 by letting the user select a marketing or screener outreach type. The marketing type will prepopulate the survey with an empty HTML question, and the screener outreach type will use the same default form content as research surveys. This also exposes the blurb concept in #670 for admins to set. This does not yet allow selecting cohorts or uploading images.

https://github.com/broadinstitute/juniper/assets/7257391/ba81cf22-6362-4082-a595-0c94ceb7519c

* Launch Admin API
* Create a new outreach: https://localhost:3000/hearthive/studies/hh_registry/env/sandbox/forms
* Test that both the marketing and screener types work correctly
* Test that blurbs can be set (although they currently aren't displayed anywhere)

### DIFF
--- a/ui-admin/src/study/surveys/CreateSurveyModal.tsx
+++ b/ui-admin/src/study/surveys/CreateSurveyModal.tsx
@@ -29,9 +29,10 @@ const CreateSurveyModal = ({ studyEnvContext, onDismiss, type }:
   // outreach defaults to a template with an HTML question. Users can edit that
   // HTML from the survey editor. Alternatively, we could allow them to design the
   // content within this modal and insert the content into the survey on their behalf.
+  const randomSuffix = Math.random().toString(36).substring(2, 15)
   const defaultTemplateJson = type === 'RESEARCH' || isOutreachScreener ?
     '{"pages":[]}' :
-    '{"pages":[{"elements":[{"type":"html","name":"outreach_content"}]}]}'
+    `{"pages":[{"elements":[{"type":"html","name":"outreach_content_${randomSuffix}"}]}]}`
 
   const createSurvey = async () => {
     doApiLoad(async () => {
@@ -102,7 +103,7 @@ const CreateSurveyModal = ({ studyEnvContext, onDismiss, type }:
               icon={faUsersViewfinder}
               title="Screener"
               description="Screener opportunities allow you to send a questionnaire to your participants
-              so you can follow up and engage with qualified participants."
+              so you can follow up with qualified participants."
               onSelect={() => setIsOutreachScreener(true)}
               isSelected={isOutreachScreener}
             />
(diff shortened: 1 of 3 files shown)


ui-admin/src/study/surveys/CreateSurveyModal.tsx
```diff
@@ -19,13 +22,23 @@ const CreateSurveyModal = ({ studyEnvContext, onDismiss, type }:
...
+  const [isOutreachScreener, setIsOutreachScreener] = useState(false)
+  const [surveyBlurb, setSurveyBlurb] = useState<string>()
+
+  //Screeners and research surveys default to an empty form, but marketing
+  // outreach defaults to a template with an HTML question. Users can edit that
+  // HTML from the survey editor. Alternatively, we could allow them to design the
+  // content within this modal and insert the content into the survey on their behalf.
+  const defaultTemplateJson = type === 'RESEARCH' || isOutreachScreener ?
+    '{"pages":[]}' :
+    '{"pages":[{"elements":[{"type":"html","name":"outreach_content"}]}]}'
```
devonbush: I think we might want to add a random suffix to the name here so it doesn't conflict with any uniqueness constraints (which IIRC don't apply to html questions yet, but might someday)

devonbush (approved):

ui-admin/src/study/surveys/CreateSurveyModal.tsx
```diff
@@ -75,6 +88,33 @@ const CreateSurveyModal = ({ studyEnvContext, onDismiss, type }:
...
+              title="Marketing"
+              description="Marketing opportunities allow you to display messages in the participant dashboard."
+              onSelect={() => setIsOutreachScreener(false)}
+              isSelected={!isOutreachScreener}
+            />
+            <CardButton
+              icon={faUsersViewfinder}
+              title="Screener"
+              description="Screener opportunities allow you to send a questionnaire to your participants
+              so you can follow up and engage with qualified participants."
```
denniscunningham: Nit" "and engage" seems redundant with "follow up", so can probably be removed.

ui-core/src/types/forms.ts
```diff
@@ -14,6 +14,7 @@ export type SurveyType = 'RESEARCH' | 'OUTREACH'
 
 export type Survey = VersionedForm & {
   surveyType: SurveyType
+  blurb?: string
```
denniscunningham, line 17: Should we allow this to be empty?
I forget from the other PR if there is a placeholder if this is not specified.

denniscunningham (approved):
Looks good.
I'm having trouble seeing how users will get started with these, but looking forward to seeing the follow-up PRs.

ui-core/src/types/forms.ts
MatthewBemis: It's a nullable column and it may not exist for research surveys. Since we don't yet have a way for users to set/display blurbs for those research surveys, I've left it optional for now, but that could change as this feature matures.

PR #673, opened 2023-12-12
[no ticket yet] Cohort builder WIP

MatthewBemis:
#### DESCRIPTION

I'm out the next few days, but here's where I got with the cohort builder. It's still a bit rough in some areas but I think gives us something to work off of/refine for now.

https://github.com/broadinstitute/juniper/assets/7257391/ccac2029-363f-49f6-b6e6-665c9cd88971

TODOs:

* Doesn't yet populate the dropdown with all of the questions for the selected survey
* Doesn't support anything other than ANDing all of the rules
* The state in the `CriteriaRow` component is still quite rough
* Implement the UX for building a cohort that spans multiple studies

### DIFF
--- a/ui-admin/src/HomePage.tsx
+++ b/ui-admin/src/HomePage.tsx
@@ -8,12 +8,14 @@ import { faPlus } from '@fortawesome/free-solid-svg-icons'
 import { Button } from './components/forms/Button'
 import CreateNewStudyModal from './study/CreateNewStudyModal'
 import { useUser } from './user/UserProvider'
+import CreateNewCohortModal from './study/CreateNewCohortModal'
 
 /** Shows a user the list of portals available to them */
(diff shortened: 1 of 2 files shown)


devonbush:
This is super cool.  Have you looked at any existing javascript/react rule builders?  e.g https://react-querybuilder.js.org/.   We might want to start using one of those now, even if we're only including a small subset of functionality

MatthewBemis:
@devonbush I think that's the right move. Even with a minimal set of features we need to build, it was quickly approaching the point of being silly to do it ourselves. I'll look at a few others but the one you linked looks well-supported and supports writing to a number of different useful formats.

MatthewBemis:
Closing this PR for now, revisiting this with a 3rd party rule builder.

PR #674, opened 2023-12-15
JN-774 adding more "Add question" buttons

devonbush:
#### DESCRIPTION

In creating the "add html question" ticket, I found it was frustrating to keep clicking back and forth to be able to add a new question.  This adds an "add question" button to the question panel display, and also an "add next question" to individual questions, to make it easier to create/test surveys sequentially.  (Note, this counts as a member's choice ticket, but I'm doing it now since next week I'll probably mostly be helping Connor and/or working in Matt's changes)

image

1. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms/surveys/oh_oh_basicInfo?selectedElementPath=pages%5B0%5D.elements%5B3%5D
2. Confirm you're on the ourhealth basics survey, editing the `oh_oh_basic_mghPatient` question
3. press "Add next question"
4. complete the new question dialog
5. confirm your new question is created immediately following the oh_oh_basic_mghPatient question, and that it is brought into focus

### DIFF
--- a/ui-admin/src/forms/FormDesigner.tsx
+++ b/ui-admin/src/forms/FormDesigner.tsx
@@ -1,7 +1,7 @@
-import { flow, get, identity, set, update } from 'lodash/fp'
-import React, {useState} from 'react'
+import { flow, identity, set, update } from 'lodash/fp'
+import React, { useState } from 'react'
 
-import {FormContent, FormContentPage, FormElement, FormPanel, HtmlElement, Question} from '@juniper/ui-core'
+import { FormContent, FormContentPage, FormPanel, HtmlElement, Question } from '@juniper/ui-core'
 
 import { HtmlDesigner } from './designer/HtmlDesigner'
 import { PageDesigner } from './designer/PageDesigner'
@@ -11,9 +11,10 @@ import { QuestionTemplatesDesigner } from './designer/QuestionTemplatesDesigner'
 import { FormTableOfContents } from './FormTableOfContents'
 import { PageListDesigner } from './designer/PageListDesigner'
 import { useSearchParams } from 'react-router-dom'
-import {Modal} from "react-bootstrap";
-import {NewQuestionForm} from "./designer/NewQuestionForm";
-import _cloneDeep from "lodash/cloneDeep";
+import { Modal } from 'react-bootstrap'
+import { NewQuestionForm } from './designer/NewQuestionForm'
+import _cloneDeep from 'lodash/cloneDeep'
+import { getContainerElementPath, getCurrentElementIndex, getSurveyElementFromPath } from './designer/designer-utils'
 
 type FormDesignerProps = {
   readOnly?: boolean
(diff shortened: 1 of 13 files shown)


ui-admin/src/forms/FormDesigner.tsx
```diff
@@ -11,23 +11,79 @@ import { QuestionTemplatesDesigner } from './designer/QuestionTemplatesDesigner'
...
+    } else {
+      // add the new question to the end of the page/panel
+      newQuestionIndex = containerToUpdate.elements.length
+      containerToUpdate.elements.push(newQuestion)
+    }
+    onChange(newValue)
+    // we want to view the new question, but we need to wait for the state change to propagate before it will exist
+    window.setTimeout(() => {
+      setSelectedElementPath(`${containerElementPath}.elements[${newQuestionIndex}]`)
+    }, 0)
```
MatthewBemis: If this is set with a timeout of 0 millis, is this actually reliably accomplishing anything? Would this be cleaner with a state callback or effect that can guarantee that the state has actually been updated?

MatthewBemis (commented):
This worked well when I tested it out. But I do have one comment/question about the code

ui-admin/src/forms/FormDesigner.tsx
```diff
@@ -11,23 +11,79 @@ import { QuestionTemplatesDesigner } from './designer/QuestionTemplatesDesigner'
...
   const [searchParams, setSearchParams] = useSearchParams()
   const selectedElementPath = searchParams.get('selectedElementPath') ?? 'pages'
   const selectedElement = getSurveyElementFromPath(selectedElementPath, value)
   const setSelectedElementPath = (path: string) => {
     searchParams.set('selectedElementPath', path)
     setSearchParams(searchParams)
   }
+
+  let selectedElementType: SelectedElementType = 'none'
+  if (selectedElementPath === 'pages') {
```
denniscunningham, line 42: I think this clause is trying to improve readability of subsequent code, but a comment describing the role of selectedElementPath and selectedElement would help readability of the logic here. (IOW, its unclear why the logic below leads to each assignment.)

ui-admin/src/forms/FormDesigner.tsx
```diff
@@ -11,23 +11,79 @@ import { QuestionTemplatesDesigner } from './designer/QuestionTemplatesDesigner'
...
+    }
+  }
+
+
+  const insertQuestionAtCursor = (newQuestion: Question) => {
+    if (['pages', 'questionTemplates', 'none'].includes(selectedElementType)) {
+      // we don't know what to add the question to
+      return
+    }
+    const newValue = _cloneDeep(value)
```
denniscunningham: It is unclear from the FormDesigner function (lack of) comments what `value` is and why it needs a deep copy.

ui-admin/src/forms/designer/PanelDesigner.tsx
```diff
@@ -3,22 +3,36 @@ import React from 'react'
...
   value: FormPanel
   onChange: (newValue: FormPanel, removedElement?: FormElement) => void
     setSelectedElementPath: (path: string) => void
     selectedElementPath: string
+    addNextQuestion?: () => void
 }
 
 /** UI for editing a panel in a form. */
 export const PanelDesigner = (props: PanelDesignerProps) => {
-  const { readOnly, value, onChange, setSelectedElementPath, selectedElementPath } = props
```
denniscunningham, line 17: As elsewhere what `value` contains is unclear.
Am I missing a JS convention of some sort?

ui-admin/src/forms/designer/HtmlDesigner.tsx
```diff
@@ -3,20 +3,31 @@ import React from 'react'
...
+  const { element, readOnly, onChange, addNextQuestion } = props
 
   return (
     <div className="d-flex flex-column h-100">
-      <h2>{element.name}</h2>
+      <div className="d-flex align-items-center justify-content-between">
+        <h2>{element.name} (html)</h2>
+        {addNextQuestion && <div>
+          <Button variant="secondary" className="ms-auto" onClick={addNextQuestion}>
+            <FontAwesomeIcon icon={faPlus}/> Add question
```
denniscunningham: It's unclear why this is not 'Add new question'. Is it because HTML is not technically a question? Will users see it that way? (We include 'Html' as a question type in the new question dropdown.) 
IOW, might this be confusing? (Am I adding a question to this HTML element?)

denniscunningham (approved):

ui-admin/src/forms/FormDesigner.tsx
devonbush: agree, I think 'value' can be renamed to "formContent"

ui-admin/src/forms/designer/PanelDesigner.tsx
devonbush: I think "value" is supposed to connote "the main thing this component operates on".  But I agree it's not particularly clear.  Will change

ui-admin/src/forms/designer/HtmlDesigner.tsx
devonbush: good catch, this should be "Add next question"

ui-admin/src/forms/FormDesigner.tsx
devonbush: setting a timeout of 0 guarantees that setState-y operations will be done, although you're right it's not particularly elegant.  I'll see if I can come up with something better

devonbush:
@MatthewBemis  thanks for the nudge to find something better than window.setTimeout for setState callbacks.

MatthewBemis (approved):

devonbush:
java ci failure is unrelated false positive

PR #676, opened 2023-12-19
[JN-777] Sort mailing list by joined date by default

connorlbark:
#### DESCRIPTION

Before this ticket, you could sort on the mailing list page, but there was no default sort. Now, the default sort is by the joined date, with most recently joined users appearing at the top.

Add multiple users to the mailing list. Ensure that they are sorted by their join date, descending.

### DIFF
--- a/ui-admin/src/portal/MailingListView.test.tsx
+++ b/ui-admin/src/portal/MailingListView.test.tsx
@@ -16,7 +16,7 @@ const contacts: MailingListContact[] = [{
   id: 'id2',
   name: 'person2',
   email: 'fake2@test.com',
-  createdAt: 0
+  createdAt: 1
 }]
 
 test('renders a mailing list', async () => {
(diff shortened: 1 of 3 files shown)


ui-admin/src/portal/MailingListView.test.tsx
```diff
@@ -73,3 +73,20 @@ test('delete button shows confirmation', async () => {
...
+      setupRouterTest(<MailingListView portalContext={portalContext} portalEnv={portalEnv}/>)
+  render(RoutedComponent)
+  await waitFor(() => {
+    expect(screen.getByText('person1')).toBeInTheDocument()
+  })
+
+  const joined = screen.getByText('Joined')
+
+  // "Joined" text is two parents away from the <th> element
+  expect(joined.parentElement?.parentElement?.getAttribute('aria-sort')).toBe('descending')
```
devonbush: using `closest` will make it so you don't have to hardcode the parent distance
```suggestion
 expect(joined.closest('[aria-sort]')?.getAttribute('aria-sort')).toBe('descending')
```

devonbush (approved):
Congrats on the first code PR!  One suggestion on using a different selector in the test

ui-admin/src/portal/MailingListView.test.tsx
connorlbark: Ah smart. I felt like there should have been a better way. Thanks!

MatthewBemis (approved):
Nice!

PR #678, opened 2023-12-19
[JN-714] fixing test transaction wrapping

devonbush:
#### DESCRIPTION

There was a test that wasn't transactionally wrapped with rollback, which I think might have been causing the increase in test failures due to race conditions

1. reset your local database `local-dev/run_postgres.sh start` 
2. run java tests `./gradlew test`
3. log into test database.  `psql -U test_dbuser -d pearl_test -h localhost`.  (password is `dbpwd`)
4. confirm `select * from study;` returns no results

### DIFF
--- a/populate/src/test/java/bio/terra/pearl/populate/PopulateHeartHiveTest.java
+++ b/populate/src/test/java/bio/terra/pearl/populate/PopulateHeartHiveTest.java
@@ -66,6 +66,8 @@ public class PopulateHeartHiveTest extends BasePopulatePortalsTest {
             withdrawnEnrolleeService.withdrawEnrollee(prodEnrollee.enrollee());
             // confirm we can now repopulate with overwrite
             portalPopulator.populate(new FilePopulateContext("portals/hearthive/portal.json"), true);
+            // now roll back the whole transaction so that the test doesn't actually persist the changes
+            status.setRollbackOnly();
             return null;
         });
     }
--- a/core/src/test/java/bio/terra/pearl/core/service/export/EnrolleeExportServiceTests.java
+++ b/core/src/test/java/bio/terra/pearl/core/service/export/EnrolleeExportServiceTests.java
@@ -45,7 +45,8 @@ public class EnrolleeExportServiceTests extends BaseSpringBootTest {
     private SurveyFactory surveyFactory;
 
     @Test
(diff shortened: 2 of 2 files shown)


populate/src/test/java/bio/terra/pearl/populate/PopulateHeartHiveTest.java
```diff
@@ -66,6 +66,8 @@ public void testPopulateHeartHive() {
             withdrawnEnrolleeService.withdrawEnrollee(prodEnrollee.enrollee());
             // confirm we can now repopulate with overwrite
             portalPopulator.populate(new FilePopulateContext("portals/hearthive/portal.json"), true);
+            // now roll back the whole transaction so that the test doesn't actually persist the changes
+            status.setRollbackOnly();
```
denniscunningham, line 70: Still curious what transaction manager this uses. 
That might have helped catch this in the prior PR.

denniscunningham (approved):

populate/src/test/java/bio/terra/pearl/populate/PopulateHeartHiveTest.java
devonbush: My understanding is that this uses the Spring transaction manager (AFAIK we don't have any other Tx manager), and is functionally identical to using the @Transaction annotation, with the exception that by invoking it manually, you get a handle that you can use to rollback

devonbush:
@connorlbark   adding you for a review, since the test steps are something you probably haven't done yet.  And it's a good thing to know how we wrap tests in transactions and what happens when someone (me) forgets to wrap one correctly :)

connorlbark (approved):
One of the fixed tests failed for me on first run:
```
PopulateHeartHiveTest > testPopulateHeartHive() FAILED
    org.jdbi.v3.core.statement.UnableToExecuteStatementException at PopulateHeartHiveTest.java:35
        Caused by: org.postgresql.util.PSQLException at PopulateHeartHiveTest.java:35
```
It happened a couple times on this branch but now I cannot recreate it, so it might have just been a fluke. Otherwise, the provided steps worked for me!

devonbush:
Hmm... that makes me nervous that it failed a couple of times.  I wonder if there are some more missing transactions we'll have to hunt down in the future

PR #680, opened 2023-12-20
[JN-676] Spring Boot 3 reboot

MatthewBemis:
#### DESCRIPTION

When this was first attempted, TDR hadn't yet started publishing a Jakarta client, which was a blocker for completing the Spring Boot 3 upgrade. They've since started [dual-publishing](https://broadinstitute.jfrog.io/ui/packages/gav:%2F%2Fbio.terra:datarepo-jakarta-client?name=datarepo&type=packages) Javax and Jakarta clients so we're unblocked.

Most of this PR is just wrangling dependencies, changing the word "javax" to "jakarta" in imports, and getting logging/tracing from terra-common-lib (TCL) to play nicely with our app. TCL assumes that you want to do tracing in your app, and since we don't trace yet, we need to be a bit more assertive in disabling it.

There's not any actual functionality changes so: 
* Try booting up the admin and participant API apps, check that you don't see anything fishy in the logs during/shortly after boot
* Try clicking through different parts of the app à la [demo smoke test](https://broadworkbench.atlassian.net/wiki/spaces/PEARL/pages/2699558978/Demo+smoke+test+checklist)


connorlbark (approved):
LGTM! Took a while to get the new deps but the logs looked good and the smoke test steps worked for me.

devonbush (approved):
yup, takes a long time to get all the dependencies into IntelliJ, but worked great locally for me.  Thanks for tackling this!

PR #681, opened 2023-12-20
[JN-749] Kit request task

denniscunningham:
#### DESCRIPTION
This PR includes a couple main areas::

1. Creates a kit request task when a kit is shipped to a participant, and marks that task complete when the kit is received back.
2. Provides a notification on the participant dashboard when a kit is shipped, and removes the notification when the kit is received back.

1. Update api-admin, api-participant, ui-participant;
2. Populate OurHealth;
3. Goto https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/requested/created;
4. Mash refresh button until jsalk has a sent kit.
image

1. Goto jsalk participant detail page (https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHSALK/tasks), and verify a new kit request task is now listed.
image

1. Goto https://sandbox.ourhealth.localhost:3001/ and login as jsalk@test.com.
2. Verify the dashboard is showing the kit shipped notification.
image

1. Return to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/requested/sent.
2. Mash the refresh button until jsalk has no sent kits.
3. Goto to the jsalk participant page (https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHSALK/tasks), and verify that the kit request tasks are now complete.

1. Return to https://sandbox.ourhealth.localhost:3001/hub
2. Refresh the page and verify there are no kit sent notifications.
image


core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeDao.java
```diff
@@ -117,18 +117,27 @@ public Optional<Enrollee> findByPreEnrollResponseId(UUID preEnrollResponseId) {
...
         }
+        enrollee.getParticipantNotes().addAll(participantNoteDao.findByEnrollee(enrollee.getId()));
+        return loadEnrolleeDetails(enrollee);
+    }
+
+    /**
+     * Load enrollee tasks, profile, and kit requests
+     * (See loadForAdminView description for performance information)
+     */
+    public Enrollee loadEnrolleeDetails(Enrollee enrollee) {
```
devonbush: should we call this "loadForParticipantDashboard" to match loadForAdminView?

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -277,20 +277,33 @@ protected String stringifyPepperAddress(PepperKitAddress kitAddress) {
...
-        KitRequestStatus priorStatus = kit.getStatus();
+    private void saveKitStatus(KitRequest kitRequest, PepperKit pepperKit, Instant pepperStatusFetchedAt) {
+        KitRequestStatus priorStatus = kitRequest.getStatus();
         try {
-            kit.setExternalKit(objectMapper.writeValueAsString(pepperKit));
-            kit.setExternalKitFetchedAt(pepperStatusFetchedAt);
-            kit.setStatus(PepperKitStatus.mapToKitRequestStatus(pepperKit.getCurrentStatus()));
-            dao.update(kit);
+            kitRequest.setExternalKit(objectMapper.writeValueAsString(pepperKit));
+            kitRequest.setExternalKitFetchedAt(pepperStatusFetchedAt);
```
devonbush, line 284: nice, I'm glad we're pulling these out

core/src/main/java/bio/terra/pearl/core/service/kit/pepper/StubPepperDSMClient.java
```diff
@@ -100,4 +102,30 @@ private String getNextStatus(KitRequest kit) {
...
+        Instant scanTime = labelTime.plus(5, ChronoUnit.MINUTES);
+
+        pepperKit.setLabelDate(labelTime.toString());
+        pepperKit.setScanDate(scanTime.toString());
+
+        if (status == PepperKitStatus.RECEIVED) {
+            Instant receivedTime = scanTime.plus(7, ChronoUnit.DAYS);
+            pepperKit.setReceiveDate(receivedTime.toString());
+        }
+        return pepperKit;
```
devonbush, line 129: this is great, thanks for enhancing the mocking capability

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
```diff
@@ -114,6 +114,10 @@ public Enrollee loadForAdminView(Enrollee enrollee) {
         return dao.loadForAdminView(enrollee);
     }
 
+    public Enrollee loadEnrolleeDetails(Enrollee enrollee) {
```
devonbush: same comment about name -- or maybe this belongs in an ExtService in api-participant.  Not sure the best way to connote "this method is the core loading mechanism for the participant dashboard"

core/src/main/resources/db/changelog/changesets/2023_12_08_participant_task_kit.yaml
```diff
@@ -0,0 +1,11 @@
+databaseChangeLog:
+  - changeSet:
+      id: participant_task_kit
+      author: cunningh
+      changes:
+        - addColumn:
+            tableName: participant_task
+            columns:
+              - column: { name: kit_request_id, type: uuid, constraints: {
+                nullable: true, unique: true, foreignKeyName: fk_task_kit_request, references: kit_request(id)
```
devonbush, line 10: good call on the unique constraint.

ui-participant/src/hub/HubPage.tsx
```diff
@@ -93,6 +94,7 @@ const StudySection = (props: StudySectionProps) => {
   return (
     <>
       <h1 className="mb-4">{matchedStudy.name}</h1>
+      {enrollee.kitRequests.length > 0 && <KitTask kitRequests={enrollee.kitRequests} />}
```
devonbush: it seems odd just from a naming perspective to pass a list of requests to a component called KitTask.  (I was expecting it to take a ParticipantTask as an arugment).  Maybe rename 'KitTask' to 'KitActions' or 'KitBanner' or something like that?

ui-participant/src/hub/kit/KitTask.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  const iconProps = {
+    color: 'var(--brand-color)',
+    fontSize: '30px'
+  }
+
+  const iconClass = 'col-1'
+  const eventTextClass = 'col-8'
+  const eventDateClass = 'col-3 text-end'
+  const kitEventClass = 'row mb-3 pt-3'
```
devonbush: you could pull these to before the component defintiion so it's clearer they are not dependent on args

ui-participant/src/hub/kit/KitTask.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  const hasKitRequests = kitRequests.length > 0
+  if (!hasKitRequests) {
+    return null
+  }
+
+  const unreturnedKit = (kitRequest: KitRequest) => {
```
devonbush: I'd pick a name that more clearly indicates this is a function.  'isUnreturnedKit' or 'unreturnedKitFilter'

ui-participant/src/hub/kit/KitTask.tsx
```diff
@@ -0,0 +1,101 @@
...
+    return kitRequest.status === 'SENT'
+  }
+
+  const renderHeader = () => {
```
devonbush: this is simple enough that I would inline it in the return statement below.  Alternatively, pull it outside the component definition

ui-participant/src/hub/kit/KitTask.tsx
```diff
@@ -0,0 +1,101 @@
...
+    width: '100%',
+    color: '#595959'
+  }
+
+  const kitHeaderProps = {
+    borderBottom: '1px solid #e4e4e4',
+    width: '100%',
+    color: '#595959'
+  }
+
```
devonbush: kitEventProps and kitHeaderProps are only used once, so I would inline them

ui-participant/src/hub/kit/KitTask.tsx
```diff
@@ -0,0 +1,101 @@
...
+          </div>
+        </div>
+      })}
+    </div>
+  )
+}
+
+// TODO: this is identical to a method in ui-admin/src/util/timeUtils.tsx
+// move that to ui-core and use it here (there are other methods that should
+// also be moved so: JN-781 ) -DC
```
devonbush: thanks for creating the issue.  We haven't thought through whether the time/date stuff will be identical across both or whether there will be weird differences due to how i18n is handled, so I appreciate that you didn't try to do it in this PR

in the meantime, I'd suggest moving this function to a new util/timeUtils.tsx file in ui-participant, so it's easier to find for that future refactor

devonbush (changes requested):
This is great stuff.  Tested it locally and it works great and the dashboard display looks really sharp.  
 
Requesting changes to:
1. Add KIT_REQUEST to the `taskTypeOptions` in NotificationConfigView.tsx, so that we can configure email reminders (or maybe that's a separate ticket/PR?)
2. Fix the icon/text overlapping on mobile phones
image

 Then there's a few ticky-tack codestyle suggestions, but those aren't blocking

core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeDao.java
denniscunningham: Sure. That name seems pretty specific for a general purpose layer, but are method names a way of identifying special use cases (like those that do joins)?
Also, it looks like we made a decision to do quasi joins at the DAO layer, rather than the service layer? Is that so we have the option to do actual joins at the DAO layer in the future?

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
denniscunningham: Will do.
Do you imagine that the dashboard will be the only user of this method?

Also, I see a couple instances where we call the DAO layer directly from ext services (in admin and participant), but do not see guidance on that here: https://broadworkbench.atlassian.net/wiki/spaces/PEARL/pages/2841509953/Example+code+-+Mailing+list+administration.
What are your thoughts on that?

ui-participant/src/hub/kit/KitTask.tsx
denniscunningham: I figured this would look odd, and not surprisingly these are cut-and-paste from the kit card featrue I demonstrated. In that case, they make more sense, since they were reused a handful of times.
I kept them here for two reasons:

1. We may enhance this banner at some point: (e.g. Thank you for returning your kit....).
2. We may do something like the kit card feature at some point.

IOW, these are common stylings for kit info, and I'd like a way to pull them into one place. For the kit card development made playing with styling and keeping things consistent easier, and I think it helped readability of that code. These of course do not rise to a central style sheet level, but is there another construct that fulfills the use case I described?

denniscunningham (commented):
@devonbush Thx for your review! Much appreciated.
I left some questions.

core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeDao.java
devonbush: yeah, I think the theoretical best way to do these from a separation-of-concerns point of view is to make them `getEnrolleeWithRelations(id, relationNames)` so the caller can specify which joins they want.  But that obviously makes the methods more complex.  So yeah, explicitly calling out "this just loads the thing the PArticipant view needs" was my way of flagging that the separation of concerns wasn't clean.  

And yes, the quasi-joins in the DAO were for exactly the reason you say--we'll want to do them in SQL as soon as perf. becomes an issue (or maybe even before then, just as we realize the endpoints are relatively constant)

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
devonbush: IMO we should generally avoid it.  It's not horrific to not use a service where it's just doing loading or some other simple operation that's unlikely to need audit/other logic inserted, but  it's best to always go through a service (even if for no reason other than consistency).  If there are cases where we didn't, it's probably just due to accidentally missing that step, or some code predating a refactor.

ui-participant/src/hub/kit/KitTask.tsx
devonbush: agree, that makes sense, so I'd pull them up out of the component along with the other stuff

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
devonbush: yeah, I'd imagine the dashboard will be the only user.  It might be loaded in the admin tool if we ever add a "preview dashboard" feature, but that's a ways off

denniscunningham: Thx for that info. I will change the name rather than move to an ext service, since after reading your comment I think I misunderstood.

devonbush (changes requested):
Still seeing the icon overlap issue on mobile (disregard previous comment about date not appearing -- I hadn't properly redeployed from another branch)

image

devonbush:
also, please let me know if you can add the KIT_REQUEST to the list of taskTypeOptions in NotificationConfigView.tsx, so that we can configure email reminders (or maybe that's a separate ticket/PR?)

denniscunningham:
Erf. Missed the reminder notifications. Thx for spotting that.

I'll play with the layout to get the icon overlap fixed. Curious about how you get a local branch to work on your phone. I've been simulating smaller screens with a browser on my Mac, but I'm guessing that's not completely accurate. Again, thx for spotting that. Mobile was not on my checklist.

devonbush:
For testing phone and other resolutions, there's a button on the top-left of the Chrome debugger that activates phone/tablet simulation mode. 
image
That blue (or gray when it's off) icon that looks like a phone in front of a laptop is the button you want (or maybe that's what you're saying you already use)

devonbush:
FWIW, I've looked for ways to test local development against my actual phone, and haven't found anything good.  In past places, I used ngrok (an ingress enabler that lets you run a shell command to map localhost to `ngrok.io/some-guid` so that it's accessible via external devices), but that didn't seem to work on our Broad computers due to BITS restrictions.  I asked around and no one had an alternative way (I'm guessing since we're the only really mobile-oriented application in DSP, it just doesn't come up for everyone else)

devonbush:
Thanks for reminding me of that "test-on-actual-phone" need.  Created https://broadworkbench.atlassian.net/browse/JN-784

devonbush (approved):
nice, thanks!

ui-participant/src/hub/kit/KitBanner.tsx
```diff
@@ -0,0 +1,87 @@
...
+          <h2 className="fs-6 text-uppercase mb-0">Sample collection kits</h2>
+          <div className='container'>
+            {renderHeader()}
+            {renderSentStatus(kitRequest)}
+            <div className="list-unstyled pt-4"/>
+          </div>
+        </div>
+      })}
+    </div>
+  )
```
MatthewBemis: This case may be nonsense right now for current studies, but if there is more than one kit, the headers are repeated for each:

Screenshot 2023-12-21 at 2 50 25 PM

This would solve that and clean up the div/list structure a little bit as well:

```suggestion
  return (
    <>
      <h2 className="fs-6 text-uppercase mb-0">Sample collection kits</h2>
      {renderHeader()}
      <ul className="list-unstyled">
        {unreturnedKits.map(kitRequest => {
          return <li className="container" key={kitRequest.id}>
            {renderSentStatus(kitRequest)}
          </li>
        })}
      </ul>
    </>
  )
```

MatthewBemis (approved):
This looks great! I have one minor suggestion for some frontend code cleanup but won't hold approval for that 👍

ui-participant/src/hub/kit/KitBanner.tsx
denniscunningham: Thx for the suggestion!
That changes the padding a bit between the kit banner and surveys, but that is probably an improvement.

PR #684, opened 2024-01-02
[JN-769] Persist events 

connorlbark:
#### DESCRIPTION

Persist events in the database when they occur. Show these events on the same page in the Admin UI that you can see notifications. Now, what used to be the "notifications" page is more like a "timeline" of all the granular things that has happened with an enrollee.

image

1. Restart database and fill out with `ourhealth` data
2. In admin ui, navigate to: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHBASC/timeline
3. Ensure that the appropriate events are in the table and ordered properly in relation to notifications
4. Check postgres `event` table to make sure appropriate fields are filled out for each event


devonbush:
This looks great, and works well when I test it locally.  I'll go through the code tomorrow, but a couple of meta-notes on the PR:
1. For UX changes, please always include a screenshot
2. for the steps, instead of providing a click sequence, just provide the url of where to go (unless the click sequence is important to see behavior).  So for this, you could just say "go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHBASC/notifications`"

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -746,7 +762,7 @@ paths:
         - { name: portalShortcode, in: path, required: true, schema: { type: string } }
         - { name: studyShortcode, in: path, required: true, schema: { type: string } }
         - { name: envName, in: path, required: true, schema: { type: string } }
-        - { name: configId, in: path, required: true, schema: { type: string, format: uuid} }
+        - { name: configId, in: path, required: true, schema: { type: string, format: uuid } }
```
devonbush, line 765: thanks for linting this file

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -637,6 +637,22 @@ paths:
           content: { application/json: { schema: { type: object } } }
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/portals/v1/{portalShortcode}/studies/{studyShortcode}/env/{envName}/enrollees/{enrolleeShortcode}/events:
+    get:
+      summary: List of events created by enrollee
```
devonbush:
```suggestion
      summary: List of events relating to the specified enrollee
```

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -156,7 +156,7 @@ paths:
         - { name: portalShortcode, in: path, required: true, schema: { type: string } }
         - { name: studyShortcode, in: path, required: true, schema: { type: string } }
         - { name: envName, in: path, required: true, schema: { type: string } }
-        - { name: id, in: path, required: true, schema: { type: string, format: uuid} }
+        - { name: id, in: path, required: true, schema: { type: string, format: uuid } }
```
connorlbark, line 159: My IDE reformatted a bunch of stuff. They seem like good changes, but I'm happy to remove if we don't wanna include that. (For future PRs I reconfigured so it only reformats the lines I changed)

core/src/main/java/bio/terra/pearl/core/model/workflow/Event.java
```diff
@@ -0,0 +1,23 @@
...
+import bio.terra.pearl.core.model.BaseEntity;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.experimental.SuperBuilder;
+
+import java.util.UUID;
+
+/**
+ * Simple record of an event occurring, primarily for debugging and customer support.
```
devonbush:
```suggestion
 * Simple record of an application event occurring, used primarily for debugging and customer support.
 * Note that these are records of application events (e.g. participant signup, kit received, etc...) as opposed to LogEvents, which log warnings and errors and is essentially just storing parts of the console log in the database.  
 * It might be the case that a single incident could trigger both an "Event" and a "LogEvent"
```

core/src/test/java/bio/terra/pearl/core/service/workflow/EventServiceTests.java
```diff
@@ -0,0 +1,134 @@
...
+                bundle.portalParticipantUser());
+
+        List<Event> createdEvents = eventService.findAll();
+        Assertions.assertEquals(1, createdEvents.size());
+
+        Event createdEvent = createdEvents.get(0);
+        Assertions.assertNotNull(createdEvent.getCreatedAt());
+        Assertions.assertEquals(EventClass.ENROLLEE_CREATION_EVENT, createdEvent.getEventClass());
+        Assertions.assertEquals(bundle.enrollee().getId(), createdEvent.getEnrolleeId());
+        Assertions.assertEquals(bundle.portalParticipantUser().getPortalEnvironmentId(), createdEvent.getPortalEnvironmentId());
```
devonbush: Let's create a helper method for these assertions -- they seem to be repeated in all the test "confirmSavedEventMatches(...)"

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
```diff
@@ -0,0 +1,147 @@
+import React, { useEffect, useState } from 'react'
+import Api, { Enrollee, Event, Notification, NotificationConfig } from 'api/api'
+import { notificationConfigPath, StudyEnvContextT } from '../../StudyEnvironmentRouter'
+import LoadingSpinner from 'util/LoadingSpinner'
+import { instantToDefaultString } from 'util/timeUtils'
+import NotificationConfigTypeDisplay from '../../notifications/NotifcationConfigTypeDisplay'
+import { Link } from 'react-router-dom'
+
+const eventClassToHumanReadable = (eventClass: string): string => {
+  const output = eventClass
```
devonbush: I think you can replace this function with lodash
`_startCase(_camelCase(str))`

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
```diff
@@ -0,0 +1,147 @@
...
+    .split('_')
+    .map(str => str.toLowerCase())
+    .join(' ')
+
+  // capitalize first letter of output
+  return output[0].toUpperCase() + output.substring(1)
+}
+
+type EventNotificationTableEntry = {
```
devonbush: could this be declared instead as 
`type EventNotificationTableEntry = Notification | Event` ?

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
```diff
@@ -0,0 +1,147 @@
...
+            setIsLoading(false)
+          })
+      })
+  }, [enrollee.shortcode])
+
+  useEffect(() => {
+    const newTableData: EventNotificationTableEntry[] = []
+
+    newTableData.push(...notifications.map(notificationToTableEntry))
+    newTableData.push(...events.map(eventToTableEntry))
```
devonbush:
```suggestion
    const newTableData = [...(notifications as EventNotificationTableEntry[]), ...(events as EventNotificationTableEntry[])]
```
And if you change the type declaration of EventNotificationTableEntry to be Event | Notification you might not even need the `as` casting

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+    const sorted = newTableData.sort((obj1, obj2) => {
```
devonbush: We now use tanstack react-table in our app wherever we have a table so that we don't have to rewrite sorting logic.  See MailingListView.tsx (visible athttps://localhost:3000/ourhealth/studies/ourheart/env/sandbox/mailingList) for the simplest example. Note how line 29 of that file sets the default initial sort, which is what you'll want to do to have this be sorted.

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
```diff
@@ -0,0 +1,147 @@
...
+      return 0
+    })
+
+    setTableData(sorted)
+  }, [events, notifications])
+
+  return <div>
+    <h5>Timeline</h5>
+    <LoadingSpinner isLoading={isLoading}>
+      <table className="table table-striped">
```
devonbush: yeah, if you use react-table, you can replace all this markup with `basicTableLayout(table)`

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
```diff
@@ -0,0 +1,147 @@
...
+            const matchedConfig = currentEnv.notificationConfigs
+              .find(cfg => cfg.id === row.notificationConfigId)
+            return <tr key={row.id}>
+              <td>
+                {row.notificationConfig && <NotificationConfigTypeDisplay config={row.notificationConfig}/>}
+                {row.eventClass && eventClassToHumanReadable(row.eventClass)}
+              </td>
+
+              <td>
+                {row.deliveryType && row.deliveryType}
```
devonbush: IMO using the nullish coalesce operator is clearer at communicating "show this, but show blank if undefined"
```suggestion
                {row.deliveryType ?? ''}
```

devonbush (changes requested):
Great stuff -- I'm pleased at how quickly this came together and how much potential there is here.  Some styling suggestions.  My main blocker is that we should update this to use the same table component we've now standardized on elsewhere in the admin tool.

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
connorlbark: Genius, I have much more JS experience than TS experience so stuff like this is actually really helpful to call out. Thanks!

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  /** matches each notification to a corresponding config by id */
+  function attachConfigsToNotifications(rawNotifications: Notification[]) {
+    rawNotifications.forEach(notification => {
+      notification.notificationConfig = currentEnv.notificationConfigs
+        .find(config => config.id === notification.notificationConfigId)
+    })
+  }
+
+  useEffect(() => {
```
devonbush: this will be cleaner if you use our `useLoadingEffect` hook.  See line 89 of MailingListView.tsx for an example -- that way the service calls will get wrapped in error handling, and we'll better communicate in code that this is loading the main data for the component to display.  It will also handle the isLoading logic for you

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
```diff
@@ -0,0 +1,112 @@
...
+    Api.fetchEnrolleeNotifications(portal.shortcode, study.shortcode, currentEnv.environmentName, enrollee.shortcode)
```
devonbush: Also, consider using Promise.all so the events and notifications are fetched in parallel.  Something like
```
const results = await Promise.all(
  Api.fetchNotifications(...),
  Api.fetchEnrolleeEvents(...)
)
```
Although performance is definitely not critical here, so if it's complicated or makes the code unwieldy, don't worry about it

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
```diff
@@ -0,0 +1,112 @@
...
+  }, [enrollee.shortcode])
+
+  useEffect(() => {
+    const newTableData: (Event | Notification)[] = []
+
+    newTableData.push(...notifications)
+    newTableData.push(...events)
+
+    setTableData(newTableData)
+  }, [events, notifications])
```
devonbush: This feels weird to me to have a useEffect that's just combining the two arrays, it's essentially storing the same data in two different state variables.  

Let's eliminate the separate state variables for setEvents and setNotifications, and do this combination and calling of `setTableData` in the hook above this one

core/src/test/java/bio/terra/pearl/core/service/workflow/EventServiceTests.java
```diff
@@ -0,0 +1,129 @@
...
+        List<Event> createdEvents = eventService.findAll();
+        Assertions.assertEquals(1, createdEvents.size());
+
+        Event createdEvent = createdEvents.get(0);
+        Assertions.assertNotNull(createdEvent.getCreatedAt());
+        Assertions.assertEquals(EventClass.PORTAL_REGISTRATION_EVENT, createdEvent.getEventClass());
+        Assertions.assertNull(createdEvent.getEnrolleeId());
+        Assertions.assertEquals(portalEnv.getId(), createdEvent.getPortalEnvironmentId());
+    }
+
```
devonbush, line 122: nice!

core/src/main/java/bio/terra/pearl/core/service/workflow/EventService.java
```diff
@@ -104,19 +118,52 @@ public EnrolleeCreationEvent publishEnrolleeCreationEvent(Enrollee enrollee, Por
...
+        savePublishedEvent(EventClass.ENROLLEE_CREATION_EVENT, ppUser.getPortalEnvironmentId(), enrollee);
         applicationEventPublisher.publishEvent(enrolleeEvent);
         return enrolleeEvent;
     }
 
-    /** adds ruleData to the event, and also ensures the enrollee task list is refreshed */
+    /**
+     * Saves a record of the published event. If the event does not involve a specific enrollee/portalEnv,
+     * they can be set to null.
+     */
```
devonbush, line 129: I don't think I mentioned this in the last review, but this was a great choice to make this method private -- that really clearly communicates that saving is handled internally

devonbush (changes requested):
Looking good -- the table upgrade is a big win for readability.  One more pattern you should use that I missed the first time -- `useLoadingEffect`.  Thanks again for upgrading this component -- this was a very early component that was slapped together before we had a lot of patterns and utilities in place

api-admin/src/main/resources/api/openapi.yml
MatthewBemis: Agreed, these are great formatting changes

core/src/main/java/bio/terra/pearl/core/service/workflow/EventService.java
connorlbark: If there are two things I got from my Northeastern education, it's to:
- Never make anything public
- Make so many tests you go dizzy from looking at them

MatthewBemis (approved):
Worked well when tested, code looks great too! Nice job!

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
connorlbark: This cleaned everything up a lot! thanks

ui-admin/src/study/participants/enrolleeView/EnrolleeTimeline.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const { isLoading } = useLoadingEffect(async () => {
+    await Promise.all([
+      Api.fetchEnrolleeNotifications(portal.shortcode, study.shortcode, currentEnv.environmentName, enrollee.shortcode),
+      Api.fetchEnrolleeEvents(portal.shortcode, study.shortcode, currentEnv.environmentName, enrollee.shortcode)
+    ]).then(([notifications, events]) => {
+      attachConfigsToNotifications(notifications)
+      setTableData(tableData.concat(...notifications, ...events))
+    })
+  }, [enrollee.shortcode])
```
devonbush: Let's not mix `await` and `then` -- you should use one or the other, (and generally prefer `await`).  Also, using `concat` in setTableData seems like it would duplicate entries if this component ever reloaded the list from the server, so I think it would be better (and cleaner), to just make a new array.
```suggestion
    const [notifications, events] = await Promise.all([
      Api.fetchEnrolleeNotifications(portal.shortcode, study.shortcode, currentEnv.environmentName, enrollee.shortcode),
      Api.fetchEnrolleeEvents(portal.shortcode, study.shortcode, currentEnv.environmentName, enrollee.shortcode)
    ])
    attachConfigsToNotifications(notifications)
    setTableData([...notifications, ...events])
   
  }, [enrollee.shortcode])
```

devonbush (approved):
One more change, and then this is good to go.  going ahead and approving now so you don't have to wait for my re-approval after the minor fixup.  thanks!

PR #688, opened 2024-01-06
[JN-767] Create Juniper specific kit request data

denniscunningham:
#### DESCRIPTION
This PR accomplishes a couple functional aims: removes raw Pepper request status information from the admin UI, and adds the proper support for kit request status fields in the admin UI. It also stops transmitting inappropriate kit request details to clients (e.g. GP personnel email addresses). Theses changes also apply to data export.

The main work was creating support for a client specific KitRequest DTO, KitRequestDetails. This DTO is used for both ApiAdmin and AdminParticipant, but includes a JSON 'details' field which can be used to contain (lesser) kit request info intended only for admins. Currently the 'details' field is present for both admin and participant, but can be easily erased for the participant API in the future. The field is shown as-is as an info popup in the admin participant details page.

The conversion from KitRequest to KitRequestDetails occurs in KitRequestService. I originally coded this closer to the Web interface, at KitController then KitExtService, but that introduced complexity. Happy to discuss the details.

I moved three kit related methods from EnrolleeDao to EnrolleeService, and modified them to use KitRequestService. I believe these belong in EnrolleeService, since they were making calls into other DAO classes. These methods perhaps can be modified further to call services rather than the DAO classes. 

Also fixed 'syncKitStatusFromPepper` since it did not work properly. (It is not actually called from an API.)

1. Restart ApiAdmin and ApiParticipant.
2. Populate OurHealth.
3. Goto https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/requested/queued.

image

1. Observe the kit request column content on each of the tabs, using the refresh button to move the kit request status to the next in the sequence.
2. Goto https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHSALK.

image

1. Observe the kit request section. 
2. Click on the info popups in the 'Details' column, and observe the details provided.
3. Return to the kit requests page: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/requested/queued.
4. Press refresh until there is a kit request with 'SENT' status.
5. Goto data export page: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/export/dataBrowser
6. Observe the kit request data export items.

image

8. Goto participant portal and login as jsalk@test.com: https://sandbox.ourhealth.localhost:3001/hub.
9. You should see notification of a sent kit.

image


core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestDetails.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Holds set of kit request data that is relevant to users
+ */
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+@Slf4j
+public class KitRequestDetails {
```
devonbush: "Details" as a suffix doesn't quite seem to fit here -- this doesn't have any more detail than a KitRequest, in fact it has less.  Maybe `KitRequestDto`, or `KitRequestUiDto` to indicate that this is the DTO for sending to the frontend?

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestDetails.java
```diff
@@ -0,0 +1,85 @@
...
+  private Instant labeledAt;
+  private Instant sentAt;
+  private Instant receivedAt;
+  private String trackingNumber;
+  private String returnTrackingNumber;
+  private String errorMessage;
+  /**
+   * JSON blob for additional kit request details
+   */
+  private String details;
```
devonbush: Related to the naming concern above -- it's confusing if a class "KitRequestDetails" has a field "details".  Then what else is in the class?

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestDetails.java
```diff
@@ -0,0 +1,85 @@
...
+  private String enrolleeShortcode;
+
+  public KitRequestDetails(KitRequest kitRequest, KitType kitType, String enrolleeShortcode,
+                           ObjectMapper objectMapper) {
+    this.id = kitRequest.getId();
```
devonbush: Perhaps `BeanUtils.copyProperties(src, dest, except...)` would make this section easier to read? (or see See https://stackoverflow.com/questions/5079458/copy-specific-fields-by-using-beanutils-copyproperties if we want properties to be excluded by default). 

I expect we'll start making more DTOs like this in other parts of the application, so I'd like to establish some good patterns for declaratively communicating "this is how this DTO is different/same from the database model"

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestDetails.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+@Slf4j
+public class KitRequestDetails {
+  private UUID id;
+  private Instant createdAt;
+  private KitType kitType;
```
devonbush: if this is for display purposes, could we just send kitTypeName?

devonbush (changes requested):
Works great, and appreciate the upgrades to the parsing, testing, and stub infrastructure.  I have reservations about the name `KitRequestDetails` and how we're assembling the DTO that I want to at least discuss first

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestDetails.java
connorlbark: I totally agree on `Dto`, I was about to comment that before I saw this. My only concern is that we use `Dto` in the populate side a lot, so it might be confusing, but it still communicates intent better.

devonbush: yeah, for clarity, we try to suffix the populate Dtos with  "PopDto". So either we could start calling these "UiDtos" or we just assume DTOs are for the UI unless otherwise named (I'd lean toward this).

denniscunningham: Wide open on naming. 'KitRequestDetails' was mostly a placeholder since I did not see a pattern like this in the code. Just tell me what you like.

connorlbark: I'd vote for `KitRequestDto` personally, I agree that by default if I saw `Dto` I'd assume it was for serializing for the API/UI.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestDetails.java
denniscunningham: The use of KitType precedes these changes. 
I attempted to code this change, but we use this DTO for UI (`displayName`) and export (`name`). Those usages made sense so I decided not to make this change. Of course we could include both `displayName` and `name`, but it seems like letting clients deconstruct the class instance for a particular use makes sense.
Let me know if you see this differently.

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestDetails.java
denniscunningham: I experimented with this a bit, but the result seemed more confusing than the current implementation.
I don't think there is a lot of danger in how this is currently implemented, but perhaps I am missing something.
In any case, feel free to change this to suit your concerns.

denniscunningham (commented):
Thanks for your review!
See my responses to your comments.

devonbush (approved):

connorlbark (approved):

PR #689, opened 2024-01-08
[JN-789] Cohort builder modal

MatthewBemis:
#### DESCRIPTION

Wanted to just get this out into review because I've been sitting on it for a bit. This will need integration work with the backend once the rules engine and everything is in place, so for now it's just visible to superusers.

During demo, it was mentioned that it would be nice to make the field selector searchable, allowing the user to easily find a field if there are hundreds to sift through. I attempted to override the `fieldSelector` component of react-querybuilder , but React-Select didn't play too nicely with the internal rules builder. I'm sure we can get it to work, but it was starting to be a bit of a time sink and I didn't want to invest too much time before knowing if this was definitely the path we wanted to go down. For now, I think it's okay if HeartHive has to sift through a few fields- they have fewer surveys and fields than OurHealth.

1) Go to the portal list: https://localhost:3000/
2) Select a portal and study and try adding some criteria (note this uses live to gather the fields, so if the live environment for your selected study isn't initialized then it will be lacking options)
3) You can't save it yet, so you'll see a "Not yet implemented" message when you click Create

### DIFF
--- a/ui-admin/src/study/CreateNewCohortModal.tsx
+++ b/ui-admin/src/study/CreateNewCohortModal.tsx
@@ -8,8 +8,7 @@ import useReactSingleSelect from '../util/react-select-utils'
 import Api, { ExportData } from '../api/api'
 import { useLoadingEffect } from 'api/api-utils'
 import { Field, QueryBuilder } from 'react-querybuilder'
-import 'react-querybuilder/dist/query-builder.css'
-import { QueryBuilderBootstrap } from '@react-querybuilder/bootstrap'
+import 'react-querybuilder/dist/query-builder.scss'
 import { Store } from 'react-notifications-component'
 import { failureNotification } from '../util/notifications'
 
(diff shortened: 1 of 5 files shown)


ui-admin/src/study/CreateNewCohortModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const [selectedPortal, setSelectedPortal] = useState<Portal>()
+  const [selectedStudy, setSelectedStudy] = useState<PortalStudy>()
+  const [participantFields, setParticipantFields] = useState<ExportData>()
+
+  useLoadingEffect(async () => {
+    if (!selectedStudy || !selectedPortal) { return }
+    const response = await Api.exportEnrollees(
+      selectedPortal.shortcode,
+      selectedStudy.study.shortcode,
+      'live', { fileFormat: 'JSON', limit: 0 })
```
MatthewBemis, line 32: This seemed like a quick and easy way to get all of the fields that someone might want to match on. In the future we'll probably want a dedicated endpoint for populating the field selector that also returns the field datatype. Then we can customize the valueEditor portion of the react-querybuilder to have better handling for booleans (checkbox), dates (date picker), etc. Right now it just treats everything as strings.

PR #690, opened 2024-01-08
[JN-789] Cohort builder modal

MatthewBemis:
#### DESCRIPTION

Wanted to just get this out into review because I've been sitting on it for a bit. This will need integration work with the backend once the rules engine and everything is in place, so for now it's just visible to superusers.

During demo, it was mentioned that it would be nice to make the field selector searchable, allowing the user to easily find a field if there are hundreds to sift through. I attempted to override the `fieldSelector` component of react-querybuilder , but React-Select didn't play too nicely with the internal rules builder. I'm sure we can get it to work, but it was starting to be a bit of a time sink and I didn't want to invest too much time before knowing if this was definitely the path we wanted to go down. For now, I think it's okay if HeartHive has to sift through a few fields- they have fewer surveys and fields than OurHealth.

Screenshot 2024-01-08 at 9 50 42 AM

1) Go to the portal list: https://localhost:3000/ and click "Create a cohort"
2) Select a portal and study and try adding some criteria (note this uses live to gather the fields, so if the live environment for your selected study isn't initialized then it will be lacking options)
3) You can't save it yet, so you'll see a "Not yet implemented" message when you click Create

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,5 +29,8 @@
   },
   "scripts": {
     "lint": "eslint --max-warnings=0 e2e-tests ui-admin/src ui-core/src ui-participant/src"
+  },
+  "dependencies": {
+    "react-querybuilder": "^6.5.4"
   }
 }
--- a/package.json
+++ b/package.json
@@ -29,8 +29,5 @@
   },
   "scripts": {
     "lint": "eslint --max-warnings=0 e2e-tests ui-admin/src ui-core/src ui-participant/src"
-  },
-  "dependencies": {
-    "react-querybuilder": "^6.5.4"
   }
 }
--- a/ui-admin/src/study/CreateNewCohortModal.tsx
+++ b/ui-admin/src/study/CreateNewCohortModal.tsx
@@ -18,8 +18,10 @@ export default function CreateNewCohortModal({ onDismiss }: {onDismiss: () => vo
   const initialQuery = { combinator: 'and', rules: [] }
   const [query, setQuery] = useState(initialQuery)
   const [cohortName, setCohortName] = useState('')
-  const [selectedPortal, setSelectedPortal] = useState<Portal>()
-  const [selectedStudy, setSelectedStudy] = useState<PortalStudy>()
+  const initialPortal = portalList.length == 1 ? portalList[0] : undefined
+  const [selectedPortal, setSelectedPortal] = useState<Portal | undefined>(initialPortal)
+  const initialStudy = selectedPortal?.portalStudies.length == 1 ? selectedPortal?.portalStudies[0] : undefined
+  const [selectedStudy, setSelectedStudy] = useState<PortalStudy | undefined>(initialStudy)
   const [participantFields, setParticipantFields] = useState<ExportData>()
 
   useLoadingEffect(async () => {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -29,5 +29,8 @@
   },
   "scripts": {
     "lint": "eslint --max-warnings=0 e2e-tests ui-admin/src ui-core/src ui-participant/src"
+  },
+  "dependencies": {
```
devonbush: This should be specified in the ui-admin/package.json instead

ui-admin/src/study/CreateNewCohortModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+        </div>
+      </div>
+      <h5>Cohort Source</h5>
+      <p className="text-muted mb-2">Choose the source of your cohort participants.</p>
+      <div className="my-3">
+        <label htmlFor={selectPortalInputId}>Portal</label>
+        <Select inputId={selectPortalInputId} options={portalOptions}
+          value={selectedPortalOption} onChange={portalOnChange}/>
+        <label htmlFor={selectStudyInputId} className="mt-3">Study</label>
+        <Select isDisabled={!selectedPortal} inputId={selectStudyInputId} options={studyOptions}
```
devonbush, line 87: I like that you put the study selector here so this component can get reused in other places.

ui-admin/src/study/CreateNewCohortModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Returns a cohort builder modal
+ */
+export default function CreateNewCohortModal({ onDismiss }: {onDismiss: () => void}) {
+  const { portalList } = useNavContext()
+  const initialQuery = { combinator: 'and', rules: [] }
+  const [query, setQuery] = useState(initialQuery)
+  const [cohortName, setCohortName] = useState('')
+  const [selectedPortal, setSelectedPortal] = useState<Portal>()
+  const [selectedStudy, setSelectedStudy] = useState<PortalStudy>()
```
devonbush: selectedStudy and selectedPortal should each default if there's only one study/portal available for that user

ui-admin/src/study/CreateNewCohortModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+      selectedPortal.shortcode,
+      selectedStudy.study.shortcode,
+      'live', { fileFormat: 'JSON', limit: 0 })
+    const result = await response.json()
+    if (!response.ok) {
+      Store.addNotification(failureNotification('Failed to load cohort criteria options', result.message))
+    } else {
+      setParticipantFields(result)
+    }
+  }, [selectedStudy])
```
devonbush: You can just add a 3rd argument to useLoadingEffect to customize the error message, you don't have to check for the error yourself
```suggestion
      setParticipantFields(result)
  }, [selectedStudy], 'failed to load cohort criteria options')
```

devonbush (changes requested):
Good stuff!  A few suggestions, the blocking one is moving the package.json dependency.

devonbush (approved):

connorlbark (approved):
Works well for me! I'm wondering if we should (in the future) override the styling to make it fit more in-line with the theming of our application, but that's definitely out of the scope of this.

PR #691, opened 2024-01-08
[JN-754] Add DSM env vars to participant app

MatthewBemis:
#### DESCRIPTION

This moves the DSM env vars into the participant app so we don't run into problems in the future when we eventually need them. The participant and admin apps will have access to the same variables once this terra-helmfile PR is merged: https://github.com/broadinstitute/terra-helmfile/pull/4927

1. Run `bash ./local-dev/render_environment_vars.sh ApiParticipantApp foo@bar.org` and ensure that the DSM signing secret is now included


devonbush (approved):
nice! thanks for linking the helmfile update

pegahtah (approved):
LGTM!

PR #693, opened 2024-01-09
[JN-805] Avoid duplicating outreach tasks

MatthewBemis:
#### DESCRIPTION

I'm fairly new to events/tasks, but my understanding is:

1. A participant views an outreach activity
2. This triggers an "EnrolleeSurveyEvent" to be published from EventService.java
3. The createSurveyTasks event listener in SurveyTaskDispatcher.java picks up this event and attempts to create a new task
4. The check in `isDuplicateTask` was only de-duplicating `SURVEY` task types, not `OUTREACH` task types

1. Reboot participant API and participant UI
2. Log in to HeartHive
3. Scroll to the bottom of the dashboard and click on the `Foobar study` outreach activity
4. Refresh the page
5. You should not see a duplicate of this outreach activity

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/model/workflow/EventClass.java
+++ b/core/src/main/java/bio/terra/pearl/core/model/workflow/EventClass.java
@@ -2,6 +2,7 @@ package bio.terra.pearl.core.model.workflow;
 
 import bio.terra.pearl.core.service.consent.EnrolleeConsentEvent;
 import bio.terra.pearl.core.service.kit.KitStatusEvent;
+import bio.terra.pearl.core.service.survey.EnrolleeSurveyEvent;
 import bio.terra.pearl.core.service.workflow.BaseEvent;
 import bio.terra.pearl.core.service.workflow.EnrolleeCreationEvent;
 import bio.terra.pearl.core.service.workflow.PortalRegistrationEvent;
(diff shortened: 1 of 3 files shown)


core/src/main/java/bio/terra/pearl/core/model/workflow/EventClass.java
```diff
@@ -2,14 +2,15 @@
...
+import bio.terra.pearl.core.service.survey.EnrolleeSurveyEvent;
 import bio.terra.pearl.core.service.workflow.BaseEvent;
 import bio.terra.pearl.core.service.workflow.EnrolleeCreationEvent;
 import bio.terra.pearl.core.service.workflow.PortalRegistrationEvent;
 
 public enum EventClass {
     ENROLLEE_CREATION_EVENT(EnrolleeCreationEvent.class),
     PORTAL_REGISTRATION_EVENT(PortalRegistrationEvent.class),
-    ENROLLEE_SURVEY_EVENT(EnrolleeCreationEvent.class),
+    ENROLLEE_SURVEY_EVENT(EnrolleeSurveyEvent.class),
```
MatthewBemis, line 13: Unrelated to the fix in this PR, but I noticed this while tracing through the code

devonbush (approved):
Nice!  thanks for digging into this

connorlbark (approved):

PR #694, opened 2024-01-09
[JN-797] Audit Profile Service

connorlbark:
#### DESCRIPTION

This PR makes ProfileService automatically audit all CRUD operations by extending DataAuditedService. This changes how profile is/has been auditing; previously, a DataChangeRecord was made for every single field that changes, but now  the entire object is saved before and after. As a consequence, this ticket also significantly upgrades the DataChangesRecords table under enrollee view to automatically scrub for all changes if the DataChangeRecord's new/oldValue fields are JSON objects, as in the case of Profile audits now.

I think I got them right, but I would appreciate double checking I got the right IDs (admin/participant/portal participant/enrollee ids) in the DataAuditInfo objects in the right places.

1. Restart db and fill out with ourhealth data
2. Create a new participant in ourhealth (https://sandbox.ourhealth.localhost:3001/)
3. Accept consent
4. Fill out basic information including mailing address information
5. Go to the new enrollee's audit log and ensure that all the of changes are there (https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/<your participant shortcode\>/changeRecords)
6. Edit the mailing address and name
7. Go back to the audit log and ensure that the updates are logged


core/src/main/java/bio/terra/pearl/core/model/workflow/DataAuditInfo.java
```diff
@@ -21,4 +24,30 @@ public static DataAuditInfo fromAdminUserId(UUID responsibleAdminUserId) {
                 .operationId(UUID.randomUUID())
                 .build();
     }
+
```
devonbush: The only logic these constructors are adding is setting the operation id.  So let's just make that operationId set by default always.  Then we can get rid of these constructors (including the one above) and just use the builders instead

```
@Builder.Default
private UUID operationId = UUID.randomUUID();
```

core/src/main/java/bio/terra/pearl/core/service/participant/ProfileService.java
```diff
@@ -36,16 +39,38 @@ public Optional<Profile> loadWithMailingAddress(UUID profileId) {
...
+    public void delete(UUID profileId, DataAuditInfo auditInfo) {
         Profile profile = dao.find(profileId).get();
-        dao.delete(profileId);
+        super.delete(profileId, auditInfo);
         if (profile.getMailingAddressId() != null) {
             mailingAddressDao.delete(profile.getMailingAddressId());
         }
     }
+
+    @Override
```
devonbush, line 68: I'm not sure we need to do this.  if the mailingAddress is null, then it hasn't been changed.  So we don't need to include it in the audit trail

core/src/main/java/bio/terra/pearl/core/service/survey/AnswerProcessingService.java
```diff
@@ -42,46 +48,36 @@ public AnswerProcessingService(ObjectMapper objectMapper, ProfileService profile
...
-        ObjectWithChangeLog<Profile> profileChanges = mapValuesToType(answers, profileMappings,
+        mapValuesToType(answers, profileMappings,
                 profile, AnswerMappingTargetType.PROFILE);
-        profileService.updateWithMailingAddress(profile);
-        return profileChanges;
+
+        DataAuditInfo auditInfo = DataAuditInfo.fromEnrolleeId(enrolleeId, ppUser.getId(), responsibleUserId);
+        auditInfo.setSurveyId(surveyId);
+
+        profileService.updateWithMailingAddress(profile, auditInfo);
```
devonbush: it's a bummer this method doesn't support returning a handle to the changeRecords produced.  But that's a task for another day...

core/src/test/java/bio/terra/pearl/core/service/survey/AnswerProcessingServiceTests.java
```diff
@@ -74,9 +97,9 @@ public void testNoOpWithNoMappings() {
                 "testSurvey_q1", "myFirstName",
                 "testSurvey_q2", "addressPart1"
         ));
-        List<DataChangeRecord> changeRecords = answerProcessingService.processAllAnswerMappings(answers,
+        answerProcessingService.processAllAnswerMappings(answers,
                 new ArrayList<>(), null, UUID.randomUUID(), null, null);
-        assertThat(changeRecords, hasSize(0));
+        // TODO: assertThat(changeRecords, hasSize(0));
```
devonbush: don't forget this TODO!

devonbush (approved):
Good stuff, and I really appreciate the test coverage -- I think the tests are at a good sweet spot of testing the functionality, but not spending too much time on tests that might change next month.  

Some code cleanup suggestions to try, otherwise looks good.  As discussed, we'll want to think about cleaning up how we handle mailingAddress nested changes, but that's a task for another day

core/src/main/java/bio/terra/pearl/core/service/participant/ProfileService.java
connorlbark: The issue that this was fixing is that when the DataAuditedService ran `find` to get the old version of the model, `find` wouldn't include the MailingAddress. Because of that, every data audit made it look like the mailing address was being added anew, because it would go from null -> having a value instead of being updated or even not changed. Not 100% confident about this solution, but this seemed like an easy way to ensure that every profile audit (new and old value) was guaranteed to have the correct mailing address.

core/src/main/java/bio/terra/pearl/core/model/workflow/DataAuditInfo.java
connorlbark: Sounds good - in my head, I was using these functions as a way to direct developers but there could be so many variations of these functions it probably makes more sense to beef up the documentation on how DataAuditInfo should be used, so I will do that as well.

core/src/test/java/bio/terra/pearl/core/service/kit/KitRequestServiceTest.java
```diff
@@ -118,14 +125,14 @@ void testUpdateKitStatus(TestInfo testInfo) throws Exception {
...
-            .juniperKitId(kitRequest.getId().toString())
-            .currentStatus(PepperKitStatus.SENT.pepperString)
-            .scanDate(sentDate)
-            .errorMessage(errorMessage)
-            .build();
+                .juniperKitId(kitRequest.getId().toString())
+                .currentStatus(PepperKitStatus.SENT.pepperString)
+                .scanDate(sentDate)
+                .errorMessage(errorMessage)
+                .build();
```
pegahtah, line 132: Everything looks good, I'm wondering if the whitespace changes are correct here? (I'm not implying that they are not, just asking)

pegahtah (approved):

core/src/test/java/bio/terra/pearl/core/service/kit/KitRequestServiceTest.java
connorlbark: Honestly I have no idea what is preferred, I think this was just my default IDE settings. I think I changed it to where my IDE now should only reformat lines that changed , so I should have less of these random extra formatting changes in the future.

PR #697, opened 2024-01-16
[JN-820] Fix flaky test

connorlbark:
#### DESCRIPTION

Fixes test which failed erroneously on an earlier PR I pushed up. I believe it is due to the fact that there were database objects being created outside of the @Transcational annotation, which lead to race conditions.

All tests pass

### DIFF
--- a/core/src/test/java/bio/terra/pearl/core/service/kit/pepper/LivePepperDSMClientTest.java
+++ b/core/src/test/java/bio/terra/pearl/core/service/kit/pepper/LivePepperDSMClientTest.java
@@ -9,7 +9,10 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import org.apache.http.entity.ContentType;
-import org.junit.jupiter.api.*;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
 import org.junit.jupiter.api.function.Executable;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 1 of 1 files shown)


core/src/test/java/bio/terra/pearl/core/service/kit/pepper/LivePepperDSMClientTest.java
```diff
@@ -65,7 +70,11 @@ void startup(TestInfo info) throws Exception {
 
     @Transactional
     @Test
-    public void testBadErrorResponseFromPepper() throws Exception {
+    public void testBadErrorResponseFromPepper(TestInfo info) throws Exception {
+        Enrollee enrollee = enrolleeFactory.buildPersisted(info.getDisplayName());
```
devonbush: we've now got a helper for consistency here
```suggestion
        Enrollee enrollee = enrolleeFactory.buildPersisted(getTestName(info));
```

core/src/test/java/bio/terra/pearl/core/service/kit/pepper/LivePepperDSMClientTest.java
```diff
@@ -22,17 +25,23 @@
...
-    private Enrollee enrollee;
-    private KitRequest kitRequest;
-    private PepperKitAddress address;
 
+    @Autowired
+    private EnrolleeFactory enrolleeFactory;
+    @Autowired
+    private KitRequestFactory kitRequestFactory;
+    @Autowired
+    private ObjectMapper objectMapper;
```
devonbush, line 44: thanks for moving these to the top -- our general practice is to only have them at the bottom if it's an extremely long list

devonbush (approved):
thanks for fixing this -- one minor fixup around using a helper method for testNames

core/src/test/java/bio/terra/pearl/core/service/kit/pepper/LivePepperDSMClientTest.java
devonbush: and you'll want to update the others too

pegahtah (approved):

PR #699, opened 2024-01-16
JN-761 Delete Study button

pegahtah:
#### DESCRIPTION

Adding a delete button for a created study

**UPDATE Jan 22:** added some changes to reduce the race conditions in the tests. The tests pass more easily now, however still in some rare cases a race condition might still occur. 

Go to Juniper Home, create a new study, 
Then click on ... button, select delete, write the delete prompt and press delete. The study should be gone and Juniper should allow you to reuse the same study shortcode again


api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyExtService.java
```diff
@@ -62,6 +69,29 @@ public Study create(String portalShortcode, StudyCreationDto study, AdminUser op
     return newStudy;
   }
 
+  @Transactional
+  public void delete(String portalShortcode, String studyShortcode, AdminUser operator) {
+    if (!operator.isSuperuser()) {
+      throw new PermissionDeniedException("You do not have permission to create studies");
```
devonbush: typo 
```suggestion
      throw new PermissionDeniedException("You do not have permission to delete studies");
```

api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyExtService.java
```diff
@@ -62,6 +69,29 @@ public Study create(String portalShortcode, StudyCreationDto study, AdminUser op
...
+    NotFoundException notFoundException =
+        new NotFoundException(
+            "Portal "
+                + portalShortcode
+                + " was not found as one of the portals for study "
+                + studyShortcode);
+    Portal portal =
+        portalService
+            .findOneByShortcodeOrHostname(portalShortcode)
+            .orElseThrow(() -> notFoundException);
```
devonbush: I don't think you need to reuse the same exception across both cases
```suggestion
            .orElseThrow(() -> new NotFoundException(
            "Portal % was not found.".formatted(portalShortcode, studyShortcode);
```

api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyExtService.java
```diff
@@ -62,6 +69,29 @@ public Study create(String portalShortcode, StudyCreationDto study, AdminUser op
...
+    PortalStudy portalStudy =
+        portalStudyService
+            .findStudyInPortal(studyShortcode, portal.getId())
+            .orElseThrow(() -> notFoundException);
```
devonbush:
```suggestion
            .orElseThrow(() -> new NotFoundException(
            "Portal % was not found as one of the portals for study %s.".formatted(portalShortcode, studyShortcode);
```

ui-admin/src/api/api.tsx
```diff
@@ -487,6 +487,15 @@ export default {
     return await this.processJsonResponse(response)
   },
 
+  async deleteStudy(portalShortcode: string, studyShortcode: string): Promise<Response> {
+    const url = `${API_ROOT}/portals/v1/${portalShortcode}/studies/${studyShortcode}`
```
devonbush: we've got a helper function for some of this url
```suggestion
    const url = `${basePortalUrl(portalShortcode)}/studies/${studyShortcode}`
```

ui-admin/src/study/adminTasks/DeleteStudyModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const deleteStudy = async () => {
+    setIsLoading(true)
+
+    await Api.deleteStudy(portal.shortcode, study.shortcode).catch(() =>
+      Store.addNotification(failureNotification('Error deleting study'))
+    )
+    reload()
+    setIsLoading(false)
+    onDismiss()
```
devonbush: we have a helper function that handles error/loading stuff.  So you can simplify the above to:
```
await doApiLoad( async () => Api.deleteStudy(portal.shortcode),
{ setIsLoading} )
reload()
onDismiss()
```

ui-admin/src/study/adminTasks/DeleteStudyModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return <Modal show={true} onHide={onDismiss}>
+    <Modal.Header closeButton>
+      <Modal.Title>Delete Study</Modal.Title>
+      <div className="ms-4">
+        {study.name}
+      </div>
+    </Modal.Header>
+    <Modal.Body>
+      <div className="mb-3">
+        Are you sure you want to delete the <strong>{study.name}</strong> study? </div>
```
devonbush:
```suggestion
        Are you sure you want to delete the <strong>{study.name}</strong> study?  This
        will permanently delete the study, and all forms, configuration, and participants, across all environments and portals </div>
```

devonbush (approved):
looks good!  Congrats on your first substantive Juniper PR.  Just a few non-blocking style and content suggestions.

connorlbark (changes requested):
It looks like adding a parameter to the constructor of `StudyExtService` broke one the tests. Luckily it's a really easy fix. Basically, some tests require you to mock every single parameter to any objects you use. So, all you have to do is add the mock. The failing test is: `GlobalExceptionHandlerTest`. Add `  @MockBean private PortalService portalService;` to the class's fields and you should be all set. Then all the tests should pass

pegahtah:
> looks good! Congrats on your first substantive Juniper PR. Just a few non-blocking style and content suggestions.

Thanks Devon for pointing me to the helper functions.

core/src/main/java/bio/terra/pearl/core/service/EnvironmentService.java
```diff
@@ -16,4 +17,14 @@ public EnvironmentService(EnvironmentDao dao) {
     public Optional<Environment> findOneByName(EnvironmentName name) {
         return dao.findByName(name);
     }
+
+    @Transactional
+    public Environment buildPersisted(Environment.EnvironmentBuilder builder, String testName) {
```
connorlbark: I like this solution a lot! But I am not sure how I feel about these living outside of the tests. Hmm. Is it possible to make to make (or does it already exist?) a `TestDao` or something where Dao calls that are exclusively for tests can live? That way we only have to have the special extra SQL stuff when we need it.

connorlbark (commented):
one small thing then I'm 👍 !

core/src/main/java/bio/terra/pearl/core/service/EnvironmentService.java
pegahtah: @connorlbark To help me understand better, are you more concerned about the changes in the `EnvironmentDao` and `getCreateQuerSql`, or about the way `buildPersisted` is being called instead of `create`?
If it's the former, I thought changing the sql query for this specific dao to do nothing in times of conflicts on env `name` is something that makes sense in other parts of the code as well.is this assumption wrong?

connorlbark: I see what you're saying about the fact that it shouldn't do anything in other cases, but I would argue that it is correct behavior for it to error if the application attempted to make a new environment with the same name, simply because it shouldn't happen and I would say it's a bug/unintentional call if it does. But I'm open to changing that view of things. I am mostly worried that, since the tests didn't work, we're changing application-level code, even though (imo) that code was totally fine and was doing the intended behavior. In addition, `buildPersisted` is a testing concept, and only exists within the tests - if we decide we want to not error on conflicts, which I am cool with, then I would just replace that call with a call to just the `create` function and remove `buildPersisted` for consistency.

core/src/main/java/bio/terra/pearl/core/dao/EnvironmentDao.java
```diff
@@ -21,4 +23,26 @@ public EnvironmentDao(Jdbi jdbi) {
     public Optional<Environment> findByName(EnvironmentName environmentName) {
         return findByProperty("name", environmentName.toString());
     }
+
+    @Transactional
```
devonbush: DAOs should never use transactional annotations -- transactions are a service-layer concern

core/src/main/java/bio/terra/pearl/core/dao/EnvironmentDao.java
```diff
@@ -21,4 +23,26 @@ public EnvironmentDao(Jdbi jdbi) {
...
+    public Environment buildPersisted(Environment.EnvironmentBuilder builder, String testName) {
```
devonbush: this method doesn't seem to belong here -- it should be in the tests/factories

core/src/testFixtures/java/bio/terra/pearl/core/factory/EnvironmentFactory.java
```diff
@@ -27,10 +27,7 @@ public Environment buildPersisted(String testName, EnvironmentName envName) {
      * if the given environment exists, and either return it or a new one as needed.
      */
     public Environment buildPersisted(Environment.EnvironmentBuilder builder, String testName) {
-        Optional<Environment> environment = environmentService.findOneByName(builder.build().getName());
-        if (environment.isPresent()) {
-            return environment.get();
-        }
-        return environmentService.create(builder.build());
+
+        return environmentService.buildPersisted(builder, testName);
```
devonbush: I don't understand why you're moving this outside the factory

devonbush (changes requested):
let's talk more about those new EnvironmentService methods -- I'm not sure why they're needed

core/src/main/java/bio/terra/pearl/core/service/EnvironmentService.java
pegahtah: Ah now I see what you mean... That makes sense.

devonbush (approved):
Looks good, I don't think we need the sql update in the EnvironmentDao any more though

populate/src/main/java/bio/terra/pearl/populate/service/BasePopulator.java
```diff
@@ -31,13 +31,14 @@ public T populate(P context, boolean overwrite) {
...
         D popDto = readValue(fileString);
-        return populateFromDto(popDto, context, overwrite);
+        return this.populateFromDto(popDto, context, overwrite);
     }
 
     public D readValue(String popDtoString) throws IOException {
         return objectMapper.readValue(popDtoString, getDtoClazz());
     }
 
+    @Transactional
```
connorlbark, line 41: minor nit: I am not sure we need this, as the parent populate function is Transactional.

connorlbark (approved):
looks good!

PR #700, opened 2024-01-17
[JN-817] Fix request kits bug

connorlbark:
#### DESCRIPTION

The response entity that the endpoint was returning was not a java bean, so it was freaking out when it tried to serialize it. I made it a standard java bean format and now it works.

- Restart admin api and ui
- Fill out ourhealth
- Go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/eligible
- Check a couple participants and hit "Send sample collection kit"
- If it either works or has an address validation error, then the endpoint works properly.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
@@ -14,6 +14,10 @@ import bio.terra.pearl.core.service.study.StudyService;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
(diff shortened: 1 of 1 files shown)


connorlbark (commented):
@devonbush these kinds of errors are pretty easy to miss and are making me wonder if we should do more controller-level testing. is there a reason we don't? if not, I think I know my next member's choice activity :)

devonbush:
hmm... so this controller *used* to work, I wonder how/when it got broken.  In any case, agree that we could use some controller testing.  The reason that we didn't before was because most of our controllers were so trivial that tests wouldn't add a lot of value, so we focused our effort elsewhere.  You're right that at least we should have a habit of putting tests on controllers that aren't just passing a core model back and forth

devonbush (approved):

pegahtah (approved):

PR #706, opened 2024-01-25
[JN-762] survey config refactor

devonbush:
#### DESCRIPTION

This refactors survey configuration to the survey object itself, rather than living on the StudyEnvironmentSurvey as it did before.  There are virtually no functional changes in this PR, except that now marking a survey as required is done via a new "Options" menu from the survey editor.  The UX for that is still in-progress -- I need to run it by Erin

image

1. redeploy apiAdminApp
2. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/forms
3. note that the first two surveys are still marked as required
4. click on a non-required survey to edit it
5. use the "options" button to mark the survey as required, then save the survey, creating a new version
6. go back to the forms list, reload the page, and confirm the survey is now required


core/src/main/resources/db/changelog/changesets/2024_01_22_survey_fields.yaml
```diff
@@ -0,0 +1,52 @@
...
+                allow_participant_start = study_environment_survey.allow_participant_start, 
+                allow_participant_reedit = study_environment_survey.allow_participant_reedit,
+                prepopulate = study_environment_survey.prepopulate, 
+                recur = study_environment_survey.recur, 
+                recurrence_interval_days = study_environment_survey.recurrence_interval_days, 
+                days_after_eligible = study_environment_survey.days_after_eligible, 
+                required = study_environment_survey.required
+              from study_environment_survey
+              where survey.id = study_environment_survey.survey_id;
+        - dropColumn:
```
connorlbark, line 32: is there a good way to make ensure that this works? like, test it on the dev database or something? also, do we need to check in `study_envrionment_survey` that the SES is still active? i just had an experience from duos where we had a complex changeset like this totally break stuff and i am so scared now haha

populate/src/main/java/bio/terra/pearl/populate/service/SurveyPopulator.java
```diff
@@ -81,10 +81,7 @@ public Optional<Survey> findFromDto(SurveyPopDto popDto, PortalPopulateContext c
 
     @Override
     public Survey overwriteExisting(Survey existingObj, SurveyPopDto popDto, PortalPopulateContext context) throws IOException {
-        existingObj.setContent(popDto.getContent());
-        existingObj.setName(popDto.getName());
-        existingObj.setFooter(popDto.getFooter());
-        existingObj.setPublishedVersion(popDto.getPublishedVersion());
+        BeanUtils.copyProperties(popDto, existingObj, "id", "stableId", "version", "createdAt", "lastUpdatedAt", "answerMappings", "portalId");
```
connorlbark, line 84: maybe i'm confused, but doesn't it also need to copy `content`, `name`, `footer` and `publishedVersion`?

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -836,7 +855,7 @@ paths:
           content: { application/json: { schema: { type: object } } }
         '500':
           $ref: '#/components/responses/ServerError'
-  /api/portals/v1/{portalShortcode}/env/{envName}/triggers/{configId}/test:
+  /api/portals/v1/{portalShortcode}/studies/{studyShortcode}/env/{envName}/triggers/{configId}/test:
```
connorlbark: Was this broken? I am confused because it seems to be working on development and also in this PR, but there is no corresponding change in `api.tsx` so I feel like I'm missing something.

ui-admin/src/study/surveys/FormOptions.test.tsx
```diff
@@ -0,0 +1,82 @@
+import { render, screen, waitFor } from '@testing-library/react'
+import React from 'react'
+import { mockStudyEnvContext, mockSurvey, mockSurveyVersionsList } from 'test-utils/mocking-utils'
+import FormOptions from './FormOptions'
+import { select } from 'react-select-event'
+import userEvent from '@testing-library/user-event'
+
+jest.mock('api/api', () => ({
```
connorlbark, line 8: this is cool

ui-admin/src/study/surveys/FormOptions.tsx
```diff
@@ -0,0 +1,163 @@
...
+        onDismiss={onDismiss}/>
+    </Modal.Body>
+  </Modal>
+}
+
+
+/**
+ *
+ */
+export const VersionSelector = ({
```
connorlbark, line 92: Every time i previewed something on the version selector it disappeared

ui-admin/src/study/surveys/FormOptions.tsx
```diff
@@ -0,0 +1,163 @@
...
+      </div>}/></Modal.Title>
+    </Modal.Header>
+    <Modal.Body>
+      { isSurvey &&
+        <label className="form-label d-block">
+          <input type="checkbox" checked={(workingForm as Survey).required}
+            onChange={e => updateWorkingForm({
+              ...workingForm, required: e.target.checked
+            })}
+          /> Required
```
connorlbark: I made a required survey but it didn't update when I went back to the options; it was just unchecked.

connorlbark (changes requested):
I had some issues with the options modal, but otherwise it seems good!

core/src/main/resources/db/changelog/changesets/2024_01_22_survey_fields.yaml
devonbush: yeah, this script will run when you checked out the branch for local testing, and will also run when it gets merged to demo, so that will be our test.  (I also ran it locally a few times).  Theoretically we shouldn't need to check that the SES is still active -- we want this script to run on both active and inactive SES's.  The inactive ones are needed to update older versions of the surveys with correct configs

populate/src/main/java/bio/terra/pearl/populate/service/SurveyPopulator.java
devonbush: we do, the syntax of 'copyProperties' is that it copies all properties *except* the ones explicitly listed

api-admin/src/main/resources/api/openapi.yml
devonbush: hmm... I'm not sure -- maybe this is a merge artifact.  I'll double-check

ui-admin/src/study/surveys/FormOptions.tsx
devonbush: hmm... we'll have to do a screenshare.  It works great on my machine :)

populate/src/main/java/bio/terra/pearl/populate/service/SurveyPopulator.java
connorlbark: Ahhhhh, that is a confusing function haha

ui-admin/src/study/surveys/FormOptions.tsx
devonbush: I believe that's intentional (I agree it's confusing).  The idea is that you can't open a preview of something you're already previewing.  In any case, it predates this change.

api-admin/src/main/resources/api/openapi.yml
devonbush: yup, just did a merge and it cleaned up, thanks for calling that out

ui-admin/src/study/surveys/FormOptions.tsx
MatthewBemis: @connorlbark did you also save the survey afterwards? That part tripped me up- I forgot to save after toggling.

Requiring a save is consistent with other parts of self-service, so I don't think the behavior should necessarily change right now, but I think I'd advocate for the information in the tooltip to be on the modal itself to make it more clear.

ui-admin/src/study/surveys/SurveyEditorView.tsx
```diff
@@ -191,14 +167,19 @@ const SurveyEditorView = (props: SurveyEditorViewProps) => {
...
           studyEnvContext={studyEnvContext}
           visibleVersionPreviews={visibleVersionPreviews}
           setVisibleVersionPreviews={setVisibleVersionPreviews}
-          stableId={currentForm.stableId}
-          setShow={setShowVersionSelector}
-          show={showVersionSelector}
-          isConsentForm={isConsentForm}/>
+          workingForm={{ ...currentForm, ...draft }}
+          updateWorkingForm={(props: SaveableFormProps) => {
+            setDraft({ ...draft, ...props, date: Date.now() })
```
MatthewBemis, line 179: Nice, glad the settings are captured in drafts too!

MatthewBemis (commented):
I think there may be some style glitches on the FormOptions modal which look different than the screenshot in your PR description: I don't see the "Configuration" and "Actions" headers, and the "Other versions" tooltip is displayed directly on the modal:

Screenshot 2024-01-29 at 9 31 36 AM

devonbush:
Good catch, Matt, I had a commit get put on the wrong branch.  Just updated so the style is consistent with the screenshot

devonbush:
Just fixed merge conflict and pulled the "you must save" reminder into the dialog itself -- thanks for the suggestion Matt.  
image

MatthewBemis (approved):

ui-admin/src/study/surveys/FormOptions.tsx
connorlbark: Hmm, yeah I am finding this whole flow (editing, saving, etc.) very confusing but that is totally outside of the scope of this PR.

connorlbark (approved):

PR #708, opened 2024-01-26
JN-813 Proxy data model

pegahtah:
#### DESCRIPTION

This is the draft PR for the proxy data model. This PR definitely needs cleaning up and more comments and tests, but I would really appreciate it if I can get a first round of reviews on it while I polish it up

Currently in the draft version the register flow has changed with a hardcoded `isProxy = true` to go through the registeration and enrollment flow of a proxy and their child. So at this moment you can just spin up any environment and try to register.

This will change before I make the actual PR.


ui-participant/src/login/RedirectFromOAuth.tsx
```diff
@@ -54,11 +54,12 @@ export const RedirectFromOAuth = () => {
 
           const email = auth.user.profile.email as string
           const accessToken = auth.user.access_token
+          const isProxy = true
```
pegahtah: This was for the sake of testing the flow through the frontend, I will change this before making this an actual PR

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -125,6 +131,32 @@ private RegistrationResult register(String portalShortcode, EnvironmentName envi
         return new RegistrationResult(user, ppUser);
     }
 
+    public RegistrationResult registerGovernedUser(String portalShortcode, ParticipantUser proxy) {
```
pegahtah, line 146: Before this method is called, the parent/proxy user has been registered and exists in participant user. In time of enrollment, if the server sees that the frontend wants to enroll someone else as the enrollee (for example the child of the proxy) this method registers the child (inserts a record in the participant_user) before enrolling them

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -125,6 +131,32 @@ private RegistrationResult register(String portalShortcode, EnvironmentName envi
...
+        mainUser = participantUserService.create(mainUser);
+
+        PortalParticipantUser ppUser = new PortalParticipantUser();
+        ppUser.setPortalEnvironmentId(portalEnv.getId());
+        ppUser.setParticipantUserId(mainUser.getId());
+
+        Profile profile = new Profile();
+        profile.setContactEmail(mainUser.getUsername());
+        profile.setGivenName("Governed by proxy "+proxy.getUsername());
+        profile.setFamilyName("Governed by proxy "+proxy.getUsername());
```
pegahtah: Just mocking some data up at this point to have a prettier db, will change this

core/src/main/resources/db/changelog/changesets/proxy_data_model.yaml
```diff
@@ -0,0 +1,27 @@
+databaseChangeLog:
+  - changeSet:
+      id: create_enum_relationship
+      author: pegah
+      changes:
+        - sql:
+            sql: CREATE TYPE relationship AS ENUM ('proxy', 'guardian', 'parent', 'mother', 'father', 'other');
```
devonbush: I've never worked in a place that created types for enums, my suspicion was always that it wasn't worth the effort and I didn't like having the source-of-truth for the acceptable value set be in two places.  Would love to get your experience with them

core/src/main/resources/db/changelog/changesets/proxy_data_model.yaml
```diff
@@ -0,0 +1,27 @@
...
+                nullable: false, foreignKeyName: fk_participant_enrollee_relation_participant_user, references: participant_user(id)} }
+              - column:  { name: enrollee_id, type: uuid, constraints: { #enrollee id is the id of the user who is actually enrolled in the study
+                nullable: false, foreignKeyName: fk_enrollee_enrollee_relation_participant_user, references: enrollee(id)
+              }  }
+              - column: { name: relationship_id, type: uuid, defaultValueComputed: gen_random_uuid(), constraints: { nullable: false} }
+              - column: { name: relationship_type, type: relationship, constraints: { nullable: false} }
+              - column: { name: proxy, type: boolean, defaultValueBoolean: false, constraints: { nullable: false } }
+              - column: { name: beginDate, type: datetime, defaultValue: now(), defaultValueComputed: true, constraints: { nullable: false } }
+              - column: { name: endDate, type: datetime, constraints: { nullable: true } }
+              - column: { name: created_at, type: datetime, constraints: { nullable: false } }
```
devonbush: move the lastUpdated and createdAt definitions to below id, just for easier finding

core/src/main/resources/db/changelog/changesets/proxy_data_model.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - changeSet:
+      id: "proxy_data_model"
+      author: pegah
+      changes:
+        - createTable:
+            tableName: enrollee_relation
+            columns:
+              - column: { name: id, type: uuid, defaultValueComputed: gen_random_uuid(), constraints: { nullable: false, primaryKey: true } }
+              - column: { name: participant_user_id, type: uuid, constraints: {#participant user id is the user who is the proxy but not the enrollee
+                nullable: false, foreignKeyName: fk_participant_enrollee_relation_participant_user, references: participant_user(id)} }
```
devonbush: for naming foreign keys, fk_[tableName]_[relationName]
```suggestion
                nullable: false, foreignKeyName: fk_enrollee_relation_participant_user, references: participant_user(id)} }
```

core/src/main/resources/db/changelog/changesets/proxy_data_model.yaml
```diff
@@ -0,0 +1,27 @@
...
+              - column:  { name: enrollee_id, type: uuid, constraints: { #enrollee id is the id of the user who is actually enrolled in the study
+                nullable: false, foreignKeyName: fk_enrollee_enrollee_relation_participant_user, references: enrollee(id)
```
devonbush: same comment re foreing key name

core/src/main/resources/db/changelog/changesets/proxy_data_model.yaml
```diff
@@ -0,0 +1,27 @@
...
+              }  }
+              - column: { name: relationship_id, type: uuid, defaultValueComputed: gen_random_uuid(), constraints: { nullable: false} }
```
devonbush: how is this different than `id` above?

core/src/main/java/bio/terra/pearl/core/model/participant/EnrolleeRelation.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class EnrolleeRelation extends BaseEntity {
+    private UUID participantUserId;
+    private UUID enrolleeId;
+    private UUID relationshipId;
```
devonbush: I don't think you need this column

core/src/main/java/bio/terra/pearl/core/model/participant/EnrolleeRelation.java
```diff
@@ -0,0 +1,34 @@
...
+    private RelationshipType relationshipType;
+    private boolean isProxy;
+    @Builder.Default
+    private List<SurveyResponse> surveyResponses = new ArrayList<>();
+    @Builder.Default
+    private List<ConsentResponse> consentResponses = new ArrayList<>();
+
+    public enum RelationshipType{
```
devonbush: this enum belongs in its own file.  And I wouldn't put "OTHER" in it yet, just leave PROXY as the only option for now

core/src/main/java/bio/terra/pearl/core/model/participant/EnrolleeRelation.java
```diff
@@ -0,0 +1,34 @@
...
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class EnrolleeRelation extends BaseEntity {
+    private UUID participantUserId;
+    private UUID enrolleeId;
+    private UUID relationshipId;
+    private RelationshipType relationshipType;
+    private boolean isProxy;
```
devonbush: I think this field is redundant -- we already have the enum.  You can make an 'isProxy' method on the enum for easy boolean logic.  If you want a field that will downstream logic easier, you could add an "hasProxy" boolean on Enrollee.  But I'd lean towards not having any fields like this until we find out we really need them, because cache fields are expensive to maintain

core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeDao.java
```diff
@@ -90,6 +87,10 @@ public Optional<Enrollee> findByParticipantUserId(UUID userId, UUID studyEnviron
                 "study_environment_id", studyEnvironmentId);
     }
 
+    public Optional<Enrollee> findByEnrolleeId(UUID enrolleeId) {
```
devonbush: what is this method doing?  There is no "enrollee_id" property in this table

api-participant/src/main/java/bio/terra/pearl/api/participant/service/AuthUtilService.java
```diff
@@ -16,32 +17,45 @@
...
     }
     return enrolleeOpt.get();
   }
 
+  private Enrollee authParticipantUserToGovernedEnrollees(
+      UUID participantUserId, String enrolleeShortcode) {
+    Optional<Enrollee> governedEnrolee =
+        enrolleeRelationService.findGovernedEnrollees(participantUserId).stream()
+            .filter(enrollee -> enrollee.getShortcode().equals(enrolleeShortcode))
+            .findAny();
```
devonbush: Instead of finding an optional, and then checking it, you can just use `orElseThrow`
```suggestion
            .findFirst().orElseThrow(() -> new PermissionDenied...;
```

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
```diff
@@ -0,0 +1,37 @@
...
+@Service
+public class EnrolleeRelationService extends CrudService<EnrolleeRelation, EnrolleeRelationDao> {
+    EnrolleeService enrolleeService;
+
+    public EnrolleeRelationService(EnrolleeRelationDao enrolleeRelationDao, EnrolleeService enrolleeService) {
+        super(enrolleeRelationDao);
+        this.enrolleeService = enrolleeService;
+    }
+
+    public List<EnrolleeRelation> findEnrolleeRelationsByParticipantUserId(UUID participantUserId) {
```
devonbush: no need to put the model name in the method -- it's implied
```suggestion
    public List<EnrolleeRelation> findByParticipantUserId(UUID participantUserId) {
```

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
```diff
@@ -0,0 +1,37 @@
...
+        return dao.findByParticipantUserId(participantUserId);
+    }
+
+    public List<Enrollee> findGovernedEnrollees(UUID participantUserId) {
```
devonbush, line 34: this feels like it needs to be portal-scoped.  e.g. if I'm logging into ourhealthstudy.org, where I'm a proxy for someone, this method should only return enrollees that I'm a proxy for in that portal.  If I'm also a proxy for someone in another portal, should it return them?

core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeRelationDao.java
```diff
@@ -0,0 +1,31 @@
...
+public class EnrolleeRelationDao extends BaseMutableJdbiDao<EnrolleeRelation> {
+
+    public EnrolleeRelationDao(Jdbi jdbi) {
+        super(jdbi);
+    }
+    @Override
+    protected Class<EnrolleeRelation> getClazz() {
+        return EnrolleeRelation.class;
+    }
+
```
devonbush, line 23: might as well add an extra method here `findByParticipantUserId(UUID participantUserId, RelationType type) { return findByTwoProperties... }`
so you can simplify the service logic

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
```diff
@@ -0,0 +1,37 @@
...
+        List<EnrolleeRelation> enrolleeRelationsList =
+                findEnrolleeRelationsByParticipantUserId(participantUserId).stream().filter(EnrolleeRelation::isProxy)
+                        .collect(Collectors.toList());
+        List<Enrollee> governedEnrollees = new ArrayList<>();
+        enrolleeRelationsList.forEach(enrolleeRelation ->
+            enrolleeService.findOneByEnrolleeId(enrolleeRelation.getEnrolleeId()).ifPresent(governedEnrollees::add));
+        return governedEnrollees;
+
```
devonbush:
```suggestion
        return enroleeService.findAll(enrolleeRelations.stream().map(EnrolleeRelation::getId).toList())
```

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
```diff
@@ -0,0 +1,37 @@
...
+        super(enrolleeRelationDao);
+        this.enrolleeService = enrolleeService;
+    }
+
+    public List<EnrolleeRelation> findEnrolleeRelationsByParticipantUserId(UUID participantUserId) {
+        return dao.findByParticipantUserId(participantUserId);
+    }
+
+    public List<Enrollee> findGovernedEnrollees(UUID participantUserId) {
+        List<EnrolleeRelation> enrolleeRelationsList =
```
devonbush: in general, you don't need to add 'list' at the end of variable names unless there's a bunch of lists, maps, sets, and you need to distinguish them
```suggestion
        List<EnrolleeRelation> enrolleeRelations =
```

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
```diff
@@ -96,6 +96,9 @@ public EnrolleeService(EnrolleeDao enrolleeDao,
     public Optional<Enrollee> findOneByShortcode(String shortcode) {
         return dao.findOneByShortcode(shortcode);
     }
+    public Optional<Enrollee> findOneByEnrolleeId(UUID enrolleeId) {
```
devonbush: you don't need this method -- `find(UUID enrolleeId)` does exactly this

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -140,13 +172,32 @@ protected PreregistrationResponse validatePreRegResponseId(UUID preRegResponseId
         return preRegResponse;
     }
 
+    public String generateGUID(String prefix, EnvironmentName environmentName) {
```
devonbush: This should get consolidated in with generateShortcode in EnrolleeService as a general "generateSeureRandString(int length)" method.  Also, I would not put numbers in it to avoid 1/l 0/O issues

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -125,6 +131,32 @@ private RegistrationResult register(String portalShortcode, EnvironmentName envi
         return new RegistrationResult(user, ppUser);
     }
 
+    public RegistrationResult registerGovernedUser(String portalShortcode, ParticipantUser proxy) {
+        PortalEnvironment portalEnv = portalEnvService.findOne(portalShortcode, proxy.getEnvironmentName()).get();
+        ParticipantUser mainUser = new ParticipantUser();
+        mainUser.setEnvironmentName(proxy.getEnvironmentName());
+        String mainUserName = proxy.getUsername() + "-proxied-";
+        String guid = generateGUID(mainUserName, proxy.getEnvironmentName());
+        mainUser.setUsername(mainUserName + guid);
```
devonbush: these sorts of things are generally clearer with formatters
```suggestion
        String mainUserName =  "%s-proxied-%s".formatted(proxy.getUsername(), generateGUID(mainUserName, proxy.getEnvironmentName());
```

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -125,6 +131,32 @@ private RegistrationResult register(String portalShortcode, EnvironmentName envi
...
+        mainUser = participantUserService.create(mainUser);
+
+        PortalParticipantUser ppUser = new PortalParticipantUser();
+        ppUser.setPortalEnvironmentId(portalEnv.getId());
+        ppUser.setParticipantUserId(mainUser.getId());
+
+        Profile profile = new Profile();
```
devonbush: for this and the other constructions, use the Builders
```suggestion
        Profile profile =  Profile.builder() 
           .contactEmail(...)
           ...
           .build()
```

devonbush (commented):
The overall approach seems solid.  Lots of nitpicks and style stuff.  I think the main next step is to create synthetic proxy users.

Add two new users in populate/src/main/resources/seed/portals/demo/participants (one for the proxy and one for the enrollee)
Then add an enrollee  in populate/src/main/resources/seed/portals/demo/studies/heartdemo/enrollees.  
You'll then want to add a `List<EnrolleeRelation> relations` to Enrollee, and `List<EnrolleeRelationPopDto> relationPopDtos` to EnrolleePopDto so that you can specify the proxy relationship by username (see e.g. how EnrolleePopDto has a "linkedUsername" property to allow linking of an enrollee to a participantUser

core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
```diff
@@ -114,6 +134,33 @@ public HubResponse<Enrollee> enroll(ParticipantUser user, PortalParticipantUser
         return hubResponse;
     }
 
+    private Enrollee enrollGovernedUser(String portalShortcode, ParticipantUser proxyUser, StudyEnvironment studyEnv, PortalParticipantUser ppUser, UUID preEnrollResponseId) {
+        // Before this, at time of registration we have registered the proxy as a participant user, but now we need to both register and enroll the child they are enrolling
+        RegistrationService.RegistrationResult registrationResult = registrationService.registerGovernedUser(portalShortcode, proxyUser);
+        log.info("Governed user {} registered", registrationResult.participantUser().getId());
+        ParticipantUser mainUser = registrationResult.participantUser();
+        PortalParticipantUser ppMainUser = registrationResult.portalParticipantUser();
```
MatthewBemis: Is there a consistent naming scheme that we can adopt to more quickly differentiate between the different user types? Between `proxyUser` `ppUser`, `mainUser`, `ppMainUser`, and `governedUser`, it took me a bit of time to understand what was happening here.

This applies generally too, not just within this method

MatthewBemis (commented):
Agreed that the approach looks good. I had one additional comment about naming that I think might clean things up a bit too

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
pegahtah: I see your point, it raises the question for me that does it need to be portal scoped or study_portal scoped?

core/src/main/resources/db/changelog/changesets/proxy_data_model.yaml
pegahtah: I totally understand what you mean, I haven't had a lot of experience but in my experience, the decision often depends on the scale and complexity of the project. In large-scale applications, especially if new contributors are added to the project a lot, defining types for enums can be very beneficial for code quality, maintainability, and ensuring strong typing.

However, in smaller projects or scripts where speed of development is more important than strict type safety, the additional effort might not be necessary. We can argue that any project in DDP is of the second kind, but Juniper still gets new devs every quarter because of rotations. However, I think taking into account what you said, it looked like it made more sense for me to remove the enum type from here.

core/src/main/java/bio/terra/pearl/core/service/participant/ParticipantUtilService.java
```diff
@@ -0,0 +1,29 @@
+package bio.terra.pearl.core.service.participant;
+
+import java.security.SecureRandom;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+public class ParticipantUtilService {
```
devonbush: Consider renaming to "RandomUtilService"-- that's more specific since basically everything in Juniper touches participants

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
devonbush: Don't forget this comment!

ui-participant/src/login/RedirectFromOAuth.tsx
devonbush: don't forget to remove this!

devonbush: You could pull this from an environment variable "ENROLL_AS_PROXY"  if you don't want to worry about accidentally leaving it set.  you could add that in the .env file defaulted to false.

core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
```diff
@@ -83,8 +97,8 @@ public PreEnrollmentResponse createAnonymousPreEnroll(
     }
 
     @Transactional
-    public HubResponse<Enrollee> enroll(ParticipantUser user, PortalParticipantUser ppUser, EnvironmentName envName,
-                                            String studyShortcode, UUID preEnrollResponseId) {
+    public HubResponse<Enrollee> enroll(String portalShortcode, ParticipantUser user, PortalParticipantUser ppUser, EnvironmentName envName,
+                                            String studyShortcode, UUID preEnrollResponseId, boolean isProxy) {
```
devonbush: Let's reshuffle these arguments so that the portal/study/env args are together.  
```suggestion
    public HubResponse<Enrollee> enroll(String portalShortcode, EnvironmentName envName, String studyShortcode, ParticipantUser user, PortalParticipantUser ppUser, UUID preEnrollResponseId, boolean isProxy) {
```

core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
```diff
@@ -83,8 +97,8 @@ public PreEnrollmentResponse createAnonymousPreEnroll(
...
         log.info("creating enrollee for user {}, study {}", user.getId(), studyShortcode);
         StudyEnvironment studyEnv = studyEnvironmentService.findByStudy(studyShortcode, envName).get();
```
devonbush:
```suggestion
        StudyEnvironment studyEnv = studyEnvironmentService.findByStudy(studyShortcode, envName).orElseThrow(() -> new NotFoundException("Study environment %s %s not found".formatted(studyShortcode, envName)));
```

core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
```diff
@@ -114,6 +134,33 @@ public HubResponse<Enrollee> enroll(ParticipantUser user, PortalParticipantUser
         return hubResponse;
     }
 
+    private Enrollee enrollGovernedUser(String portalShortcode, ParticipantUser proxyUser, StudyEnvironment studyEnv, UUID preEnrollResponseId) {
+        // Before this, at time of registration we have registered the proxy as a participant user, but now we need to both register and enroll the child they are enrolling
+        RegistrationService.RegistrationResult registrationResult = registrationService.registerGovernedUser(portalShortcode, proxyUser);
+        Portal portal = portalService.findOneByShortcode(portalShortcode).orElseThrow();
```
devonbush:
```suggestion
        Portal portal = portalService.findOneByShortcode(portalShortcode).orElseThrow(() -> new NotFoundException("portal %s not found", portalShortcode)));
```

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -125,6 +136,32 @@ private RegistrationResult register(String portalShortcode, EnvironmentName envi
...
+    public RegistrationResult registerGovernedUser(String portalShortcode, ParticipantUser proxy) {
+        PortalEnvironment portalEnv = portalEnvService.findOne(portalShortcode, proxy.getEnvironmentName()).get();
+        ParticipantUser governedUser = new ParticipantUser();
+        governedUser.setEnvironmentName(proxy.getEnvironmentName());
+        String guid = generateGUID(governedUser.getUsername(), proxy.getEnvironmentName());
+        String governedUserName = proxy.getUsername() + GOVERNED_USER_INDICATOR + guid;
+        governedUser.setUsername(governedUserName);
+        governedUser = participantUserService.create(governedUser);
+
+        PortalParticipantUser ppUser = new PortalParticipantUser();
```
devonbush:
```suggestion
        PortalParticipantUser governedPpUser = new PortalParticipantUser();
```

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -39,13 +45,19 @@ public class RegistrationService {
     private EventService eventService;
     private ObjectMapper objectMapper;
 
+    private static final String CHARACTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+    private static final String GOVERNED_USER_INDICATOR = "-proxied-";
+    private static final int LENGTH = 10;
```
devonbush: more descriptive name here, and we definitely don't need 10 chars, 4 is plenty given that this is being appended to study/env/proxy specific email
```suggestion
    private static final int GOVERNED_EMAIL_SUFFIX_LENGTH = 4;
```

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -39,13 +45,19 @@ public class RegistrationService {
     private EventService eventService;
     private ObjectMapper objectMapper;
 
+    private static final String CHARACTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+    private static final String GOVERNED_USER_INDICATOR = "-proxied-";
```
devonbush: shortening for easier readability in admin tool
```suggestion
    private static final String GOVERNED_USERNAME_INDICATOR = "-prox-";
```

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -39,13 +45,19 @@ public class RegistrationService {
     private EventService eventService;
     private ObjectMapper objectMapper;
 
+    private static final String CHARACTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
devonbush:
```suggestion
    private static final String GOVERNED_USERNAME_SUFFIX_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
```

core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
```diff
@@ -114,6 +134,33 @@ public HubResponse<Enrollee> enroll(ParticipantUser user, PortalParticipantUser
...
+        log.info("Governed user {} registered", registrationResult.participantUser().getId());
+        ParticipantUser governedUser = registrationResult.participantUser();
+        PortalParticipantUser ppMainUser = registrationResult.portalParticipantUser();
```
devonbush: calling this "main" really threw me off
```suggestion
        PortalParticipantUser governedPpUser = registrationResult.portalParticipantUser();
```

core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
```diff
@@ -114,6 +134,33 @@ public HubResponse<Enrollee> enroll(ParticipantUser user, PortalParticipantUser
...
+
+        EnrolleeRelation enrolleeRelation = EnrolleeRelation.builder()
+                .enrolleeId(enrollee.getId())
+                .participantUserId(proxyUser.getId())
+                .relationshipType(RelationshipType.PROXY)
+                .portalId(portal.getId())
+                .build();
+
+        enrolleeRelationService.create(enrolleeRelation);
+        return  enrollee;
```
devonbush: for clarity and easier testing, I would create a record that contains everything this service method created, and then return that.
```suggestion
        return  new GovernedEnrolleeCreationRecord(enrollee, enrolleeRelation, governedPpUser, governedUser);
```

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -140,13 +177,26 @@ protected PreregistrationResponse validatePreRegResponseId(UUID preRegResponseId
         return preRegResponse;
     }
 
+    public String generateGUID(String userName, EnvironmentName environmentName) {
```
devonbush: this method could be renamed -- it's not really generating a GUID
```suggestion
    public String generateGovernedUsername(String proxyUserName, EnvironmentName environmentName) {
```

devonbush (changes requested):
nice!  just a bunch of style changes, hopefully nothing that takes too long

core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeRelationDao.java
```diff
@@ -0,0 +1,35 @@
...
+    public List<EnrolleeRelation> findByParticipantUserIdAndPortalId(UUID participantUserId, UUID portalId){
+        return findAllByTwoProperties("participant_user_id", participantUserId, "portal_id", portalId);
+    }
+
+    public List<EnrolleeRelation> findByEnrolleeIdAndPortalId(UUID enrolleeId, UUID portalId){
+        return findAllByTwoProperties("enrollee_id", enrolleeId, "portal_id", portalId);
+    }
+
+    public Optional<EnrolleeRelation> findByParticipantUserId(UUID participantUserId, RelationType type) {
```
MatthewBemis: Should this be called `findByParticipantUserIdAndRelationshipType`?

MatthewBemis: Also, is `RelationType` the correct type? It looks like it's importing `javax.management.relation.RelationType`. Should this be `RelationshipType`?

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
```diff
@@ -237,6 +238,7 @@ public void deleteByStudyEnvironmentId(UUID studyEnvironmentId, Set<CascadePrope
     public Enrollee create(Enrollee enrollee) {
         if (enrollee.getShortcode() == null) {
             enrollee.setShortcode(generateShortcode());
+
```
MatthewBemis:
```suggestion
```
nit: remove extra new line

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -125,6 +137,33 @@ private RegistrationResult register(String portalShortcode, EnvironmentName envi
...
+        String governedUserName = GOVERNED_USERNAME_INDICATOR.formatted(proxy.getUsername(), governedUsernameSuffix);//a@b.com-prox-guid
+        governedUser.setUsername(governedUserName);
+        governedUser = participantUserService.create(governedUser);
+
+        PortalParticipantUser governedPpUser = new PortalParticipantUser();
+        governedPpUser.setPortalEnvironmentId(portalEnv.getId());
+        governedPpUser.setParticipantUserId(governedUser.getId());
+
+        Profile governedProfile =  Profile.builder()
+                .contactEmail(governedUser.getUsername())
```
MatthewBemis: Should the contact email be set to the contact email of the proxy?

core/src/main/resources/db/changelog/changesets/proxy_data_model.yaml
```diff
@@ -0,0 +1,25 @@
+databaseChangeLog:
+#  - changeSet:
+#      id: create_enum_relationship
+#      author: pegah
+#      changes:
+#        - sql:
+#            sql: CREATE TYPE relationship AS ENUM ('PROXY', 'guardian', 'parent', 'mother', 'father', 'other');
```
MatthewBemis:
```suggestion
```

core/src/test/java/bio/terra/pearl/core/service/workflow/EnrollmentWorkflowTests.java
```diff
@@ -146,6 +156,91 @@ public void testParticipantWorkflow(TestInfo info) {
     }
 
 
+    @Test
+    @Transactional
+    public void testGovernedUserEnrollment(){
+        PortalEnvironment portalEnv = portalEnvironmentFactory.buildPersisted("testGovernedUserEnrollment");
+        StudyEnvironment studyEnv = studyEnvironmentFactory.buildPersisted(portalEnv, "testGovernedUserEnrollment");
+        ParticipantUserFactory.ParticipantUserAndPortalUser userBundle = participantUserFactory.buildPersisted(portalEnv,"testGovernedUserEnrollment");
```
MatthewBemis: `getTestName` can be used to pass in the test name instead of having to pass in the static string everywhere

MatthewBemis (commented):
It's exciting to see this coming together! Just have a few small comments/questions

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
devonbush: yes.

devonbush (changes requested):
Tested it locally, looks good, pending Matt's comments.  The blockers for me are the contactEmail of the governedUser should be the email of the proxy (otherwise we'll have to refactor a lot of logic to always check for proxies before emailing) and the RelationType improt Matt found

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
pegahtah: Great catch!! I missed that! I have made a fix and a test to check it

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -125,6 +137,33 @@ private RegistrationResult register(String portalShortcode, EnvironmentName envi
         return new RegistrationResult(user, ppUser);
     }
 
+    public RegistrationResult registerGovernedUser(String portalShortcode, ParticipantUser proxy) {
```
devonbush, line 146: this method should have @Transactional on it for the case where it's called on it's own (e.g. if a proxy is registering a second governed user, and so this isn't being called from within another registration transaction

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -125,6 +137,33 @@ private RegistrationResult register(String portalShortcode, EnvironmentName envi
...
+
+        PortalParticipantUser governedPpUser = new PortalParticipantUser();
+        governedPpUser.setPortalEnvironmentId(portalEnv.getId());
+        governedPpUser.setParticipantUserId(governedUser.getId());
+
+        Profile governedProfile =  Profile.builder()
+                .contactEmail(proxy.getUsername())
+                .givenName(null)
+                .familyName(null)
+           .build();
```
devonbush: We want to use the contact email, not the username.  this will cover the case where the proxy has changed their contact email, and then registers a governed user. (e.g. a parent registers a kid, changes address, then registers another kid)
```suggestion
        PortalParticipantUser proxyPpUser = portalParticipantUserService.findOne(proxy.getId(), portalShortcode).orElseThrow(
                () -> new NotFoundException("Portal user not found for proxy in portal %s".formatted(portalShortcode))
        );
        Profile proxyProfile = profileService.find(proxyPpUser.getProfileId()).orElseThrow();
        Profile governedProfile =  Profile.builder()
                .contactEmail(proxyProfile.getContactEmail())
                .givenName(null)
                .familyName(null)
           .build();
```

core/src/main/resources/db/changelog/changesets/proxy_data_model.yaml
```diff
@@ -0,0 +1,19 @@
...
+            tableName: enrollee_relation
+            columns:
+              - column: { name: id, type: uuid, defaultValueComputed: gen_random_uuid(), constraints: { nullable: false, primaryKey: true } }
+              - column: { name: created_at, type: datetime, constraints: { nullable: false } }
+              - column: { name: last_updated_at, type: datetime, constraints: { nullable: false } }
+              - column: { name: participant_user_id, type: uuid, constraints: {#participant user id is the user who is the proxy but not the enrollee
+                nullable: false, foreignKeyName: fk_enrollee_relation_participant_user, references: participant_user(id)} }
+              - column:  { name: enrollee_id, type: uuid, constraints: { #enrollee id is the id of the user who is actually enrolled in the study
+                nullable: false, foreignKeyName: fk_enrollee_relation_enrollee, references: enrollee(id)}  }
+              - column: { name: portal_id, type: uuid, constraints: { nullable: false, foreignKeyName: fk_enrollee_relation_portal, references: portal(id) } }
```
devonbush: Sorry, this portal_id column is on the ERD, but I just realized that we don't need it.  And I see that you're not using it anywhere.  So let's remove it

devonbush (changes requested):
getting closer!   thanks for your persistence on this.   Sorry, two small things that I missed in earlier reviews, and an update so that we're setting the governed email to the proxy's contact email, rather than their username.  thanks!

core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
pegahtah: Ah! great catch!

devonbush (approved):
Nice! Thanks again for your persistence in seeing this through -- and sorry again about the ERD having bad info in it.  I just went through and updated some other parts of the ERD that had gotten out of date

core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeRelationDao.java
```diff
@@ -0,0 +1,39 @@
...
+        return EnrolleeRelation.class;
+    }
+
+    public List<EnrolleeRelation> findByParticipantUserIdAndPortalId(UUID participantUserId, UUID portalId){
+        return findAllByTwoProperties("participant_user_id", participantUserId, "portal_id", portalId);
+    }
+
+    public List<EnrolleeRelation> findByEnrolleeIdAndPortalId(UUID enrolleeId, UUID portalId){
+        return findAllByTwoProperties("enrollee_id", enrolleeId, "portal_id", portalId);
+    }
```
MatthewBemis: I think this is what's causing a few of your unit test failures- since `portal_id` is no longer a property of `EnrolleeRelation`, these cause a SQL exception to be thrown

MatthewBemis (approved):

PR #710, opened 2024-01-29
[JN-831] Fix undefined on profile

connorlbark:
#### DESCRIPTION

If you created an enrollee but did not fill out basics or consent, the profile page could render certain rows as `undefined` instead of "Not provided". This fixes that.

Before filling out basic
image

After
image

- Create new participant but do not consent or start basics survey
- Go to the profile for that participant
- Ensure it says `Not Provided` for everything

### DIFF
--- a/ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
+++ b/ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
@@ -96,13 +96,22 @@ export function ReadOnlyProfile(
   }
 ) {
   const mailingAddress = profile.mailingAddress
+
   return <>
-    <ReadOnlyRow title={'Name'} values={[`${profile.givenName} ${profile.familyName}`]}/>
+    <ReadOnlyRow title={'Name'}
+      values={[
+        !isEmpty(profile.givenName) || !isEmpty(profile.familyName)
+          ? `${profile.givenName || ''} ${profile.familyName || ''}`
+          : ''
+      ]}/>
     <ReadOnlyRow title={'Birthdate'} values={[dateToDefaultString(profile.birthDate)]}/>
     <ReadOnlyRow title={'Primary Address'} values={[
       mailingAddress.street1,
       mailingAddress.street2,
-      `${mailingAddress.city}, ${mailingAddress.state} ${mailingAddress.postalCode}`
+      (!isEmpty(mailingAddress.city) || !isEmpty(mailingAddress.state) || !isEmpty(mailingAddress.postalCode))
+        ? `${mailingAddress.city || '(not provided)'}, ` +
+        `${mailingAddress.state || '(not provided)'} ${mailingAddress.postalCode || '(not provided)'}`
+        : ''
     ]}/>
     <ReadOnlyRow title={'Email'} values={[profile.contactEmail]}/>
     <ReadOnlyRow title={'Phone'} values={[profile.phoneNumber]}/>
--- a/ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
+++ b/ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
@@ -157,12 +157,12 @@ export function EditableProfile(
       <div className='row'>
         <div className='col'>
           <input className="form-control" type="text" value={profile.givenName || ''}
-            placeholder={'Given Name'}
+            placeholder={'Given Name'} aria-label={'Given Name'}
             onChange={e => onFieldChange('givenName', e.target.value)}/>
         </div>
         <div className='col'>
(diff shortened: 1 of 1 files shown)


ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
```diff
@@ -96,13 +96,22 @@ export function ReadOnlyProfile(
...
 ) {
   const mailingAddress = profile.mailingAddress
+
   return <>
-    <ReadOnlyRow title={'Name'} values={[`${profile.givenName} ${profile.familyName}`]}/>
+    <ReadOnlyRow title={'Name'}
+      values={[
+        !isEmpty(profile.givenName) || !isEmpty(profile.familyName)
+          ? `${profile.givenName || ''} ${profile.familyName || ''}`
+          : ''
```
devonbush:
```suggestion
        `${profile.givenName || ''} ${profile.familyName || ''}`.trim()
```

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
```diff
@@ -96,13 +96,22 @@ export function ReadOnlyProfile(
...
+      ]}/>
     <ReadOnlyRow title={'Birthdate'} values={[dateToDefaultString(profile.birthDate)]}/>
     <ReadOnlyRow title={'Primary Address'} values={[
       mailingAddress.street1,
       mailingAddress.street2,
-      `${mailingAddress.city}, ${mailingAddress.state} ${mailingAddress.postalCode}`
+      (!isEmpty(mailingAddress.city) || !isEmpty(mailingAddress.state) || !isEmpty(mailingAddress.postalCode))
```
devonbush: this is kind of gross -- and addresses are only going to get more complicated over time.  I'd just make a separate "ReadOnlyAddress" component

devonbush (approved):
Looks good, one suggestion that might be out of scope for this PR.  Also, I didn't catch this in the previous PR, but we'll want to make sure the title/values in ReadOnlyRow use `label` elements so they are accessible to visually impaired users.  If that's a quick fix, you can do it here, otherwise, please add a ticket in the upcoming sprint to fix it

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
connorlbark: I agree - I will fix that as part of this PR.

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
```diff
@@ -286,7 +323,7 @@ export function FormRow(
   }
 ) {
   return <>
-    <div className="w-25 fw-bold mb-4 mt-2">
+    <div className="w-25 fw-bold mb-4 mt-2" aria-label={title}>
       {title}
     </div>
     <div className="w-75 mb-4">
```
devonbush, line 329: Nice work on the aria-labels up above.

I don't think adding the aria label here addresses the accessibility concern though.  The accessibility concern is that if you're just using divs, there's nothing that tells a blind reader that the title is connected to the content.  So what you want to do is use labels and htmlFor so the relation between the two elements is clear.  So you can use  `const contentId =  useId()` and then:

```suggestion
    <label className="w-25 fw-bold mb-4 mt-2" htmlFor={contentId}>
      {title}
    </div>
    <div className="w-75 mb-4" id={contentId}>
```

devonbush (approved):

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
connorlbark: Ahhh, I understand. I have not done accessibility stuff before so this is new to me. Thanks

connorlbark: Does the aria id need to be on the `input` ? Or is your code enough to make it accessible?

devonbush: yes, that's right, the id should be on the input -- good catch!

devonbush:
don't forget to add a second reviewer (Matt)

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
connorlbark: Hmm, then there's an issue: There are multiple inputs to one label, e.g., `Name` has `Given Name` and `Family Name`. As far as I can tell, without doing some field weirdness (`<fieldset>`), `aria-label` seems like the best we can do.

MatthewBemis (approved):

PR #713, opened 2024-01-30
[JN-658] Add Spanish and Dev translations to Demo Basics survey

MatthewBemis:
#### DESCRIPTION

This adds 2 new languages to the basics survey for the demo study. The "dev" language was added so we can more easily confirm that translations are working as expected, without having to translate anything back to English.

Spanish translations were hallucinated courtesy of GitHub Copilot (although I speak some Spanish and for the most part they actually looked fairly okay to me).

https://github.com/broadinstitute/juniper/assets/7257391/743edc24-e0ff-492c-a2bc-22aa4073e574

1. Start up AdminApiApp
2. Repopulate the demo study: `./scripts/populate_portal.sh demo`
3. View the Basics survey: https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms/surveys/hd_hd_basicInfo?readOnly=false
4. Go to the JSON Editor tab to make sure it renders
5. Go to the Preview tab, and on the right you can toggle between the languages by typing in "es" for Spanish and "dev" for the fake dev language

Out of scope for this PR: The visual question editor doesn't like the new structure of the survey JSON. It will be need to be updated to handle the fact that every translated field is now a struct rather than a string. In the meantime, those fields will display as [object Object]. This doesn't cause any errors and in the meantime, the JSON editor still works as intended.

### DIFF
--- a/populate/src/main/resources/seed/portals/demo/studies/heartdemo/surveys/basic.json
+++ b/populate/src/main/resources/seed/portals/demo/studies/heartdemo/surveys/basic.json
@@ -24,7 +24,11 @@
     "questionStableId": "hd_hd_basic_birthSex", "targetType": "PROFILE", "targetField": "sexAtBirth", "mapType": "STRING_TO_STRING"
   }],
   "jsonContent": {
-    "title": "The Basics",
+    "title": {
+      "default": "The Basics",
+      "es": "Los Basicos",
+      "dev": "DEV_The Basics"
+    },
     "showQuestionNumbers": "off",
     "showProgressBar": "bottom",
     "pages": [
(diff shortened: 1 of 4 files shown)


devonbush (approved):
awesome, really like how you did the DEV language.  It'll be nice when we can have the locale selector in the preview be a dropdown instead of freetext, but I think that should wait until we add the "supported languages" concept to portals to do that

connorlbark (approved):
~~it looks great! But when you go back to the fancy editor, for the question it just says `[object Object]` since the layout is more complex now.~~
~~That fix might not be in the scope of this PR, so I'll approve, but definitely something we'll have to figure out a solution for.~~
EDIT: Just read your comment on your PR, in that case 👍 !

PR #714, opened 2024-01-31
JN-823 survey task updating

devonbush:
#### DESCRIPTION
This enables admins to update the versions assigned to participants of a given survey.  The concrete use case is to enable OurHealth to update the versions assigned to their soft-launch participants. 

As part of this, this includes some new designs from Erin for Form management pages:
image
This is definitely in-progress design -- starting the process of moving away from isolated sandbox/irb pages, and moving towards more holistic management.

image

This also updates ParticipantTasks to be audited entities.  

This PR is very long, but most of it is relatively boring bookkeeping related to making the participant tasks audited, and also adding some new populate stuff for demo so that we have an environment where multiple people are assigned to different versions of a survey.  For the review, focus on the changes to ParticipantTaskDao, ParticipantTaskService, ParticipantTaskExtService, and then the frontend changes to the forms UX.  

1. redeploy apiAdminApp
2. repopulate demo
3. go to `https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms`
4. click "v3" of the Social Determinants of Health survey
5. See the listing of the versions and tasks
6. Clikc "update all to version 3"
7. confirm the participants are all updated


api-admin/src/main/resources/api/openapi.yml
```diff
@@ -609,6 +626,40 @@ paths:
...
+      parameters:
+        - { name: portalShortcode, in: path, required: true, schema: { type: string } }
+        - { name: studyShortcode, in: path, required: true, schema: { type: string } }
+        - { name: envName, in: path, required: true, schema: { type: string } }
+      requestBody:
+        required: true
+        content: { application/json: { schema: { type: object } } }
+      responses:
+        '200':
+          description: BasicMetricDatum objects
```
MatthewBemis: Looks like `BasicMetricDatum` is leftover copy/paste

api-admin/src/main/resources/application.yml
```diff
@@ -19,7 +19,7 @@ env:
     billingProfileId: ${BILLING_PROFILE_ID:550212cf-45dc-478f-9fe6-9b3290e2a1fe}
     tdrExportEnabled: ${TDR_EXPORT_ENABLED:false}
   dsm:
-    useLiveDsm: ${USE_LIVE_DSM:false}
+    useLiveDsm: ${USE_LIVE_DSM:true}
```
MatthewBemis: Is this intentionally flipped?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConfiguredSurveyController.java
```diff
@@ -30,6 +31,27 @@ public ConfiguredSurveyController(
     this.surveyExtService = surveyExtService;
   }
 
+  /**
+   * gets all StudyEnvironmentSurveys for the given study, along with their associated Survey (minus
+   * content)
+   */
+  @Override
+  public ResponseEntity<Object> findWithNoContent(
```
connorlbark, line 39: Unsure how I feel about the name of this function - feels a bit ambiguous, but otherwise looks good

ui-admin/src/components/forms/Button.tsx
```diff
@@ -115,3 +116,14 @@ export const IconButton = (props: IconButtonProps) => {
     </Button>
   )
 }
+
+/**
+ *
```
connorlbark: nit: fill out or remove comment

ui-admin/src/study/surveys/SurveyEnvironmentTable.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+
+/**
+ * shows a list of surveys and the version currently live in each environment
+ */
+export default function SurveyEnvironmentTable(props: SurveyTableProps) {
+  if (props.stableIds.length === 0) {
+    return <div className="fst-italic fw-light pb-3 ps-2">None</div>
+  }
+  return <table className="table table-striped table-bordered">
```
connorlbark: Is there a reason not to use tanstack react table here? You would have to generate the column definitions on the fly... but I think it could work.

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcher.java
```diff
@@ -49,10 +50,14 @@ public void createSurveyTasks(EnrolleeEvent enrolleeEvent) {
                 enrolleeEvent.getPortalParticipantUser(),
                 enrolleeEvent.getEnrolleeRuleData(),
                 studyEnvSurveys);
+        DataAuditInfo auditInfo = DataAuditInfo.builder()
+                .systemProcess(getClass().getSimpleName() + ".createSurveyTasks")
```
connorlbark, line 54: Maybe make a helper for formatting the systemProcess, since it will be pretty common?

core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
```diff
@@ -85,18 +57,55 @@ public Optional<StudyEnvironmentSurvey> findActiveBySurvey(UUID studyEnvId, UUID
...
+                if (surveyStableId != null) {
+                    query = query.bind("stableId", surveyStableId);
+                }
+                return query.mapTo(clazz)
+                        .list();
+        });
+    }
+
+    /** gets all the study environment surveys and attaches the relevant survey objects in a batch */
+    public List<StudyEnvironmentSurvey> findAllWithSurvey(UUID studyEnvId, Boolean active) {
```
connorlbark, line 91: should we add a couple tests to ensure attachSurveys works?

core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
```diff
@@ -280,6 +280,21 @@ protected List<T> findAllByTwoProperties(String column1Name, Object column1Value
         );
     }
 
+    protected List<T> findAllByTwoProperties(String column1Name, Object column1Value,
+                                             String column2Name, Collection<?> column2Values) {
```
connorlbark, line 284: Should both objects be collections to make them consistent? I see what you mean about needing more composable methods... Maybe a generic `findByAllProperties` that works like creating a map; `findByAllProperties('prop1', List.of('val1', 'val2'), 'prop2', List.of('val3', 'val4'), ...)`

ui-admin/src/study/surveys/SurveyEnvironmentTable.tsx
```diff
@@ -0,0 +1,129 @@
...
+    { detailEnv && <SurveyEnvironmentDetailModal
+      stableId={stableId}
+      studyEnvParams={{ ...studyEnvParams, envName: detailEnv }}
+      onDismiss={() => setDetailEnv(undefined)}
+    />}
+  </tr>
+}
+
+/**
+
```
connorlbark: nit: extra comment

connorlbark (commented):
Looks great! I love the changes a lot, already a big improvement. Mostly curious about using tanstack vs raw HTML table

core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
```diff
@@ -85,18 +57,55 @@ public Optional<StudyEnvironmentSurvey> findActiveBySurvey(UUID studyEnvId, UUID
...
+        return studyEnvSurvs;
+    }
+
+    public List<StudyEnvironmentSurvey> findAllWithSurveyNoContent(List<UUID> studyEnvironmentIds, String stableId, Boolean active) {
+        List<StudyEnvironmentSurvey> studyEnvSurveys = findAll(studyEnvironmentIds, stableId, active);
+        attachSurveys(studyEnvSurveys, false);
+        return studyEnvSurveys;
+    }
+
+    protected void attachSurveys(List<StudyEnvironmentSurvey> studyEnvSurveys, boolean withContent) {
```
MatthewBemis: TOL: I've noticed that we often use booleans to modify the behavior of functions. IMO this makes it harder to quickly understand what the behavior is when it's being called: what does `true` or `false` actually mean and how does it modify the behavior? So I end up having to dig into the implementation anyway.

I wonder if we could make these types of parameter lists a bit more readable by either using Enums with descriptive names, or doing away with them entirely in favor of more focused methods with more descriptive names.

Scala has a nice feature that lets you prefix the parameter name in the function call to reduce ambiguity, but it doesn't look like that's made it to Java yet, i.e. `attachSurveys(studyEnvSurveys, withContent: true)` Maybe one day...

Anyway, this isn't a new pattern brought in by this PR so it's absolutely not blocking, but it crossed my mind while reviewing.

MatthewBemis (approved):
Just a few comments, none blocking

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConfiguredSurveyController.java
devonbush: concur, I'm not a fan, but can't think of anything better

ui-admin/src/study/surveys/SurveyEnvironmentTable.tsx
devonbush: that's a good question.  I hadn't anticipated any need for sorting/filtering capability, but it can't hurt

core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
devonbush: there already is a findAllTwoProperties that takes two collections--see immediately below this.  The idea is that you call findAllByTwoProperties with what you have, and then based on whether it's a collection or not, the right method will get invoked to construct the right query

core/src/main/java/bio/terra/pearl/core/dao/study/StudyEnvironmentSurveyDao.java
devonbush: that's a good point -- and yeah, named parameters are awesome.  I like your suggestion of an enum, I'll see how that looks

core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
connorlbark: Oh, I understand. That makes more sense in context.

ui-admin/src/study/surveys/SurveyEnvironmentTable.tsx
devonbush: this was a great suggestion -- moving it to a tanstack table didn't add any functionality, but I think it made the code easier to follow

connorlbark (approved):
Looks awesome! Excited for more changes like this on the study creation/update workflow

PR #715, opened 2024-01-31
[JN-833] Create stubbed address validation client

connorlbark:
#### DESCRIPTION

Creates a stub client for address validation.

I operated under the assumption that we are likely going to use Google's address validation API, as this is what Pepper currently uses. As such, I modeled our Dto object as a very simplified version of [Google's validation response](https://developers.google.com/maps/documentation/address-validation/understand-response), just covering the fields we care about and translating them into our own objects (MailingAddress).

I assumed there were really 3 main possible responses:
- Invalid address
- Improvable address that user can override (i.e. address was valid but could be better)
- Valid address

But perhaps this assumption is wrong - let me know what you think.

- None


devonbush (approved):
Looks good, I like the setup you've got for stubbing improvable and errors.  Doesn't have to be in this PR, but you'll eventually want to add synthetic participants with addresses that trigger each of those modes, similar to the "badAddress" enrollee (`populate/src/main/resources/seed/portals/ourhealth/studies/ourheart/enrollees/badAddress.json`)

core/src/main/java/bio/terra/pearl/core/service/address/AddressValidationClientStub.java
```diff
@@ -0,0 +1,88 @@
...
+                                    .state("MA")
+                                    .country("USA")
+                                    .postalCode("02142")
+                                    .build())
+                    .sessionId(sessionId)
+                    .build();
+        } else if (address.getStreet1().contains(ERROR_INDICATOR)) {
+            int errorCode = 500;
+            try {
+                errorCode = Integer.parseInt(address.getPostalCode());
```
pegahtah, line 71: I know this is a stub, but I don't understand this line, how/why is the error code relevant to the postal code? In case of server error on Juniper side (like the case in the catch) do you want to return 404 or 500?

pegahtah: Oh also something that may be useful, Postal codes in Canada are a six-character string consisting of three letters and three digits. For example, the postal code for the CN Tower in Toronto is M5V 3L9.

connorlbark: It allows you to customize the error returned so you can test different scenarios. E.g.

```
123 ERROR St
Boston, MA 404
```
returns a 404 error

pegahtah: Gotcha

pegahtah (approved):

PR #718, opened 2024-02-01
[JN-829] treating missing body as bad request

devonbush:
#### DESCRIPTION

We're getting missing POST body errors in the logs, presumably from bots hitting our participant log endpoint.  I did some code tracing, and I really can't find any way that would happen in our UI.  This updates the error classes in our default error handlers to include missing POST body (and other malformed http) as BAD_REQUEST.

This also updates us to the latest version of bio.terra.terra-java.  I was investigating whether that update would impact whether POST bodies are required, and it doesn't, but it seems like a good update for hygiene anyway, given that it works as a drop-in-replacement

1. redeploy apiParticipantApp
2. edit ui-participant.  api.tsx  -  delete line 453 so that the log request does not include a body
3. go to `sandbox.ourhealth.localhost:3001`
4. confirm that a bad log request appears in the IntelliJ console, but at log level "INFO"
```
2024-02-01T15:06:57.541-05:00 6dB5lGVo  INFO 70738 --- [nio-8081-exec-7] b.t.p.a.p.c.GlobalExceptionHandler       : Exception: org.springframework.http.converter.HttpMessageNotReadableException: Required request body is missing: public org.springframework.http.ResponseEntity<java.lang.String> bio.terra.pearl.api.participant.controller.LoggingController.log(java.lang.Object)
Request: POST /api/public/log/v1/log 400
```

### DIFF
--- a/api-admin/build.gradle
+++ b/api-admin/build.gradle
@@ -14,7 +14,7 @@ apply from: 'publishing.gradle'
 
 dependencies {
 
-    implementation 'bio.terra:terra-common-lib:0.1.2-SNAPSHOT'
+    implementation 'bio.terra:terra-common-lib:0.1.10-SNAPSHOT'
     implementation project(':core')
     implementation project(':populate')
     implementation 'org.apache.commons:commons-dbcp2'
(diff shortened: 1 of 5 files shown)


MatthewBemis (approved):
Worked for me 👍

connorlbark (approved):

PR #719, opened 2024-02-05
[JN-841] Address validation in Admin UI

connorlbark:
#### DESCRIPTION

Adds a `Validate` button which will hit the stub endpoint. Note that `Validate` is not called when you hit `Next: Justification` - my thought on this was that admins should be able to ignore validation if they wanted to (e.g., if they just wanted to change the participant's name). I also put some of the core logic into `ui-core` so that when we move to the participant side, we only need to worry about the UI flow and not the logic of validation.

I also edited the result DTO to better map with the service we have decided upon.

Per discussion with Erin, validation is 1/2 of the issue - the other half is making it easy to enter the correct address. As such, the next ticket I will be working on adding an autocomplete address stub in the same client (Smarty supports this also).

- Restart admin api
- Rebuild core ui
- Go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHBASC/profile or any other participant
- Go into edit mode on profile
- Try validating - it should pass and all fields go green
- Put `BAD` in the first address line and click Validate to observe how bad addresses are handled
- Keep `BAD` in the address line and delete other fields to observe how missing fields are handled
- Replace the fields and put `IMPROVABLE` in the first address line to observe the behavior of an address which could be improved
- Accept or deny the new address and ensure the correct functionality
- Try also `INFERENCE`


ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
```diff
@@ -332,3 +300,141 @@ export function FormRow(
...
+        </div>
+        <div className='col'>
+          <input
+            className={formatClassName('country')}
+            type="text" value={mailingAddress.country || ''} placeholder={'Country'}
+            onChange={e => onFieldChange('country', e.target.value)}/>
+        </div>
+      </div>
+      <LoadingSpinner isLoading={isLoadingValidation}>
+        <button className="btn btn-link" onClick={validateAddress}>Validate</button>
```
connorlbark: I wasn't sure how to make this button only show up in dev/local. Is there any easy way to do that?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/address/AddressValidationController.java
```diff
@@ -0,0 +1,42 @@
...
+    UUID sessionUUID;
+    if (Objects.nonNull(sessionId) && !sessionId.isEmpty()) {
+      sessionUUID = UUID.fromString(sessionId);
+    } else {
+      sessionUUID = UUID.randomUUID();
+    }
+
+    MailingAddress mailingAddress = objectMapper.convertValue(body, MailingAddress.class);
+
+    return ResponseEntity.ok(client.validate(sessionUUID, mailingAddress));
```
devonbush: Let's split this into two lines.  It's nice to be able to see the return type in the controller, and have an easy post-call place to put a breakpoint.
```suggestion
    AddressValidationResultDto result = client.validate(sessionUUID, mailingAddress)
    return ResponseEntity.ok(result);
```

core/src/main/java/bio/terra/pearl/core/service/address/AddressValidationClientStub.java
```diff
@@ -85,4 +89,56 @@ public AddressValidationResultDto validate(UUID sessionId, MailingAddress addres
 
         return AddressValidationResultDto.builder().valid(true).sessionId(sessionId).build();
     }
+
```
devonbush, line 118: this seems like a really good amount of stub rules to enable testing, without going overboard. nice

core/src/test/java/bio/terra/pearl/core/service/address/AddressValidationClientStubTest.java
```diff
@@ -44,7 +45,47 @@ public void testInvalidAddressResponse() {
...
+        AddressValidationResultDto result = clientStub.validate(
+                sessionId,
+                MailingAddress
+                        .builder().street1("BAD St").state("CO").country("USA").postalCode("12345").build());
+
+        Assertions.assertFalse(result.isValid());
+
+        Assertions.assertEquals(List.of("BAD", "St"), result.getUnresolvedTokens());
+        Assertions.assertEquals(List.of(AddressComponent.HOUSE_NUMBER, AddressComponent.CITY), result.getMissingComponents());
+
```
devonbush, line 51: there are 3 different validate calls in this method -- either split them into separate tests, or add a comment at the top of each to indicate what it's trying to test

ui-admin/src/address/SuggestBetterAddressModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+    onDismiss: () => void
+  }
+) {
+  const changes = findDifferencesBetweenObjects(inputtedAddress, improvedAddress)
+    .filter(diff => !irrelevantFields.includes(diff.fieldName))
+
+  return <Modal show={true}>
+    <Modal.Header>
+      <Modal.Title>
+        Is this your address?
```
devonbush: "Is this the correct address?" might be more appropriate language

ui-admin/src/address/SuggestBetterAddressModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+      </Modal.Title>
+    </Modal.Header>
+    <Modal.Body>
+      <p>
+        We think there might be some improvements which could be made to your address. Would you like to accept them?
```
devonbush: Shorter language: "Accept these improvements to the address [suggested by validation service]?"  (I don't know if we wan tto name which address validation service we use

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
```diff
@@ -332,3 +300,141 @@ export function FormRow(
   </>
 }
 
+
+function EditMailingAddressRow(
```
devonbush, line 302: this is complex enough to be worth putting in its own file

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
devonbush: ah, yeah, this is where we need a feature flag.   We've talked about putting in feature flag infrastructure before, and it wasn't worth it back then given the already-outrageously large amount of configuration.  But this is a good case for it, we might want to think about it

In the meantime, you can just restrict this feature to superusers with `  const { user } = useUser()` and then checking user.superuser.

devonbush (changes requested):
Super cool UX!  A few minor touchups.  The blocking change is to add the address validation integration to the integration dashboard at `https://localhost:3000/integrations`.  So you'll need to add the address validation config to the  ConfigExtService.buildConfigMap() method, and then make an "AddressValidationIntegrationDashboard" component similar to  KitIntegrationDashboard.  it doesn't need to be anything fancy, just a place where you can confirm whether the server is using the stub or real validator

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
connorlbark: Agreed. Thinking about it more, I think eventually this might be made more flexible and put in core so that address validation doesn't have to be remade everywhere in the app.

core/src/main/java/bio/terra/pearl/core/model/address/AddressValidationResultDto.java
```diff
@@ -1,41 +1,36 @@
...
-    // resolve, e.g. "12345678" in "12345678 Main St"
-    private List<String> unresolvedTokens;
+    // Indicates if the suggested address required
+    // inferring components of the address that
+    // were either missing or incorrect
+    private Boolean hasInferredComponents;
 
-    // When validating the same address over and over again,
-    // use the same sessionId
-    private UUID sessionId;
```
connorlbark, line 40: @devonbush fyi after playing around with the Java client for Smarty I decided to go ahead and change our result dto to match as it wouldn't be too much work. Should be similar-ish behavior to what you reviewed before. The only real new concept is `hasInferredComponents`. I'm not using the `vacant` field as of yet, but it could be used as a warning in the future (we could expand our concept of warnings, as Smarty has a lot of not-quite-valid but not-quite-invalid states it can return, vacant just being one of them).

Also, after talking with Erin, we're thinking about having the street1 input be a search which autocompletes the rest of the address. The service we're using supports that so I created a ticket to stub that and add it to the admin ui. After all that prep work (whew) it _should_ be pretty easy to connect up the UI to the real service with the testing code I was playing around with this morning.

devonbush: nice, 100% agree

devonbush (approved):
nice!

core/src/main/java/bio/terra/pearl/core/service/address/AddressValidationClientStub.java
```diff
@@ -1,36 +1,44 @@
...
-     * - Example missing component data will be returned
+     * - Example missing component data will be returned based on simple rules
      * - Real-world responses are not guaranteed have missingComponents or unresolvedTokens
      * <br>
      * State 2: Improvable address
      * - Put the word "IMPROVABLE" in street1, e.g. 415 IMPROVABLE St
      * - Returns valid = true with a suggested address of 415 Main St
      * <br>
+     * State 3: Required inference
+     * - Put the word "INFERENCE" in street1, e.g. 415 INFERENCE St
```
MatthewBemis, line 39: I think this should read:

```suggestion
     * - Put the word "INFERRED" in street1, e.g. 415 INFERRED St
```

ui-admin/src/integration/AddressValidationIntegrationDashboard.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { useState } from 'react'
+import { useLoadingEffect } from 'api/api-utils'
+import Api, { InternalConfig } from 'api/api'
+import LoadingSpinner from 'util/LoadingSpinner'
+
+
+/** shows controls and debug info for testing kit request (Pepper) integrations */
```
MatthewBemis: nit: copy/paste leftover

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
```diff
@@ -12,6 +12,7 @@ import { findDifferencesBetweenObjects } from 'util/objectUtils'
 import { Store } from 'react-notifications-component'
 import { successNotification } from 'util/notifications'
 import { doApiLoad } from 'api/api-utils'
+import EditMailingAddress from '../../../address/EditMailingAddress'
```
MatthewBemis:
```suggestion
import EditMailingAddress from 'address/EditMailingAddress'
```

MatthewBemis (approved):
Super slick, well done. Just had a few nits

PR #721, opened 2024-02-05
JN-790 Enabling manual assignment of surveys

devonbush:
#### DESCRIPTION
This enables study staff to manually assign a survey to participants in an environment.  For now the UX is just a single "assign to everyone" button, but the API supports more sophisticated operations. The idea is that this will serve as the infrastructure for enabling advanced survey triggers in later work.  
image

To enable testing, and also because it will be generally useful, a new property has been added to Survey: `assignToAllNewEnrollees`.  This defaults to true.  If false, the survey will not be assigned by default to new participants.  

LATER WORK:
  UX improvements to the survey table around the 'unpublished badge'
  Adding "autoUpdateToLatestVersion" as configurable survey property

1. redeploy ApiAdminApp
2. repopulate demo
3. go to `https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms`
4. Click on the 'sandbox' version of the Sandbox survey
5. note that no participants are currently assigned to it (the modal should look like the above screenshot)
6. click "Assign..."
7. Confirm the participants were all assigned the sandbox survey
8. enroll. a new participant in demo sandbox.  
9. Confirm they are *not* assigned the sandbox survey

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcher.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcher.java
@@ -78,7 +78,8 @@ public class SurveyTaskDispatcher {
                 taskOpt = Optional.of(buildTask(enrollees.get(i), ppUsers.get(i),
                         studyEnvironmentSurvey, studyEnvironmentSurvey.getSurvey()));
             } else {
-                taskOpt = buildTaskIfApplicable(enrollees.get(i), ppUsers.get(i), enrolleeRuleDatas.get(i),
+                List<ParticipantTask> existingTasks = participantTaskService.findByEnrolleeId(enrollees.get(i).getId());
+                taskOpt = buildTaskIfApplicable(enrollees.get(i), existingTasks, ppUsers.get(i), enrolleeRuleDatas.get(i),
                         studyEnvironmentSurvey, studyEnvironmentSurvey.getSurvey());
             }
             if (taskOpt.isPresent()) {
@@ -101,7 +102,7 @@ public class SurveyTaskDispatcher {
                                                    UUID studyEnvironmentId) {
         if (assignDto.assignAllUnassigned()) {
             return enrolleeService.findUnassignedToTask(studyEnvironmentId,
-                    assignDto.targetStableId(), assignDto.targetAssignedVersion());
+                    assignDto.targetStableId(), null);
         } else {
             return enrolleeService.findAll(assignDto.enrolleeIds());
         }
@@ -121,6 +122,7 @@ public class SurveyTaskDispatcher {
         for (StudyEnvironmentSurvey studyEnvSurvey: studyEnvSurveys) {
             if (studyEnvSurvey.getSurvey().isAssignToAllNewEnrollees()) {
                 Optional<ParticipantTask> taskOpt = buildTaskIfApplicable(enrolleeEvent.getEnrollee(),
+                        enrolleeEvent.getEnrollee().getParticipantTasks(),
                         enrolleeEvent.getPortalParticipantUser(), enrolleeEvent.getEnrolleeRuleData(),
                         studyEnvSurvey, studyEnvSurvey.getSurvey());
                 if (taskOpt.isPresent()) {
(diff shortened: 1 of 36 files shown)


ui-admin/src/study/surveys/SurveyEnvironmentDetailModal.tsx
```diff
@@ -37,6 +40,61 @@ export default function SurveyEnvironmentDetailModal(props: SurveyEnvironmentDet
...
+    }]
+  }, [stableId, participantTasks.length])
+
+  const table = useReactTable({
+    data: configuredSurveys,
+    columns,
+    enableRowSelection: true,
+    getCoreRowModel: getCoreRowModel(),
+    getSortedRowModel: getSortedRowModel(),
+    debugTable: true
```
MatthewBemis:
```suggestion
    getSortedRowModel: getSortedRowModel()
```

nit: remove debug statements

MatthewBemis (approved):

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcher.java
```diff
@@ -33,68 +38,139 @@
...
-            log.info("Task creation: enrollee {}  -- task {}, target {}", enrolleeEvent.getEnrollee().getShortcode(),
-                    task.getTaskType(), task.getTargetStableId());
-            task = participantTaskService.create(task, auditInfo);
-            enrolleeEvent.getEnrollee().getParticipantTasks().add(task);
+        List<StudyEnvironmentSurvey> studyEnvSurveys = studyEnvironmentSurveyService
+                .findAllByStudyEnvIdWithSurvey(enrolleeEvent.getEnrollee().getStudyEnvironmentId());
+
+        for (StudyEnvironmentSurvey studyEnvSurvey: studyEnvSurveys) {
+            if (studyEnvSurvey.getSurvey().isAssignToAllNewEnrollees()) {
+                Optional<ParticipantTask> taskOpt = buildTaskIfApplicable(enrolleeEvent.getEnrollee(),
```
connorlbark, line 124: nit: Not sure how I feel about `buildTaskIfApplicable`, since it's just a simple check, maybe doing

`if (isEligibleForStudy) { buildTask() }`

would be a bit nicer so we can avoid the optional & make it easier to follow

core/src/test/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcherTest.java
```diff
@@ -65,4 +94,51 @@ void testOutreachIsDuplicate() {
...
+        List<ParticipantTask> participantTasks = participantTaskService.findTasksByStudyAndTarget(sandboxBundle.getStudyEnv().getId(), List.of(survey.getStableId()));
+        assertThat(participantTasks, hasSize(1));
+
+        // shouldn't create a duplicate task even if the enrolleeId is provided manually, since eligibility override is false
+        assignDto = new ParticipantTaskAssignDto(TaskType.SURVEY, survey.getStableId(), survey.getVersion(), List.of(sandbox1.enrollee().getId()), false, false );
+        surveyTaskDispatcher.assign(assignDto, sandboxBundle.getStudyEnv().getId(), operator);
+        participantTasks = participantTaskService.findTasksByStudyAndTarget(sandboxBundle.getStudyEnv().getId(), List.of(survey.getStableId()));
+        assertThat(participantTasks, hasSize(1));
+
+        // if overriding eligibility is specified, then a duplicate task should be created
```
connorlbark, line 137: What is the use case for doing this? I see how overriding eligibility could be nice for situations where maybe they wouldn't get a task because of conditions based on responses they made, but it seems weird to intentionally give duplicates

connorlbark (approved):

core/src/test/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcherTest.java
devonbush: Yeah, TBH I could see anywhere from five to zero use cases for this.  The ones that I imagined was their prior task was corrupted or they wanted to be able to take the survey again even though they were techinially within the recurrence window (e.g. the survey is supposed to be taken once a year, but the participant needs to take it after only 11 months have passed).  Super speculative in both ways...

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcher.java
devonbush: oh, now I remember why it was built that way.  the logic to  figure out if it's a duplicate is non-trivial (and will likely become more complex over time) and requires building the task in advance (which is also non-trivial).  So making the isEligible() a separate check that doesn't return a task would be expensive.  I'll keep it as-is for now, but ponder if there's a better structure for the next change

PR #722, opened 2024-02-05
[JN-834] updating Hearthive populate to match prod

devonbush:
#### DESCRIPTION

This has zero code changes.  this just updates the populate files for HeartHive to match production using the results of the extract tool.  As these extracts get bigger and bigger we might want to explore other places to keep them (e.g. a shared folder somewhere) so that they don't pollute our repo too much.  We might also want to investigate a "extract only latest" function that doesn't download all past versions of everything.  But for now this is fine -- it doesn't take too long to populate and it's nice to have a very prod-like study to test against (e.g. where you can see how something behaves with 67 past versions...

TO REVIEW:  do a quick skim to confirm the only updated files are in the populate hearthive directories.

1. run `./scripts/populate_portal.sh demo`
2. confirm it populates, and you can log into the sandbox as gelion@test.com


devonbush:
closing, will redo after Matt's site content change makes it to prod

PR #729, opened 2024-02-09
[JN-867] Improved question editing UI

devonbush:
#### DESCRIPTION

This condenses a lot of the question editing UI to make it friendlier for browsing and less wall-of-texty.  The main goal was putting multiple choices in a tabular format that makes vertical scanning easier.  Erin and I still aren't happy with how many info icons there are, the long term solution will probably be to extract all the info icon content into a zendesk page linked from the top of the deisgner, but this is good enough for now. 

BEFORE:
image

AFTER:
image

1. this has no functional changes, so just go to forms and surveys, pick a survey, and click on some questions and see how it looks.


devonbush:
oops, hearthive files got in this PR by mistake...

MatthewBemis (approved):
Nice, great changes!

connorlbark (approved):
side by side with dev it's a simple but huge improvement!

PR #730, opened 2024-02-09
[JN-856] Live Smarty connection for US addresses

connorlbark:
#### DESCRIPTION

This is 1/2 of the Smarty API connection. They use separate APIs for US address validation and international validation - this is just the US half. If you give it an international address it will not validate properly. That will be tackled in an upcoming ticket.

Currently, I only have a trial API key. You can set up your own free 30 day trial of the API at https://www.smarty.com/ or you can message me to get the API keys I've been using. You just need to put them into your Intellij run config environment variables as `SMARTY_AUTH_ID` and `SMARTY_AUTH_TOKEN`. I assume once we get actual production keys we can put those in the vault like everything else.

1. Acquire (from me or your own account) the API keys and put them in your environment as above
2. Edit admin api `application.yml` line 14: to be `addrValidationClientClass: ${ADDR_VALIDATION_CLIENT_CLASS:SmartyAddressValidationClient}`
3. Restart admin api
4. Rebuild core
6. Go to any participant, e.g. https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHBASC
7. Try entering: 415 Main St
8. Click on validate and ensure flow seems good / has not regressed from stub
9. Try misspelling some aspects of the address
11. Smarty has a bunch of example address for different validation situations: https://www.smarty.com/docs/cloud/us-street-api#footnotes Try them out and see if any have unexpected behavior


devonbush (approved):
Looks good, in this or a future PR, please add the smartyAuthId and smartyAuthToken (redacted using `maskSecret`) to the configMap so we can display them in the Integration dashboard

core/src/main/java/bio/terra/pearl/core/service/address/SmartyClientProvider.java
```diff
@@ -0,0 +1,25 @@
+package bio.terra.pearl.core.service.address;
+
+import com.smartystreets.api.ClientBuilder;
+import com.smartystreets.api.StaticCredentials;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SmartyClientProvider {
```
devonbush: I'm not sure we really need this to be a component?  I wonder if it should be a utility method "buildStaticCredentials()"

connorlbark: I agree with you - I didn't originally code it that way. But doing it this way made it possible to make the tests. I couldn't figure out how to do mocking when I constructed the smarty client inside of the constructor of our address validation client. I think the number of similarly-named clients within this package bothers me, but I haven't seen a clear way to make it less ambiguous.

Maybe we could rename `AddressValidationClient` to `AddressValidationService` so that client in this context JUST means smarty's client. And make a wrapper around Smarty's clients called `SmartyValidationClient`? Then that client has a `validateUS` and a `validateInternational` function. Hmm. I think I like that more and would allow us to get rid of the SmartyClientProvider class because we would have a wrapper we could mock.

MatthewBemis (approved):
Nice, worked great locally!

connorlbark:
@devonbush I was able to get rid of the client provider, but I am still unhappy about the naming convention of the classes / the way they flow - there are so many classes named `client` and I am not sure the best way to structure it or even just rename them to be less ambiguous. What do you think? Do you like the idea of changing it to be named `AddressValidationService`, since its primary job is translating the results of smarty to a DTO, or does that complicate the idea of a service?

devonbush:
Yeah, I think AddressValidationService makes sense, since there's no CRUD-ish AddressValidationService for it to be confused with

devonbush (approved):

devonbush:
approving -- up to you whether you want to tackle the renames in this or a subsequent PR

PR #732, opened 2024-02-09
[JN-845] fixing flaky task test

devonbush:
#### DESCRIPTION

The test was inadvertently using an order-specific assertion but the underlying dao method does not guarantee any order in which the enrollees are returned

1. confirm tests pass for this PR


MatthewBemis (approved):
:shipit:

connorlbark (approved):
passed many times in a row locally & investigation makes sense!

PR #734, opened 2024-02-12
[JN-864] Add international address validation

connorlbark:
#### DESCRIPTION

Right now, there are only 4 supported countries - UK, Mexico, Canada, & Australia. I could add more, but I figured we wanted to do it slowly and these seem like the 'MVP' international countries we want to support. I created another ticket to support as many of the countries that A-T shipped to as we can so we can focus on that separately, as I imagine there will be a LOT of edge cases.

I also used a react-select component to select supported countries or allow you to input your own. Validation will only be possible when you select one that already exists. This simplifies A LOT of the flow - having a freeform field adds complication as they could put in, e.g. "United Kingdom" "UK" "GB" "GBR" "UK (Scotland)" etc. etc. 

I started the work of internationalizing addresses for the participant UI by having the EditMailingAddress component take in a language. This language is used by the browser to display the correct country names. However - there is no internationalization of the labels and placeholder texts. I imagine this will be tackled when we work on the participant profile editing.

Also refactors AddressValidationClient to be AddressValidationService to disambiguate from SmartyClient.

- Obtain trial API key for international address validation from smarty.com (could be the same as old account key if you have one, just need to also enable international address trial)
- Provide as `SMARTY_AUTH_ID` and `SMARTY_AUTH_TOKEN` in run config
- Go back to any participant's profile
- Try entering US address like before, ensure no regressions
- Try entering random address in: Mexico, Great Britain, Canada
- Be sure to validate multiple times to ensure no weird formatting changes (e.g. if you accept the changes, if you hit validate again it shouldn't suggest a different address again, it should just go green).
- Try different situations - namely, addresses with apartments. Here are some examples with apartments:
UK: Flat 9 Wheatstone House 650-654 Chiswick High Road London W4 5BB
Australia: 7 / 4 Spade St Port Douglas QLD 4877
Canada: 107-6807 92 Ave NW Edmonton AB T6B 0S3
Mexico: I could not find an example online, but everything I can find says that it should work the same as US/AUS (apartment on the same line).
- Try PO boxes. Note that PO Boxes don't seem to be a part of the data, so you can put really any number in and it will work (unfortuantely) but it will still validate the locality/postalcode/state information.
UK: PO Box 7169 Poole BH15
Australia: PO Box 105. North Sydney NSW 2060. AUSTRALIA.
Mexico: Again, could not find one :*(
- Test manually setting a different default language at `EditMailingAddress.tsx:56` and verify that the country names update to match

See https://www.loqate.com/en-us/address-verify/international-address-formats/ for some guidance on formats


api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
```diff
@@ -72,6 +77,12 @@ public ResponseEntity<ErrorReport> internalErrorExceptionHandler(Exception ex) {
     return buildErrorReport(ex, HttpStatus.INTERNAL_SERVER_ERROR, request);
   }
 
+  @ExceptionHandler({AddressValidationException.class})
+  public ResponseEntity<ErrorReport> addressValidationExceptionHandler(
+      AddressValidationException ex) {
+    return buildErrorReport(ex, ex.getHttpStatusCode(), request);
+  }
```
connorlbark, line 84: Is this the right way to add error handling for this exception? It primarily throws 502s, which I think but am not sure if it is the correct code to return.

connorlbark:
I just discovered https://stackoverflow.com/questions/29919596/how-do-i-get-a-list-of-countries-in-my-website so I will update this to display country names instead of hardcoding them. This will allow us to be sensitive to preferred language. I'll put it back to a draft until I get it working.

connorlbark:
@devonbush @MatthewBemis After finding and fixing 15 edge cases and 5 different ways this could go wrong... This is now tentatively under review 🤕

core/src/main/java/bio/terra/pearl/core/service/address/SmartyClient.java
```diff
@@ -16,18 +16,47 @@ public class SmartyClient {
...
+            this.internationalClient.send(lookup);
+        } catch (Exception e) {
+            handleException(e);
+            lookup.setResult(null);
+        }
     }
+
+    void handleException(Exception e) throws AddressValidationException {
+        for (String key : INVALID_ADDRESS_EXCEPTIONS) {
+            if (e.getMessage().contains(key)) {
```
devonbush, line 51: wouldn't we want to throw a BAD REQUEST error in these situtations?

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/GlobalExceptionHandler.java
devonbush: if it's throwing that because the upstream service is unavailable then yes, that's correct.  I asked a question below about how we're going to handle BAD_REQUEST errors

core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
```diff
@@ -0,0 +1,217 @@
...
+        this.client = client;
+    }
+
+    // see: https://www.iban.com/country-codes
+    // if a country is given that is not one of these countries, then
+    // validation will always error with an IllegalArgumentException
+    public static final List<String> SUPPORTED_COUNTRIES = List.of(
+            "CA",
+            "GB",
+            "MX"
```
devonbush: this list would only change in response to code changes, right?  e.g. to add support for DE, we'd have to add some edge-case handlign code somewhere to deal with German addresses?   I think that there's no reason/benefit to put this list in the DB or in an environment variable, but want to check

core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
```diff
@@ -0,0 +1,217 @@
...
+    );
+
+    @Override
+    public AddressValidationResultDto validate(MailingAddress address) throws AddressValidationException {
+
+        if (SUPPORTED_COUNTRIES.stream().noneMatch(val -> val.equalsIgnoreCase(address.getCountry()))) {
```
devonbush, line 45: will we eventually need the same type of "alternate spelling" checks here you put in for the U.S.?  Or is the assumption that we'll always populate country with a dropdown if it's international?

core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
```diff
@@ -0,0 +1,217 @@
...
+        lookup.setAddress1(addr.getStreet1());
+        lookup.setAddress2(addr.getStreet2());
+        lookup.setAdministrativeArea(addr.getState());
+        lookup.setLocality(addr.getCity());
+        lookup.setPostalCode(addr.getPostalCode());
+        lookup.setCountry(addr.getCountry());
+
+        return lookup;
+    }
+
```
devonbush, line 72: I like how you've largely mirrored the internal structure of the USSmarty service.  I could see making an abstract "BaseSmartyService" with a bunch of abstract protected methods that the US and International services override.  (I like your decision to make these non-public, which is why we couldn't use an interface.)

ui-admin/src/address/EditMailingAddress.tsx
```diff
@@ -40,6 +51,20 @@ export default function EditMailingAddress(
     onMailingAddressFieldChange(field, value)
   }
 
+  const calcCountryOptionsForLang = () => {
```
devonbush: this is a great function -- let's move it out into a `language-util.tsx` or `i18n-utils` file in the utils directory

ui-admin/src/address/EditMailingAddress.tsx
```diff
@@ -55,25 +80,25 @@ export default function EditMailingAddress(
     }, { setIsLoading: setIsLoadingValidation })
   }
 
-  const formatClassName = (field: keyof MailingAddress) => {
+  const formatClassName = (field: keyof MailingAddress, baseClasses = 'form-control') => {
```
devonbush: Let's rework this so that this function is `isFieldValid(field: keyof MailingAddress)` and just returns a boolean.   Then you can use `classNames` (which we already import) and call that function from the jsx.  See https://www.npmjs.com/package/classnames

devonbush (approved):
Looks good, a few suggestions. The one that I feel strongest about is that we should use `classNames` to separate out the validation logic from the css concatenation

ui-admin/src/address/EditMailingAddress.tsx
connorlbark: Oooo this is very cool. That makes sense to me.

core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
connorlbark: That was my thought. Even if a country doesn't need code changes, it would still need to be reviewed that it works with the code and that we are handling its format correctly. I considered adding them to the database but slightly leaned towards hardcoding them; happy to change it either way though.

Edit: I think the advantage of putting them in the DB would be that, in the future, if we needed more complex formatting, we could achieve that by saving the format strings in the database. the other advantage would be in internationalizing the form, some websites will change the order of address forms based upon the country. Both of those might be overkill for our needs.

core/src/main/java/bio/terra/pearl/core/service/address/SmartyClient.java
connorlbark: My thought is that most invalid states should be in the validation result and only errors which signal that we are interfacing with smarty wrong should be exceptions. These (`"GET request lacked required fields"` and `"GET request lacked a street field"`) happen in situations based on invalid user input. Without this code, if the user misspells a country (which, to be fair, shouldn't happen with the dropdown) it would throw a bad request and just have an error toast which I found confusing.

core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
connorlbark: I went back and forth. If you feel strongly, I am happy to add more alternate spellings in. Smarty itself doesn't do alternate spelling checks, and it made it a lot easier to assume that it is always the 2 letter country code as opposed to allowing any kind of input, but making it more flexible shouldn't be too hard. I don't think there will be many situations where someone types their own country, but maybe if they do, we should warn them?

core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
devonbush: yeah, let's just keep this for now -- it'll be faster to iterate

core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
devonbush: cool, yeah, let's assume the country code is validate client-side for now

core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
connorlbark: I agree and would like to do it that way but Smarty has made the confusing decision to have multiple classes of the same name (`Candidate`, `Lookup`) but they are in different packages. So, the US `Lookup` and US `Candidate` is not compatible with the international one, so we can't share these internal methods :'(

core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
connorlbark: @devonbush The discussion in demo about making the countries that show up at the top / configuring which ones are visible makes me think that maybe we should go ahead and engineer them to be a part of the database. What are your thoughts? We could have all the countries we want to show up and just have a "Validatable" boolean on them to denote if we have done the work for validation. I'm worried about keeping a list of countries, though. We would have to maintain that and it seems tricky.

MatthewBemis (approved):
:shipit:

PR #735, opened 2024-02-12
JN-815 (draft) Populate proxies and governed users for heart demo study

pegahtah:
#### DESCRIPTION

PR for adding synthetic data to the demo study by populating 

Run `./scripts/populate_portal.sh demo` , then in the DB look at participant_user, enrollee and enrollee_relation tables. There should be 3 relationships in the  enrollee_relatio, between username `proxy@test.com` and `proxy@test.com-prox-GOVR`, `proxyFamily2@test.com` and `proxyFamily2@test.com-prox-ABCD` and `proxyFamily2@test.com` and `proxyFamily2@test.com-prox-BCDE`


api-participant/src/main/java/bio/terra/pearl/api/participant/service/AuthUtilService.java
```diff
@@ -35,7 +35,7 @@ public Enrollee authParticipantUserToEnrollee(UUID participantUserId, String enr
     // for now, a user is only allowed to access an enrollee if it's themself.  Later, we'll add
     // proxies
     Optional<Enrollee> enrolleeOpt =
-        enrolleeService.findByEnrolleeId(participantUserId, enrolleeShortcode);
```
pegahtah, line 38: seems like this naming was wrong

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
```diff
@@ -31,14 +29,7 @@ public List<EnrolleeRelation> findByParticipantUserId(UUID participantUserId) {
         return dao.findByEnrolleeIdAndRelationshipType(enrolleeId, relationshipType);
     }
 
-    public List<Enrollee> findGovernedEnrollees(UUID participantUserId) {
-        List<EnrolleeRelation> enrolleeRelations=
-                findByParticipantUserId(participantUserId).stream().filter(enrolleeRelation -> isProxy(enrolleeRelation))
-                        .collect(Collectors.toList());
-        return enrolleeService.findAll(enrolleeRelations.stream().map(EnrolleeRelation::getId).toList());
-    }
-
```
pegahtah, line 40: to avoid circular services, I broke this and moved part of it to EnrolleeService, it also makes more sense since it is supposed to return a list of Enrollees

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
```diff
@@ -51,6 +42,19 @@ public Enrollee newGovernedEnrolleeCreationRecord(Enrollee enrollee, Participant
...
 
         this.create(enrolleeRelation);
-        return  enrollee;
+        return enrollee;
+    }
+
+    @Override
+    @Transactional
+    public void delete(UUID enrolleeRelationId, Set<CascadeProperty> cascades) {
+        dao.delete(enrolleeRelationId);
```
pegahtah, line 51: I feel like there should be some checks or guards here, but I'm not sure what is appropriate.

ui-participant/src/hub/HubPage.tsx
```diff
@@ -18,6 +18,16 @@ export default function HubPage() {
...
+  const [activeEnrolleeId, setActiveEnrolleeId] = useState(enrollees[0]?.id)
+
+  // Function to change the active enrollee
+  const handleTabClick = (enrolleeId: string)  => {
+    setActiveEnrolleeId(enrolleeId)
+  }
+
+  // Find the active enrollee based on activeEnrolleeId
+  const activeEnrollee = enrollees.find(enrollee => enrollee.id === activeEnrolleeId)
+
```
pegahtah, line 31: This should most probably moved to another branch, I was using the data to come up with a design for the dashboard for a proxy. My initial design was to use tabs and to use notification alerts as a signal for a task waiting. I don't have approval from Erin so I will move these changes to another local branch and revert here, but I thought since it is already here I might as well attach a screenshot here too
Screenshot 2024-02-08 at 1 29 59 PM

api-participant/src/main/java/bio/terra/pearl/api/participant/service/CurrentUserService.java
```diff
@@ -82,6 +88,11 @@ public UserWithEnrollees loadFromUser(ParticipantUser user, String portalShortco
     PortalParticipantUser portalUser = portalParticipantUser.get();
     user.getPortalParticipantUsers().add(portalUser);
     List<Enrollee> enrollees = enrolleeService.findByPortalParticipantUser(portalUser);
+    List<EnrolleeRelation> enrolleeRelations =
+        enrolleeRelationService.findByParticipantUserId(user.getId()).stream()
+            .filter(enrolleeRelation -> enrolleeRelationService.isProxy(enrolleeRelation))
+            .collect(Collectors.toList());
+    enrollees.addAll(enrolleeService.findGovernedEnrollees(enrolleeRelations));
```
devonbush, line 95: I'm not sure whether it makes more sense to return governed enrollees as part of the enrollees collection, or whether this should return an object that has enrollees and governedEnrollees as separate collections

api-participant/src/main/java/bio/terra/pearl/api/participant/service/CurrentUserService.java
```diff
@@ -82,6 +88,11 @@ public UserWithEnrollees loadFromUser(ParticipantUser user, String portalShortco
     PortalParticipantUser portalUser = portalParticipantUser.get();
     user.getPortalParticipantUsers().add(portalUser);
     List<Enrollee> enrollees = enrolleeService.findByPortalParticipantUser(portalUser);
+    List<EnrolleeRelation> enrolleeRelations =
+        enrolleeRelationService.findByParticipantUserId(user.getId()).stream()
```
devonbush, line 92: BLOCKER: we only want to find governedEnrollees related to the current portal.  I'd suggest making a new method in EnrolleeRelationService `findByPortalUserAndRelationshipType` that then delegates to a same-named DAO method that pulls out the appropriate EnrolleeRelations.

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
devonbush: that's a great question -- let's ask the team.  I suspect the right answer is to make EnrolleeRelationService extend `DataAuditedService` so that we can allow deletes, but we'll always have a record of the changes.  And then let's talk tomorrow about what withdrawal looks like for proxies

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
```diff
@@ -279,6 +284,11 @@ public String generateShortcode() {
         return shortcode;
     }
 
+    @Transactional
+    public List<Enrollee> findGovernedEnrollees(List<EnrolleeRelation> enrolleeRelations) {
```
devonbush, line 288: since this doesn't care what the relation type is
```suggestion
    public List<Enrollee> findRelatedEnrollees(List<EnrolleeRelation> enrolleeRelations) {
```

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
```diff
@@ -279,6 +284,11 @@ public String generateShortcode() {
...
+        return findAll(enrolleeRelations.stream().map(EnrolleeRelation::getId).toList());
```
devonbush, line 289: BLOCKER: I dont' think this works as written
```suggestion
        return findAll(enrolleeRelations.stream().map(EnrolleeRelation::getEnrolleeId).toList());
```

populate/src/main/java/bio/terra/pearl/populate/service/EnrolleeRelationPopulator.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public EnrolleeRelation createNew(EnrolleeRelationPopDto popDto, StudyPopulateContext context, boolean overwrite) throws IOException {
+        EnrolleeRelation enrolleeRelation = buildEnrolleRerlation(popDto, context);
+        enrolleeRelationService.create(enrolleeRelation);
+        return enrolleeRelation;
+    }
+
+    public EnrolleeRelation buildEnrolleRerlation(EnrolleeRelationPopDto popDto, StudyPopulateContext context){
```
devonbush, line 71: typo in name
```suggestion
    public EnrolleeRelation buildEnrolleRelation(EnrolleeRelationPopDto popDto, StudyPopulateContext context){
```

populate/src/main/java/bio/terra/pearl/populate/dto/participant/EnrolleeRelationPopDto.java
```diff
@@ -0,0 +1,17 @@
...
+import bio.terra.pearl.core.model.participant.RelationshipType;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Getter
+@Setter
+@NoArgsConstructor
+public class EnrolleeRelationPopDto extends EnrolleeRelation {
+    private String linkedEnrolleeUsername;
```
devonbush, line 13: why does this need the linkedEnrolleeUsername?  Shouldn't the enrollee shortcode be sufficient?

populate/src/main/java/bio/terra/pearl/populate/dto/participant/EnrolleeRelationPopDto.java
```diff
@@ -0,0 +1,17 @@
...
+    private String enrolleeShortCode;
+    private String linkedParticipantUsername;
```
devonbush, line 15: the 'linked' is kind of redundant
```suggestion
    private String participantUsername;
```

populate/src/main/java/bio/terra/pearl/populate/service/EnrolleeRelationPopulator.java
```diff
@@ -0,0 +1,91 @@
...
+        ParticipantUser attachedGovernedUser = participantUserService
+                .findOne(popDto.getLinkedEnrolleeUsername(), context.getEnvironmentName()).get();
```
devonbush, line 73:
```suggestion
       Enrollee governedEnrollee = enrolleeService.findOneByShortcode(popDto.getEnrolleeShortcode()).orElseThrow();
        ParticipantUser attachedGovernedUser = participantUserService
                .find(governedEnrollee.getParticipantUserId());
```

populate/src/main/resources/seed/portals/demo/participants/proxyFamily1.json
```diff
@@ -0,0 +1,18 @@
+{
+  "participantUser": {
+    "username": "proxy@test.com",
+    "lastLoginHoursAgo": 2
+  },
+  "profile": {
+    "givenName": "Proxy",
+    "familyName": "FAMLILY",
```
devonbush, line 8:
```suggestion
    "familyName": "Family1",
```

populate/src/main/java/bio/terra/pearl/populate/dto/StudyEnvironmentPopDto.java
```diff
@@ -31,6 +31,8 @@ public class StudyEnvironmentPopDto extends StudyEnvironment {
     private List<String> kitTypeNames = new ArrayList<>();
     @Builder.Default
     private List<String> enrolleeFiles = new ArrayList<>();
+    @Builder.Default
+    private List<String> enrolleeRelationFiles = new ArrayList<>();
```
devonbush, line 35: Having these off the studyEnvPopDto will make these much easier to lose track of than if they were attached to the governedEnrollee.  I would put these on EnrolleePopDto instead.  Then you also won't need to have them be separate files, they can just be specified in the enrollee file. 

Populating will be a little trickier, since you need to make sure all enrollees are populated before populating the relations. so where you currently modified StudyPopulator to iterate over the enrolleeRelationFiles, instead you'll want to invoke a new method on enrolleePopulator

```
for (String enrolleeFile : studyPopEnv.getEnrolleeFiles()) {
            enrolleePopulator.populateRelations(context.newFrom(enrolleeFile), overwrite);
        }
```
and then that method in EnrolleePopulator can re-read the enrollee file, find any relations, and invoke EnrolleeRelationPopulator as needed

ui-participant/src/hub/HubPage.tsx
devonbush: looks reasonable, and yes, please move to another branch :)

devonbush (changes requested):
good stuff, nice work teasing through all the populate logic.  Let's chat tomorrow, looking at this and thinking ahead makes me want to make some tweaks to the overall proxy data model.

populate/src/main/java/bio/terra/pearl/populate/service/EnrolleeRelationPopulator.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    protected Class<EnrolleeRelationPopDto> getDtoClazz() {
+        return EnrolleeRelationPopDto.class;
+    }
+
+    @Override
+    public Optional<EnrolleeRelation> findFromDto(EnrolleeRelationPopDto popDto, StudyPopulateContext context) {
+        EnrolleeRelation enrolleeRelation = buildEnrolleRerlation(popDto, context);
```
MatthewBemis, line 45: I think there's a non-ASCII character here (viewing in IntelliJ makes it more apparent)

core/src/main/java/bio/terra/pearl/core/model/participant/EnrolleeRelation.java
```diff
@@ -18,6 +18,7 @@
 @SuperBuilder
 @NoArgsConstructor
 public class EnrolleeRelation extends BaseEntity {
+    private UUID id;
```
MatthewBemis, line 21: I don't think this field is explicitly needed here, since it's part of the `BaseEntity` class that EnrolleeRelation extends

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
```diff
@@ -51,6 +42,19 @@ public Enrollee newGovernedEnrolleeCreationRecord(Enrollee enrollee, Participant
...
+    }
+
+    public void deleteByEnrolleeId(UUID enrolleeId) {
+        List<EnrolleeRelation> enrolleeRelations = dao.findByEnrolleeId(enrolleeId);
+        for (EnrolleeRelation enrolleeRelation : enrolleeRelations) {
+            dao.delete(enrolleeRelation.getId());
+        }
```
MatthewBemis, line 58: This could be simplified to one query by adding a `deleteByEnrolleeId` method to the `EnrolleeRelationDao`, which would call `deleteByProperty("enrollee_id", enrolleeId)`

MatthewBemis (commented):
Cool! Just a few additional comments, and a big +1 to Devon's comment about putting the enrollee relation populate files on the `EnrolleePopDto`

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
devonbush: True, although once we start attaching stuff to EnrolleeRelations (surveyResponses, etc...) we'll need to do the deletes individually anyway so that we can clear out all those things first

pegahtah: Yes I don't expect delete to stay like this for long. Right now the only real functionality of it is in the populate classes, but eventually if we want to have that delete ability for the users and throughout the system then we need to dig much deeper into what delete vs withdrawal means.

populate/src/main/java/bio/terra/pearl/populate/service/EnrolleeRelationPopulator.java
pegahtah: Wow nice catch! 👀

pegahtah:
New PR #743

PR #739, opened 2024-02-16
[JN-759] Participant Edit Profile from Dashboard

connorlbark:
#### DESCRIPTION

Enables participants to edit their profile from their dashboard.

This required updating a lot of internal cruft. Namely, the login packet now includes `profile` and `portalParticipantUser`, which is now provided when you call `useUser()`. Many thanks to Devon for helping me figure that out.

The mailing address editing component has been move to core so it can be the same across ui-admin and ui-participant - this means that, internally, this new profile editing has the infrastructure to handle address validation, but **_does not currently do any validation_**. I figure we want to make those connections once we have an actual production smarty key.

Finally, this data doesn't show up in the data audit table, but the data is in fact audited. I will fix this and a few other data audit table bugs in my next ticket. This one is big enough as-is 😬 

Screenshot 2024-02-23 at 10 50 49 AM
Screenshot 2024-02-23 at 10 51 04 AM
Screenshot 2024-02-23 at 10 51 08 AM

- Rebuild core
- Restart admin api and participant api
- Populate ourhealth
- Go to ourhealth website
- Make a new user (ensure registration process has no regressions)
- Click on user icon in top right
- Click on Profile
- Observe `Not Provided` on all fields
- Fill out basics
- Go back to Profile
- Observe the data you entered
- Use the pencil button to update all of the fields
- Try hard refreshing to ensure data still shows
- Go to admin api for the new user, observe that the new data you entered looks good
  - Note that this won't be reflected in the data audit table (the records do exist tho, will fix the table in next ticket)


ui-core/package.json
```diff
@@ -9,6 +9,7 @@
     "classnames": "^2.3.2",
     "downshift": "^7.6.0",
     "inputmask": "3.3.10",
+    "react-select": "^5.4.0",
```
devonbush, line 12: If you're going to put this here, you should take it out of the ui-admin package.json as well

ui-participant/src/components/ThemedModal.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { usePortalEnv } from '../providers/PortalProvider'
+import { BrandConfiguration, brandStyles } from '../util/brandUtils'
+
+/**
+ * react-bootstrap creates modals outside the <body> tag, which
+ * means that they are outside the div which provides brand
+ * styling. This modal manually adds theming back to any of the
+ * contents of the modal.
+ */
+export default function ThemedModal(
```
devonbush, line 13: nice!

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/participant/ProfileController.java
```diff
@@ -0,0 +1,74 @@
...
+    this.requestUtilService = requestUtilService;
+    this.profileExtService = profileExtService;
+    this.objectMapper = objectMapper;
+    this.request = request;
+  }
+
+  @Override
+  public ResponseEntity<Object> updateProfile(
+      String portalShortcode, String envName, UUID ppUserId, Object body) {
+    Optional<ParticipantUser> participantUserOpt = requestUtilService.getUserFromRequest(request);
```
devonbush: this should just use requireUser

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/participant/ProfileController.java
```diff
@@ -0,0 +1,74 @@
...
+            ppUserId,
+            profile);
+
+    return ResponseEntity.ok(updatedProfile);
+  }
+
+  @Override
+  public ResponseEntity<Object> findProfile(String portalShortcode, String envName, UUID ppUserId) {
+
+    Optional<ParticipantUser> participantUserOpt = requestUtilService.getUserFromRequest(request);
```
devonbush: same, use requireUser

api-participant/src/main/java/bio/terra/pearl/api/participant/service/AuthUtilService.java
```diff
@@ -42,6 +42,16 @@ public Enrollee authParticipantUserToEnrollee(UUID participantUserId, String enr
     return enrolleeOpt.get();
   }
 
+  public PortalParticipantUser authParticipantUserToPortalParticipantUser(
+      UUID participantUserId, UUID portalParticipantUserId) {
+    return portalParticipantUserService
+        .find(portalParticipantUserId)
+        .filter(
+            portalParticipantUser ->
+                portalParticipantUser.getParticipantUserId().equals(participantUserId))
```
devonbush: BLOCKER: this needs to be broadened to cover proxies too.  Which is going to be a bit involved since proxies are at the enrollee level.  check out lines 100-106 of CurrentUserService for how to get all the proxy relationships for a portalParticipantUser.  Then you'll have to check whether the PortalParticipantUserId is associated with the ParticipantUserId of any of those proxied enrollees.

api-participant/src/main/java/bio/terra/pearl/api/participant/service/ProfileExtService.java
```diff
@@ -0,0 +1,85 @@
...
+            .findOne(portalShortcode, envName)
+            .orElseThrow(NotFoundException::new);
+
+    if (!ppUser.getPortalEnvironmentId().equals(portalEnvironment.getId())) {
+      throw new PermissionDeniedException("User not in portal environment");
+    }
+
+    return profileService
+        .loadWithMailingAddress(ppUser.getProfileId())
+        .orElseThrow(NotFoundException::new);
```
devonbush, line 83: This is really more of an illegal state -- all ppUsers should have a profile. 
 Let's go ahead and make a new Exception class ProfileMissing that extends IllegalStateException.  (see StudyEnvironmentMissing for a similar example)

ui-participant/src/participant/EditParticipantProfileModals.tsx
```diff
@@ -0,0 +1,331 @@
...
+import { javaLocalDateToJsDate, jsDateToJavaLocalDate, MailingAddress } from '@juniper/ui-core'
+import ThemedModal from '../components/ThemedModal'
+import Modal from 'react-bootstrap/Modal'
+import EditAddress from '@juniper/ui-core/build/components/EditAddress'
+
+/**
+ * Set of modals that are automatically picked based upon the `showFieldModal` prop,
+ * which denotes which field the user would like to edit.
+ */
+export default function EditParticipantProfileFieldModals(
```
devonbush: this component now doesn't add any functionality or state -- it's just a switch statement.  So let's turn it into an ordinary function and put it directly in ParticipantProfile.tsx

ui-participant/src/participant/EditParticipantProfileModals.tsx
```diff
@@ -0,0 +1,331 @@
...
+// common props for all profile edit modals
+type EditModalProps = {
+  profile: Profile,
+  dismissModal: () => void,
+  save: (p: Profile) => void
+
+}
+
+// internal profile editing methods
+const useProfileEditMethods = (props: EditModalProps) => {
```
devonbush, line 44: sweet hook -- nice!

devonbush (changes requested):
Looks good, worked well locally when I tested it.  Some suggestions, the only blocker is making sure we support proxies in the auth method.

ui-participant/src/participant/ParticipantProfile.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faPencil } from '@fortawesome/free-solid-svg-icons'
+import { useUser } from 'providers/UserProvider'
+import { dateToDefaultString } from '@juniper/ui-core'
+import Api, { MailingAddress, Profile } from 'api/api'
+import EditParticipantProfileFieldModals from './EditParticipantProfileModals'
+import { isEmpty } from 'lodash'
+
+/**
+ *
```
MatthewBemis: Can you fill out the jsdocs for these new components?

ui-participant/src/participant/ParticipantProfile.tsx
```diff
@@ -0,0 +1,170 @@
...
+  return <>
+    {address.street1 && <ProfileTextRow text={address.street1}/>}
+    {address.street2 && <ProfileTextRow text={address.street2}/>}
+    {(address.city || address.postalCode || address.state) &&
+        <ProfileTextRow text={`${address.city} ${address.state} ${address.postalCode}`.split(/\s+/).join(' ')}/>}
+    {address.country && <ProfileTextRow text={address.country}/>}
+  </>
+}
+
+const Bar = () => {
```
MatthewBemis: Always throws me off when there's a legitimate use of "foo" or "bar" left in code. :) Rename to `HorizontalBar` or something?

ui-core/src/components/EditAddress.tsx
```diff
@@ -0,0 +1,220 @@
...
+/**
+ *
+ */
+export default function EditAddress(
+  {
+    mailingAddress,
+    setMailingAddress,
+    validationResult,
+    showLabels = true,
+    language = 'en' // todo: accept internationalized labels
```
MatthewBemis: Just filed a ticket for this. Can you attach [JN-910](https://broadworkbench.atlassian.net/browse/JN-910) to this comment?

ui-participant/src/participant/ParticipantProfile.tsx
connorlbark: Haha, I did not even consider that, but that does make sense.

ui-core/src/components/EditAddress.tsx
connorlbark: This is very partially completed, I made the navbar `Profile` button internationalized but the form itself isn't.

api-participant/src/main/java/bio/terra/pearl/api/participant/service/AuthUtilService.java
```diff
@@ -42,6 +49,29 @@ public Enrollee authParticipantUserToEnrollee(UUID participantUserId, String enr
...
+    // participant user was not the ppUser, but could be a proxy.
+    // check if proxy exists on any of the enrollees; if so, then
+    // they have access to this ppUser.
+    List<Enrollee> enrolleeList = enrolleeService.findByPortalParticipantUser(ppUser);
+
+    if (enrolleeRelationService.isUserProxyForAnyOf(participantUserId, enrolleeList)) {
+      return ppUser;
+    }
+
+    throw new NotAuthorizedException("Cannot act on behalf of this user.");
```
MatthewBemis: I think this is more accurately a `ForbiddenException` (the credentials are valid, but the user doesn't have permission to access the resource). But I would actually instead return a `NotFoundException("Portal participant user not found.")` here to match the case where the user doesn't exist. As-is, this theoretically allows someone to probe for the existence of a particular user ID. I get that that's pretty unlikely and doesn't reveal any PII, but if someone can't act on behalf of a user then they probably don't need to know that it exists.

MatthewBemis (approved):
Looks great! I just have one suggestion but otherwise 👍

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
```diff
@@ -33,6 +34,11 @@ public List<EnrolleeRelation> findByTargetEnrolleeId(UUID enrolleeId) {
         return dao.findByTargetEnrolleeId(enrolleeId);
     }
 
+    public boolean isUserProxyForAnyOf(UUID participantUserId, List<Enrollee> enrolleeList) {
```
devonbush: Since this method only needs the enrolleeIds, I might change the signature to just take a list of EnrolleeIds instead of full Enrollee objects

devonbush (approved):
nice!

PR #741, opened 2024-02-18
[JN-825] cohort rule evaluation

devonbush:
#### DESCRIPTION

This adds a Cohort rule evaluator.  Still only supporting basic =, !=, && and ||, but this probably covers 90% of the rules people need.  gradle seems to be complaining about an ambiguous execution order, I'll try to fix that in follow-on PRs

1. This doesn't have any functional changes to the application.  So just run the unit tests


core/src/main/java/bio/terra/pearl/core/rule/CohortRuleEvaluator.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public Object evaluateTerm(CohortRuleParser.TermContext ctx) throws RuleEvaluationException, RuleParsingException {
+        if (ctx.BOOLEAN() != null) {
+            return Boolean.parseBoolean(ctx.BOOLEAN().getText());
+        } else if (ctx.STRING() != null) {
+            String rawString = ctx.STRING().getText();
+            // trim off brackets and any whitespace immediately inside the brackets
+            return rawString.substring(1, rawString.length() - 1).trim();
+        } else if (ctx.NUMBER() != null) {
+            return Double.parseDouble(ctx.NUMBER().getText());
```
connorlbark, line 68: is there any possible issue with it parsing all numbers as doubles? i worry about integer comparison / edge cases, but maybe that's unfounded since the tests seem to work OK.

core/src/test/java/bio/terra/pearl/core/rule/CohortRuleEvaluatorTests.java
```diff
@@ -0,0 +1,78 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+public class CohortRuleEvaluatorTests {
```
connorlbark, line 11: i like these tests, very clean

connorlbark (approved):
really cool stuff! excited to see where this goes.

core/src/main/java/bio/terra/pearl/core/rule/CohortRuleEvaluator.java
devonbush: We'd definitely have to worry if we were a bank -- but since we're just dealing with small, low-precision numbers, we should be fine.  FWIW our numbers mostly come from the frontend where everything is a double in Javascript

MatthewBemis (approved):

PR #744, opened 2024-02-20
[JN-875] enrollee rule evaluation

devonbush:
#### DESCRIPTION

This plugs the rule evaluator from the prior PRs into our actual task assignment machinery.  This doesn't yet yield any functional changes -- the default rule still applies everywhere, and we don't allow customers to specify rules yet.  But now we're actually capable of applying logic to properties of the enrollee.  Upcoming PRs will tackle:

1. figuring out the best time to check for rules that apply to assign additional tasks.  Right now, you could create a survey with a rule `profile.sexAtBirth = 'M'` but it would never be assigned to anyone, since profile.sexAtBirth is always null on EnrolleeCreation, and that's the only time we currently check for tasks to assign
2. figuring out how to support rules based on answers.  e.g. `{ enrollee.surveyResponses.baseline.cardiomyopathy } = true`
3. translating rules to SearchFacets so that cohorts can be identified in bulk

1. no functional changes, so just run unit tests.


core/src/main/java/bio/terra/pearl/core/service/rule/EnrolleeRuleEvaluator.java
```diff
@@ -70,19 +94,22 @@ public Object evaluateTerm(CohortRuleParser.TermContext ctx) throws RuleEvaluati
             String rawString = ctx.VARIABLE().getText();
             // trim off brackets and any whitespace immediately inside the quotes
             String variableName = rawString.substring(1, rawString.length() - 1).trim();
-            if (variableMap.containsKey(variableName)) {
-                Object value = variableMap.get(variableName);
+            try {
+                Object value = PropertyUtils.getNestedProperty(ruleData, variableName);
```
connorlbark, line 98: Nice! This is really cool stuff.

core/src/main/java/bio/terra/pearl/core/service/rule/EnrolleeRuleEvaluator.java
```diff
@@ -1,36 +1,60 @@
...
         CommonTokenStream tokens = new CommonTokenStream(lexer);
         CohortRuleParser parser = new CohortRuleParser(tokens);
         CohortRuleParser.ExprContext exp = parser.expr();
         return evaluateExpression(exp);
     }
 
+    // the default rule is that the action is applied if the enrollee is a subject
+    protected boolean applyDefaultRule() {
+        return ruleData.getEnrollee().isSubject();
+    }
```
connorlbark, line 56: Is `isSubject()` something we always want to be a part of the rule unless they specify otherwise? Or maybe we just need to be careful to add documentation around the fact that proxies are included by default which is something they might not want.

connorlbark (approved):
looks great, I am excited to see where this goes

core/src/main/java/bio/terra/pearl/core/service/rule/EnrolleeRuleEvaluator.java
devonbush: that's a really good question -- essentially, is this a default rule, or is it also a default condition that gets added to every rule unless the rule explicitly mentions subject?  will put some thought into that -- we don't want A-T to add a rule "sexAtBirth = 'M'" to a survey thinking that will limit it to male subjects, and then suddenly the survey gets assigned to proxies too.

MatthewBemis (approved):

PR #745, opened 2024-02-21
[JN-892] Support top 10 countries

connorlbark:
#### DESCRIPTION

Adds more address validation support. It was very hard to find real address examples online and I have only a limited number of API calls before I run out :'(. I tried my best to try one 'complex' address per country but it's really hard to find real addresses that are more than just a house number (e.g., apartments, etc.). I do think there are gonna be edge cases users are gonna find, so we'll have to make sure to allow them to override any errors or suggestions they get. 

- edit admin ui `application.yml:14` to be `addrValidationServiceClass: ${ADDR_VALIDATION_CLIENT_CLASS:SmartyAddressValidationService}`
- enter your api keys into your intellij runtime environment vars as `SMARTY_AUTH_ID` and `SMARTY_AUTH_TOKEN`
- Restart admin api / ui
- Go to any enrollee's profile
- Ensure all of the countries we should support with this make the validate button go blue
- Try a variety of random addresses you can find on google maps (I think at the end of the day the best testing for this is gonna be real users, unfortunately)

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
@@ -36,7 +36,17 @@ public class SmartyInternationalAddressValidationService implements AddressValid
             "CA",
             "GB",
             "MX",
-            "AU"
+            "AU",
+            "TR",
+            "ES",
+            "PL",
+            "DE",
+            "FR", // todo
+            "IT", // todo
+            "CZ",
+            "BR",
+            "SE",
+            "CH" // todo
     );
 
     @Override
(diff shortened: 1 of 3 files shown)


core/src/main/java/bio/terra/pearl/core/service/address/SmartyInternationalAddressValidationService.java
```diff
@@ -115,8 +128,11 @@ private boolean isValid(Candidate candidate) {
 
     private MailingAddress suggestedAddress(MailingAddress input, Candidate candidate) {
 
-        if (input.getCountry().equals("GB")) {
+        // many countries put the apartment on its own line
+        if (List.of("GB", "FR", "IT", "CH").contains(input.getCountry())) {
```
devonbush: let's pull this list into a constant

devonbush (approved):
one small tweak, otherwise looks good

MatthewBemis (approved):

PR #746, opened 2024-02-22
[JN-875] enabling search by answer values

devonbush:
#### DESCRIPTION
This enables a search facet that matches based on answer values.  Yet another piece of infrastructure we'll need for cohort building.  This PR has no application-visible changes -- this search facet is not exposed in the UI.  

Following PRs will include:
1. Enabling the rule evaluator to produce search facets.  In essence, if a rule isn't capable of being evaluated based on just the simple enrollee/profile information we already had loaded, it will get converted to SearchFacets, and then run as a "search" with an extra facet of the enrollee id.  And then if the search returns a row, the result will be true. 
2. Making search facets more robust -- right now they just support ANDed facets and equality operators.  We'll need ORs and inequality operators too.  (this is the big advantage of this overall approach -- by consolidating search and Rule evaluation, we can make both better at the same time)

1. no functional changes, so just run some searches in the participant list and confirm nothing got broken.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/participant/search/facets/AnswerFacetValue.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/participant/search/facets/AnswerFacetValue.java
@@ -1,0 +1,67 @@
+package bio.terra.pearl.core.service.participant.search.facets;
+
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+
+/**
+ * Much of this class mimics some of the type processing found in Answer.java.  However, we don't support querying
+ * objectValue stuff yet.
+ */
+@Getter
+@Setter
+@NoArgsConstructor
+public class AnswerFacetValue implements FacetValue {
+  private String stringValue;
+  private Boolean booleanValue;
+  private Double numberValue;
+  private String questionStableId;
+  private String surveyStableId;
+
+  public AnswerFacetValue(String surveyStableId, String questionStableId, String value) {
+        this.stringValue = value;
+        this.questionStableId = questionStableId;
+      this.surveyStableId = surveyStableId;
+  }
+
+    public AnswerFacetValue(String surveyStableId, String questionStableId, Boolean value) {
+        this.booleanValue = value;
+        this.questionStableId = questionStableId;
+        this.surveyStableId = surveyStableId;
+    }
+
+    public AnswerFacetValue(String surveyStableId, String questionStableId, Double value) {
+        this.numberValue = value;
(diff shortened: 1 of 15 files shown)


core/src/main/java/bio/terra/pearl/core/service/participant/search/facets/AnswerFacetValue.java
```diff
@@ -0,0 +1,67 @@
+package bio.terra.pearl.core.service.participant.search.facets;
+
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+
+/**
+ * Much of this class mimics some of the type processing found in Answer.java.  However, we don't support querying
+ * objectValue stuff yet.
```
MatthewBemis, line 10: Is this going to nudge us to convert the `objectValue` column to `jsonb`?

MatthewBemis (approved):
I really like how this approach for building out rule evaluation concentrates effort on facets.

core/src/test/java/bio/terra/pearl/core/dao/participant/EnrolleeSearchDaoTests.java
```diff
@@ -235,4 +236,100 @@ public void testTaskSearch(TestInfo info) throws Exception {
...
+    // check that a search for a nonexistent question returns no results
+    facet = new SqlSearchableFacet(
+            new AnswerFacetValue(survey.getStableId(), "question3", false), new AnswerFacetSqlGenerator());
+    result = enrolleeSearchDao.search(studyEnv.getId(), List.of(facet));
+    assertThat(result.stream().map(EnrolleeSearchResult::getEnrollee).toList(), hasSize(0));
+  }
+
+  @Autowired
+  private AnswerService answerService;
 }
```
connorlbark, line 335: nit: move to top?

connorlbark (approved):
I don't fully understand all the code here, but it looks good from what I understand!

core/src/main/java/bio/terra/pearl/core/service/participant/search/facets/AnswerFacetValue.java
devonbush: yeah, probably

core/src/test/java/bio/terra/pearl/core/dao/participant/EnrolleeSearchDaoTests.java
devonbush: I'm going to do that in the upcoming PR I'm working on now -- I've got more tests added, so that way it won't merge conflict

PR #747, opened 2024-02-22
[JN-869] Internationalize application texts

MatthewBemis:
#### DESCRIPTION

Adds a core language texts table to store application-related texts. These are not editable by users for the time being, and the thought is that we'll own translating these as new languages come up in studies. Eventually, we could allow study staff to overwrite these in their custom portal language texts (which will contain survey names in an upcoming PR).

Note that this does not internationalize every single application text. The primary purpose of this PR is to get the infra in place with some illustrative examples of how it will be used. The rest of the texts will be updated as part of JN-872.

1. Repopulate demo`./scripts/populate_portal.sh demo`
2. Change the site language and verify that the navbar buttons are translated https://sandbox.demo.localhost:3001/ 
3. Check the dashboard and make sure the substitutions are working correctly
4. Repopulate ourhealth`./scripts/populate_portal.sh ourhealth` 
5. Verify that the navbar buttons and dashboard still displays in English


core/src/main/java/bio/terra/pearl/core/model/i18n/CoreLanguageText.java
```diff
@@ -0,0 +1,17 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class CoreLanguageText extends BaseEntity {
+    private String messageKey;
```
connorlbark: Nit: I find `messageKey` slightly confusing, because it makes me think the text is for translations of an email or some kind of chat message.

ui-participant/src/Navbar.tsx
```diff
@@ -21,14 +22,14 @@ type NavbarProps = JSX.IntrinsicElements['nav']
 
 /** renders the navbar for participants */
 export default function Navbar(props: NavbarProps) {
-  const portalEnv = usePortalEnv()
-  const { localContent } = portalEnv
+  const { portal, portalEnv, reloadPortal, localContent } = usePortalEnv()
+  const { i18n } = useI18n()
```
connorlbark, line 26: This is so nice!!

ui-participant/src/providers/I18nProvider.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { createContext, useContext, useEffect, useState } from 'react'
+import { LanguageText } from '@juniper/ui-core'
+import Api from 'api/api'
+import { useUser } from './UserProvider'
+
+const I18nContext = createContext<I18nContextT | null>(null)
+
+export type I18nContextT = {
+  languageTexts: LanguageText[]
```
connorlbark: nit: transform this to a map? I could see the `i18n` function being called many times in one page and making it a map would make it a constant-time operation.

I'm also _slightly_ worried about the fact that the languageTexts could get very big - we could maybe have sandboxed `i18n` providers, like e.g. `<I18nProvider area='homepage'>` which will grab just the translations for the homepage, etc. for each area of the application. But that is likely total overengineering and we don't need to do that. Just a thought - I think it's in a great place right now.

ui-participant/src/providers/I18nProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  const i18n = (key: string) => {
+    return languageTexts.find(text => text.messageKey === key)?.text || `{${key}}`
+  }
+
+  return <>
+    {isLoading && <div className="bg-white h-100 w-100">
+      <div className="position-absolute top-50 start-50 translate-middle">Loading...</div>
+    </div>}
```
connorlbark, line 59: This popup is kinda jarring when switching between languages. Maybe we keep the old language up until the new translation has loaded, so it doesn't make the whole screen go white.

connorlbark (approved):
This is awesome! Works great locally and I think it's designed really well. I just have non-blocking thoughts and comments.

ui-participant/src/providers/I18nProvider.tsx
MatthewBemis: I totally agree, I initially shared that same worry and started down that path in the original implementation, but yeah it ended up feeling like overkill for where we're at right now. It's definitely something to keep an eye on though as app functionality expands over time

core/src/main/java/bio/terra/pearl/core/model/i18n/CoreLanguageText.java
MatthewBemis: Good point, I'll rename!

core/src/main/resources/db/changelog/changesets/2024_02_16_language_texts.yaml
```diff
@@ -0,0 +1,18 @@
+databaseChangeLog:
+  - changeSet:
+      id: add_core_language_text
+      author: mbemis
+      changes:
+        - createTable:
+            tableName: core_language_text
```
devonbush: why do we call this "core_language_text" and not just "language_text"?  I'd just as soon keep the name short since we'll be using it in a lot of places.  (I could also see `i18n_text` or `lang_text` or `locale_text` as possible names)

core/src/main/resources/db/changelog/changesets/2024_02_16_language_texts.yaml
```diff
@@ -0,0 +1,18 @@
...
+            columns:
+              - column: { name: id, type: uuid, defaultValueComputed: gen_random_uuid(), constraints: { nullable: false, primaryKey: true } }
+              - column: { name: created_at, type: datetime, constraints: { nullable: false } }
+              - column: { name: last_updated_at, type: datetime, constraints: { nullable: false } }
+              - column: { name: i18n_key, type: text, constraints: { nullable: false } }
```
devonbush: just 'key' used to be a SQL reserved word (and also a special react property, but maybe `key_name` would be easier on the fingers than i18n_key?

api-admin/src/main/java/bio/terra/pearl/api/admin/config/InitialPopulate.java
```diff
@@ -23,5 +25,11 @@ public void populateSeedIfNeeded() throws IOException {
     } else {
       log.info("Existing admin users found, skipping seed populate");
     }
+    if (coreLanguageTextService.count() == 0) {
```
devonbush: until we have a language text editing UX and really lean into the DB being the source-of-truth, we might just want to always repopulate the core language texts so that we can do text updates/additions without writing a migration script every time

ui-participant/src/Navbar.tsx
```diff
@@ -96,7 +97,7 @@ export default function Navbar(props: NavbarProps) {
                   )}
                   to="/hub"
                 >
-                    Log In
+                  {i18n('navbarLogin')}
```
devonbush, line 100: yeah, this is really smooth developer UX -- nice

populate/src/main/resources/seed/i18n/dev/languageTexts.json
```diff
@@ -0,0 +1,14 @@
+[
```
devonbush, line 1: I like the amount of scoping you did on the keys for these.  they communicate where they are/what they do, while still being concise and readable

ui-participant/src/hub/HubPage.test.tsx
```diff
@@ -57,6 +57,14 @@ jest.mock('../providers/UserProvider', () => {
   }
 })
 
+jest.mock('providers/I18nProvider', () => {
```
devonbush, line 60: let's make a test-util for this so we don't have to repeat it everywhere.  Something like `useMockI18n(mockTexts)` .  And this pattern might be so common we might just want to make a "renderWithI18n" helper that takes the component and an (optional) list of message keys, similar to how we have renderWithRouter helpers.

ui-participant/src/providers/I18nProvider.tsx
devonbush: there's something deeply meta about having to have a hardcoded "loading" text on the i18n provider :).    I think you should be able to use the `PageLoadingIndicator` component to do a full screen loading.  Alternatively, I could see just replacing the language selection control text with a loading spinner while the language is loading

core/src/main/resources/db/changelog/changesets/2024_02_16_language_texts.yaml
```diff
@@ -0,0 +1,18 @@
...
+              - column: { name: text, type: text, constraints: { nullable: false } }
+              - column: { name: language, type: text, constraints: { nullable: false } }
```
devonbush, line 14: by far our most common case is going to be pulling all the messages out for a given language. I think it would be worth 5mins seeing if there's a postgresql index type we can put on the `language` column that would make that retrieval faster.  I think a regular index won't help, since IIRC postgresql ignores indexes if it suspects the query will return more than ~1% of the total rows.  But there might be some sort of partition/sorting index that will force postgres to store all the messages in a given language together (at the cost of higher insert time, but we don't care about that)

ui-participant/src/providers/I18nProvider.tsx
devonbush: agree we don't need section-specific modules yet.  What might be easy to do though is to have the language texts be sent from the server to the client already as a map.  check out the `json_object_agg` function of postgresql -- you might be able to make a DAO method that just returns the texts already in a json hash by key.  Alternatively, you could just put them in a hash in Java, and then use Spring caching to cache the hash

devonbush (changes requested):
very cool, works great when testing locally.  A bunch of suggestions, some of which might be best deferred to later PRs.  The ones that I want to talk about before this gets merged are the column and table names

ui-participant/src/providers/I18nProvider.tsx
MatthewBemis: Agreed that it's a bit hilarious to have the hardcoded text here, of all places.

The issue here is that switching languages doesn't just cause the language texts to reload, it also causes the portal to reload because we need to fetch the site content in the new language. So the main source of the white "Loading..." flash is actually from the PortalProvider. This provider mirrors the portal loading screen which should prevent an _additional_ flash from occurring.

The flash irritated me too but then I remembered that users probably aren't going to be rapidly cycling through languages like I was while testing. I do think we can and should improve loading indicators, but I'd like to keep it outside of the scope of this PR and file a ticket for it, if that's okay.

core/src/main/resources/db/changelog/changesets/2024_02_16_language_texts.yaml
MatthewBemis: I'd buy something like `i18n_text`. The reason for prefixing with `core` was to indicate that these are system/application-owned texts and they aren't owned by any particular portal environment. Admittedly, "core" might not be the best name.

I waffled back and forth on if these should live in the same table as the eventual portal-specific texts for things like survey names. I could easily imagine it either way. The separation makes the data a bit straightforward to work with and interpret IMO, but throwing everything in one table and adding a nullable portal_env_id on there wouldn't be awful.

core/src/main/resources/db/changelog/changesets/2024_02_16_language_texts.yaml
MatthewBemis: Yeah, the cardinality of `language` kind of makes a lot of the index options worthless. I looked a bit to see if there was a way to get Postgres to cluster them together and didn't find anything that looked appealing at the moment. But if we hit a performance wall in the future, we could maybe use the `CLUSTER` command on this table every time the application-related texts repopulate on boot to keep entries for each language near each other. I don't know if that's crazy, my DB skills are rusty. But there's I imagine there's something we can do with more research.

Hopefully the Spring-level caching alleviates some of the performance concerns in the meantime.

devonbush (approved):
seems like we're on the same page, thanks for the extra discussion

PR #750, opened 2024-02-23
[JN-909] Show portal participant user records in data audit table

connorlbark:
#### DESCRIPTION

Shows portal participant user records in data audit table.

Also, groups together related changes into one row on the table. This fixes the weird rendering of the participant task records. This table could probably be significantly improved, but it's a start.
image

Right now, I do not believe there is a way to create portal participant user records in the UI. Therefore, ensure that the data change record table has not regressed.


core/src/main/java/bio/terra/pearl/core/service/workflow/DataChangeRecordService.java
```diff
@@ -2,23 +2,42 @@
...
+        this.portalParticipantUserService = portalParticipantUserService;
+    }
+
+    // There can be records which exist for an enrollee which, for one reason or another,
+    // are tied only to the PortalParticipantUser. This grabs all of them
+    // to ensure that we are missing nothing for a given enrollee.
+    public List<DataChangeRecord> findAllRecordsForEnrollee(Enrollee enrollee) {
+        PortalParticipantUser ppUser = portalParticipantUserService
+                .findByProfileId(enrollee.getProfileId())
+                .orElseThrow(() -> new NotFoundException("Invalid enrollee"));
```
devonbush: This is an IllegalStateException.  Also, this "lookup up ppUser by Enrollee" is common enough -- let's make an explicit "findForEnrollee(Enrollee enrollee)" method in PortalParticipantUserService, that then delegates to the the findByPRofileId() method.  That will make the code a bit easier to read.

devonbush (approved):
looks good -- minor exception-handling suggestions

pegahtah (approved):
LGTM

PR #751, opened 2024-02-23
[JN-843] Rename site image and add delete button to media page

connorlbark:
#### DESCRIPTION

Adds a delete endpoint to delete site media. Also, renames site image to site media.

- Repopulate ourhealth to ensure populate works
- Ensure no regressions from name change (compiles/runs locally, participant side works, CRUD operations on media still work)
- Extract and populate as zip to ensure extract functionality works as expected
- Go to `https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/media`
- Try out the delete button


devonbush:
Thanks for tackling this!  you'll want to rename the three indexes on the table too 
```
"site_image_pkey" PRIMARY KEY, btree (id)
    "site_image_clean_file_name_version" UNIQUE CONSTRAINT, btree (clean_file_name, version, portal_shortcode)
Foreign-key constraints:
    "fk_site_image_portal_shortcode" FOREIGN KEY (portal_shortcode) REFERENCES portal(shortcode)
```

ui-admin/src/portal/media/SiteMediaDeleteModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+      Store.addNotification(successNotification('file successfully deleted'))
+      onSubmit()
+    }, { setIsLoading })
+  }
+
+  return <Modal show={true} onHide={onDismiss}>
+    <Modal.Header closeButton>
+      <Modal.Title>Delete Media</Modal.Title>
+    </Modal.Header>
+    <Modal.Body>
```
devonbush, line 35: BLOCKER: This is an irreversible delete and we don't have assurances that the image isn't being used on the 'live' site.  I think until we develop that check (which won't be trivial, since it will involve parsing a lot of SiteContent objects and so is way outside the scope of this ticket), we should restrict deletes to superusers only.  For non-superusers, this dialog can just tell them to email support and include [[url of medial to delete]]"

Alternatively, I could see allowing non-superusers to do this, but giving more explanatory text about "don't delete images that have been published, you don't need to delete images to update them, etc..."   Might be worth pinging Erin to see what she thinks we can trust users with

ui-admin/src/portal/media/SiteMediaDeleteModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <Modal.Footer>
+      <LoadingSpinner isLoading={isLoading}>
+        <button className='btn btn-secondary' onClick={onDismiss}>Cancel</button>
+        <Button variant="danger" onClick={deleteImage}>Delete</Button>
+      </LoadingSpinner>
+    </Modal.Footer>
+  </Modal>
+}
+
+/**
```
devonbush: this is duplicated code from SiteMediaUploadModal -- you don't need it here

devonbush (approved):
Looks good, and worked locally -- appreciate the attention to detail in all the places this had to be updated.  One blocker about restricting this to superusers since it's a potentially prod-destructive operation without guardrails, and one copy-paste cleanup, and this will be good to go!

ui-admin/src/portal/media/SiteMediaDeleteModal.tsx
connorlbark: oops, good catch

pegahtah (approved):

PR #754, opened 2024-02-26
[JN-877] Check for browser compatibility when logging JS errors

MatthewBemis:
#### DESCRIPTION

Co-authored by @connorlbark. This adds a utility function that lets us test for browser compatibility 1) when the site first loads, and 2) when we log errors so we can log js compatibility errors with the correct log level

The easiest way to test this is to modify `isBrowserCompatible` to always throw an error, and then call `logError` from inside LandingPage.tsx. If you inspect your network calls, you should see that the error was logged as `INFO` with event name `js-compatibility`


devonbush (approved):
do we want to log an INFO with the browser type when we get an incompatible browser?

connorlbark (approved):
:shipit:

PR #755, opened 2024-02-27
[JN-872] Move the rest of application texts into languageTexts

MatthewBemis:
#### DESCRIPTION

Most texts were done as part of https://github.com/broadinstitute/juniper/pull/747

This also makes use of SurveyJS's built-in i18n so we can have various SurveyJS UI elements translated without having to internationalize them ourselves.

Restart admin API to repopulate the database with the new texts
In the participant UI, confirm that you see the updated texts
Change the language and also verify that you can see the translations for various SurveyJS UI elements (note that Next button or the page number indicators at the bottom of the survey view)

### DIFF
--- a/ui-core/src/participant/landing/sections/FrequentlyAskedQuestionsTemplate.tsx
+++ b/ui-core/src/participant/landing/sections/FrequentlyAskedQuestionsTemplate.tsx
@@ -35,6 +35,8 @@ type FrequentlyAskedQuestionsConfig = {
   questions: FaqQuestion[], // the questions
   showToggleAllButton?: boolean, // whether or not the show the expand/collapse button
   title?: string, // large heading text
+  collapseAllText?: string,
+  expandAllText?: string
 }
 
 const validateFaqQuestion = (questionConfig: unknown): FaqQuestion => {
@@ -51,6 +53,8 @@ const validateFrequentlyAskedQuestionsConfig = (config: SectionConfig): Frequent
   const title = requireOptionalString(config, 'title', message)
   const blurb = requireOptionalString(config, 'blurb', message)
   const showToggleAllButton = requireOptionalBoolean(config, 'showToggleAllButton', message)
+  const collapseAllText = requireOptionalString(config, 'collapseAllText', message)
+  const expandAllText = requireOptionalString(config, 'expandAllText', message)
 
   const questions = config.questions
   if (!Array.isArray(questions)) {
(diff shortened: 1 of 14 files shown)


ui-core/src/participant/landing/sections/FrequentlyAskedQuestionsTemplate.tsx
```diff
@@ -165,7 +173,7 @@ function FrequentlyAskedQuestionsTemplate(props: FrequentlyAskedQuestionsProps)
             <span className="d-inline-block me-2 text-center" style={{ width: 20 }}>
               <FontAwesomeIcon icon={allQuestionsAreExpanded ? faXmark : faPlus} />
             </span>
-            {allQuestionsAreExpanded ? 'Collapse' : 'Expand'} all
+            {allQuestionsAreExpanded ? collapseAllText : expandAllText}
```
MatthewBemis, line 176: These were the only 2 bits of text that users couldn't already override in the siteContent JSON

devonbush (approved):
awesome, pleased at how this is coming together

ui-core/src/participant/landing/sections/FrequentlyAskedQuestionsTemplate.tsx
connorlbark: It still said "Expand All" on the FAQ page for all languages for me - is that expected?

ui-participant/src/hub/StudyResearchTasks.tsx
```diff
@@ -61,9 +61,9 @@ export default function StudyResearchTasks(props: StudyResearchTasksProps) {
             className="btn rounded-pill ps-4 pe-4 fw-bold btn-primary"
           >
             {enrolleeHasStartedTaskType(enrollee, nextTask.taskType)
-              ? 'Continue'
-              : 'Start'}
-            {' '}{taskTypeDisplayMap[nextTask.taskType]}
+              ? i18n('taskContinue')
+              : i18n('taskStart')}
+            {' '}{i18n(taskTypeMap[nextTask.taskType])}
```
connorlbark, line 66: Probably not, but I am slightly worried that some languages might flip this around.

connorlbark (approved):

ui-core/src/participant/landing/sections/FrequentlyAskedQuestionsTemplate.tsx
MatthewBemis: These are overridable in the site content editor rather than being a part of LanguageTexts, so they'll still say Expand All by default

ui-participant/src/hub/StudyResearchTasks.tsx
MatthewBemis: Yeah, some of these may need to become full phrases instead of appending bits of strings. Pluralization is another area that will need to be treated carefully

PR #758, opened 2024-02-27
JN-843 fixing bean initialization condition

devonbush:
#### DESCRIPTION
Apparently adding two endpoints pointing to the same method can cause initialization problems depending on order of setup.  (note https://sandbox.ourhealth.ddp-dev.envs.broadinstitute.org/ is broken right now).  This fixes that by routing the legacy get to a separate method that then delegates to the real method

1.  redeploy apiParticipantApp
2. confirm images still load for participants


MatthewBemis (approved):
👍

connorlbark (approved):

PR #766, opened 2024-02-28
[JN-885] SurveyJS can validate addresses

connorlbark:
#### DESCRIPTION

New automagic `addressvalidation` surveyjs question type. Survey will validate iff you provide an addressvalidation question (which will render as nothing). You need to provide the names of the questions which populate the fields of the mailing address. When validation occurs (by default when "Next" is hit), these will go to the backend, and the custom question essentially acts as the place that the validator can put the suggested address modal into the DOM. If the user gets a suggested address, then they can either deny or accept. This will flip a flag in the question's state that will make the validator skip it, as the user should be able to progress after being suggested an address and saying yes or no.

If a question with the new addressvalidation type is not added, then the validator will do nothing. So, it will be backwards compatible, and can be added to your surveys in the regular sandbox, irb, live fashion. Also, since it takes in the fields, you can use it as many times as you want in one survey. Although, there should only be one address per page, as having more than one on one page would open the possibility of multiple modals opening at once. If needed maybe we could look at the desired UX for this.

- Rebuild core
- (optional) Complete steps to use real smarty API
    - Add smarty keys to Participant API environment vars (like in Admin API, just copy over)
    - Change `application.yml` for both admin and participant api:
         - `ADDR_VALIDATION_CLIENT_CLASS:AddressValidationServiceStub` -> `ADDR_VALIDATION_CLIENT_CLASS:SmartyAddressValidationService` 
- Restart participant and admin API
- Repopulate ourhealth (new validation question added)
- Create new ourhealth participant
- Fill out first page of basics form
- On the address validation page, try out different scenarios
    - 1) Enter valid address and make sure it goes to next page
    - 2) Enter misspelled address and accept the changes and make sure it goes to next page. Also ensure that the values update properly on the admin side.
    - 3) Enter misspelled address and deny the changes, make sure it does not stop you
    - 4) Enter invalid address and ensure it flags the right fields
 - Go to admin view of the ourhealth basics survey
 - Ensure that address validation works there as well when you disable `Ignore Validation`


ui-core/src/surveyjs/address-validation-modal-question.tsx
```diff
@@ -0,0 +1,180 @@
...
+    }
+  }
+
+  renderElement() {
+    return (
+      <>
+        {(this.value.addressValidationResult?.suggestedAddress
+            && !this.value.canceledSuggestedAddress
+            && !this.value.acceptedSuggestedAddress) &&
+            <SuggestBetterAddressModal
```
connorlbark, line 145: Note: This modal currently does not have any theming in the participant UI. Adding theming requires making it a `ThemedModal` using the participant-ui component, which, since this is in core, is a not as straightforward as just using it. For this PR, I decided to focus just on functionality, but I have this component themed and much nicer UX-wise with my next PR.

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/address/AddressValidationController.java
```diff
@@ -0,0 +1,32 @@
...
+
+@Controller
+public class AddressValidationController implements AddressValidationApi {
+
+  ObjectMapper objectMapper;
+  AddressValidationService client;
+
+  AddressValidationController(
+      ObjectMapper objectMapper, AddressValidationServiceProvider clientProvider) {
+    this.client = clientProvider.get();
```
devonbush: Let's store the clientProvider, rather than the client.  This preserves the ability of the clientProvider to change the client even after startup
```suggestion
    this.clientProvider = clientProvider;
```

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/address/AddressValidationController.java
```diff
@@ -0,0 +1,32 @@
...
+    this.objectMapper = objectMapper;
+  }
+
+  /** Validates an address using either the stub or real client depending on configuration. */
+  @Override
+  public ResponseEntity<Object> validate(Object body) {
+    MailingAddress mailingAddress = objectMapper.convertValue(body, MailingAddress.class);
+
+    AddressValidationResultDto result = client.validate(mailingAddress);
```
devonbush:
```suggestion
    AddressValidationResultDto result = clientProvider.get().validate(mailingAddress);
```

ui-core/src/surveyjs/address-validation-modal-question.tsx
```diff
@@ -0,0 +1,180 @@
...
+                this.question.value = {
+                  ...this.value,
+                  modalDismissed: true,
+                  acceptedSuggestedAddress: true
+                }
+                if (this.value.addressValidationResult.suggestedAddress) {
+                  this.accept(this.value.addressValidationResult.suggestedAddress)
+                }
+              }}
+              deny={() => {
```
devonbush: nit: `reject` I think is more commonly used as the opposite of accept?

devonbush:
I'm seeing a bug where if I sign in as an existing participant (jsalk) and then go to the basics, it errors on the address page 
image

devonbush (changes requested):
tried it locally--works great. I'm guessing we just need some code to handle existing participants who filled out the survey prior to address validation being added.  And a couple minor code changes.  Looking forward to seeing the next UX iteration!

connorlbark:
> I'm seeing a bug where if I sign in as an existing participant (jsalk) and then go to the basics, it errors on the address page image

I wasn't able to recreate this, but the error makes sense and I added more safeguards in that function which would catch this.

ui-admin/src/address/EditMailingAddress.tsx
```diff
@@ -1,11 +1,15 @@
...
+  explainAddressValidationResults,
+  findDifferencesBetweenObjects,
+  MailingAddress
+} from '@juniper/ui-core'
 import React, { useState } from 'react'
 import { doApiLoad } from '../api/api-utils'
 import Api from '../api/api'
 import LoadingSpinner from '../util/LoadingSpinner'
-import SuggestBetterAddressModal from './SuggestBetterAddressModal'
+import SuggestBetterAddressModal from '@juniper/ui-core/build/components/SuggestBetterAddressModal'
```
MatthewBemis: `SuggestBetterAddressModal` and `EditAddress` should be exported in `index.ts`, that way you'll be able to import them both directly from `@juniper/ui-core` instead of relying on the build paths.

api-participant/src/main/resources/api/openapi.yml
```diff
@@ -240,6 +240,21 @@ paths:
           content: { application/json: { schema: { type: object } } }
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/portals/v1/{portalShortcode}/env/{envName}/address/v1/validate:
```
MatthewBemis: Is the second `v1` path segment intentional here?

ui-core/src/surveyjs/address-validation-modal-question.tsx
```diff
@@ -0,0 +1,184 @@
+/**
+ * A SurveyJS question that renders a multiple selection combobox.
+ * This provides similar functionality as the "tagbox" in https://github.com/surveyjs/custom-widgets.
+ * However, this virtualizes the options list to support mahy more
+ * options while remaining performant.
+ */
```
MatthewBemis: Looks like copy-paste leftover 😃

ui-core/src/surveyjs/address-validator.tsx
```diff
@@ -0,0 +1,199 @@
...
+      || existingValidationState.modalDismissed)
+}
+
+const isSameAddress = (addr1: MailingAddress, addr2: MailingAddress): boolean => {
+  return findDifferencesBetweenObjects(addr1, addr2)
+    .filter(val => !['id', 'createdAt', 'lastUpdatedAt'].includes(val.fieldName))
+    .length === 0
+}
+
+// TODO: this is _not_ internationalized, and only works well for US address formats.
```
MatthewBemis: Can you associate this with a new Jira ticket?

ui-admin/src/forms/FormPreview.tsx
```diff
@@ -5,6 +5,8 @@ import 'survey-core/survey.i18n'
 import { FormContent, PortalEnvironmentLanguage, surveyJSModelFromFormContent, useForceUpdate } from '@juniper/ui-core'
 
 import { FormPreviewOptions } from './FormPreviewOptions'
+import Api from '../api/api'
+import { createAddressValidator } from '@juniper/ui-core/build/surveyjs/address-validator'
```
MatthewBemis: Same here- this should be imported directly from `@juniper/ui-core`

MatthewBemis (commented):
Very nice, worked great locally. Just have a few suggestions

ui-core/src/surveyjs/address-validation-modal-question.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+    const question = this.questionBase.survey.getQuestionByName(questionName)
+
+    if (question) {
+      question.value = value
+    }
+  }
+
+  renderElement() {
+    if (isNil(this.value) || isNil(this.value.addressValidationResult)) {
```
devonbush: a little clearer
```suggestion
    if (isNil(this.value.?addressValidationResult)) {
```

devonbush (approved):
tested, works great now in both pre-existing participants, and new ones.  One super-trivial suggestion.

MatthewBemis (approved):

PR #767, opened 2024-03-01
[JN-893] adding e2e mailing list test

devonbush:
#### DESCRIPTION

This adds an e2e test that follows the first few steps of our smoke test checklist, in confirming that the mailing list form works. Future steps will include:

1. figuring out when/if to run these as part of CI
2. figuring out when/if to run these manually. -- right now I've just added them as a step in the Smoke Test checklsit.

1. run `npm test` from the e2e-tests directory


MatthewBemis (approved):
nice! 🎉

connorlbark (commented):
It took more weird setup issues to get it working, but i think I figured out how to get them to run consistently.

However, I  am getting this error:

```
Error: Timed out 5000ms waiting for expect(received).toHaveText(expected)

Expected string: "Juniper Home"
Received string: "Juniper"
```
From
```
await expect(page.locator('h1')).toHaveText('Juniper Home')
```
Could it be that it's automatically going to the pariticpant list page, so it no longer has that header? That is what it seems to do when I log in

devonbush:
great catch -- yeah, you must only have a single portal populated.  I'll update the logic to be resilient to that

connorlbark (approved):
Nice! Promise.any in the test is cool - didn't know you could use it like that.

devonbush:
yeah, that was a first for me too! :)

PR #770, opened 2024-03-01
[JN-928] add permissions

devonbush:
#### DESCRIPTION
This adds a simple roles/permissions display to the admin tool, as well as a function for checking whether a given user has a given permission.  No permissions are yet enforced server-side

image

To support this, this adds a couple sample roles and permissions.  Right now these are populated by migration script.  As the list grows, we might want to instead have a RolePermissionPopulator that updates/populates based on a standard list somewhere in source code so that it's easier to see the canonical list.  Currently, the role table itself is hidden, visible only to superusers and users with the `prototype_tester` role.  Going forward, we can use this to hide other in-development features from being visible this way--this will let us then expose those features to users without requiring a code change/release.

One tricky thing to note:  The current user object retrieved by the login method does not have the full role->permission hierarchy attached.  Rather it has a property called "portalPermissions" that's a direct map of portalIds to arrays of permission names.  This is both for ease of access, and to make sure that our frontend code always branches based on permissions, rather than roles.  

REMAINING WORK (there is a looooong road ahead...):
 1. study-scoped roles (this PR adds the schema for it, but no code)
 2. server-side enforcement of permissions
 3. robust permission/role population
 4. UI for editing roles for a user, and permissions for a role

1. redeploy apiAdminApp
5. login as yourself
6. go to https://localhost:3000/ourhealth/studies/ourheart/users
7. click the staff user
8. confirm you see a table indicating the user is a study admin
9. log out
10. login as staff@ourhealth.org
11. go to https://localhost:3000/ourhealth/studies/ourheart/users
12. click the staff user
13. confirm the roles table does not appear


core/src/main/java/bio/terra/pearl/core/model/admin/AdminUser.java
```diff
@@ -22,5 +23,5 @@ public class AdminUser extends BaseEntity {
     @Builder.Default
     private boolean superuser = false;
     @Builder.Default
-    private Set<PortalAdminUser> portalAdminUsers = new HashSet<>();
```
connorlbark, line 25: Are we moving away from sets?

populate/src/main/resources/seed/portals/demo/portal.json
```diff
@@ -83,7 +83,8 @@
     }
   ],
   "adminUsers": [
-    {"username": "staff@heartdemo.org"}
+    {"username": "staff@heartdemo.org", "roleNames": ["study_admin"]},
+    {"username": "pro_staff@heartdemo.org", "roleNames": ["study_admin", "prototype_tester"]}
```
connorlbark: nit: `pro_staff` makes me think it's an upgraded 'pro version' staff account, maybe `prototype_staff` or `proto_staff`?

ui-admin/src/user/AdminUserDetail.tsx
```diff
@@ -0,0 +1,94 @@
...
+    getCoreRowModel: getCoreRowModel(),
+    getSortedRowModel: getSortedRowModel()
+  })
+  return <>
+    { basicTableLayout(roleTable) }
+  </>
+}
+
+/**
+ *
```
connorlbark: nit: leftover jsdoc

connorlbark (approved):
looks good to me! seems easy to use and well designed.

core/src/main/java/bio/terra/pearl/core/model/admin/AdminUser.java
devonbush: yes.  Lists are both what JDBI returns naturally, and don't impose business limitations (elements of a set must be unique and can't be ordered).   We don't want our POJOs enforcing constraints like that.

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -1143,43 +1143,22 @@ paths:
...
       summary: Gets an admin user
       tags: [ admin_user ]
       operationId: get
       parameters:
         - { name: id, in: path, required: true, schema: { type: string, format: uuid } }
+        - { name: portalShortcode, in: query, required: false, schema: { type: string } }
       responses:
         '200':
-          description: admin user object
-          content: { application/json: { schema: { $ref: '#/components/schemas/AdminUserDto' } } }
```
MatthewBemis, line 1156: This was the only use of `AdminUserDto` in the spec, so that can be removed now too

MatthewBemis (approved):

PR #774, opened 2024-03-05
[JN-884] Enable address validation on participant profile

connorlbark:
#### DESCRIPTION

Also includes significant UX improvements from discussions with Erin. Because everything is reused, these improvements are across all address components/pages in admin/participant.

Screenshot 2024-03-07 at 1 16 53 PM

Screenshot 2024-03-07 at 1 17 00 PM

Screenshot 2024-03-07 at 1 17 05 PM

- Enable real smarty service in Participant/Admin application.yml
- Login as jsalk
- Go to profile
- Try out different addresses, both good and bad
    - 415 Main St Cambridge MA 02142 (should update zip code, if saved, should do nothing on re-save because it's valid)
    - 415 Main Ave Cambridge MA 02142 (should update Ave -> St)
    - Try 415 Main Ave, but reject it and ensure that it still works
    - Try an apartment building to ensure it asks for unit number
- Try surveyjs experience in ourhealth basics to ensure new modal is themed / no regressions
- Try admin edit participant profile experience to ensure no regressions


devonbush (approved):
Works well.  To confirm -- this is not supposed to allow overriding the bad address, right?

connorlbark:
> Works well. To confirm -- this is not supposed to allow overriding the bad address, right?

Ah yes, good point. I actually forgot that component of this PR. We (erin and i) were thinking about displaying a support email where an admin could go in and manually set your address if it fails and you are confident it was right. I couldn't find a "support email" defined anywhere, though. I was meaning to ask / make a follow up PR if it wasn't as simple as "grab the email from a predefined source". Admins can do this from the profile editor, but they aren't (as of yet) able to edit survey responses, so maybe that is something else we would need to implement in this case.

I also am not sure the best way to "disable this" until it's ready for production - since we don't have the keys, we don't want to run address validation. We could just hard code a flag not to do it until we get the production key then change it after its set up.

MatthewBemis:
I might be failing to remember a use-case here, but do we need admin overrides for addresses in survey responses? I had assumed that the address in the survey was more informational, and the address after it gets mapped to the profile is what we actually rely on for shipping kits, which _is_ admin overridable. So I wouldn't think that this PR is blocked.

MatthewBemis (approved):
Looks good. Pre-approving this, pending result of the above discussion.

One small testing note:

In the profile editor, if you have an address like this in your profile with Apt # on the second line:

```
100 Main St
Apt 5
Boston MA 02119-3500
US
```

If you save again, it will reformat the Apt # to be on the first line (as expected), but when you accept the suggestion, the Apt # will end up on both line 1 and line 2

PR #778, opened 2024-03-07
[JN-864] Include Pre-Configured Email Templates on Study Creation

connorlbark:
#### DESCRIPTION

Also, adds the idea of a "Study Template", which populates studies from json files in populate's seed folder. I assumed, given discussions we had, the idea of pre-populated objects could go further than just the base-layer stuff we want and more into "Here's a specific kind of study you want - let's start there".

- Create a study with the "Base" template - it should have the email trigger configurations
- Create a study with the template cleared - it should have nothing


api-admin/src/test/java/bio/terra/pearl/api/admin/service/study/StudyExtServiceTests.java
```diff
@@ -115,4 +115,36 @@ public void testStudyDeletionNeedsSuperUser(TestInfo testInfo) {
...
     assertThat(portalStudyService.findByPortalId(portal.getId()), not(empty()));
   }
+
+  @Test
+  @Transactional
+  void testCreateWithTemplate(TestInfo info) {
+    Portal portal = portalFactory.buildPersisted(getTestName(info));
+    for (EnvironmentName envName : EnvironmentName.values()) {
+      environmentFactory.buildPersisted(getTestName(info), envName);
+      portalEnvironmentFactory.buildPersisted(getTestName(info), envName, portal.getId());
```
devonbush: Let's move this into a new  portalFactory method like "buildPersistedWithEnvironments"

api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyExtService.java
```diff
@@ -66,9 +80,43 @@ public Study create(String portalShortcode, StudyCreationDto study, AdminUser op
...
+    StudyPopDto studyPopDto;
+
+    try {
+      String fileContents = filePopulateService.readFile(filename, config);
+      studyPopDto = studyPopulator.readValue(fileContents);
+
+      studyPopDto.setShortcode(newStudy.getShortcode());
+      studyPopDto.setName(newStudy.getName());
+
+      studyPopulator.createPreserveExisting(newStudy, studyPopDto, config);
```
devonbush: congrats on working through the very wonky populator interfaces.  I think you want to use `studyPopulator.populateFromDto()` here though, as that picks up an extra preProcessing step (which is a no-op for studies currently, but could have code later).    I think createPreserveExisting is public from a leftover time when Populator was an interface.  createPreserveExisting should be protected (although changing it is outside of this PR)

devonbush (approved):
looks good!  couple small suggestions.  As we discussed, big things ahead for this feature!

MatthewBemis (approved):

PR #780, opened 2024-03-08
[JN-638] add ability to skip address validation 

devonbush:
#### DESCRIPTION

This adds a checkbox for admins to skip address validation, plugging into the existing Pepper API for taking a `skipAddressValidation` boolean. 

image

this updates the kit request views so you can see whether address validation was skipped as part of the 'detail'.
image
this display is pretty rough.  But we'll be updating the UX for these lists soon to add some more fields (creating admin user, etc...) and so I didn't want to do too much here.

1. redeploy apiAdminApp 
2. repopulate demo
3. go to `https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants/HDBASC`
4. confirm that when you click on the "details" info icon on the kit request you see the request.
5. click "Create kit request" and create kit requests with/without validation


ui-admin/src/study/participants/RequestKitModal.tsx
```diff
@@ -84,3 +87,21 @@ export const useKitTypeSelect = (studyEnvParams: StudyEnvParams) => {
       onChange={option => setKitType(option?.value ?? '')}/>
   }
 }
+
+/**
+ *
+ */
+export const useBadAddressOverride = (initialValue: boolean) => {
```
connorlbark, line 94: Not sure I'm a huge fan of this approach over a component - I could be convinced, but the added simplicity doesn't seem worth the cost of confusion.

connorlbark (approved):
Looks good! Not related to this PR but I noticed that the sorting on the kits table was bugged - if you try to sort on `Created` time it doesn't have the expected behavior. if it's an easy fix, maybe worth doing here

ui-admin/src/study/participants/RequestKitModal.tsx
devonbush: interesting -- to me these hooks that let the control manage its own state are less confusing since they handle everything themselves.  I'll add it to the eng. strategy.  I know there are cases where a component is best, and where a hook like this is best (e.g. the useReactTable hook), we could use some good guidance on when to use which

connorlbark: Hmm, I see what you're saying. I see hooks as being `set up internal state (that has no connection to DOM)` and components as `connect thing to your DOM`. I think `useReactTable` is less confusing because it only returns what you're supposed to add to the dom - but at that point, I feel like it should just be a component. Idk. I don't feel super strongly about it, so I will let you make the call for this case.

MatthewBemis (approved):

PR #781, opened 2024-03-08
[JN-815] adding login enrollee tests

devonbush:
#### DESCRIPTION

Previously, we had no tests validating login for users with enrollees attached.  This adds a few tests for single, multiple, and proxy enrollee situations.  this fixes a bug in how we were finding relations - along the way this switches from filtering invalid relations in the DB to doing it in Java.  The number of relations is almost always 0-1, so doing it in the DB isn't worth the extra query complexity.

1. login to participant UI


connorlbark (approved):
Nice, I could log in! New tests look good too 👍

pegahtah (approved):
Nice refactoring and tests

PR #783, opened 2024-03-11
[JN-888] Adding renaming ability to upload/populate

devonbush:
#### DESCRIPTION

This adds the ability to rename/recode portals when you are uploading or populating them (and was inspired by Connor's PR and a conversation about the template portal).  This allows you to extract a portal, and then re-upload it as a different portal, effectively "cloning it".  This takes care of assigning new stableIds to any documents attached to the portal so that the new portal is independent from the old.  

image

Future PRs will handle defaulting the b2c configuration, and allowing changing the display names of the portals/studies

1. go to the portal populate page `https://localhost:3000/populate/portal`
2. put in the path for the ourhealth study (`portals/ourhealth/portal.json`)
3. add a shortcode override
4. populate it.
5. Observe the new portal got created with a different name/shortcode. (note the participant side won't be accessible due to b2c config missing -- that's for a later PR)
image

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/internal/PopulateController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/internal/PopulateController.java
@@ -15,6 +15,7 @@ import bio.terra.pearl.populate.service.BaseSeedPopulator;
 import jakarta.servlet.http.HttpServletRequest;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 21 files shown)


MatthewBemis (approved):

connorlbark (approved):
Strangely, when I used a shortcode override, it didn't replace the portal name, but instead prepended like it should do with other shortcodes. I also think, per our discussion, long-term having the shortcodes be portal-specific seems like a more elegant solution, but it is exciting to be able to have this all working in only a couple days of work

PR #792, opened 2024-03-13
[JN-941] move i18nprovider into ui-core

MatthewBemis:
#### DESCRIPTION

This moves the I18nProvider into core so we can internationalize common components. I also moved `selectedLanguage` and `changeLanguage` out of the user provider and into the i18n provider to simplify this refactoring. Otherwise, the UserProvider would need to be refactored to handle both admin users and participant users.

I also internationalized the SiteFooter component (with the Terms of Use and Privacy Policy links) since that was a common component that also needed I18nProvider to be in ui-core.  While this sets things up to be used in the admin UI, the admin UI doesn't properly support i18n yet so you'll see the substitution keys when previewing the site in the admin site builder. The participant site will render the substitutions correctly, though.

* rebuild ui-core, restart admin and participant apps
* Repopulate Heart Demo `./scripts/populate_portal.sh demo`
* Load the participant site and make sure that the Privacy Policy and Terms of Use texts in the site footer render in the correct language https://heartdemo.ourhealth.localhost:3001/


connorlbark (approved):
great refactor! I also really like moving `selectedLanguage` to the i18n provider, I think that makes a lot of sense.

devonbush (approved):
nice!  big fan of the move of selectedLanguage out of UserProvider -- I think that's going to make a lot of concepts cleaner especially as we get deeper into proxies

PR #793, opened 2024-03-13
JN-953 fix for delete study bug

pegahtah:
#### DESCRIPTION

The frontend was passing in the latest value of the portal in the portalList instead of getting the right value for the study. 

Try deleting a study from admin api

### DIFF
--- a/ui-admin/src/HomePage.tsx
+++ b/ui-admin/src/HomePage.tsx
@@ -45,7 +45,7 @@ function HomePage() {
                     aria-expanded="false" aria-label="configure study"/>
                   <div className="dropdown-menu">
                     <ul className="list-unstyled">
-                      <li key={`delete-${portal.shortcode}-${study.shortcode}`}>
+                      <li>
                         <button className="dropdown-item"
                           onClick={
                             () => {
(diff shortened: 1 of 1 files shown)


ui-admin/src/HomePage.tsx
```diff
@@ -43,17 +45,18 @@ function HomePage() {
                     aria-expanded="false" aria-label="configure study"/>
                   <div className="dropdown-menu">
                     <ul className="list-unstyled">
-                      <li>
+                      <li key={`delete-${portal.shortcode}-${study.shortcode}`}>
```
connorlbark: Should this be `delete-${selectedPortal.shortcode}-${selectedStudy.shortcode}`?

pegahtah: Oh I think this could actually go away

pegahtah: it should not be selected* because then the keys for all of items in the list would change based on the click, but we don't really need a key here because there is no need for a memory in the list here.

connorlbark (approved):
worked for me

devonbush (approved):

PR #794, opened 2024-03-13
[JN-954] Allow admin user to be added to more than one portal

MatthewBemis:
#### DESCRIPTION

This fixes a bug where an AdminUser couldn't be added to a new portal after initial creation.

Restart admin API
Add a new admin user to OurHealth
Add the same admin user to HeartHive
Verify that the admin user has gets to both portals


connorlbark (approved):
nice!

devonbush (approved):

PR #802, opened 2024-03-15
[JN-944] prototype enrollee importer

devonbush:
#### DESCRIPTION

adds a prototype for an enrolleeImportService.  Doesn't do much except read in a username, but wanted to get thoughts on how the importer is integrated with the export formatters.

1. no functional changes to the app, so just run the tests


core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
```diff
@@ -0,0 +1,100 @@
...
+        this.registrationService = registrationService;
+        this.enrollmentService = enrollmentService;
+        this.profileService = profileService;
+    }
+
+    @Transactional
+    /**
+     * exports the specified number of enrollees from the given environment
+     * The enrollees will be returned most-recently-created first
+     * */
```
MatthewBemis, line 47: nit: "exports" leftover in copy/paste

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.*;
+
+@Service
+public class EnrolleeImportService {
+    /** for now, we only support importing from a specific style of export. */
+    ExportOptions IMPORT_OPTIONS = new ExportOptions(false, true, true, ExportFileFormat.TSV, null);
```
MatthewBemis, line 31: This is already pretty closely based on Peppers export options right?

MatthewBemis (approved):

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
connorlbark, line 31: Nit: should we rename `ExportOptions` if it's also being used for imports? I could also see the argument that we are importing data exported with these `ExportOptions`... Maybe `SerializationOptions`?

connorlbark (approved):
looks good!!

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
devonbush: that's a good idea -- I'll see if these diverge or not over upcoming PRs, and if they stay together, I like "serializationOptions" as the name

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
devonbush: good catch, will fix in later PR to avoid up-the-chain merge conflicts

PR #810, opened 2024-03-22
[JN-944] proof-of-concept form parser

devonbush:
#### DESCRIPTION

minimal PoC for parsing a pepper form into a Juniper survey file that could be read by our populate logic.   All code will likely be changed, so no need to review the code, just get a sense of what's happening generally.   We're reading Pepper survey .conf files (which I've added to our source for now, later we could just read them off the classpath) and parsing them--the only tricky bit is getting the language and variable substitutions right (which we definitely aren't getting all the way right, but just 30% of the way there.

1. run `./gradlew pepper-import:bootRun` 
2. confirm a new file `pepper-import/out/prequal.conf` is created that has a survey-like structure


connorlbark (approved):
nice start! i have to write a lot of lines to catch up on the contributions list 🤔

MatthewBemis (approved):
Clicking into a new PR and seeing something like this: 
Screenshot 2024-03-25 at 11 39 00 AM

followed by "no need to review the code" is always a huge relief 😄 

Looks good though. Worked when I tried it out and seems like a great first step.

devonbush:
yes, this project was definitely about padding my github stats :)

PR #811, opened 2024-03-22
[JN-621] Redirect to B2C pages using users preferred language

MatthewBemis:
#### DESCRIPTION

The important change here is that the various B2C redirects (sign-in, register, change password) take in a `ui_locales` query parameter that honors the `selectedLanguage` that's been provided by `I18nProvider`. This is in preparation of applying the new B2C polices in https://github.com/broadinstitute/terraform-ap-deployments/pull/1419 

While I was mucking with some mocks, I also went and fixed a bunch of console error logs that were being produced by other tests. I also lifted the `asMockedFn` test util from Terra UI so we could get rid of the `@ts-expect-error "TS doesn't know about mocks"` notation that we had in several unit tests.

* Boot up the participant UI and populate Heart Demo (make sure you're booting with B2C configured)
* Select Spanish in the Navbar
* Click the sign in button
* You should be redirected to a B2C sign in page with _some_ (but not all) of the content in Spanish - the aforementioned b2c policies need to be applied before all of the content will be translated
* Log in as an existing user and verify that you see some Spanish translations for the B2C Change Password screen
* Register as a new user and verify that you see some Spanish translations for the B2C registration page (you won't be able to complete the registration flow in Spanish - the new B2C policies and assets need to be applied before this will fully work)


ui-participant/src/Navbar.tsx
```diff
@@ -283,3 +223,81 @@ export function LanguageDropdown({ languageOptions, selectedLanguage, changeLang
       </li>) : null
   )
 }
+
+/**
+ * User account dropdown menu, with options to edit profile, change password, and log out
+ */
+export const AccountDropdown = () => {
```
devonbush: nice refactor

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.test.tsx
```diff
@@ -22,6 +22,13 @@ jest.mock('user/UserProvider', () => {
   }
 })
 
+jest.mock('api/api', () => ({
+  ...jest.requireActual('api/api'),
+  fetchEnrolleeAdminTasks: jest.fn(() => Promise.resolve([])),
```
devonbush: there's a shorthand for this that I just learned about...
```suggestion
  fetchEnrolleeAdminTasks: jest.fn().mockResolvedValue([]),
```

devonbush (approved):
looks great -- just one code style touchup suggestion

connorlbark (approved):
nice!

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.test.tsx
MatthewBemis: TIL. I just updated everywhere to use this instead, so Copilot and devs can take advantage

PR #812, opened 2024-03-23
[JN-946] enrollee invitation proof-of-concept

devonbush:
#### DESCRIPTION

This enables invitation emails to be sent to participants who have data in Juniper (such as from a migration), but do not have b2c accounts, allowing them to create their linked b2c account

image

To enable easy testing of this feature, I needed a way to repeatedly create an enrollee who did not have a b2c account.  So I introduced the concept of "keyed" populate users.  This lets us be sure that every time we populate demo, the `HDINVI` enrollee will be created with a new username and so will be able to be invited without conflicting with an existing b2c account.  The email generated looks like [your username]+[key]-[randomchars]@broadinstitute.org.

FUTURE WORK:
 1. b2c configuration to prefill the email address field (and maybe make it read-only)
 2. not hardcoding all "AD_HOC" emails to prompt for subject/body text
 3. some sort of indication in the data model (other than lastLogin) that a participantUser is (a) from a migration and (b) does not yet have a b2c account

1. redeploy ApiAdminApp
2. repopulate demo
3. make sure the ui-participant is running with b2c auth *on*
4. go to sandbox participant list `https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants`
5. check-select the HDINVI enrollee, and press "Send email"
6. In the "email template" dropdown, select "Invitation email"
7. ignore the text fields, and press "send to 1 participant"
8. Confirm you get an email that looks like the screenshot above
9. click the link, confirm you are taken (after a redirect) to the "Verify your email" screen
10. (you won't be able to actually verify the email since demo b2c isn't yet configured)

### DIFF
--- a/ui-participant/src/landing/registration/InvitationPage.tsx
+++ b/ui-participant/src/landing/registration/InvitationPage.tsx
@@ -25,7 +25,7 @@ export default function InvitationPage() {
     auth.signinRedirect({
       redirectMethod: 'replace',
       extraQueryParams: {
-        option: 'signup2',
+        option: 'signup',
         // eslint-disable-next-line camelcase
         login_hint: accountName
       }
--- a/ui-participant/src/landing/registration/InvitationPage.tsx
+++ b/ui-participant/src/landing/registration/InvitationPage.tsx
@@ -1,0 +1,38 @@
+import React, { useEffect } from 'react'
+import { useNavigate, useSearchParams } from 'react-router-dom'
+import { useAuth } from 'react-oidc-context'
+import { PageLoadingIndicator } from 'util/LoadingSpinner'
+
+/** Page for participants who already have enrollee data in Juniper (from a migration or admin action), and need
+ * to join to link their account */
+export default function InvitationPage() {
+  const auth = useAuth()
+  const [searchParams] = useSearchParams()
+  const accountName = searchParams.get('accountName') || ''
+  const navigate = useNavigate()
+
+  const register = () => {
+    if (process.env.REACT_APP_UNAUTHED_LOGIN) {
+      /**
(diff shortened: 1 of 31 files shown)


connorlbark (approved):
Worked well for me - I assume it's expected that if you go through the full invitation process it doesn't work (I got a 500 error on every page since there wasn't a portal participant user). 👍  pending the failing test being fixed.

devonbush:
hmmm... @connorlbark  where are you seeing there's no portal participant user?  when I do the populate demo, a portal participant user gets created for the 'invited' participant.  I wonder if there's some condition that can cause that populate to be flaky?

ui-participant/src/landing/registration/InvitationPage.tsx
```diff
@@ -0,0 +1,40 @@
...
+      // we don't need to create a b2c account, just log the user in
+      const result = await Api.unauthedLogin(accountName)
+      loginUser(result, result.user.token)
+      navigate('/hub')
+      return
+    }
+    auth.signinRedirect({
+      redirectMethod: 'replace',
+      extraQueryParams: {
+        option: 'signup2',
```
MatthewBemis: Is `signup2` a typo?

MatthewBemis (approved):
Looks good. One minor question

And as another data point, the portal participant user was successfully created when I tested this out

ui-participant/src/landing/registration/InvitationPage.tsx
devonbush: leftover from testing, good catch!

devonbush:
security hotspot is using random number generator in populate file.  psuedo random is fine there since it's dev-only

PR #815, opened 2024-03-25
[JN-963] Fix file upload validation

MatthewBemis:
#### DESCRIPTION

The existing file type validation was not particular reliable. Icon files can potentially have multiple valid mime types, which was causing issues when Whitney was trying to add a new favicon for the template study. Uploading a `.jpeg` was also broken (but `jpg` was fine). Instead of capturing every possible mime type for each allowed file type, we'll just validate using file extension. This is in line with the backend validation anyway.

Restart admin UI
Try to upload a .jpeg file and a .ico file (confirm it succeeds)
Try to upload a .zip file (confirm that this is still disallowed and that validation still works)


devonbush (approved):
I like the simplicity, nice!

connorlbark (approved):
nice!

PR #818, opened 2024-03-26
[JN-940] Add eligibility rule to survey

connorlbark:
#### DESCRIPTION

Adds eligibility rule to surveys which restricts the survey to only assign to enrollees that match the given expression

- Create a survey with the rule `{profile.givenName} = 'Jonas'` and auto assign to existing
- Make sure the participant assignment is only one person 
- Go to OHSALK's task list to ensure it has the new survey
- Ensure no other participant has it in their task list
- Try creating or updating a survey with an invalid eligibility rule and ensure it does not allow.


core/src/main/java/bio/terra/pearl/core/service/rule/EnrolleeProfileBundle.java
```diff
@@ -4,9 +4,12 @@
...
 import lombok.AllArgsConstructor;
 import lombok.Getter;
+import lombok.experimental.SuperBuilder;
 
-@Getter @AllArgsConstructor
-public class EnrolleeRuleData {
+@Getter
+@AllArgsConstructor
+@SuperBuilder
+public class EnrolleeProfileBundle {
```
connorlbark, line 12: I am not sure if this is the best name for this class. It was originally `EnrolleeRuleData`. I wanted to delete it in favor of `EnrolleeSearchContext`, but I noticed it was being used in a bunch of unrelated places to rules.  I thought making it a `Bundle` with a corresponding `EnrolleeBundleService` could be more generic, but I am not 100% sure about it.

core/src/main/java/bio/terra/pearl/core/service/rule/EnrolleeRuleData.java
```diff
@@ -6,8 +6,11 @@
...
 import lombok.Getter;
 
-/** minimal set of data that we pre-load when processing enrollees for business logic and/or emails */
-@Getter @AllArgsConstructor
+/**
+ * minimal set of data that we pre-load when processing enrollees for business logic and/or emails
+ */
+@Getter
+@AllArgsConstructor
 public class EnrolleeRuleData {
```
connorlbark, line 14: I think we should rename this class - I almost tried to delete it because I thought it was only related to enrollee rules but then realized it was being used in a LOT of places. I was thinking `EnrolleeBundle`/`EnrolleeBundleService` would make it a bit clearer

connorlbark: Update: Out of scope, so I made a ticket in the eng tech debt backlog to rename this.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/SurveyExtService.java
```diff
@@ -94,6 +99,8 @@ public Survey create(String portalShortcode, Survey survey, AdminUser operator)
     if (existing.size() > 0) {
       throw new IllegalArgumentException("A survey with that stableId already exists");
     }
+    // ensure the rule can be parsed (if null/empty, this will be a no-op)
```
devonbush, line 102: nice!

core/src/main/java/bio/terra/pearl/core/service/consent/ConsentTaskDispatcher.java
```diff
@@ -98,12 +103,14 @@ public static boolean checkIsEnrolleeConsented(List<ParticipantTask> participant
...
     /** builds the consent tasks, does not add them to the event or persist them */
     public List<ParticipantTask> buildTasks(Enrollee enrollee,
-                                                   EnrolleeRuleData enrolleeRuleData,
+                                            EnrolleeRuleData enrolleeRuleData,
                                                    UUID portalParticipantUserId,
                                                    List<StudyEnvironmentConsent> studyEnvConsents) {
         List<ParticipantTask> tasks = new ArrayList<>();
         for (StudyEnvironmentConsent studyConsent : studyEnvConsents) {
-            if (EnrolleeRuleEvaluator.evaluateRule(studyConsent.getEligibilityRule(), enrolleeRuleData)) {
+            if (enrollee.isSubject() && enrolleeSearchExpressionParser
```
devonbush, line 113: I think the logic we want is  "if rule is blank, apply default rule, otherwise parse rule."  (See `evaluateRuleChecked` in EnrolleeRuleEvaluator).  Otherwise we've foreclosed the ability to have forms that apply to non-subjects.

core/src/main/java/bio/terra/pearl/core/service/search/EnrolleeSearchExpressionParser.java
```diff
@@ -46,7 +48,7 @@ public EnrolleeSearchExpressionParser(EnrolleeDao enrolleeDao, AnswerDao answerD
 
 
     public EnrolleeSearchExpression parseRule(String rule) throws RuleParsingException {
-        if (StringUtils.isBlank(rule)) {
+        if (Objects.isNull(rule) || StringUtils.isBlank(rule)) {
```
devonbush: this is redundant -- StringUtils.isBlank checks for nulls

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcher.java
```diff
@@ -189,8 +194,10 @@ public Optional<ParticipantTask> buildTaskIfApplicable(Enrollee enrollee,
         return Optional.empty();
     }
 
-    public static boolean isEligibleForSurvey(String eligibilityRule, EnrolleeRuleData enrolleeRuleData) {
-        return EnrolleeRuleEvaluator.evaluateRule(eligibilityRule, enrolleeRuleData);
+    public boolean isEligibleForSurvey(String eligibilityRule, EnrolleeRuleData enrolleeRuleData) {
```
devonbush, line 197: same comment as with consent forms -- we need to apply a default rule (that it only applies to subjects) if nothing is specified

core/src/test/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcherTest.java
```diff
@@ -142,4 +141,28 @@ public void testAssignDoesntDuplicate(TestInfo testInfo) {
         assertThat(participantTasks, hasSize(2));
     }
 
+    @Test
+    @Transactional
```
devonbush, line 145: great test, let's add another one to confirm we don't assign surveys to non-subjects by default.

devonbush (changes requested):
overall looks good -- just needs some logic tweaks to handle default rules

core/src/main/java/bio/terra/pearl/core/service/consent/ConsentTaskDispatcher.java
connorlbark: Yeah, I agree with you that it shouldn't be hardcoded, but I want to sync with Erin on how these rules should work regarding isSubject. Previously, the isSubject check was just in the enrollee rule evaluator, but I am not sure how I feel about it being a default rule - let's say you wanted to give the survey to both enrollees and their proxies. Then, you have to do `{enrollee.subject} = true and {enrollee.subject} != true`? Or, if we only do the isSubject if the rule is empty, then does that mean we have to prepend every single rule with `{enrollee.subject} = true`? Both feel weird/arcane to me.

Thinking about it, it feels like these subject checks should exist outside of search expressions, at least in the context of survey eligibility rules. It feels like, to me, there should be two checkboxes above the eligibility rule, `Give survey to enrollees` and `Give survey to proxies` (in the case that the study has proxies), and you can either select one or both. By default, only enrollees would be checked. Then there isn't any magic happening in the backend that you have to work around and it's super clear who is and is not being given this survey.

connorlbark: Also, I made the assumption that consent surveys would only be for enrollees, because they're the ones consenting to the survey, but maybe this is not always true.

devonbush (approved):
after discussion, agree the logic is good enough for now, but please add the extra tests

MatthewBemis (approved):

PR #821, opened 2024-03-27
[JN-620] Form name i18n

MatthewBemis:
#### DESCRIPTION

Opening this for early feedback. There are still a few wrinkles to iron out.

This internationalizes forms by using the built in SurveyJS "title" field. This required a refactor of how we populate the task list in a participant dashboard because we need to also retrieve the form with the task so we can parse the title in the appropriate language. We were already doing this type of retrieval for Outreach tasks, so this rolls Consents and Surveys into the same endpoint.

This is currently a bit rough in a few areas:

* We still ask for and persist the "name" field on surveys when they're created. In the future, setting the "name" field when creating a new survey should automatically inject the title into the survey JSON upon creation.
* Until consents and surveys are unified, some of the backend code is a little bit rough
* Likely needs some harmonization with currently-open proxy PRs

Screenshot 2024-03-27 at 5 47 08 PM

* Repopulate demo
* Login as jsalk
* Confirm that you see Spanish and DEV form titles

### DIFF
--- a/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/task/ParticipantTaskController.java
+++ b/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/task/ParticipantTaskController.java
@@ -5,42 +5,33 @@ import bio.terra.pearl.api.participant.service.ParticipantTaskExtService;
 import bio.terra.pearl.api.participant.service.RequestUtilService;
 import bio.terra.pearl.core.model.EnvironmentName;
 import bio.terra.pearl.core.model.participant.ParticipantUser;
-import bio.terra.pearl.core.model.workflow.TaskType;
-import bio.terra.pearl.core.service.participant.EnrolleeService;
 import jakarta.servlet.http.HttpServletRequest;
-import java.util.List;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
 
 @Controller
 public class ParticipantTaskController implements ParticipantTaskApi {
-  private EnrolleeService enrolleeService;
-  private ParticipantTaskExtService participantTaskExtService;
-  private RequestUtilService requestUtilService;
-  private HttpServletRequest request;
+  private final ParticipantTaskExtService participantTaskExtService;
+  private final RequestUtilService requestUtilService;
+  private final HttpServletRequest request;
 
   @Autowired
   public ParticipantTaskController(
-      EnrolleeService enrolleeService,
       ParticipantTaskExtService participantTaskExtService,
       RequestUtilService requestUtilService,
       HttpServletRequest request) {
-    this.enrolleeService = enrolleeService;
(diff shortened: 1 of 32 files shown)


api-participant/src/main/java/bio/terra/pearl/api/participant/controller/task/ParticipantTaskController.java
```diff
@@ -5,42 +5,33 @@
...
-    this.enrolleeService = enrolleeService;
     this.participantTaskExtService = participantTaskExtService;
     this.requestUtilService = requestUtilService;
     this.request = request;
   }
 
   @Override
-  public ResponseEntity<Object> listTasksWithSurveys(
-      String portalShortcode, String envName, String taskType) {
+  public ResponseEntity<Object> listTasksWithSurveys(String portalShortcode, String envName) {
```
devonbush, line 30: rename this method?
```suggestion
  public ResponseEntity<Object> listTasksWithForms(String portalShortcode, String envName) {
```

api-participant/src/main/java/bio/terra/pearl/api/participant/service/ParticipantTaskExtService.java
```diff
@@ -1,67 +1,94 @@
...
+
+    return IntStream.range(0, tasks.size())
+        .mapToObj(i -> new TaskAndForm(forms.get(i), tasks.get(i)))
+        .toList();
   }
 
-  public record TaskAndSurvey(Survey survey, ParticipantTask task) {}
+  public record TaskAndForm(VersionedForm form, ParticipantTask task) {}
+
+  public record TasksAndForms(
```
devonbush, line 90: I don't think there's any need to have a DTO that breaks these out -- they can be sorted/binned client-side

devonbush (commented):
hmm... I worry about the performance impact of this.  I like that using the surveyJs title attribute keeps things simpler, but this means that we have to load and parse every survey content into javascript just to load the dashboard.  And once we have customers with long surveys translated into a bunch of languages, that's going to be BIG.

MatthewBemis:
@devonbush yeah that's a very reasonable concern. I suppose we could parse the title server-side and return it in the users preferred language. we'd still have to load all of the content from the DB, but this saves a bunch data transfer and parsing on the client side.

maybe this wasn't the right approach. i was hoping to keep survey content self-contained by leaning into the built-in `title` attribute but it might not have been worth it. i'll think about it some more

devonbush:
What if, when a survey is saved, we automatically parse the title attribute into LanguageTexts?  (kind of mirroring how we create SurveyQuestionDefinitions automatically when a survey is saved)
Then they'd be fetched automatically with the rest of the Portal i18n content, and assuming we assembled the keys in a predictable way, the frontend could figure out the text for a given survey stableId/version.  We'd have to add portal and environment scoping to LanguageTexts to avoid returning the titles to all surveys, but that scoping was probably something we'd have to do later anyway.

MatthewBemis:
Closing in favor of a new branch

PR #823, opened 2024-03-29
[JN-902] Dashboard background is configurable

connorlbark:
#### DESCRIPTION

Paired with Matt

Makes dashboard background configurable in UI. 

- Restart admin API
- Do not populate - ensure that ourhealth dashboard background looks good
- Populate ourhealth & demo - ensure that demo has its own, unique background
- Try changing background and ensure that it still works - NOTE it breaks multi-lang, so you have to stay on English for this. There could be an argument to be made that these brandings should be outside of the local site content, but we figured that was out the scope of this PR to determine if that made sense.


devonbush (approved):
Nice!  I agree that we'll eventually want a concept of "these things are on localSiteContent, but are probably language-invariant", but for now I heartily agree that putting it on LocalizedSiteContent was the right choice

MatthewBemis (approved):

PR #825, opened 2024-03-29
[JN-801] Admins can add users to the mailing list (members choice)

MatthewBemis:
#### DESCRIPTION

Allows study staff to add new contacts to the mailing list. Useful if they collect contact info from an in-person event. Also has an "Import CSV" feature which will automatically populate the table.

https://github.com/broadinstitute/juniper/assets/7257391/afd0b7cf-f57d-4345-a3a1-44d2c17fd940

* Restart admin API
* Repopulate OurHealth
* Go to the mailing list view: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/mailingList
* Add a few contacts manually
* Try importing a csv file with a bunch of contacts


ui-admin/src/portal/MailingListView.tsx
```diff
@@ -118,11 +114,14 @@ export default function MailingListView({ portalContext, portalEnv }:
           <RowVisibilityCount table={table}/>
         </div>
         <div className="d-flex">
-          <Button onClick={download}
-            variant="light" className="border m-1" disabled={!numSelected}
-            tooltip={numSelected ? 'Download selected contacts' : 'You must select contacts to download'}>
-            <FontAwesomeIcon icon={faDownload} className="fa-lg"/> Download
```
MatthewBemis, line 124: I switched download to use the generic DownloadControl component from tableUtils. It does introduce a behavior change. Does an admin actually need to be able to select individual contacts to download or is downloading the entire table fine?

devonbush: The reason we supported select+download was in case they wanted to filter the list (only email people who signed up this month?).  But I'm not sure we have any actual use case for that, so getting rid of it is fine.

core/src/main/java/bio/terra/pearl/core/service/portal/MailingListContactService.java
```diff
@@ -22,6 +22,16 @@ public Optional<MailingListContact> findByPortalEnv(UUID portalEnvId, String ema
         return dao.findByPortalEnv(portalEnvId, emailAddress);
     }
 
+    @Transactional
+    public List<MailingListContact> bulkCreate(UUID portalEnvId, List<MailingListContact> contacts) {
+        findByPortalEnv(portalEnvId).forEach(existing ->
```
devonbush, line 33:
```suggestion
        // remove contacts from the submitted list if they already exist in this environment
        findByPortalEnv(portalEnvId).forEach(existing ->
```

ui-admin/src/portal/AddMailingListUsersModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+      <button className="btn btn-primary"
+        onClick={() =>
+          setContacts([...contacts, { name: '', email: '' }])
+        }>
+        Add
+      </button>
+    </Modal.Body>
+    <Modal.Footer>
+      <LoadingSpinner isLoading={isLoading}>
+        {FileChooser}
```
devonbush: I think it's confusing to have this button here.  How about directly underneath the "Add" button?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/MailingListExtService.java
```diff
@@ -47,6 +49,43 @@ public List<MailingListContact> getAll(
...
+    Portal portal = authUtilService.authUserToPortal(user, portalShortcode);
+    PortalEnvironment portalEnv =
+        portalEnvironmentService
+            .findOne(portal.getShortcode(), envName)
+            .orElseThrow(() -> new NotFoundException("Portal environment not found"));
+
+    List<MailingListContact> newContacts =
+        mailingListContactService.bulkCreate(portalEnv.getId(), contacts);
+
+    List<DataChangeRecord> changeRecords =
```
devonbush: I'm not sure we need to audit mailing list records, but if we do, we should just change MailingListService to extend DataAuditedService, and then you'll get a bulkCreate method for free :).  This will mean switching MailingListRecords from immutable to mutable, but that's fine since they're now audited

devonbush (changes requested):
Looks great!  a couple minor suggestion.  the blocking one is that if we're going to audit mailingList records, I think we should go all the way.  this might mean we need a new ResponsibleEntity type of "anonymousParticipant"

ui-admin/src/portal/AddMailingListUsersModal.tsx
MatthewBemis: Done!

Screenshot 2024-04-01 at 2 13 30 PM

core/src/main/java/bio/terra/pearl/core/service/portal/MailingListContactService.java
```diff
@@ -22,6 +27,17 @@ public Optional<MailingListContact> findByPortalEnv(UUID portalEnvId, String ema
...
+                contacts.removeIf(contact -> contact.getEmail().equals(existing.getEmail())));
+
+        contacts.forEach(contact -> contact.setPortalEnvironmentId(portalEnvId));
+        bulkCreate(contacts, auditInfo);
+        return contacts;
+    }
+
     @Transactional
     public void deleteByPortalEnvId(UUID portalEnvId) {
```
connorlbark, line 42: since it's data audited, should this delete take in a DataAuditInfo and be audited?

core/src/main/java/bio/terra/pearl/core/model/audit/ResponsibleEntity.java
```diff
@@ -33,4 +34,10 @@ public ResponsibleEntity(String systemProcess) {
         }
         this.systemProcess = systemProcess;
     }
+    public ResponsibleEntity(Boolean anonymousUser) {
```
connorlbark: nit: should this just be `ResponsibleEntity()` to lean into the "anonymous"-ness? I would find `new ResponsibleEntity(true)` a bit arcane...

populate/src/main/java/bio/terra/pearl/populate/service/PortalPopulator.java
```diff
@@ -142,7 +143,7 @@ private void populateMailingList(PortalEnvironmentPopDto portalEnvPopDto, Portal
             contact.setPortalEnvironmentId(savedEnv.getId());
             contact.setEmail(contact.getEmail());
             contact.setName(contact.getName());
-            mailingListContactService.create(contact);
+            mailingListContactService.create(contact, null);
```
connorlbark: should this pass in a system process data audit info?

connorlbark (approved):
Some non-blocking changes, looks good! I like the flow, it feels really clean.

core/src/main/java/bio/terra/pearl/core/service/portal/MailingListContactService.java
MatthewBemis: I could go either way on this. Since we'd only ever use this if we're destroying the portal, I'm not sure it's worth keeping a record of each contact that was also destroyed, especially since this is sort of internal-only during portal repopulate. But if you or Devon feel strongly, I can totally make the change

devonbush (approved):

PR #826, opened 2024-03-29
[JN-979] Portal specific shortcodes

connorlbark:
#### DESCRIPTION

For member's choice, I wanted to do a spike to see how hard it would be to have portal specific shortcodes. This would enable more advanced templating, since different portals could import the same study template and not have to worry about the stable ids conflicting.

However, since they are not study specific, for multi-study portals, we do have to be careful about template stable ids. I'm thinking, when we "clone" studies, we could have an optional `stable_id_prefix` that would only be needed in the multi-study case. Or, we could make shortcodes study-specific 😵‍💫. I suspect this will be possible but harder than portal specific shortcodes.
 
- Repopulate ourhealth and demo
- Ensure no functional regressions
- Create a survey in demo with the shortcode `oh_oh_basicInfo`. Assign to all participants and ensure that ourhealth and demo both display the correct surveys.


core/src/main/resources/db/changelog/changesets/2024_03_29_portal_scoped_stable_ids.yaml
```diff
@@ -0,0 +1,76 @@
...
+            tableName: survey
+
+        - addNotNullConstraint:
+            columnName: portal_id
+            tableName: consent_form
+        #
+        #        - addNotNullConstraint:
+        #            columnName: portal_id
+        #            tableName: email_template
+
```
connorlbark: All of the other versioned objects already had portal ids by default, so they didn't need any changes, except for `email_template`. This one has a single not-portal-specific template for admin users. Right now, that is considered to be a part of the `null` portal, as in `findByShortcode('shortcode', null)`. I don't know if that's the right approach since it is the odd one out for not having a not-null constraint on portal_id.

core/src/main/resources/db/changelog/changesets/2024_03_29_portal_scoped_stable_ids.yaml
```diff
@@ -0,0 +1,76 @@
+databaseChangeLog:
+  - changeSet:
+      id: "portal_scoped_stable_ids"
+      author: connorlbark
+      changes:
+        - dropForeignKeyConstraint:
```
devonbush: I suspect they're funcitonally identical, but for hygiene
```suggestion
        - dropUniqueConstraint:
```

core/src/main/resources/db/changelog/changesets/2024_03_29_portal_scoped_stable_ids.yaml
devonbush: that makes sense -- we've thought about having a guaranteed to exist "juniper" or "internal" Portal where we could hang templates like this.

devonbush (approved):

devonbush:
looks good!

core/src/main/resources/db/changelog/changesets/2024_03_29_portal_scoped_stable_ids.yaml
connorlbark: Should that `internal` portal be a ticket or is this null portal good?

core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
```diff
@@ -338,6 +358,39 @@ protected List<T> findAllByTwoProperties(String column1Name, List<?> column1Valu
...
+    protected List<T> findAllByThreeProperties(String column1Name, List<?> column1Values,
+                                               String column2Name, List<?> column2Values,
+                                               String column3Name, List<?> column3Values) {
+        if (column1Values.size() != column2Values.size()) {
+            throw new IllegalArgumentException("column1Values and column2Values must be the same size");
+        }
+        if (column1Values.isEmpty()) {
+            // short circuit this case because bindList errors if list is empty
+            return new ArrayList<>();
+        }
```
MatthewBemis: Do these error-checks need to be updated to handle 3 value lists?

core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
```diff
@@ -338,6 +358,39 @@ protected List<T> findAllByTwoProperties(String column1Name, List<?> column1Valu
         });
     }
 
+    protected List<T> findAllByThreeProperties(String column1Name, List<?> column1Values,
```
MatthewBemis, line 361: Out of scope for this PR, but I do think we should consider using varargs for these types of DAO methods so we can handle an arbitrary number of properties, something like this (this code hasn't been thoroughly tested, I just had Copilot mock something up quickly):

```
public record ColumnProperty(String columnName, List<?> columnValues) {}

protected List<T> findAllByProperties(ColumnProperty... properties) {
    if (Arrays.stream(properties).map(p -> p.columnValues().size()).distinct().count() > 1) {
        throw new IllegalArgumentException("columnValues must all be the same size");
    }
    if (Arrays.stream(properties).anyMatch(p -> p.columnValues().isEmpty())) {
        // short circuit this case because bindList errors if list is empty
        return new ArrayList<>();
    }

    String whereClause = IntStream.range(0, properties[0].columnValues().size())
            .mapToObj(i -> Arrays.stream(properties)
                    .map(p -> "%s = :%sValue%d and ordering = %d".formatted(p.columnName(), p.columnName(), i, i))
                    .collect(Collectors.joining(" and ")))
            .collect(Collectors.joining(" or "));

    String sortValues = IntStream.range(0, properties[0].columnValues().size())
            .mapToObj(i -> Arrays.stream(properties)
                    .map(p -> "(:%sValue%d, %d)".formatted(p.columnName(), i, i))
                    .collect(Collectors.joining(", ")))
            .collect(Collectors.joining(", "));

    String joinClause = """
        join (values %s) as sortJoin %s
        on %s
        """.formatted(sortValues,
            Arrays.stream(properties)
                    .map(p -> "(%s, ordering)".formatted(p.columnName()))
                    .collect(Collectors.joining(", ")),
            Arrays.stream(properties)
                    .map(p -> "%s = sortJoin.%s".formatted(p.columnName(), p.columnName()))
                    .collect(Collectors.joining(" and ")));

    return jdbi.withHandle(handle -> {
        Query query = handle.createQuery("select * from %s %s where %s order by sortJoin.ordering"
                .formatted(tableName, joinClause, whereClause));
        for (int i = 0; i < properties[0].columnValues().size(); i++) {
            for (ColumnProperty p : properties) {
                query = query.bind("%sValue%d".formatted(p.columnName(), i), p.columnValues().get(i));
            }
        }
        return query.mapTo(clazz).list();
    });
}
```

MatthewBemis (approved):
Just a couple of small comments/suggestions

core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
connorlbark: I like this idea a lot - I vote we make it a ticket to do. We could then do like Map.of/Map.ofEntries, and have helper methods like:

```
findAllByTwoProperties(String column1, Object val1, String column2, Object val2 ) {
     return findAllByProperties(new ColumnProperty(column1, val1), new ColumnProperty(column2, val2)
}
```

etc. so it's fully backwards compatible. Map.of has this up to 10 entries long, which could be cool to have that flexibility for our own DAO methods.

PR #829, opened 2024-04-01
[JN-980] save preferred language during registration

MatthewBemis:
#### DESCRIPTION

This sets the preferredLanguage profile field upon registration. Previously, we did not honor a user's preferred language immediately after registration (they would have to re-select it to actually set the profile field). This resulted in some of the content being loaded in the default language (english) and they'd also receive the welcome email in english.

* In the *demo* study, select a language other than English
* Register a new user (try this twice: once with b2c enabled and once with it disabled - it should work for both)
* Verify that when you're redirected to the dashboard, the content is in the language that you originally selected prior to the b2c registration screens
* Also confirm that you receive the welcome email in the correct language

### DIFF
--- a/api-participant/src/main/resources/api/openapi.yml
+++ b/api-participant/src/main/resources/api/openapi.yml
@@ -377,6 +377,7 @@ paths:
       parameters:
         - { name: portalShortcode, in: path, required: true, schema: { type: string } }
         - { name: envName, in: path, required: true, schema: { type: string } }
+        - { name: preferredLanguage, in: query, required: false, schema: { type: string } }
         - { name: preRegResponseId, in: query, required: false, schema: { type: string, format: uuid } }
       requestBody:
         content:
--- a/api-participant/src/main/resources/api/openapi.yml
+++ b/api-participant/src/main/resources/api/openapi.yml
@@ -357,6 +357,7 @@ paths:
       parameters:
         - { name: portalShortcode, in: path, required: true, schema: { type: string } }
         - { name: envName, in: path, required: true, schema: { type: string } }
+        - { name: preferredLanguage, in: query, required: false, schema: { type: string } }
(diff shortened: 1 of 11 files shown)


ui-participant/src/api/api.tsx
```diff
@@ -289,13 +290,12 @@ export default {
...
+  async register({ preRegResponseId, email, accessToken, preferredLanguage }: {
+    preRegResponseId: string | null, email: string, accessToken: string, preferredLanguage: string | null
+  }): Promise<LoginResult> {
     bearerToken = accessToken
-    let url = `${baseEnvUrl(false)}/register`
-    if (preRegResponseId) {
-      url += `?preRegResponseId=${preRegResponseId}`
-    }
+    const params = queryString.stringify({ preRegResponseId, preferredLanguage })
+    const url = `${baseEnvUrl(false)}/register?${params}`
```
devonbush, line 298: nice cleanups with stringify

api-participant/src/main/resources/api/openapi.yml
```diff
@@ -357,6 +357,7 @@ paths:
       parameters:
         - { name: portalShortcode, in: path, required: true, schema: { type: string } }
         - { name: envName, in: path, required: true, schema: { type: string } }
+        - { name: preferredLanguage, in: query, required: false, schema: { type: string } }
```
devonbush, line 360: This seems to me like something that belongs in the body, rather than the url, but it could go either way

devonbush (approved):
Looks good to me!  One suggestion that could go either way.

connorlbark (approved):
looks great! outside of the scope, but I noticed the verification email was still in english - is that expected? everything else B2c was spanish and worked super well.

MatthewBemis:
@connorlbark yeah the B2c verification code email is out of scope. that work is captured in https://broadworkbench.atlassian.net/browse/JN-959

PR #831, opened 2024-04-02
[JN-794] Fix dangling foreign key when deleting notifications

MatthewBemis:
#### DESCRIPTION

Fixes a dangling foreign key reference that caused repopulate to fail

* Repopulate demo
* Ensure at least one sendgrid_event references a notification in demo (you can just set the notification_id for a sendgrid_event manually)
* Repopulate demo, confirm it doesn't through a SQL exception


devonbush (approved):
Looks good.  As we talked about, don't merge this until we've had a chance to test the exception truncation fix

connorlbark (approved):

PR #832, opened 2024-04-03
[JN-917] truncating long exceptions

devonbush:
#### DESCRIPTION

Exceptions were being logged as string blobs because they were longer than Azure's 16K limit.  This truncates long exceptions to stay under the limit.  this caused the populate errors during team testing today to not generate log alerts

1. after this is merged, try repopulating demo (prior to merging Matt's fix) on the demo server
2. confirm exception is formed as a proper error and logged to azure/slack


MatthewBemis (approved):

connorlbark (approved):
code looks good :shipit:

PR #833, opened 2024-04-03
[JN-969] fix portal icon

connorlbark:
#### DESCRIPTION

Changing the favicon does in fact update the participant website, but it did not reflect on the admin side because they were hardcoded to look for the V1 of the media. I changed the method/endpoint to be able to get the latest version so that the admin side could be more responsive.

- Repopulate ourhealth
- Upload new favicon
- Observe that it is reflected in the admin tools in all the appropriate places as well as the participant side


devonbush (approved):
nice fix!  Now that I think about it, we probably should have favicon be configured on SiteContent to match how we configure everything else, and so people can test favicon changes.  But that would be a big change requiring upgrade scripts and populate/extract changes, so I like how you've done this

MatthewBemis (approved):

PR #834, opened 2024-04-04
[JN-904] sign in for unenrolled users logic

devonbush:
#### DESCRIPTION

This updates our oauth redirect to better handle cases where a user signs into a portal that only has a single study accepting enrollment -- it will now auto-enroll them.

This also adds logic to check if a participant came in via an invite, in which case it will not attempt to register them, even though their b2c account is new, since they will already have a participantUser created.

This also substantially cleans up some of the redirect logic, and adds tests.  The tests were a huge pain to write, and I tried adding more, but failed hard because my jest-mock skills weren't up to it.  I couldn't figure out good ways to use jest.mock (which has to be at the root level of the file) but have the mock change between tests

1. repopulate demo portal 
2.  launch participant ui with dev-mode login enabled
3.  log into the participant ui as `unenrolled@test.com`
4. confirm you get enrolled in the study
5. Try lots of different sign-in/enroll/registration combos to check for regressions


ui-participant/src/login/LoginUnauthed.tsx
```diff
@@ -1,22 +1,38 @@
...
     setIsError(false)
-    Api.unauthedLogin(emailAddress).then((result: LoginResult) => {
-      loginUser(result, result.user.token)
-    }).catch(() => {
+    try {
+      const loginResult = await Api.unauthedLogin(emailAddress)
+      loginUser(loginResult, loginResult.user.token)
+
+      // Enroll in the default study if not already enrolled
+      if (defaultEnrollStudy && !userHasJoinedStudy(defaultEnrollStudy, loginResult.enrollees)) {
```
connorlbark: should it also make sure that there isn't an enrollee that exists for any study? I could see in the future weird edge cases where an enrollee has joined an "unjoinable" study then all of the sudden get added to a different study because of this logic.

connorlbark (approved):
works well for me!

ui-participant/src/login/LoginUnauthed.tsx
devonbush: that's a good point, will update

MatthewBemis (approved):

PR #837, opened 2024-04-05
[JN-943] Proxy profile mappings

connorlbark:
#### DESCRIPTION

Allows you to add proxy profile questions as answer mappings.

Also, refactors how pre enrollment surveys work. During enrollment, those answers are backfilled so that we can use all the same answer mapping / etc infrastructure that all other surveys use. This enables the ability to use answer mappings in pre enrollment surveys. It also allows us to simplify how these are displayed in the UI, as there doesn't need to be any parsing of the "fullData" field - they're just totally normal survey responses now. 

- Repopulate demo
- Register new user as proxy, fill out governed user and proxy names in pre enroll survey
- Observe that these names are reflected properly in the database / admin tool

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/AnswerProcessingService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/AnswerProcessingService.java
@@ -100,7 +100,7 @@ public class AnswerProcessingService {
         if (proxyProfileMappings.isEmpty() || !hasTargetedChanges(proxyProfileMappings, answers, AnswerMappingTargetType.PROXY_PROFILE)) {
             return;
         }
-        // grab the operator (or the proxy's) profile to update it
+        // grab the operator (which is the proxy) profile to update it
         Profile profile = profileService.loadWithMailingAddress(operator.getProfileId()).get();
         mapValuesToType(
                 answers,
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/AnswerProcessingService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/AnswerProcessingService.java
@@ -107,6 +107,7 @@ public class AnswerProcessingService {
                 proxyProfileMappings,
                 profile,
                 AnswerMappingTargetType.PROXY_PROFILE);
+        profileService.updateWithMailingAddress(profile, auditInfo);
     }
 
     /**
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/AnswerProcessingService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/AnswerProcessingService.java
@@ -107,8 +107,6 @@ public class AnswerProcessingService {
                 proxyProfileMappings,
                 profile,
                 AnswerMappingTargetType.PROXY_PROFILE);
-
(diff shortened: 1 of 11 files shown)


core/src/main/java/bio/terra/pearl/core/service/survey/AnswerProcessingService.java
```diff
@@ -47,33 +48,68 @@ public AnswerProcessingService(ObjectMapper objectMapper, ProfileService profile
...
+            List<Answer> answers,
+            List<AnswerMapping> mappings,
+            PortalParticipantUser operator,
+            DataAuditInfo auditInfo) {
+        List<AnswerMapping> proxyProfileMappings = mappings.stream().filter(mapping ->
+                mapping.getTargetType().equals(AnswerMappingTargetType.PROXY_PROFILE)).toList();
+        if (proxyProfileMappings.isEmpty() || !hasTargetedChanges(proxyProfileMappings, answers, AnswerMappingTargetType.PROXY_PROFILE)) {
+            return;
+        }
+        // grab the operator (or the proxy's) profile to update it
```
devonbush:
```suggestion
        // grab the operator (which is the proxy) profile to update it
```

core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
```diff
@@ -11,67 +11,72 @@
...
+    private final PortalParticipantUserService portalParticipantUserService;
+    private final ParticipantUserService participantUserService;
+    private final EnrolleeService enrolleeService;
+    private final EnrolleeRelationService enrolleeRelationService;
+    private final RegistrationService registrationService;
+    private final EventService eventService;
+    private final ObjectMapper objectMapper;
+    private final AnswerMappingDao answerMappingDao;
+    private final SurveyResponseService surveyResponseService;
+    private final AnswerProcessingService answerProcessingService;
```
devonbush, line 65: thanks for updating all to final

core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
```diff
@@ -154,6 +167,49 @@ public HubResponse<Enrollee> enroll(EnvironmentName envName, String studyShortco
         return hubResponse;
     }
 
+    private void backfillPreEnrollResponse(PortalParticipantUser operator, Enrollee enrollee, PreEnrollmentResponse preEnrollResponse) {
```
devonbush, line 170: this is very cool.

devonbush (approved):
This is awesome -- great approach to backfill preEnrollment into answers.  One thing to consider is that previously, SurveyResponses always corresponded to a ParticipantTask and a Survey with a StudyEnvironmentSurvey.  That's no longer true, which is (I think) fine for now.  But we might want to (in a later ticket) think about:

1. whether we should create a ParticipantTask for preEnrollment
2. whether we want to export pre-enrollment answers. 
3. related to the above, whether we want to make PRE_ENROLL a type of Survey, so that the pre-enrollment survey doesn't need a special place in the StudyEnvironment data model.  (I think if we did this, we'd get export working for free)

MatthewBemis (approved):

PR #839, opened 2024-04-05
[JN-894] refactoring consents to be surveys

devonbush:
#### DESCRIPTION

The plan for replacing ConsentForms with Surveys:

1. this PR contains code so that Surveys can be created with type "CONSENT" and they will behave exactly like consent forms do.  This PR does not break any existing consent form functionality, except it removes the button to create new ConsentForms in the admin tool, replacing it with one that creates Surveys of type CONSENT
2. This PR also contains code to migrate consent forms and responses to surveys and surveyResponses (but not deleting the existing forms and responses, just making them invisible to the study staff/participants)
3. Once this PR is released, we can use the new populate command infrastructure to execute the above migration during a scheduled downtime
4. After the migration everything should work, and then we can submit a follow-up PR deleting the tables and entities and consent code at our convenience

1. redeploy ApiAdminApp and ApiParticipantApp
2. confirm consent functionality, view, print, complete, etc.. still works as normal
3. go to `https://localhost:3000/populate/runCommand` and type the command "CONVERT_CONSENTS"
image

5. confirm consent functionality, view, print, complete, etc.. still works as normal
6. The only noticeable difference will be the admin tool Forms & Surveys view.  Prior to migration there will be an extra "none" below the consent forms, and after migration that will be gone and the updated survey actions table will appear
BEFORE:
image
AFTER:
image


MatthewBemis (approved):
Looks great, and worked great when I tried it out. Glad this ended up being cleaner than expected!

ui-admin/src/study/StudyContent.tsx
```diff
@@ -119,15 +123,27 @@ function StudyContent({ studyEnvContext }: {studyEnvContext: StudyEnvContextT})
...
-                  { !isReadOnlyEnv && <button className="btn btn-secondary" data-testid={'addConsent'} onClick={() => {
-                    setShowCreateConsentModal(!showCreateConsentModal)
-                  }}>
-                    <FontAwesomeIcon icon={faPlus}/> Add
-                  </button> }
-                </li>
               </ul>
             </div>
+            <div className="flex-grow-1 pt-3">
+              <SurveyEnvironmentTable stableIds={consentSurveyStableIds}
```
connorlbark, line 129: I love that it "Just Works™" 😄

ui-admin/src/study/surveys/CreateSurveyModal.tsx
```diff
@@ -77,7 +80,7 @@ const CreateSurveyModal = ({ studyEnvContext, onDismiss, type }:
 
   return <Modal show={true} onHide={onDismiss} className={type === 'OUTREACH' ? 'modal-lg' : 'modal'}>
     <Modal.Header closeButton>
-      <Modal.Title>Create New {type === 'RESEARCH' ? 'Research Survey' : 'Outreach'}</Modal.Title>
+      <Modal.Title>Create new {type.toLowerCase()}</Modal.Title>
```
connorlbark: nit: I find "Create new research" an awkward phrase - I think we should keep "survey" in there for at least research surveys, if not all of them.

connorlbark (approved):
works great! very easy to convert and completely seamless from the user experience.

PR #841, opened 2024-04-09
[JN-863] Add defaultLanguage to PortalEnvironment

MatthewBemis:
#### DESCRIPTION

This adds a new defaultLanguage property to the PortalEnvironmentConfig. Study admins can view and set this setting in the "Site Settings" page.

* Restart everything
* Repopulate demo
* Demo sandbox environment is set to Spanish by default now, so test the following:
* Site content editor loads in Spanish: https://localhost:3000/demo/studies/heartdemo/env/sandbox/siteContent
* Email templates load in Spanish: https://localhost:3000/demo/studies/heartdemo/env/sandbox/notificationContent
* Form Preview tab loads in Spanish (the designer will still load in English- that's because of how the SurveyJS JSON is modeled): https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms/surveys/hd_hd_basicInfo
* Also try changing the default language in sandbox and confirm everything still works as expected: https://localhost:3000/demo/studies/heartdemo/env/sandbox/settings

Now, as a participant:
* Load Demo sandbox participant site: https://sandbox.demo.localhost:3001/
* Delete the `selectedLanguage` entry from local storage
* Refresh the page, it should load in Spanish

### DIFF
--- a/ui-admin/src/App.tsx
+++ b/ui-admin/src/App.tsx
@@ -48,7 +48,7 @@ function App() {
       <ConfigConsumer>
         { config =>
           <AuthProvider {...getOidcConfig(config.b2cTenantName, config.b2cClientId, config.b2cPolicyName)}>
-            <I18nProvider>
+            <I18nProvider defaultLanguage={'en'}>
               <UserProvider>
                 <div className="App d-flex flex-column min-vh-100">
                   <IdleStatusMonitor maxIdleSessionDuration={30 * 60 * 1000} idleWarningDuration={5 * 60 * 1000}/>
--- a/ui-admin/src/portal/useDefaultPortalLanguage.ts
+++ b/ui-admin/src/portal/useDefaultPortalLanguage.ts
@@ -1,10 +1,10 @@
 import { useContext } from 'react'
 import { PortalContext } from './PortalProvider'
 import { useParams } from 'react-router-dom'
-import { StudyParams } from '../study/StudyRouter'
+import { StudyParams } from 'study/StudyRouter'
 
 /**
- *
+ * Returns the default language for the current portal environment.
  */
 export function useDefaultLanguage() {
   const { portal } = useContext(PortalContext)
--- a/ui-admin/src/portal/useDefaultPortalLanguage.ts
+++ b/ui-admin/src/portal/useDefaultPortalLanguage.ts
@@ -1,0 +1,28 @@
+import { useContext } from 'react'
+import { PortalContext } from './PortalProvider'
+import { useParams } from 'react-router-dom'
+import { StudyParams } from '../study/StudyRouter'
+
+/**
+ *
+ */
+export function useDefaultLanguage() {
+  const { portal } = useContext(PortalContext)
+  const params = useParams<StudyParams>()
+  const envName: string | undefined = params.studyEnv
+
+  const defaultLanguage = portal?.portalEnvironments.find(env =>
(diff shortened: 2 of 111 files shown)


ui-admin/src/App.tsx
```diff
@@ -48,7 +48,7 @@ function App() {
       <ConfigConsumer>
         { config =>
           <AuthProvider {...getOidcConfig(config.b2cTenantName, config.b2cClientId, config.b2cPolicyName)}>
-            <I18nProvider>
+            <I18nProvider defaultLanguage={'en'}>
```
MatthewBemis, line 51: I think there's some more work to do here, but I wasn't ready to include it in this PR. I'd like to see if we can have a common PortalProvider interface that can be used across the ui-admin and ui-participant, kinda like how we have the common ApiProvider. Without this, the I18nProvider needs to be handled slightly differently depending on the context that it's in

ui-admin/src/portal/useDefaultPortalLanguage.ts
```diff
@@ -0,0 +1,28 @@
...
+import { PortalContext } from './PortalProvider'
+import { useParams } from 'react-router-dom'
+import { StudyParams } from 'study/StudyRouter'
+
+/**
+ * Returns the default language for the current portal environment.
+ */
+export function useDefaultLanguage() {
+  const { portal } = useContext(PortalContext)
+  const params = useParams<StudyParams>()
```
devonbush: we have a helper for this -- `useStudyEnvParamsFromPath`

devonbush (approved):
thanks for updating 111 files.  :).  One small suggestion to use a helper method

ui-core/src/participant/I18nProvider.tsx
```diff
@@ -32,12 +32,15 @@ const SELECTED_LANGUAGE_KEY = 'selectedLanguage'
...
+export function I18nProvider({ defaultLanguage, portalShortcode, children }: {
+  defaultLanguage: string, portalShortcode?: string, children: React.ReactNode
+}) {
   const Api = useApiContext()
   const [isLoading, setIsLoading] = useState(true)
   const [isError, setIsError] = useState(false)
   const [languageTexts, setLanguageTexts] = useState<Record<string, string>>({})
-  const [selectedLanguage, setSelectedLanguage] = useState(localStorage.getItem(SELECTED_LANGUAGE_KEY) || 'en')
+  const [selectedLanguage, setSelectedLanguage] = useState(
+    localStorage.getItem(SELECTED_LANGUAGE_KEY) || defaultLanguage)
```
connorlbark, line 43: Nit: The backend can take `null` values now, right? Should the i18n provider just send null instead of a pre-set default? I assume, in the future, we're going to have more internationalized admin participant-preview stuff, so I could see the `defaultLanguage='en'` being the cause of some confusing bugs, and, especially since most portals are english, it could be easy to overlook.

connorlbark (approved):
Worked great!

ui-core/src/participant/I18nProvider.tsx
MatthewBemis: That's a great point- the backend _can_ take nulls for portal-related endpoints, but not for the language texts. Since most of those are global, there's not really an equivalent concept of defaultLanguage, which is why I've just settled on English. Fortunately the number of places where this is hardcoded are minimal but I hear you that it would be better to have _none_

PR #842, opened 2024-04-09
[JN-985] Add survey autocomplete button to irb and sandbox environments

MatthewBemis:
#### DESCRIPTION

I waste wayyyy too much time manually filling out preenroll and consent forms while developing. This adds a new button to the top-right of the form view to automatically fill out every question in the form with one click. Only available in sandbox and irb environments.

It probably doesn't handle every question type properly, and might insert things into the survey response that are a bit nonsensical based on the question type/question requirements, but it works well enough for the existing Heart Demo surveys which was my main motivation, and it's really just a developer productivity thing anyway.

Screenshot 2024-04-09 at 10 19 47 PM

* Repopulate Heart Demo
* Register as a new user on sandbox or irb
* Click the auto-complete button on the pre-enroll form
* Verify that the form is filled out and that you're navigated to the last page of the form
* Try doing the same on a live environment, and verify that you don't see the auto-complete button (or the review mode button, since I brought that into the consent and pre-enroll views)

### DIFF
--- a/ui-participant/src/hub/consent/ConsentView.tsx
+++ b/ui-participant/src/hub/consent/ConsentView.tsx
@@ -98,11 +98,9 @@ function RawConsentView({ form, enrollee, resumableData, pager, studyShortcode, 
 }
 
 /** handles paging the form */
-function PagedConsentView({ form, responses, enrollee, studyShortcode }:
-                              {
-                                form: StudyEnvironmentConsent, responses: ConsentResponse[], enrollee: Enrollee,
-                              studyShortcode: string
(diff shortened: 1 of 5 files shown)


devonbush (approved):
YES!  Where has this been all my life?!?!?!  I tested it on OurHealth, smooth as butter

connorlbark (approved):
Matt... this is probably the best thing that has ever happened. I genuinely jumped up in the air when I tried it for the first time. This is awesome. Sometimes it really is the little things in life haha

PR #843, opened 2024-04-10
[JN-987] Survey designer plays nicely with internationalized surveys

MatthewBemis:
#### DESCRIPTION

The survey designer will now display the content from the `default` language instead of `[object Object]`.

Other languages cannot be viewed or edited in the designer, but they can be previewed in the "Preview" tab. If you try to edit the value of an element that's internationalized, it will work for the `default` language, but will revert to an un-internationalized version. I think this is okay for now.

* Repopulate demo
* Load the Basics survey and confirm that you see the actual text values and that there aren't any [object Object]s in the text boxes: https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms/surveys/hd_hd_basicInfo


ui-core/src/types/forms.ts
```diff
@@ -111,6 +111,19 @@ export type FormContentPage = BaseElement & {
   elements: FormElement[]
 }
 
+/**
+ *  Certain SurveyJS elements can take on multiple forms. For example, the "title" field
+ *  for a question could either be a string, or an object mapping language codes to strings.
+ *  "default" is always present; the other languages are arbitrary.
+ */
+export type I18nMapElement = {
```
MatthewBemis: I don't like this name, but I haven't yet decided what I'm going to rename it to

ui-admin/src/util/juniperSurveyUtils.tsx
```diff
@@ -141,3 +142,15 @@ export function questionFromRawText(rawText: string): QuestionObj {
 ReactQuestionFactory.Instance.registerQuestion('addressvalidation', props => {
   return React.createElement(SurveyQuestionAddressValidation, props)
 })
+
+/**
+ * Takes in an I18nSurveyElement and either returns the default value (if I18nSurveyElement is in the
+ * internationalized format), or the string itself (if I18nSurveyElement is a simple string).'
+ * Once the survey designer is fully
+ */
+export const getI18nSurveyElement = (element?: I18nSurveyElement, language?: string) => {
```
devonbush: making the argument default cleans up the ternary below
```suggestion
export const getI18nSurveyElement = (element?: I18nSurveyElement, language = 'default') => {
```

ui-core/src/types/forms.ts
devonbush: I would just call this `I18nMap`

ui-admin/src/util/juniperSurveyUtils.tsx
devonbush: I might also rename this to `i18nSurveyText` to be a bit shorter

devonbush (approved):
works well locally.  Just some style suggestions

connorlbark (approved):

PR #844, opened 2024-04-10
[JN-997] populate_portal.sh can accept multiple portal shortcodes

MatthewBemis:
#### DESCRIPTION

Just a quick dev QOL tweak that allows you to specify multiple portal shortcodes when running the populate script

Confirm that the single case still works `./scripts/populate_portal.sh ourhealth` still works
Confirm that the multi-portal case works `./scripts/populate_portal.sh ourhealth hearthive demo`


devonbush (approved):
I see you are wise in the ways of shell...

wluken (approved):

PR #846, opened 2024-04-12
[JN-1000] Export proxy relationships

connorlbark:
#### DESCRIPTION

Adds proxy relations to the data export. In addition, you can now request to include proxies as distinct rows in the data export, unchecked by default. If checked, then the `enrollee.subject` row will be included, otherwise it won't.

- Export data with and without enrollees and ensure data looks as expected

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/export/ExportOptions.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/export/ExportOptions.java
@@ -1,18 +1,25 @@
 package bio.terra.pearl.core.service.export;
 
-import lombok.Builder;
+import lombok.Getter;
+import lombok.experimental.SuperBuilder;
 
-public record ExportOptions(
-        boolean splitOptionsIntoColumns,
-        boolean stableIdsForOptions,
-        boolean onlyIncludeMostRecent,
-        boolean includeProxiesAsRows,
-        ExportFileFormat fileFormat,
-        Integer limit) {
+
+@SuperBuilder
+@Getter
+public final class ExportOptions {
+    private final boolean splitOptionsIntoColumns;
+    private final boolean stableIdsForOptions;
+    private final boolean onlyIncludeMostRecent;
+    private final boolean includeProxiesAsRows;
+    private final ExportFileFormat fileFormat;
+    private final Integer limit;
+
     public ExportOptions() {
-        this(false, false, true, false, ExportFileFormat.TSV, null);
+        this.splitOptionsIntoColumns = false;
+        this.stableIdsForOptions = false;
+        this.onlyIncludeMostRecent = true;
+        this.includeProxiesAsRows = false;
+        this.fileFormat = ExportFileFormat.TSV;
+        this.limit = null;
     }
-
-    @Builder
-    public ExportOptions {}
 }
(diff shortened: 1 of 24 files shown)


core/src/main/java/bio/terra/pearl/core/service/export/ExportOptions.java
```diff
@@ -1,13 +1,16 @@
 package bio.terra.pearl.core.service.export;
 
-import bio.terra.pearl.core.service.export.ExportFileFormat;
 import lombok.Builder;
 
-public record ExportOptions (boolean splitOptionsIntoColumns, boolean stableIdsForOptions, boolean onlyIncludeMostRecent,
-                             ExportFileFormat fileFormat,
-                             Integer limit) {
+public record ExportOptions(
```
devonbush: this constructor is getting unwieldy with booleans.  How about we convert this to a class so we can have defaults and use a lombok builder?

core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/EnrolleeFormatter.java
```diff
@@ -11,11 +10,16 @@
...
     public static final String ENROLLEE_MODULE_NAME = "enrollee";
-    public static final List<String> INCLUDED_PROPERTIES = List.of("shortcode", "consented", "createdAt");
+    public static final List<String> INCLUDED_PROPERTIES = List.of(
+            "shortcode", "consented", "createdAt"
+    );
 
     public EnrolleeFormatter(ExportOptions exportOptions) {
         itemFormatters = INCLUDED_PROPERTIES.stream().map(propName -> new PropertyItemFormatter<Enrollee>(propName, Enrollee.class))
                 .collect(Collectors.toList());
+        if (exportOptions.includeProxiesAsRows()) {
```
devonbush: Let's always include this column.  Even though most of the time it will always just be "true" it's a useful reminder

core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/EnrolleeRelationFormatter.java
```diff
@@ -0,0 +1,44 @@
...
+        itemFormatters = INCLUDED_PROPERTIES.stream()
+                .map(propName -> new PropertyItemFormatter<>(propName, EnrolleeRelation.class))
+                .collect(Collectors.toList());
+        // we have to handle kitType separately because we'll need to match it to the kitType name
+        moduleName = MODULE_NAME;
+        displayName = "Proxy Relations";
+    }
+
+    @Override
+    public Map<String, String> toStringMap(EnrolleeExportData enrolleeData) {
```
devonbush, line 30: let's add a comment at the top of this method that this is a near-duplicate of toStringMap in KitRequestFormatter so that if we end up with another formatter like this, we can extract helper methods

core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/EnrolleeRelationFormatter.java
```diff
@@ -0,0 +1,44 @@
...
+        List<EnrolleeRelation> relations = enrolleeData.getEnrolleeRelations();
+        // sort the kits oldest first
```
devonbush: copypasta

devonbush (approved):
Looks good!  a couple of small suggestions, and we definitely want to make ExportOptions have a proper builder, but it's ok by me if you want a separate PR for that

MatthewBemis (approved):

PR #847, opened 2024-04-12
[JN-1002] Fix admin welcome emails

MatthewBemis:
#### DESCRIPTION

Quick fix for admin welcome emails, which would throw an exception:

```
Unexpected exception occurred invoking async method:
public void bio.terra.pearl.core.service.notification.email.AdminEmailService.sendWelcomeEmail(bio.terra.pearl.core.model.portal.Portal,bio.terra.pearl.core.model.admin.AdminUser)
java.util.NoSuchElementException: No value present
at java.base/java.util.Optional.get(Optional.java:143)
```

The query generated by `findByThreeProperties` doesn't handle null values correctly. This could also be fixed directly in the BaseJdbiDao methods, but I didn't want to assume that this was the intended behavior for everything, and it would also introduce a lot of inconsistency with the other DAO methods unless we refactored all of the queries.

* Restart Admin Api
* Try adding a new admin user: https://localhost:3000/demo/studies/heartdemo/users
* Confirm that you get the welcome email

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/notification/EmailTemplateDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/notification/EmailTemplateDao.java
@@ -4,6 +4,7 @@ import bio.terra.pearl.core.dao.BaseVersionedJdbiDao;
 import bio.terra.pearl.core.model.notification.EmailTemplate;
 
 import java.util.List;
+import java.util.Optional;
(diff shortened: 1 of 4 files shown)


devonbush (approved):
nice, yeah, better handling of nulls in queries is something we might be able to get for free as we move to JOOQ

connorlbark (approved):

PR #853, opened 2024-04-19
[JN-1010] Add proxy enrollment flag

connorlbark:
#### DESCRIPTION

Adds flag which enables and disables proxy enrollment. Right now, it just throws an error if you try to enroll as proxy with it disabled, but future features (add new participant button, etc.) will be dependent on this.

- Enable and disable proxy enrollment
- When enabled, ensure proxy enrollment flow is normal
- When disable, ensure error is thrown when attempt to enroll as proxy (this shouldn't ever happen unless a study staff leaves the proxy question erroneously. It does leave the user in a hanging state where they are registered but have no enrollees, but since this all happens post-registration I can't imagine any other expected behavior.)

### DIFF
--- a/ui-admin/src/study/StudySettings.tsx
+++ b/ui-admin/src/study/StudySettings.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react'
 import { StudyEnvContextT } from './StudyEnvironmentRouter'
-import { LoadedPortalContextT } from '../portal/PortalProvider'
-import PortalEnvConfigView from '../portal/PortalEnvConfigView'
+import { LoadedPortalContextT } from 'portal/PortalProvider'
+import PortalEnvConfigView from 'portal/PortalEnvConfigView'
 import { PortalEnvironment } from '@juniper/ui-core'
 import { doApiLoad } from 'api/api-utils'
 import { Button } from 'components/forms/Button'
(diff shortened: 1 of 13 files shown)


ui-admin/src/study/StudySettings.tsx
```diff
@@ -76,6 +76,16 @@ export function StudyEnvConfigView({ studyEnvContext, portalContext }:
       </label>
     </div>
 
+    <div>
+      <label className="form-label">
+        accepting proxy enrollment <InfoPopup content={
```
devonbush: can you start a zendesk page on proxy enrollment?  You'll probably just want to link that page, since the proxy handling will be up in the air for a while

devonbush (approved):

ui-admin/src/study/StudySettings.tsx
connorlbark: Good point - should this be behind the `prototype user` feature until we get it ready for production? I'm leaning towards making adding the zendesk a TODO as so much is in flight that this page would probably need to be rewritten by the end.
Edit: I decided to put this behind a prototype user flag so it won't be visible and I made a ticket to add a zendesk article here.

MatthewBemis (approved):

PR #857, opened 2024-04-23
[JN-967] Add new participant functionality

connorlbark:
#### DESCRIPTION

Adds the functionality in place for an add new participant button, but does not add the button.

In addition, the code is multi-study compatible - with the code in place, you can:
- Enroll an existing governed user in a new study (will create new proxy enrollee if needed for this study env)
- Enroll a new governed user in a study you already proxy for
- Enroll a new governed user in a study you don't already proxy for (will create new proxy enrollee for this study env)

- Create a new participant in demo - with or without a proxy
- Go to https://sandbox.demo.localhost:3001/studies/heartdemo/join/preEnroll?isProxyEnrollment=true (For team testing: go to https://sandbox.demo.ddp-dev.envs.broadinstitute.org/studies/heartdemo/join/preEnroll?isProxyEnrollment=true)
- Fill out form as normal, note that the proxy question is disabled
- Submit and observe new governed user


ui-participant/src/hub/consent/ConsentView.tsx
```diff
@@ -78,6 +78,10 @@ function RawConsentView({ form, enrollee, resumableData, pager, studyShortcode,
           type: 'SUCCESS'
         }
       }
+
+      if (!response.enrollee.profile) {
+        response.enrollee.profile = enrollee.profile
+      }
       updateEnrollee(response.enrollee).then(() => {
```
connorlbark, line 85: quick fix: this updateEnrollee call during consent submission would strip the profile off the enrollee

ui-participant/src/studies/enroll/PreEnroll.tsx
```diff
@@ -65,10 +65,37 @@ export default function PreEnrollView({ enrollContext, survey }:
...
+  const disableProxyQuestion = async () => {
+    const proxyQuestion = await fetchProxyQuestion()
+    if (!proxyQuestion) {
+      return
+    }
+
+    // set proxy question to true by default so any proxy-specific questions are shown
+    surveyModel.setValue(proxyQuestion, 'true')
+    // disable proxy question so the user can't change it
+    surveyModel.getQuestionByName(proxyQuestion).readOnly = true
```
MatthewBemis: Survey autocomplete ignores the fact that the proxy question is read-only, so we might want to add something like `if (question.isReadOnly) { return }` above line 16 in `SurveyAutoCompleteButton.tsx`

ui-participant/src/studies/enroll/StudyEnrollRouter.tsx
```diff
@@ -13,20 +13,22 @@ import { useHasProvidedStudyPassword, usePreEnrollResponseId } from 'browserPers
 
 import { StudyEnrollPasswordGate } from './StudyEnrollPasswordGate'
 import { alertDefaults, AlertLevel } from '@juniper/ui-core'
-import { enrollCurrentUserInStudy } from '../../util/enrolleeUtils'
+import { enrollCurrentUserInStudy, enrollProxyUserInStudy } from '../../util/enrolleeUtils'
 import { logError } from '../../util/loggingUtils'
```
MatthewBemis:
```suggestion
import { enrollCurrentUserInStudy, enrollProxyUserInStudy } from 'util/enrolleeUtils'
import { logError } from 'util/loggingUtils'
```

api-participant/src/main/java/bio/terra/pearl/api/participant/service/EnrollmentExtService.java
```diff
@@ -0,0 +1,125 @@
...
+      ParticipantUser user,
+      PortalParticipantUser ppUser,
+      String studyShortcode,
+      EnvironmentName envName) {
+    return enrolleeService
+        .findByParticipantUserIdAndStudyEnv(ppUser.getParticipantUserId(), studyShortcode, envName)
+        .orElseGet(
+            () ->
+                enrollmentService
+                    .enroll(ppUser, envName, studyShortcode, user, ppUser, null, false)
```
MatthewBemis, line 122: for my own understanding: we only attach the pre-enrollment response if the enrollee is the subject?

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/survey/AnswerMappingController.java
```diff
@@ -0,0 +1,64 @@
...
+
+    AnswerMapping mapping =
+        answerMappingService
+            .findByTargetField(
+                portalShortcode,
+                stableId,
+                version,
+                AnswerMappingTargetType.valueOf(targetType),
+                targetField)
+            .orElseThrow(() -> new NotFoundException("Proxy question not found"));
```
MatthewBemis: This error message should probably be more generic since the endpoint itself isn't proxy-specific

MatthewBemis (approved):
Worked well for me. Just a few small comments

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/survey/AnswerMappingController.java
```diff
@@ -0,0 +1,64 @@
...
+      ObjectMapper objectMapper) {
+    this.answerMappingService = answerMappingService;
+    this.request = request;
+    this.requestUtilService = requestUtilService;
+    this.authUtilService = authUtilService;
+    this.objectMapper = objectMapper;
+  }
+
+  @Override
+  public ResponseEntity<Object> findAnswerMapping(
```
devonbush: I don't think we need this controller at all.  The pre-enrollment survey should contain the question show/hide logic within itself.   Assuming you add the "proxyProfile" surveyJs variable we talked about this afternoon, the preEnrollment survey can just have prepopulate/readOnly logic based on whether that proxyProfile variable exists.  See the `enableIf` property on surveyJs questions

ui-participant/src/studies/enroll/PreEnroll.tsx
```diff
@@ -65,10 +65,37 @@ export default function PreEnrollView({ enrollContext, survey }:
     })
   }
 
+  const fetchProxyQuestion = async (): Promise<string | null> => {
```
devonbush: as mentioned above, this is some clever code that I really hope we don't actually need

devonbush (changes requested):
Cool functionality!  Also, very much appreciate you split the functionality and UX into separate PRs.  

Requesting changes because I think there's a simpler way to accomplish making the proxy fields pre-populated/readonly, that will save us from having to create the answermapping controller.

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/survey/AnswerMappingController.java
connorlbark: ~Hmm, I see what you mean. I think the advantage of the answer mapping is that it minimizes the amount of "magic" json strings you had to include for proxy enrollment that required us to add documentation and might be points of confusion for study managers trying to add their own proxy enrollment questions, especially since this was a behavior I can't come up with a reason why someone would want to disable and might lead to some confusion when people forget to include them. But I can see how the surveyjs approach requires making fewer internal concepts visible, and eventually we would probably have a flow that sets all of this up for them anyway, so I can go ahead and remove it.~

Upon further reflection I have no objections, I think you're right we should lean on what surveyjs can do for us

api-participant/src/main/java/bio/terra/pearl/api/participant/service/EnrollmentExtService.java
connorlbark: Yes, this is accurate. That code/idea existed before I started working on proxies but I think the idea is that the proxy would only have a pre-enroll response if they enrolled themselves. That way, if the pre-enroll answers differed, or we needed to grab more information about the subject than just the name, then we could store those answers separately.

ui-participant/src/util/surveyJsUtils.tsx
```diff
@@ -81,16 +83,20 @@ export function useSurveyJSModel(
...
+    extraVariables = {}
   } = opts
 
   const { portalEnv } = usePortalEnv()
   const { i18n } = useI18n()
 
+  const { profile } = useActiveUser()
+  const { user, enrollees } = useUser()
+  const proxyProfile = enrollees.find(enrollee => enrollee.participantUserId === user?.id && enrollee.profile)?.profile
+
```
connorlbark, line 101: consequence of this: profile/proxyProfile are available on _all_ surveys by default, don't need to worry about passing it down

I also added the ability to pass more variables down. I felt the surveyjs was a lot easier to understand if `{isProxyEnrollment}` was passed down to the survey instead of not null checks on the proxy profile.

devonbush (approved):

PR #858, opened 2024-04-24
[JN-1026] pr template update

devonbush:
#### DESCRIPTION

updates PR to include screenshots

take a look, confirm language makes sense

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,4 +1,4 @@
-#### DESCRIPTION
+#### DESCRIPTION (include screenshots, and mobile screenshots for participant UX)
 
 
 


MatthewBemis (approved):

.github/pull_request_template.md
```diff
@@ -1,4 +1,4 @@
-#### DESCRIPTION
+#### DESCRIPTION (include screenshots, and mobile screenshots for participant UX)
```
connorlbark, line 1: Should we make this a whole new section? at least for me I would probably skip over it mentally but if it was a full section under description I would be less likely to skip the screenshots part. I can go either way tho so I'm gonna go ahead and approve.

connorlbark (approved):

.github/pull_request_template.md
devonbush: yeah, I was torn.  I don't like the idea of a separate screenshot section because I like it best when screenshots are integrated into the description.  But agree it could get glossed over.  Let's see how it works

PR #863, opened 2024-04-29
[JN-1035] UX for participant proxy flow

connorlbark:

Implements Erin's proxy workflow designs.

Screenshot 2024-05-01 at 11 57 14 AM
Screenshot 2024-05-01 at 11 57 19 AM
Screenshot 2024-05-01 at 11 57 26 AM
Screenshot 2024-05-01 at 11 57 29 AM

Note that you can no longer access your profile via the top right user dropdown if you have proxies.

Screenshot 2024-05-01 at 11 57 36 AM
Screenshot 2024-05-01 at 11 57 58 AM
Screenshot 2024-05-01 at 11 58 00 AM
Screenshot 2024-05-01 at 11 58 03 AM
Screenshot 2024-05-01 at 11 58 05 AM

- Sign up user with proxies
- Add more proxies via the drop down
- Select between them, fill out things on their behalf
- Go to manage profiles, edit profiles
- Do the same thing but on mobile
- Sign up normal user
- Observe no visual changes


ui-participant/src/Navbar.tsx
```diff
@@ -310,11 +310,11 @@ export const AccountOptionsDropdown = () => {
             {user.username}
           </p>
           <hr className="dropdown-divider d-none d-lg-block"/>
-          <NavLink to="/hub/profile">
+          {relations.length === 0 && <NavLink to="/hub/profile">
```
devonbush:
```suggestion
          {!relations.length && <NavLink to="/hub/profile">
```

devonbush: Let's put a link to the manage profiles page here too if there are > 0 relations.  I think it's nice to know that the profile editing capability is at least always accessible here, and then for proxy studies it's also accessible from the dashboard selector

ui-participant/src/participant/ManageProfiles.tsx
```diff
@@ -0,0 +1,60 @@
...
+        >
+          <div className='w-100 border-1 border-bottom p-3 text-center'>
+            <span className='fs-4 fw-bold'>{i18n('manageProfiles')}</span>
+          </div>
+          {ppUsers.map((ppUser, idx) => {
+            const name = useName(ppUser)
+            return (
+              <Link
+                to={`/hub/profile`}
+                state={{ ppUserId: ppUser.id }}
```
devonbush: using state for this will make it much harder to link or debug to give prod support.  Let's put the ppUserId in the path.  (it could be cool to use the enrolleeShortcode in the path, but that might result in some extra hassle in matching the shortcode to the ppUser)

ui-participant/src/participant/ManageProfiles.tsx
```diff
@@ -0,0 +1,60 @@
...
+                key={ppUser.id}
+                className={
```
devonbush: use the classNames helper here to clean up those ternaries/concat operations

ui-participant/src/util/enrolleeUtils.ts
```diff
@@ -51,3 +53,32 @@ export async function enrollProxyUserInStudy(
   await refreshLogin()
   return hubUpdate
 }
+
+/**
+ * Provides the name of the user (or 'You' if no name provided). If undefined, returns an empty string.
```
devonbush:
```suggestion
 * Provides the name of the user (or 'You' / 'Your dependent' if no name provided). If undefined, returns an empty string.
```

devonbush (changes requested):
Worked great when testing locally, nice use of i18n.  The change I'd like is to make profile pages routable, so that the url is different for different profiles -- this will help in both diagnosing prod errors, and in making test steps

populate/src/main/resources/seed/i18n/dev/languageTexts.json
```diff
@@ -67,6 +67,14 @@
   {"keyName": "suggestBetterAddressHeader", "text": "DEV_Is this the correct address?", "language": "dev"},
   {"keyName": "suggestBetterAddressBody", "text": "DEV_We were able to verify a similar address. Please select which you would like to use.", "language": "dev"},
   {"keyName": "newAddress", "text": "DEV_New Address", "language": "dev"},
-  {"keyName": "oldAddress", "text": "DEV_Old Address", "language": "dev"}
+  {"keyName": "oldAddress", "text": "DEV_Old Address", "language": "dev"},
+  {"keyName": "you", "text": "DEV_You", "language": "dev"},
+  {"keyName": "yourDependent", "text": "DEV_Your Dependent", "language": "dev"},
+  {"keyName": "youInParens", "text": "DEV_(you)", "language": "dev"},
```
MatthewBemis, line 73: As of https://github.com/broadinstitute/juniper/pull/865, the i18n helper supports substitutions. I wish I had added it sooner, but it could make some of these language texts a little more comprehensible and easier to deal with in certain languages, so we don't have to break these down to individual words. I could see this instead being something like `DEV_${participantGivenName} ${participantFamilyName} (you)`

Totally up to you though- I'm curious what you think is more readable.

connorlbark: Hmm... It would definitely be possible, but I think the logic is slightly more complex than a simple replace, so I might leave it the way it is. But this is very good to know in the future.

ui-participant/src/participant/ManageProfiles.tsx
connorlbark: Should we give ppUsers a shortcode? I like using a shortcode here for brevity but using the enrollee is somewhat indirect since, e.g., the manage profiles page isn't tied to a specific study, but rather the entire portal, so if a ppUser has more than one enrollee (or... None?) then it would either have to pick randomly or freak out and explode.

Also, more generally, I find that I keep having to do `profile = enrollees.find(profileId = ppuser.profileId).profile` - would it make sense to do a refactor of how we're passing down / thinking about profiles? It feels ppUsers should really be the ones owning profiles, it would disambiguate a lot of multi-study compatible code.

MatthewBemis (approved):

devonbush (approved):
nice, thanks for adding routing.  It occurs to me now that we should probably make the participant selector routable too -- e.g. changing the active user adds "?activeUser=xxxx" to the url params (or maybe it's `hub/{userId}`).  But that's outside the scope of this PR, because that would require upgrading a lot of different routes to preserve that param

PR #866, opened 2024-04-30
[JN-1036 & JN-1037] Fix flash on registration and make governed user default user

connorlbark:

Significantly cleans up the sign in flow, adding loading indicators / making sure everything is loaded before arrival at the hub page.

https://github.com/broadinstitute/juniper/assets/17440010/ac5eee80-869b-4e3e-a52b-8124bc8f883a

Also fixes proxy signup to default to the governed user.

- Create new governed user
- Observe loading spinners (throttle in chrome if needed)
- Observe that the hub page defaults to the governed user
- Create new regular user
- No regressions

### DIFF
--- a/ui-participant/src/hub/HubPage.tsx
+++ b/ui-participant/src/hub/HubPage.tsx
@@ -13,6 +13,7 @@ import OutreachTasks from './OutreachTasks'
 import HubPageParticipantSelector from './HubPageParticipantSelector'
 import { useActiveUser } from 'providers/ActiveUserProvider'
 import { useUser } from 'providers/UserProvider'
+import { LoadingSpinner } from '../util/LoadingSpinner'
 
 
 /** renders the logged-in hub page */
@@ -82,6 +83,10 @@ export default function HubPage() {
           style={{ background: '#fff', maxWidth: 768 }}
         >
           {relations.length > 0 && <HubPageParticipantSelector/>}
+          {/* eslint-disable-next-line react/jsx-no-undef */}
+          {enrollees.length === 0 && <div className={'d-flex justify-content-center'}>
+            <LoadingSpinner/>
+          </div>}
           {enrollees.map(enrollee => <StudySection key={enrollee.id} enrollee={enrollee} portal={portal}/>)}
         </main>
         <div className="hub-dashboard mx-auto"
--- a/ui-participant/src/hub/HubPage.tsx
+++ b/ui-participant/src/hub/HubPage.tsx
@@ -83,7 +83,6 @@ export default function HubPage() {
           style={{ background: '#fff', maxWidth: 768 }}
         >
           {relations.length > 0 && <HubPageParticipantSelector/>}
-          {/* eslint-disable-next-line react/jsx-no-undef */}
           {enrollees.length === 0 && <div className={'d-flex justify-content-center'}>
             <LoadingSpinner/>
           </div>}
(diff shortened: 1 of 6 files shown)


ui-participant/src/hub/HubPage.tsx
```diff
@@ -82,6 +83,9 @@ export default function HubPage() {
           style={{ background: '#fff', maxWidth: 768 }}
         >
           {relations.length > 0 && <HubPageParticipantSelector/>}
+          {enrollees.length === 0 && <div className={'d-flex justify-content-center'}>
+            <LoadingSpinner/>
+          </div>}
```
connorlbark: Might not be the best way to handle this - I fear if someone gets into a hanging state without an enrollee, they will be confused about the fact that it seems to be loading infinitely. But maybe this is OK because that is an invalid state and it makes it clearer "something has gone wrong". Open to ideas.

devonbush: seems like we need an indicator of whether we've loaded all the login info.  Either let's add an explicit "loaded" to ActiveUserContext, or make the enrollees property nullable, so that it can be null if we haven't loaded anything yet

devonbush (changes requested):
nice cleanup!  Let's just tidy up the loading state (as long as it's not too involved)

ui-participant/src/hub/HubPage.tsx
connorlbark: ~I am honestly not sure what is happening here. It's kind of hard to debug. With the "await" added on the study enroll router before it routes back to `/hub`, when it reaches the hub page it should actually be fully post-login/enrollment, so I don't think there's any actual network things happening when the loading indicator shows up right now. I think it is just that react needs a couple frames to refresh internal state to populate the enrollee property on the active user provider. So, maybe it doesn't make sense to add a loading indicator? It will flash, but it won't flash "All activities complete" anymore. But my local set up is not as slow as the production setup, so maybe what I am seeing is not accurate to what it will look like when it hits the demo server.~

I figured it out!

ui-participant/src/providers/UserProvider.tsx
```diff
@@ -148,15 +148,15 @@ export default function UserProvider({ children }: { children: React.ReactNode }
     const oauthAccessToken = localStorage.getItem(OAUTH_ACCRESS_TOKEN_KEY)
     const internalLogintoken = localStorage.getItem(INTERNAL_LOGIN_TOKEN_KEY)
     if (oauthAccessToken) {
-      Api.refreshLogin(oauthAccessToken).then(loginResult => {
+      return Api.refreshLogin(oauthAccessToken).then(loginResult => {
```
connorlbark: I have so much beef with async/promises 😡 This was the one return holding us back...

devonbush: ahh, good find.  Instead of return, let's refactor this to use await and try/catch.  You're right that both are confusing, so let's stick to `async/await` as much as possible

devonbush (approved):
One suggestion to update the older code to use async/await

ui-participant/src/studies/enroll/StudyEnrollRouter.tsx
```diff
@@ -140,9 +140,13 @@ function StudyEnrollOutletMatched(props: StudyEnrollOutletMatchedProps) {
     }
   }
 
+  const [isLoading, setIsLoading] = useState(false)
   // when either preEnrollment or login status changes, navigate accordingly
   useEffect(() => {
+    setIsLoading(true)
     determineNextRoute()
+      .then(() => setIsLoading(false))
+      .catch(() => setIsLoading(false))
```
devonbush:
```suggestion
    determineNextRoute()
      .finally(() => setIsLoading(false))
```

MatthewBemis (approved):

PR #868, opened 2024-04-30
[JN-904] add registerOrLogin endpoint for returns from OAuth

devonbush:

Since the OAuth handoff can be unreliable, this defaults all logins to be register or login.  This is fine since in prod our B2C tenants are all separate, so we don't have to worry about accidentally registering/enrolling someone.  If they've got a B2C account--they should be registered.  

This also cleans up some logout pathways -- previously we were doing some double-loads and unnecessary task fetching  

1. redeploy apiAdminApp
2. make sure local participant UI is running in b2c mode 
3. go to sandbox.ourhealth.localhost:3001. 
4. login as jsalk@test.com
5. confirm login works as expected
6. logout
7. confirm logout works as expected
8. register a new user, confirm works as usual.
9. Using the Azure portal, create a new B2C user in the demo dev tenant that does not exist in Juniper
10. confirm you can login as that user, and are automatically registered & enrolled


ui-participant/src/hub/OutreachTasks.tsx
```diff
@@ -110,3 +113,268 @@ const studyForTask = (task: ParticipantTask, studies: Study[]) => {
 const enrolleeForTask = (task: ParticipantTask, enrollees: Enrollee[]) => {
   return enrollees.find(enrollee => enrollee.studyEnvironmentId === task.studyEnvironmentId)!
 }
+
+
+const foo = {
```
MatthewBemis: leftover from development?

MatthewBemis (approved):
nice, worked great locally

ui-participant/src/hub/OutreachTasks.tsx
devonbush: you caught me -- was trying to pad my commit length stats...

ui-participant/src/hub/OutreachTasks.tsx
```diff
@@ -62,8 +62,11 @@ export default function OutreachTasks({ enrollees, studies }: {enrollees: Enroll
   }
 
   useEffect(() => {
-    loadOutreachActivities()
-  }, [enrollees])
+    if (enrollees.length) {
```
connorlbark, line 65: Nit: I am unfamiliar with this syntax so my instinct would be to make it clearer (`enrollees.length > 0`). Also - this might not happen anymore since we're fully reloading the participant before they go to the hub page.

connorlbark (approved):
Works great! Fun to be able to flex my new access to the dev B2c tenants 😄

ui-participant/src/hub/OutreachTasks.tsx
devonbush: I think it's a shorthand common in JS and Ruby, relying on the fact that 0 evaluates to `falsy`. I like it because once you get used to it, it's easier to read at a glance.

PR #870, opened 2024-05-01
[JN-1043] Admin proxy view

connorlbark:

Adds the ability to quickly view an enrollee's proxy. Also, cleans up the enrollee overview to not use the raw HTML inputs and instead the new fancy card that Erin uses in her design.

image

- Sign up a user with a proxy
- Go to the user's account, observe that the proxy shows up
- Go to a non-proxied account and observe correct behavior.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeRelationController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeRelationController.java
@@ -1,0 +1,48 @@
+package bio.terra.pearl.api.admin.controller.enrollee;
+
+import bio.terra.pearl.api.admin.api.EnrolleeRelationApi;
+import bio.terra.pearl.api.admin.service.AuthUtilService;
+import bio.terra.pearl.api.admin.service.enrollee.EnrolleeRelationExtService;
+import bio.terra.pearl.core.model.EnvironmentName;
+import bio.terra.pearl.core.model.admin.AdminUser;
+import bio.terra.pearl.core.model.participant.EnrolleeRelation;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import jakarta.servlet.http.HttpServletRequest;
+import java.util.List;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+
+@Controller
+@Slf4j
+public class EnrolleeRelationController implements EnrolleeRelationApi {
+  private AuthUtilService authUtilService;
+  private EnrolleeRelationExtService enrolleeRelationExtService;
+  private HttpServletRequest request;
+  private ObjectMapper objectMapper;
+
+  public EnrolleeRelationController(
+      AuthUtilService authUtilService,
+      EnrolleeRelationExtService enrolleeRelationExtService,
+      ObjectMapper objectMapper,
+      HttpServletRequest request) {
+    this.authUtilService = authUtilService;
+    this.enrolleeRelationExtService = enrolleeRelationExtService;
+    this.objectMapper = objectMapper;
+    this.request = request;
+  }
+
+  @Override
+  public ResponseEntity<Object> findRelationsByTargetIdWithEnrollees(
+      String portalShortcode, String studyShortcode, String envName, String enrolleeShortcode) {
+    AdminUser adminUser = authUtilService.requireAdminUser(request);
+    List<EnrolleeRelation> relations =
+        enrolleeRelationExtService.findRelationsByTargetIdWithEnrollees(
+            portalShortcode,
+            studyShortcode,
+            EnvironmentName.valueOfCaseInsensitive(envName),
+            adminUser,
+            enrolleeShortcode);
+    return ResponseEntity.ok(relations);
+  }
+}
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeRelationController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeRelationController.java
@@ -33,15 +33,15 @@ public class EnrolleeRelationController implements EnrolleeRelationApi {
   }
 
   @Override
-  public ResponseEntity<Object> findRelationsByTargetIdWithEnrollees(
+  public ResponseEntity<Object> findRelationsForTargetEnrollee(
       String portalShortcode, String studyShortcode, String envName, String enrolleeShortcode) {
     AdminUser adminUser = authUtilService.requireAdminUser(request);
     List<EnrolleeRelation> relations =
-        enrolleeRelationExtService.findRelationsByTargetIdWithEnrollees(
+        enrolleeRelationExtService.findRelationsForTargetEnrollee(
+            adminUser,
             portalShortcode,
             studyShortcode,
             EnvironmentName.valueOfCaseInsensitive(envName),
-            adminUser,
             enrolleeShortcode);
     return ResponseEntity.ok(relations);
   }
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeRelationExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeRelationExtService.java
@@ -5,7 +5,6 @@ import bio.terra.pearl.core.model.EnvironmentName;
 import bio.terra.pearl.core.model.admin.AdminUser;
 import bio.terra.pearl.core.model.participant.Enrollee;
 import bio.terra.pearl.core.model.participant.EnrolleeRelation;
-import bio.terra.pearl.core.model.study.StudyEnvironment;
 import bio.terra.pearl.core.service.exception.NotFoundException;
 import bio.terra.pearl.core.service.participant.EnrolleeRelationService;
 import bio.terra.pearl.core.service.participant.EnrolleeService;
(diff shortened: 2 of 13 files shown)


ui-admin/src/components/Card.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react'
+import { isEmpty } from 'lodash'
+
+
+/**
+ * Stylized card component for displaying read-only or editable data.
+ */
+export function Card({ children }: { children: React.ReactNode }) {
```
connorlbark, line 8: Erin has been using the same style card across her designs, so I felt it made sense to consolidate it into a generic card type that we can reuse. It works kind of like the Modal component, with separate Header/Body components. I don't really like the names I chose, though, so I wanted to get thoughts and opinions on how these components should be named - Card feels a bit too generic, but that _is_ what it is, so idk.

connorlbark: For context, here's what they look like: 
image

ui-admin/src/components/InfoCard.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react'
+import { isEmpty } from 'lodash'
+
+
+/**
+ * Stylized card component for displaying read-only or editable data.
+ */
+export function InfoCard({ children }: { children: React.ReactNode }) {
```
connorlbark, line 49: I like this style of card that erin seems to use in multiple places in the designs. I decided to make it a fully generic component so that hopefully we can use it across the app.

I'm not sure if `InfoCard` is the best name for it, but it's the best I could come up with. Let me know if you have any ideas.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeRelationController.java
```diff
@@ -0,0 +1,48 @@
...
+      ObjectMapper objectMapper,
+      HttpServletRequest request) {
+    this.authUtilService = authUtilService;
+    this.enrolleeRelationExtService = enrolleeRelationExtService;
+    this.objectMapper = objectMapper;
+    this.request = request;
+  }
+
+  @Override
+  public ResponseEntity<Object> findRelationsByTargetIdWithEnrollees(
```
devonbush: I think it's a little confusing to have findByTargetId in the method name, given that this doesn't take an id.  I might rename to `findTargetingRelations`, and let's definitely add a comment with an example

api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeRelationExtService.java
```diff
@@ -0,0 +1,55 @@
...
+      EnrolleeRelationService enrolleeRelationService,
+      EnrolleeService enrolleeService,
+      StudyEnvironmentService studyEnvironmentService) {
+    this.authUtilService = authUtilService;
+    this.enrolleeRelationService = enrolleeRelationService;
+    this.enrolleeService = enrolleeService;
+    this.studyEnvironmentService = studyEnvironmentService;
+  }
+
+  public List<EnrolleeRelation> findRelationsByTargetIdWithEnrollees(
```
devonbush: same renaming comment

api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeRelationExtService.java
```diff
@@ -0,0 +1,55 @@
...
+    StudyEnvironment studyEnvironment =
+        studyEnvironmentService
+            .findByStudy(studyShortcode, environmentName)
+            .orElseThrow(() -> new NotFoundException("Study environment not found"));
+
+    Enrollee found =
+        enrolleeService
+            .findOneByShortcode(enrolleeShortcode)
+            .filter(enrollee -> enrollee.getStudyEnvironmentId().equals(studyEnvironment.getId()))
+            .orElseThrow(() -> new NotFoundException("Enrollee not found"));
```
devonbush: this idea of "get me this enrollee but make sure they're in the given studyEnv" seems common enough that it might be worth pulling this into a `findByShortcodeAndEnv` method in the core EnrolleeService

api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeRelationExtService.java
```diff
@@ -0,0 +1,55 @@
...
+
+    return enrolleeRelationService.findByTargetEnrolleeIdWithEnrollees(found.getId());
```
devonbush: yeah, this method name makes sense since we're passing in the id

ui-admin/src/components/InfoCard.tsx
devonbush: hmm.. yeah, this name works for now.  I would definitely include an "example use" comment so that it's easy at-a-glance to see how the components are intended to be nested.  Also, we might want to consider whether some of these should be components or just a set of classes 
e.g. 
```
const cardHeaderClasses = ['card-header', ...]

File: ui-admin/src/study/participants/enrolleeView/EnrolleeOverview.tsx [hunk 7]
```diff
@@ -1,29 +1,72 @@
...
+import { InfoCard, InfoCardBody, InfoCardHeader, InfoCardTitle, InfoCardValue } from '../../../components/InfoCard'
+import { useLoadingEffect } from '../../../api/api-utils'
+import LoadingSpinner from '../../../util/LoadingSpinner'
 
 /** Shows minimal identifying information, and then kits and notes */
 export default function EnrolleeOverview({ enrollee, studyEnvContext, onUpdate }:
         {enrollee: Enrollee, studyEnvContext: StudyEnvContextT, onUpdate: () => void}) {
+  const [relations, setRelations] = React.useState<EnrolleeRelation[]>([])
+
+  const { isLoading } = useLoadingEffect(async () => {
```
devonbush: Let's rename in the destructure so it's clear we're not waiting on everything to load
```suggestion
  const { isLoading: isLoadingRelations } = useLoadingEffect(async () => {
```

ui-admin/src/study/participants/enrolleeView/EnrolleeProfile.tsx
```diff
@@ -18,6 +18,14 @@ import { Store } from 'react-notifications-component'
...
 import { doApiLoad } from 'api/api-utils'
 import EditMailingAddress from 'address/EditMailingAddress'
+import {
+  InfoCard,
+  InfoCardBody,
+  InfoCardHeader,
+  InfoCardRow,
+  InfoCardTitle,
+  InfoCardValue
+} from '../../../components/InfoCard'
```
devonbush:
```suggestion
} from 'components/InfoCard'
```

ui-admin/src/study/participants/enrolleeView/EnrolleeOverview.tsx
```diff
@@ -1,29 +1,72 @@
...
+    const relations = await Api.findRelationsByTargetShortcode(
+      studyEnvContext.portal.shortcode,
+      studyEnvContext.study.shortcode,
+      studyEnvContext.currentEnv.environmentName,
+      enrollee.shortcode)
+    setRelations(relations)
+  })
+
+  if (isLoading) {
```
devonbush: no need to make the whole page wait for relation-fetching

devonbush (changes requested):
Overall looks good and works well when I tested local, just some cleanup nits and suggestions in the code.

image

Also, I think the top of the "Overview" box should be flush with the top of the gray sidebar.  (also, I think the extra "Overview" header is redundant--that top box could do without a title)

ui-admin/src/study/participants/enrolleeView/EnrolleeOverview.tsx
connorlbark: Do you think no loading animation at all? I tried it with that initially, but I felt on a slow internet if there were relations it was jarring for it to all of the sudden appear. But without proxies, it feels weird to have a loading animation in the middle of the screen that leads to nothing.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeRelationExtService.java
connorlbark: Oops 🤦

connorlbark:
> also, I think the extra "Overview" header is redundant--that top box could do without a title

I feel it looks kind of weird without a title
image

But I agree `Overview` feels redundant... Perhaps `Basic Information`? A bit verbose.

My radical idea would be to just move the profile to the overview page, so you see the entire profile and it is editable on the overview page.

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/enrollee/EnrolleeRelationController.java
connorlbark: Good catch - I forgot to rename it as I was refactoring everything. I'm not sure exactly what you mean by an example, though - do you mean an example of what it returns/how it should be used?

ui-admin/src/api/api.tsx
```diff
@@ -101,6 +101,21 @@ export type Enrollee = {
...
+  relationshipType: RelationshipType,
+  targetEnrolleeId: string,
+  targetEnrollee: Enrollee
+  enrolleeId: string
+  enrollee: Enrollee
+  createdAt: number
+  lastUpdatedAt: number
+  beginDate: number
+  endDate: number
+}
```
MatthewBemis, line 117: Similar types exist in `ui-participant`, should these go into `ui-core`?

ui-admin/src/study/participants/enrolleeView/EnrolleeOverview.tsx
devonbush: ahh, now I understand why you did that.  I don't have a problem with a loading animation that leads to nothing, and I slightly prefer that to blocking the whole page.  But I could go either way--your (or Erin's) call

connorlbark: Cool, I'm in favor of compromising with the loading animation that leads to nothing.

devonbush:
Yeah, I see your point about no title looking a bit odd too.  As long as the top is aligned, I'm good with either way, or changing the title as your suggest.

  Re: profile, when we were initially laying out this page the feedback from CMI was that they did *not* want the full profile on this page -- it took up too much space when most of the time they visit this page it's for kits and notes

MatthewBemis (approved):

ui-admin/src/api/api.tsx
connorlbark: The only issue is that I would also have to move the `Enrollee` type, which would totally be worth it, but it would be a pretty big lift in this PR. I'll add it to the eng backlog.

MatthewBemis: That's a good point- that same thing is getting in my way for the admin form completion work too, so good call to keep it out of this PR. You might get it for free in a few days anyway

devonbush (approved):

PR #878, opened 2024-05-07
[JN-1054] fixing e2e tests

devonbush:

A couple of text changes broke e2e tests.  Notably, the addition of the autocomplete feature broke a test that tried to click on a "Complete" button.  I didn't realize that playwright was set to imprecise match mode, which is probably not what we want.   I also changed the name of the button to "Autofill" to avoid confusion since that button doesn't actually "Complete" the survey, it just fills it in.

1. make sure apiAdmin, ApiParticipantApp, ui-admin and ui-participant are running, and demo is populated
2. in e2e-tests run `npx playwright test`


connorlbark (approved):
They all passed, but I did get a weird browser console error:
```
Browser Console
***
"Failed to load resource: the server responded with a status of 500 (Internal Server Error)"
***
```

devonbush:
hmm... I'll take a look -- I didn't see anything like that on my machine

MatthewBemis (approved):
I'm seeing the same error as Connor

Looks like it's coming from `new-registration-age18.test.ts`, but it still passes.

devonbush:
Ah, I think the error is consent form related -- this should go away once this is merged into development with the consent form merge

PR #880, opened 2024-05-08
[JN-1023] E2e proxy test

connorlbark:

Creates an e2e test for proxy enrollment. Also, the tests needed a decent amount of refactoring to be HeartDemo compatible. However, it should be a lot easier to add more demo tests in the future.

(note branches off of devon's e2e fixes and I had to merge development, so I recommend only looking at changes from the last 3 commits. you can do this by shift+clicking them (TIL). will change to point to development once his PR is merged)

- Run e2e tests


populate/src/main/resources/seed/portals/demo/portal.json
```diff
@@ -16,7 +16,7 @@
         "password": "broad_institute",
         "initialized": true,
         "emailSourceAddress": "support@juniper.terra.bio",
-        "defaultLanguage": "es"
+        "defaultLanguage": "en"
```
connorlbark, line 19: For better or for worse, much of the existing code relied on the english translations. While possible to make it work with spanish on demo, I figured it would be easier to just not deal with it and change the default language over to english like we'd discussed.

e2e-tests/src/pages/demo/home-about-us.ts
```diff
@@ -0,0 +1,25 @@
+import { expect, Locator, Page } from '@playwright/test'
+import data from 'src/data/ourhealth-en.json'
```
MatthewBemis, line 2: should this reference demo-specific data instead of crossing over into ourhealth? is there anything that isn't study-specific that we can pull out into a common data file? (same comment applies to `home.ts` as well)

MatthewBemis (commented):
overall looks good and worked well once I remembered that I had to start up the participant UI with b2c disabled 🤦 

I am mainly interested in keeping the separation of ourhealth/demo study data in the tests. I think as we build out even more test scenarios across different studies, we'll want to avoid blurred lines otherwise it may become difficult to manage/debug

e2e-tests/src/pages/demo/home-about-us.ts
connorlbark: I suspect there's more work to do here - I think we should step back and look at it as a team and figure out how we want to approach making the common/abstract classes we have work for every study, especially in regards to multi-language testing. I suspect it will be a decent amount of work, my thought was that I would do the minimum to make it work with demo and we all get together to think about our approach, but I'm happy to move this json over in this ticket if y'all think it would be useful.

devonbush (approved):

e2e-tests/src/pages/demo/home-about-us.ts
MatthewBemis: Deferring for discussion works for me- I just want to make sure we address it in the not-too-distant future. Definitely agree that it doesn't have to be part of this PR!

MatthewBemis (approved):

e2e-tests/src/pages/demo/home-about-us.ts
connorlbark: Added this to eng strategy meeting discussion - I think a spike where someone focuses on upgrading the abstract classes / figuring out multi-lang stuff is probably the right approach.

PR #882, opened 2024-05-09
[JN-1063] Part 1 - Add needed features to search expressions to replace search facets in participant list

connorlbark:

I decided it would be easiest to separate out the functionality and the frontend for this ticket - both will be hairy, but in their own way.

Adds:
- Ability to reference tasks, e.g. `{task.oh_oh_basics.status} = 'IN PROGRESS'`
- Contains (only works for strings), e.g. `{enrollee.shortcode} contains 'JSALK'` (it being a boolean operator follows how it works in surveyjs, we still don't have functions, but we don't have a need yet)
- Enrollee terms (e.g., `enrollee.shortcode`, `enrollee.subject`, `enrollee.consented`)
- Grabbing full name from profile e.g. `{profile.name} contains 'Jonas Sa'`

Generally refactors naming to make more sense in hindsight

Note, I didn't add a `KitTerm` for two reasons: First, kits are currently being sorted in the frontend. Second, it would require us to support arrays for the first time, and that got very complex very fast so I decided to push that off to a later date when we really need it.

- Observe tests
- If you want to play around, the easiest way is probably creating surveys. I think a (for now) superuser-only "advanced filter" in the participant list where you can test out typing search expressions might make sense. That will come in a following PR.


core/src/main/java/bio/terra/pearl/core/dao/search/EnrolleeSearchExpressionDao.java
```diff
@@ -92,6 +93,16 @@ public EnrolleeSearchExpressionResult map(ResultSet rs, StatementContext ctx) th
                     );
                 }
 
+                if (columnName.startsWith("task_") && columnName.endsWith("_created_at")) {
```
devonbush, line 96: this code feels like it would be cleaner with a regex -- you're essentially looking for a match to `/^task_(.+)_created_at$/`

devonbush (approved):
Looks good -- appreciate you splitting out the backend changes.  One small optional code cleanup

MatthewBemis (approved):

PR #884, opened 2024-05-10
[JN-1044] kit study env config

devonbush:

This moves DSM configuration to the study environment level, so for a given environment a superuser can set whether kit requests are mocked, go to a dev DSM realm, or go to a study-specific DSM realm

image

1. redeploy apiAdminApp
2. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/settings
3. change the settings for the kit requests and save
4. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/participants/OHBASC
5. request a kit
6. confirm by looking at the logs that the kit request was sent to mock/DSM as indicated


core/src/main/resources/db/changelog/changesets/2024_05_08_dsm_env_config.yaml
```diff
@@ -0,0 +1,29 @@
...
+        - addColumn:
+            tableName: study_environment_config
+            columns:
+              - column: { name: use_dev_dsm_realm, type: boolean, defaultValueBoolean: true, constraints: { nullable: false} }
+              - column: { name: use_stub_dsm, type: boolean, defaultValueBoolean: true, constraints: { nullable: false} }
+        - addDefaultValue:
+            tableName: study_environment_config
+            columnName: accepting_proxy_enrollment
+            defaultValueBoolean: false
+        - sql:
```
devonbush, line 15: make sure we auto-update ourhealth live to use real kits so we don't forget to do that in product -- everything else we can set manually.

api-admin/src/main/resources/application.yml
```diff
@@ -23,7 +23,6 @@ env:
     billingProfileId: ${BILLING_PROFILE_ID:550212cf-45dc-478f-9fe6-9b3290e2a1fe}
     tdrExportEnabled: ${TDR_EXPORT_ENABLED:false}
   dsm:
-    useLiveDsm: ${USE_LIVE_DSM:false}
```
MatthewBemis, line 26: after this is merged, can you also remove these values from the [dev](https://github.com/broadinstitute/terra-helmfile/blob/04becbb3707afe1c2b1bf31151961e3a541d4264/values/app/d2p/live/ddp-azure-dev.yaml#L170) and [prod](https://github.com/broadinstitute/terra-helmfile/blob/04becbb3707afe1c2b1bf31151961e3a541d4264/values/app/d2p/live/ddp-azure-prod.yaml#L83) helmfiles?

MatthewBemis (approved):
Looks good 👍 

non-blocking: what do you think of adding some indicator on the kit request modal page if the stub is being used? Could save some potential headaches, or it could just be extra UX noise.

Screenshot 2024-05-14 at 1 20 44 PM

connorlbark (approved):
I agree with Matt that a warning based on your study configuration would be nice on the UI for sending kits, otherwise 👍

devonbush:
heartily agree on the notification.  that's a bit non-trivial because it will require some extra fetching.  I'll make a separate ticket for it

MatthewBemis:
@devonbush isn't the `StudyEnvironmentConfig` in scope already as part of `studyEnvContext`? i.e. 

```
      { studyEnvContext.currentEnv.studyEnvironmentConfig.useStubDsm &&
          <div>warning message...</div>
      }
```

PR #885, opened 2024-05-13
[JN-1067] Add UI for answer mappings

connorlbark:

Adds new tab for answer mappings in the survey editor view.

image

- Rebuild core
- Populate demo
- Try deleting mappings, creating mappings, etc.


api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/AnswerMappingController.java
```diff
@@ -0,0 +1,60 @@
...
+import bio.terra.pearl.api.admin.service.AuthUtilService;
+import bio.terra.pearl.api.admin.service.forms.AnswerMappingExtService;
+import bio.terra.pearl.core.model.admin.AdminUser;
+import bio.terra.pearl.core.model.survey.AnswerMapping;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import jakarta.servlet.http.HttpServletRequest;
+import java.util.UUID;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+
```
devonbush: I don't think this should be a separate controller -- changes to answer mappings should flow through the SurveyController.newVersion method

ui-admin/src/study/surveys/AnswerMappingEditor.tsx
```diff
@@ -0,0 +1,305 @@
...
+    setAddNewAnswerMapping(false)
+  }
+
+  return <div className='px-3 pt-1'>
+    <p>
+      Answer mappings allow you to map answers from questions in the survey to other parts of the system.
+      For example, you could map a question which collects the participant&apos;s first name to the profile&apos;s.
+      given name field. Then, any changes the participant makes to the survey will result in changes to their profile.
+    </p>
+    <table className={'table'}>
```
connorlbark: I used a regular table (rather than a `useTable`) because it was easier to achieve my desired "+" button -> "add new answer mapping" UX. I could try to implement it with a `useTable` if you all prefer.

connorlbark:
@MatthewBemis @devonbush this is now ready for review again!

ui-admin/src/forms/FormContentEditor.tsx
```diff
@@ -73,15 +84,28 @@ export const FormContentEditor = (props: FormContentEditorProps) => {
...
               }}
             />
           </ErrorBoundary>
         </Tab>
+        <Tab
+          disabled={activeTab !== 'answermappings' && !tabsEnabled}
+          eventKey="answermappings"
+          title="Answer Mappings"
+        >
+          <ErrorBoundary>
```
devonbush: hmm... I'm not sure I'd put the error boundary here.  if the answer mappings are corrupt, we probably don't want them saving the survey

ui-admin/src/study/surveys/AnswerMappingEditor.tsx
devonbush: yes, please use the tanstack-table for this.  It'll be good practice for you because you'll have to think about what you memoize and how.  There should be examples in their docs for how to do it

devonbush (changes requested):
I think we want to adopt a team posture of "let's get really good at using tanstack table" rather than writing a lot of idiosyncratic table controls every time we want a different UX.

The UX is smooth, but it's a bit loud.  I don't think the delete buttons need to be red, or if they are, just the icon should be red.  That said, I like the green button to confirm the in-progress mapping -- it calls attention to the fact that you need to press it for your input to count.  I would make the checkmark inside it bold though

connorlbark:
> I would make the checkmark inside it bold though

Hmm, I'm not actually sure how to do this. `fw-bold` didn't work.

ui-admin/src/study/surveys/AnswerMappingEditor.tsx
connorlbark: I got it working but I'm not sure it makes the code more readable. I'll defer to your final judgement here.

ui-admin/src/study/surveys/AnswerMappingEditor.tsx
```diff
@@ -0,0 +1,303 @@
...
+export default function AnswerMappingEditor(
+  {
+    formContent, initialAnswerMappings, onChange
+  } : {
+    formContent: FormContent, initialAnswerMappings: AnswerMapping[], onChange: OnChangeAnswerMappings
+  }
+) {
+  const [answerMappings, setAnswerMappings] = useState<AnswerMapping[]>(initialAnswerMappings || [])
+
+  const [openDeleteMappingModal, setOpenDeleteMappingModal] = useState<AnswerMapping | null>(null)
```
devonbush: rename so it's clearer what's being stored
```suggestion
  const [mappingSelectedForDeletion, setMappingSelectedForDeletion] = useState<AnswerMapping | null>(null)
```

ui-admin/src/study/surveys/AnswerMappingEditor.tsx
```diff
@@ -0,0 +1,303 @@
...
+    setAddNewAnswerMapping(false)
+    setQuestionStableId(undefined)
+    setTargetType(undefined)
+    setTargetField(undefined)
+    setFormatString(undefined)
+    setMapType('STRING_TO_STRING')
+  }
+
+  // state for new mapping
+  const [questionStableId, setQuestionStableId] = useState<string>()
```
devonbush: would it be clearer to make this a single "mappingToAdd" that's of type AnswerMapping rather than separate state objects for every property?

ui-admin/src/study/surveys/SurveyEditorView.tsx
```diff
@@ -229,16 +230,35 @@ const SurveyEditorView = (props: SurveyEditorViewProps) => {
...
+          onFormContentChange={(newValidationErrors, newContent) => {
             if (isEmpty(newValidationErrors)) {
               setShowErrors(false)
               setDraft({ ...draft, content: JSON.stringify(newContent), date: Date.now() })
             }
             setValidationErrors(newValidationErrors)
           }}
+          onAnswerMappingChange={(newValidationErrors, newAnswerMappings) => {
+            if (isEmpty(newValidationErrors)) {
+              console.log('updating draft')
```
devonbush: remove these leftover logs

ui-admin/src/study/surveys/AnswerMappingEditor.tsx
devonbush: that way you could pass that object as the last row, instead of `{isLastRow: true}`

devonbush (changes requested):
Nice, there's a ton of awesome effort here in making the UX smooth and easy-to-use.  

On the technical side, you don't need to make an extra copy of the answerMappings, and the column definitions don't need to be dependent on the component state. I spent some time futzing with this and  put up a branch `db-answer-mapping-update` with an approach that gets us closer to a more general editable table pattern.   Check it out and let me know.

Also, this needs some js tests :)

ui-admin/src/study/surveys/AnswerMappingEditor.tsx
```diff
@@ -0,0 +1,334 @@
...
+          (
+            AnswerMappingTargets[newAnswerMapping.targetType]
+            && AnswerMappingTargets[newAnswerMapping.targetType][newAnswerMapping.targetField]
+          ) || 'STRING_TO_STRING'
+        )
+      })
+    }
+  }, [newAnswerMapping?.targetType, newAnswerMapping?.targetField])
+
+  const onNewAnswerMappingChange = (field: keyof EditableAnswerMapping, value: string | boolean) => {
```
devonbush, line 122: nice helper

devonbush (approved):
nice! thanks for working through all the table stuff

MatthewBemis (approved):

PR #889, opened 2024-05-15
[JN-1078] trigger survey assign

devonbush:

Previously, we would only assign survey tasks on enrollee creation or on publication of a survey.  This makes it so survey completion events can trigger follow-up surveys.

NOTE: in testing this I discovered that we're not storing answerType for any answers.  I've added a temporary workaround here, and created a ticket for the full fix

1.  Confirm the unit test makes sense
2.  redeploy apiParticipantApp
3. repopulate demo
4. log in to `https://sandbox.demo.localhost:3001/` as jsalk@test.com
5. note that he does NOT have a "Lost interest survey" assigned
6. go to the mental health survey, and change the answer to the first question to "Nearly every day", then complete the survey (you don't have to answer any other questions)
7. on the dashboard, you should now see the "Lost interest" survey


connorlbark (commented):
Works, but I do wonder if this should only be on survey completion. I found it confusing as a user experience because I signed in as Jsalk, who had already completed the survey, and I changed the answer and it didn't give me the new one.

I could see the argument that, if I select that answer but don't complete the survey, I shouldn't get the new survey. But I do think if you change an answer on a completed survey it should assign you the new survey, I wouldn't expect to have to re-complete it. Otherwise it worked great!

~As an aside, I also noticed that the eligibility rule was in the DB, but it didn't show up in the lost interests UI. A cursory look didn't make the bug seem obvious but I'll take a peek and see if I can figure out why it's doing that.~ I have a pr up to fix this.

connorlbark (approved):
(i'll go ahead and approve, up to you if you wanna fix the edge case, I don't know how common it would be)

devonbush:
Connor and I chatted some more -- we'll leave as-is for now, and then open this up for discussion at next demo

MatthewBemis (approved):
:shipit:

PR #890, opened 2024-05-15
[JN-1080] fixing answer type saves

devonbush:

We just, um, weren't saving answer types.  Which isn't a big deal since they're strictly determined by the presence of the value columns.  But still... oops.   We eventually will want a DB constraint on this, but in the short term I'd rather have answers save without type than error.  

1.  redeploy participantApiApp
2. go to https://sandbox.demo.localhost:3001/
3. log in as consented@test.com
4. start filling out a survey
5. in the database, confirm that `select count(*) from answer where answer_type is null;` is 0

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/model/survey/Answer.java
+++ b/core/src/main/java/bio/terra/pearl/core/model/survey/Answer.java
@@ -40,6 +40,22 @@ public class Answer extends BaseEntity {
         setValue(value);
     }
 
+    /** infers the type based on what value was set */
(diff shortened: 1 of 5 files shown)


connorlbark (approved):
I wonder if it would be cleaner to use a SQL trigger to auto-fill this property or make it a computed column? Either way, this is a nice cleanup 👍

MatthewBemis (approved):

devonbush:
hmm... the idea of making it a computed column is interesting.  will marinate on that

PR #891, opened 2024-05-16
[JN-1081] Fix signaturepad image display

MatthewBemis:

Fixes a bug on the admin survey response view that required the question stable id to end in "signature" in order to be rendered as an image. Switches to checking the question type instead.

* Confirm that `signaturepad` image previews are still visible: https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants/HDBASC/surveys/hd_hd_consent

### DIFF
--- a/ui-admin/src/study/participants/survey/SurveyFullDataView.test.tsx
+++ b/ui-admin/src/study/participants/survey/SurveyFullDataView.test.tsx
@@ -11,7 +11,7 @@ import userEvent from '@testing-library/user-event'
 
 describe('getDisplayValue', () => {
   it('renders a plaintext value', async () => {
-    const question: Question = { isVisible: true } as Question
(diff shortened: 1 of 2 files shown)


devonbush:
Nice.  But I feel like there must have been a reason we weren't using the question type before.  Maybe because this started out as just a list of answers and we weren't matching answers to questions?  We'll want to double-check this one after release to production just to be sure

devonbush (approved):

MatthewBemis:
@devonbush that was my guess as well. I took a look through the history of this file and the initial implementation did not have the SurveyJS Question in scope, it wasn't until the question text was implemented, as you said: https://github.com/broadinstitute/juniper/blob/6e4a14ebb70d5c625c3d087cb3402a4a9b14b73e/ui-admin/src/study/participants/survey/SurveyFullDataView.tsx

connorlbark (approved):

PR #892, opened 2024-05-16
[JN-1083] Temp fix for empty search expression

connorlbark:

If you go to edit a survey with an eligibility rule, it shows as empty. If you then save, it strips the eligibility rule (yikes). This is only if you have the fancy query builder. This is because on the first render, the query builder has nothing, so it does an `onChange` call to make it empty.

Without doing fancy ref stuff, I figured just making it skip changing the eligibility rule if the string is empty is fine in the meantime, since once we have antlr it will work properly & this is only shown to prototype users/superusers anyway, so a wonky UX is fine in the meantime.

- Create a survey with an eligibility rule
- Observe that the eligibility rule still shows up when you try to edit it

### DIFF
--- a/ui-admin/src/study/surveys/SurveyEditorView.tsx
+++ b/ui-admin/src/study/surveys/SurveyEditorView.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 
 import { PortalEnvironment, Survey, VersionedForm } from 'api/api'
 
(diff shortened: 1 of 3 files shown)


ui-admin/src/study/surveys/SurveyEditorView.tsx
```diff
@@ -56,6 +56,13 @@ const SurveyEditorView = (props: SurveyEditorViewProps) => {
   const [draft, setDraft] = useState<FormDraft | undefined>(
     !readOnly ? getDraft({ formDraftKey: FORM_DRAFT_KEY }) : undefined)
 
+  useEffect(() => {
```
devonbush: remove logs

devonbush (approved):
good quick fix -- just remember to strip out the logs

ui-admin/src/study/surveys/SurveyEditorView.tsx
connorlbark: 🤦 I keep doing this lol. I blame the coffee making me jittery.

MatthewBemis (approved):

PR #894, opened 2024-05-16
[JN-1020] Admin form editing

MatthewBemis:

Adds support for admin form editing. As a result, this brings a lot of survey-related infrastructure and types into ui-core. 

**OUT OF SCOPE**: a refined admin UX. This PR is massive and challenging to review as-is, the focus here is just getting the survey infrastructure into core along with all of the related dependencies. Refined UX will come in follow-on PRs.

This needs some extra attention when testing the participant side to ensure that there weren't any regressions. This branch has also undergone several notable rebases during it's life, and accounting for newer changes during a rebase while also moving them ui-core is particularly vulnerable to dropping new functionality. Some areas worth calling out:

* Confirm that address validation still works in surveys
* Confirm that autosave behaves as expected
* Confirm that all of the recently-added proxy survey functionality still works correctly

More specific steps coming soon, but definitely focus on the above areas.


core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -197,11 +195,12 @@ protected SurveyResponse findOrCreateResponse(ParticipantTask task, Enrollee enr
         } else {
             SurveyResponse newResponse = SurveyResponse.builder()
                     .enrolleeId(enrollee.getId())
-                    .creatingParticipantUserId(participantUserId)
                     .surveyId(survey.getId())
                     .complete(responseDto.isComplete())
                     .resumeData(responseDto.getResumeData())
                     .build();
+            Optional.ofNullable(operator.getParticipantUser()).ifPresent(pUser -> newResponse.setCreatingParticipantUserId(pUser.getId()));
```
devonbush: I think a standard `if/else` is more appropriate here -- we're not doing any chaining so the  functional wrapper makes if harder to pinpoint the if logic--but maybe that's just me.  Connor, what's your take?  Better yet, add a "setResponsibleUser" method to SurveyResponse that takes a responsibleEntity and sets the ids appropriately.  That method might then be reused via interface with Answer

ui-admin/src/study/participants/survey/SurveyEditView.tsx
```diff
@@ -1,10 +1,61 @@
...
+        enrollee={enrollee}
+        response={workingResponse}
+        selectedLanguage={defaultLanguage.languageCode}
+        adminUserId={adminUserId}
+        onSuccess={() => {
+          onUpdate()
+          Store.addNotification(successNotification('Response saved'))
+        }}
+        onFailure={() => Store.addNotification(failureNotification('Response could not be saved'))}
+        updateProfile={() => {}} //eslint-disable-line @typescript-eslint/no-empty-function
```
devonbush: I think you can avoid the typescript warning by having it return null or true, or some other simple value
```suggestion
        updateProfile={() => null} 
```

ui-admin/src/study/participants/survey/SurveyEditView.tsx
```diff
@@ -1,10 +1,61 @@
...
+import { Enrollee, PagedSurveyView, useTaskIdParam } from '@juniper/ui-core'
+import { StudyEnvContextT } from 'study/StudyEnvironmentRouter'
+import { Store } from 'react-notifications-component'
+import { failureNotification, successNotification } from 'util/notifications'
+import { usePortalLanguage } from 'portal/usePortalLanguage'
 
 /** allows editing of a survey response */
-export default function SurveyEditView({ response, survey }: {response: SurveyResponse, survey: Survey}) {
-  console.log(`not implemented ${response.surveyId} ${survey.stableId}`)
+export default function SurveyEditView({ studyEnvContext, response, survey, enrollee, adminUserId, onUpdate }: {
```
devonbush: let's rename this to SurveyResponseEditor to avoid confusion with the form editor

ui-admin/src/study/participants/survey/EnrolleeSurveyView.tsx
```diff
@@ -1,20 +1,23 @@
...
 import pluralize from 'pluralize'
 import { StudyEnvContextT } from 'study/StudyEnvironmentRouter'
+import { useUser } from 'user/UserProvider'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faPencil, faX } from '@fortawesome/free-solid-svg-icons'
 
 /** Show responses for a survey based on url param */
-export default function EnrolleeSurveyView({ enrollee, responseMap, studyEnvContext }:
-  {enrollee: Enrollee, responseMap: ResponseMapT, studyEnvContext: StudyEnvContextT}) {
+export default function EnrolleeSurveyView({ enrollee, responseMap, studyEnvContext, onUpdate }:
```
devonbush: SurveyResponseView might be a better name for this component

devonbush (commented):
Nice work on all the code rearranging!  Overall, code looks good.  Some minor suggestions.  Only commenting for now because, as you said, I want to take a lot of time to test this code out

devonbush:
Initial testing seems solid -- I had tabs open to admin and participant view of the same survey and was taking turns editing both, and it always saved the answers, which were then available to the other side upon refresh.  (which is desired behavior -- CMI confirmed earlier they do not want a google-docs-like shared editing experience.  They'd rather risk accidental overwrites than have a patient jarred by seeing their data change in front of them

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
connorlbark: Hmmm... I like the `setResponsibleUser` on the survey response approach. That mirrors (iirc) the flow on the data audit info. I don't feel super strongly on `if/else`, but I agree it might make the logic clearer.

ui-core/src/index.ts
```diff
@@ -21,8 +26,10 @@ export * from './timeUtils'
 export * from './addressUtils'
 export * from './i18nUtils'
 export * from './waitForImages'
+export * from './useTaskIdParam'
 
 export * from './test-utils/asMockedFn'
+export * from './test-utils/router-testing-utils'
```
connorlbark, line 32: I see that we already had some testing stuff in core's index file, but I wonder if we should break this out into something like `@ui-core/testing`: it seems weird (to me) to have testing utils interspersed with the core app functionality. That is probably out of the scope of this PR, though.

ui-core/src/components/forms/PagedSurveyView.tsx
```diff
@@ -0,0 +1,148 @@
...
+import React, { useRef } from 'react'
+import { useAutosaveEffect } from '../../autoSaveUtils'
+import { useI18n } from '../../participant/I18nProvider'
+import { SurveyAutoCompleteButton } from './SurveyAutoCompleteButton'
+import { SurveyReviewModeButton } from './ReviewModeButton'
+import { StudyEnvParams } from 'src/types/study'
+import { Enrollee } from 'src/types/user'
+import { Profile } from 'src/types/address'
+
+const AUTO_SAVE_INTERVAL = 3 * 1000  // auto-save every 3 seconds if there are changes
```
connorlbark, line 21: I'm still wary of relying on autosave for editing participant data, since it seems like a "thing we should not do often", but CMI folks seemed to not mind, so 🤷 . It definitely makes sense in regards to admin-only forms, though.

ui-core/src/surveyUtils.tsx
```diff
@@ -74,31 +262,31 @@ type UseSurveyJsModelOpts = {
...
- * be useful for all surveys (e.g., {isProxyEnrollment} for the pre-enroll survey)
  */
 export function useSurveyJSModel(
   form: Survey,
   resumeData: SurveyJsResumeData | null,
   onComplete: () => void,
   pager: PageNumberControl,
+  envName: EnvironmentName,
+  profile?: Profile,
+  proxyProfile?: Profile,
```
connorlbark, line 279: I'm sad that we have to move these to the parameters :( these are so easy to forget, as evidenced by the preenroll breaking. it would be so cool if there was some way to allow `useUser`/`useActiveUser` to still function in core.

It feels like more and more things are moving to core as we start to allow the admins to do more of the participant experience in the admin side. As such, I suspect we might start running into more cases like this where we have to awkwardly pass down user information to things in core. it would be cool if these hooks worked in the admin app but returned nothing except for the cases that you are looking at an enrollee's page, then it would return the appropriate user.

potentially, we could rename them to `useParticipantUser` and `useActiveParticipantUser` so that it's more clear in the admin side that they are only available when you're viewing a specific participant? dunno just spitballing

ui-core/src/types/address.ts
```diff
@@ -17,3 +17,17 @@ export type AddressValidationResult = {
   hasInferredComponents?: boolean,
   vacant?: boolean
 }
+
+export type Profile = {
```
connorlbark: should this be in `user.ts`?

ui-core/src/types/user.ts
```diff
@@ -0,0 +1,40 @@
...
+    id: string
+    createdAt: number
+    lastUpdatedAt: number
+    enrolleeId: string
+    text: string
+    kitRequestId?: string
+    creatingAdminUserId: string
+}
+
+export type Enrollee = {
```
connorlbark, line 16: let's goooooooooo

ui-core/src/participant/ApiProvider.tsx
```diff
@@ -1,9 +1,22 @@
 import React, { useContext } from 'react'
+import { SurveyResponse } from 'src/types/forms'
+import { StudyEnvParams } from 'src/types/study'
+import { HubResponse } from 'src/types/user'
+import { AddressValidationResult, MailingAddress } from 'src/types/address'
 
 export type ImageUrlFunc = (cleanFileName: string, version: number) => string
 export type SubmitMailingListContactFunc = (name: string, email: string) => Promise<object>
```
connorlbark, line 8: I tried the e2e tests (and then tried it myself) and found that, oddly, I couldn't add myself to the mailing list. It wouldn't error in the participant side, but it didn't show up on the admin side. I tried on dev and it worked.

ui-admin/src/study/surveys/FormOptionsModal.tsx
```diff
@@ -92,6 +92,20 @@ export const FormOptions = ({ studyEnvContext, workingForm, updateWorkingForm }:
                   ...workingForm, autoUpdateTaskAssignments: e.target.checked
                 })}
               /> Auto-update participant tasks to the latest version of this survey after publishing
+            </label>
+            <label className="form-label d-block">
+              <input type="checkbox" checked={(workingForm as Survey).allowAdminEdit}
+                onChange={e => updateWorkingForm({
+                  ...workingForm, allowAdminEdit: e.target.checked
+                })}
+              /> Allow admin completion
```
connorlbark, line 101: The word `completion` feels weird to me, especially when half of the time we really mean `admins can edit this participant's responses, but should only edit one small thing` not `complete it on behalf of them` - Imo, this would make more sense to me as a radio group

```
opt 1 - Participant-only form (e.g., consent form) 
opt 2 - Admin-editable participant survey (e.g., demographics)
opt 3 - Admin-only survey (e.g., medical records)
```
This also eliminates the weird possibility of a form being editable by no one haha

ui-admin/src/study/participants/survey/EnrolleeSurveyView.tsx
```diff
@@ -24,50 +27,69 @@ export default function EnrolleeSurveyView({ enrollee, responseMap, studyEnvCont
...
-        {isEditing ? 'cancel' : 'update / edit'}
-      </button>
+      { configSurvey.survey.allowAdminEdit && <button className="btn btn-secondary"
+        onClick={() => {
+          setIsEditing(!isEditing)
+          if (isEditing) { onUpdate() }
+        }
+        }>
+        {isEditing ?
+          <div><FontAwesomeIcon icon={faX}/> Cancel</div> :
```
connorlbark: To me this feels more like `Complete` or `Finish` button since the changes are already saved - but that's probably more of a UX concern.

ui-participant/src/studies/enroll/PreEnroll.tsx
```diff
@@ -26,6 +29,9 @@ export default function PreEnrollView({ enrollContext, survey }:
     null,
     handleComplete,
     pager,
+    studyEnv.environmentName,
+    undefined,
+    undefined,
```
connorlbark: we actually do need to pass these in, if they exist - they are used to prefill proxy/governed user name in the case of adding a new proxy or signing yourself up as a governed user.

connorlbark (commented):
worked great! a really impressive lift, excited to see the final product. also commenting for now, ran into some bugs. i'll be gone all next week, though, so if devon gives it a thumb then consider it worthy 😄

ui-admin/src/study/surveys/FormOptionsModal.tsx
MatthewBemis: agreed- this will definitely be touched up in a followup UX PR

ui-core/src/surveyUtils.tsx
MatthewBemis: 100% agree that the parameter list on this is way too burdensome. hooks are an interesting idea here, let me see what I can do to clean it up (but i may defer that work to a followup PR though since this branch is super vulnerable to merge conflicts)

ui-admin/src/study/participants/survey/SurveyEditView.tsx
connorlbark: you an also have a commment on the inside 
```
() => {
// todo
}
```

ui-core/src/participant/ApiProvider.tsx
MatthewBemis: I'm unable to reproduce this locally or with the e2e tests.

Devon, do you see this as well?

devonbush (approved):
nice, let's get this merged so we have time to poke at it!

wluken (approved):
Admin survey editing good to merge.

PR #910, opened 2024-05-22
[JN-1027] Run e2e tests in CI

MatthewBemis:

Runs e2e tests against dockerized admin and participant servers. These take about 9 minutes to run despite the Docker build. Not horrible considering they'll only be running on merge to `development` for now.

**Note that one of the tests is disabled in CI!** The test wasn't able to locate the expected screenshots, so it kept failing. I didn't spend a ton of time debugging this. I'd rather have 7 tests working in CI _right now_ than have none and sink more time into this. Hopefully it's a straightforward fix to get that last test working.

* Note that e2e tests passed on this PR
* Run e2e tests locally and confirm they still pass

You can emulate what the github action does locally by using the dockerized versions:

`./scripts/run_dockerized.sh admin`
`./scripts/run_dockerized.sh participant`
`CI=true npm -w e2e-tests test`

Or you can run the traditional way by booting the UIs and API servers manually and then running `npm -w e2e-tests test`

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -3,10 +3,7 @@ name: End-to-end tests
 on:
   push:
     branches:
-      - mb-jn-1027-e2e-ci #TODO switch to development
-  pull_request:
-    branches:
-      - mb-jn-1027-e2e-ci #TODO probably turn this off
+      - development
 
 jobs:
   test:
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -43,9 +43,9 @@ jobs:
       run: npm ci
     - name: Build applications
       run: |
-        REACT_APP_UNAUTHED_LOGIN=true ./gradlew :api-admin:jibDockerBuild -Djib.to.image=api-participant:ci
-        docker run -d -p 8081:8080 --net=host api-participant:ci
+        REACT_APP_UNAUTHED_LOGIN=true ./gradlew :api-participant:jibDockerBuild -Djib.to.image=api-participant:ci
         REACT_APP_UNAUTHED_LOGIN=true ./gradlew :api-admin:jibDockerBuild -Djib.to.image=api-admin:ci
+        docker run -d -p 8081:8080 --net=host api-participant:ci
         docker run -d -p 8080:8080 --net=host api-admin:ci
     - name: Install Playwright Browsers
       run: npx playwright install --with-deps
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -3,10 +3,10 @@ name: End-to-end tests
 on:
   push:
     branches:
-      - mb-jn-1027-dockerized-e2e-ci #TODO
+      - mb-jn-1027-e2e-ci #TODO switch to development
   pull_request:
     branches:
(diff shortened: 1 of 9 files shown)


build.gradle
```diff
@@ -18,6 +18,12 @@ task bundleParticipantUI(type: NpmTask, dependsOn: [npmInstall, buildUICore]) {
     args = ['--workspace=ui-participant', 'run', 'build']
 }
 
+task printVersion {
+    doLast {
+        println gradle.ext.releaseVersion
+    }
+}
+
```
MatthewBemis: oddly, this gradle task wasn't returning anything when running in CI. so i switched to setting the `ci` tag when we do the initial docker build

e2e-tests/src/tests/studies/ourhealth/mailing-list.test.ts
```diff
@@ -1,9 +1,8 @@
 import { test, expect } from '@playwright/test'
-import { adminLogin, getParticipantUrl, randomChars } from 'tests/e2e-utils'
+import { adminLogin, randomChars } from 'tests/e2e-utils'
 
 test('shows mailing list dialog on direct link', async ({ page }) => {
-  const participantUrl = getParticipantUrl('demo')
-  await page.goto(`${participantUrl}/?showJoinMailingList=true `)
+  await page.goto(`${process.env.PARTICIPANT_URL}/?showJoinMailingList=true `)
```
MatthewBemis, line 5: the `PARTICIPANT_URL` env variable takes into account whether or not the test is running in CI. `getParticipantUrl` was hardcoded to assume that this was running locally

e2e-tests/playwright.config.ts
```diff
@@ -178,13 +178,13 @@ export default defineConfig({
       command: 'cd .. && ./gradlew :api-admin:bootRun',
       url: 'http://localhost:8080/status',
       timeout: 180 * 1000,
-      reuseExistingServer: !CI
+      reuseExistingServer: true
```
devonbush, line 181: nice, so this means that we're always taking ownership of spinning up whatever resources we need outside the scope of the playwright tests.  That sounds right to me given the complexity of the setup

.github/workflows/e2e-tests.yml
```diff
@@ -45,11 +43,19 @@ jobs:
       run: npm ci
     - name: Build applications
       run: |
-        ./gradlew :api-admin:build -x test
-        ./gradlew :api-admin:copyWebApp
-        ./gradlew :api-participant:build -x test
-        ./gradlew :api-participant:copyWebApp
+        REACT_APP_UNAUTHED_LOGIN=true ./gradlew :api-participant:jibDockerBuild -Djib.to.image=api-participant:ci
+        REACT_APP_UNAUTHED_LOGIN=true ./gradlew :api-admin:jibDockerBuild -Djib.to.image=api-admin:ci
```
devonbush: oooh, nice find on this `jib.to.image` opt.   How about we update the run_dockerized script to use this, and then just invoke that run_dockerized script here?  (and then we can get rid of that printVersion gradle task).

devonbush (approved):
Yes!!!   one non-blocking suggestion.  Up to you whether you remove the hardcoded push branch trigger in this PR or a follow-up

ssettipalli (approved):

PR #913, opened 2024-05-29
[JN-1063] Replace search facets with search expressions

connorlbark:

Replaces search facets with search expressions in the participant list. Deprecates a ton of code but does not actually delete it. Future work could include showing information about the linked objects that you are searching on, or allow you to conditionally include objects even without filtering on them (for example, grabbing latest kit info).

Notes:
- Filtering still has all the same features
- Adds filtering on enrollee being a subject (defaults to true, but you can change that to find proxies)
- Adds ability to filter using a custom expressions
- Adds documentation to search expressions in zendesk and links to it on the advanced filtering menu as well as on the form options modal for eligibility rules.
- Kit status is no longer a column - but it _is_ searchable. Since the searching API will return the full objects of any objects used in the query, it would be possible to show this column when the user filters on kit status, but I figured that was out of the scope of this PR / subject to closer scrutiny on how we wanted that to work.
- Changes the searchbar to be debounced instead of requiring you to hit enter to search

image

- Play around with participant list - ensure no broken functionality
- Create lots of complex filters - observe that they function as expected and pop up with the correct pills
- Ensure that the zendesk documentation is good


ui-admin/src/study/participants/participantList/facets/EnrolleeSearchFacets.tsx
```diff
@@ -1,110 +1,270 @@
...
+        <label>User type</label>
+        <Select
+          options={subjectOptions}
+          isClearable={true}
+          value={selectedSubjectOption ? selectedSubjectOption : null}
+          onChange={selectedOption => {
+            updateSearchState('subject', selectedOption?.value)
+          }}
+        />
+      </>}
```
connorlbark, line 124: I was not sure if this concept of `User Type` was the best way to introduce this concept to study staff. Open to wording changes here.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeSearchExtService.java
```diff
@@ -52,4 +54,21 @@ public Map<String, SearchValue.SearchValueType> getExpressionSearchFacets(
...
+      String studyShortcode,
+      EnvironmentName envName,
+      String expression) {
+
+    authUtilService.authUserToStudy(operator, portalShortcode, studyShortcode);
+
+    StudyEnvironment studyEnvironment =
+        studyEnvironmentService
+            .findByStudy(studyShortcode, envName)
+            .orElseThrow(() -> new NotFoundException("Study environment not found"));
```
devonbush: since we've already validated the study exists in authUserToStudy, this should be an IllegalStateException

core/src/main/java/bio/terra/pearl/core/dao/search/EnrolleeSearchExpressionDao.java
```diff
@@ -103,12 +104,19 @@ public EnrolleeSearchExpressionResult map(ResultSet rs, StatementContext ctx) th
...
 
-                if (columnName.startsWith("mailing_address_")) {
+                if (columnName.startsWith("mailing_address_") && columnName.endsWith("_created_at")) {
                     enrolleeSearchExpressionResult.setMailingAddress(
                             BeanMapper.of(MailingAddress.class, "mailing_address")
                                     .map(rs, ctx)
                     );
                 }
+
+                if (columnName.startsWith("latest_kit_") && columnName.endsWith("_created_at")) {
```
devonbush: this is four `if`s in a row with extremely similar code -- let's find a way to extract a helper function

core/src/main/java/bio/terra/pearl/core/service/search/terms/LatestKitTerm.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Override
+    public List<EnrolleeSearchQueryBuilder.SelectClause> requiredSelectClauses() {
+        return List.of(
+                new EnrolleeSearchQueryBuilder.SelectClause("latest_kit", kitRequestDao)
+        );
+    }
+
+    @Override
```
devonbush, line 59: can you add a comment here -- I was expecting this to be done with an `order by` and `limit` but I think what you've done here is better because it's less constraining on the rest of the query.

core/src/test/java/bio/terra/pearl/core/dao/search/EnrolleeSearchExpressionDaoTests.java
```diff
@@ -463,6 +471,55 @@ public void testProfileName(TestInfo info) {
 
         Assertions.assertEquals(1, resultsName.size());
         assertTrue(resultsName.stream().anyMatch(r -> r.getEnrollee().getId().equals(jsalkBundle.enrollee().getId())));
+    }
+
+    @Test
+    @Transactional
+    public void testLatestKit(TestInfo info) throws JsonProcessingException {
+        Enrollee e = enrolleeFactory.buildPersisted(getTestName(info));
```
devonbush: e is for error 😄 
```suggestion
        Enrollee enrollee = enrolleeFactory.buildPersisted(getTestName(info));
```

ui-admin/src/study/participants/participantList/facets/EnrolleeSearchFacets.tsx
devonbush: works for me, we can see what demo says

devonbush (approved):
Looks good -- a bunch of non-blocking cleanup suggestions.  Tested a bunch locally and seems solid.  Can you make sure there is a ticket in the upcoming sprint to remove the old search code?

devonbush:
also, instead of using a placeholder in the complex expression form, I would just add an "e.g. ..." label underneath.  I wanted to copy-paste the placeholder to the field, but couldn't.  Also, I missed that `search expression` was a link to the docs.  I'd add an info icon or something else to make it more obvious that there's a help link  there.

connorlbark:
> also, instead of using a placeholder in the complex expression form, I would just add an "e.g. ..." label underneath. I wanted to copy-paste the placeholder to the field, but couldn't. Also, I missed that `search expression` was a link to the docs. I'd add an info icon or something else to make it more obvious that there's a help link there.

I have a ticket to make a "Fancy search expression editor" that will rehaul all of this - if it's ok, I might incorporate this in that ticket.

ui-admin/src/study/participants/participantList/facets/EnrolleeSearchFacets.tsx
```diff
@@ -1,110 +1,270 @@
...
+  const statusOptions = [
+    { label: 'New', value: 'NEW' },
+    { label: 'Created', value: 'CREATED' },
+    { label: 'Queued', value: 'QUEUED' },
+    { label: 'Sent', value: 'SENT' },
+    { label: 'Received', value: 'RECEIVED' },
+    { label: 'Errored', value: 'ERRORED' },
+    { label: 'Deactivated', value: 'DEACTIVATED' },
+    { label: 'Unknown', value: 'UNKNOWN' }
+  ]
```
MatthewBemis, line 240: Nice, I'm glad this isn't a freetext search field anymore!

MatthewBemis (approved):
Overall looks great! 

From testing, I was able to more reliably reproduce that mysterious bug I mentioned to you in person:

https://github.com/broadinstitute/juniper/assets/7257391/8dc13239-4e89-479e-a04a-eb828ce670f2

I don't immediately see what's causing it. (fwiw I'm not able to reproduce it on demo)

PR #925, opened 2024-06-05
[JN-1112] create core family linkage models

connorlbark:

Make sure you agree with how I modeled / linked families and enrollees.

Note that `familyRelationship` is freetext - these relationships will be defined by study staff. I think it would be a can of worms to make it an enum.

- ensure study config can enable and disable family linkage (otherwise no functional behavior)


core/src/main/java/bio/terra/pearl/core/model/participant/RelationshipType.java
```diff
@@ -1,7 +1,7 @@
 package bio.terra.pearl.core.model.participant;
 
 public enum RelationshipType{
-    PROXY;
+    PROXY, FAMILY;
```
connorlbark, line 4: There are many places (afaik only in the UI) that treat _any_ enrollee relation as providing full access; fixing that is the subject of a future PR.

core/src/test/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcherTest.java
```diff
@@ -250,7 +247,7 @@ public void testAssignOnSurveyEvent(TestInfo testInfo) {
         assertThat(participantTaskService.findByEnrolleeId(bundle.enrollee().getId()), hasSize(2));
         tasks = participantTaskService.findByEnrolleeId(bundle.enrollee().getId());
         // now the task should be added
-        assertThat(tasks.stream().map(ParticipantTask::getTargetStableId).toList(), contains("medForm", "followUp"));
+        assertThat(tasks.stream().map(ParticipantTask::getTargetStableId).toList(), containsInAnyOrder("medForm", "followUp"));
```
connorlbark: drive by flaky test fix

core/src/main/java/bio/terra/pearl/core/service/participant/FamilyService.java
```diff
@@ -0,0 +1,50 @@
...
+
+import bio.terra.pearl.core.dao.participant.FamilyDao;
+import bio.terra.pearl.core.model.participant.Family;
+import bio.terra.pearl.core.service.CrudService;
+import bio.terra.pearl.core.service.exception.internal.InternalServerException;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+public class FamilyService extends CrudService<Family, FamilyDao> {
```
devonbush: This should extend DataAuditedService so we can track creation/deletion of families

core/src/main/java/bio/terra/pearl/core/service/participant/FamilyService.java
```diff
@@ -0,0 +1,50 @@
...
+            family.setShortcode(generateShortcode());
+        }
+        Family savedFamily = dao.create(family);
+        return savedFamily;
+    }
+
+    /**
+     * Generate a unique shortcode for a family. Operates exactly like the generateShortcode method in
+     * EnrolleeService.
+     */
```
devonbush: let's pull this into a utlity method, and also let's make sure that Enrollee and Family shortcodes have the same length and allowed chars.   I think the right approach is to have a ShortcodeUtils  class with the LENGTH and ALLOWED_CHARS utility method that takes as an argument a `shortcodeAlreadyExists` String -> boolean function.   

Should we prefix Family shortcodes with `F_` or some similar prefix so they're distinguishable at a glance from Enrollee shortcodes?

devonbush (changes requested):
Overall, agree with the approach.  The blocker is that this needs to be audited.  And we should do some thinking about the best way to handle the shortcodes

core/src/main/java/bio/terra/pearl/core/service/participant/ShortcodeUtilService.java
```diff
@@ -0,0 +1,47 @@
...
+
+import bio.terra.pearl.core.service.exception.internal.InternalServerException;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.Optional;
+import java.util.function.Function;
+
+@Service
+public class ShortcodeUtilService {
```
devonbush: Generally, something is either a utility or a service.  I think of utilities as non-Spring components that encapsulate pure business logic.  So I'd rename this to ShortcodeService to connote that this hits the database  and has some spring stuff injected into it

ui-participant/src/test-utils/test-proxy-environment.tsx
```diff
@@ -1,5 +1,12 @@
...
-import { EnrolleeRelation, ParticipantUser, PortalParticipantUser } from '../api/api'
+import {
+  Enrollee,
+  Portal
+} from '@juniper/ui-core'
+import {
+  EnrolleeRelation,
+  ParticipantUser,
+  PortalParticipantUser
+} from '../api/api'
```
devonbush:
```suggestion
} from 'api/api'
```

devonbush (approved):
nice, and thanks for remembering to update the enrollee withdraw logic!

MatthewBemis (approved):
:shipit:

PR #927, opened 2024-06-06
[JN-1099] Roles/permissions populate refactor + power_user role

MatthewBemis:

Moves roles and permissions into json files that update the database whenever the admin app boots, similar to how language texts work.

Also adds a new `power_user` role with the `tdr_export` permission. The Terra Data Repo page in the sidebar is now gated by this role.

Verify roles and permissions:

* Restart admin API and then check the `role`, `permission`, and `role_permission` database tables to make sure everything looks good
* Modify some of the permissions and roles by updating their descriptions, adding/removing permissions from roles
* Restart admin API
* Verify that the changes were reflected in the DB

Verify that the Terra Data Repo can only be seen by superusers or power_users (note: roles can only be granted by making database queries):

* Repopulate demo
* Log in as power_user@heartdemo.org and confirm that you can see Terra Data Repo in the study sidebar. Viewing the datasets list should not yield any errors
* Log in as staff@heartdemo.org and confirm that you can't see Terra Data Repo in the study sidebar, and going to https://localhost:3000/demo/studies/heartdemo/env/live/export/dataRepo/datasets throws an error telling you that you need the `tdr_export` permission


api-admin/src/main/java/bio/terra/pearl/api/admin/service/AuthUtilService.java
```diff
@@ -76,6 +74,28 @@ public Portal authUserToPortal(AdminUser user, String portalShortcode) {
     throw new NotFoundException("Portal %s not found".formatted(portalShortcode));
   }
 
+  public Portal authUserToPortalWithPermission(
```
MatthewBemis, line 77: I experimented with a few different types of helper methods, but this one was the cleanest and fit in with our existing patterns.

One other option I tried out was a helper function that wraps around the calls that you want to gate behind a permission-check. This is a pattern that I've used in the past on some Scala projects and I like it because it makes it easier to read nested permission checks (we have some of this now with portal and study permissions checking). But I don't know of a way to write it in Java that doesn't result in ugly call-sites. Since this isn't a huge pain point right now and might not be in the future, I didn't push on it, but here's what I came up with if anyone's curious:

Helpers:

```
public <T> T withPortalPermission(AdminUser user, UUID portalId, String permission, Supplier<T> action) {
  if (userHasPermission(user, portalId, permission)) {
    return action.get();
  } else {
    throw new NotFoundException("User does not have required permission on portal, or it may not exist.");
  }
}

private boolean userHasPermission(AdminUser user, UUID portalId, String permission) {
  return adminUserService
      .findByUsernameWithPermissions(user.getUsername())
      .map(
          userWithPermissions ->
              userWithPermissions
                  .portalPermissions()
                  .getOrDefault(portalId, new HashSet<>())
                  .stream()
                  .anyMatch(perm -> perm.equals(permission)))
      .orElse(false);
}
  ```

Usage:

```
return authUtilService.withPortalPermission(user, portal.getId(), "tdr_export", () -> {
  StudyEnvironment studyEnv = studyEnvironmentService.findByStudy(studyShortcode, environmentName).get();
  return dataRepoExportService.listDatasetsForStudyEnvironment(studyEnv.getId());
});
```

File: api-admin/src/main/java/bio/terra/pearl/api/admin/service/ScheduledDataRepoExportService.java [hunk 2]
```diff
@@ -31,8 +31,7 @@ public void pingDataRepoStatus() {
     if (isTdrConfigured()) {
       log.info("Pinging Terra Data Repo. Up: " + dataRepoExportService.getServiceStatus());
     } else {
-      log.warn(
-          "Error: Skipping TDR status ping, as TDR has not been configured for this environment.");
+      log.info("Skipping TDR status ping, as TDR has not been configured for this environment.");
```
MatthewBemis, line 34: Getting rid of "error" in the log message so these don't have to be filtered out of error log queries

api-admin/src/main/java/bio/terra/pearl/api/admin/service/AuthUtilService.java
devonbush: thanks for exploring this! Totally agree that this method signature is the correct one.  This inspired me to create a custom annotation for this method.  Let me know what you think of the PR

api-admin/src/main/java/bio/terra/pearl/api/admin/service/AuthUtilService.java
```diff
@@ -76,6 +74,28 @@ public Portal authUserToPortal(AdminUser user, String portalShortcode) {
...
+      AdminUser user, String portalShortcode, String permission) {
+    Portal portal = authUserToPortal(user, portalShortcode);
+    if (user.isSuperuser()) {
+      return portal;
+    }
+    adminUserService
+        .findByUsernameWithPermissions(user.getUsername())
+        .ifPresent(
```
devonbush: if no user is found, this auth check will succeed. Granted, it's unlikely the user will have been deleted in between the time the AdminUser was fetched and this call, but this is auth code so it's better to be safe.  let's use ifPresentOrElse to throw an exception if there's no user found

api-admin/src/main/java/bio/terra/pearl/api/admin/service/ScheduledDataRepoExportService.java
devonbush: THANK YOU!

api-admin/src/test/java/bio/terra/pearl/api/admin/service/admin/RoleServiceTests.java
```diff
@@ -0,0 +1,34 @@
...
+  public void updateRoleDescription(TestInfo testInfo) {
+    Role initialRole = roleFactory.buildPersisted(getTestName(testInfo));
+    roleService.find(initialRole.getId());
+
+    assertThat(initialRole.getDescription(), equalTo(null));
+
+    initialRole.setDescription("Study staff role");
+    Role updated = roleService.update(initialRole);
+    roleService.find(updated.getId());
+
```
devonbush, line 31: thanks for adding these simple tests!

devonbush (changes requested):
Looks great, and really appreciate the smooth populate functionality.  Just one super-narrow security hole to plug, and an annotation PR to consider

api-admin/src/main/java/bio/terra/pearl/api/admin/service/AuthUtilService.java
MatthewBemis: oh yikes, good catch!

populate/src/main/resources/seed/iam/permissions.json
```diff
@@ -0,0 +1,32 @@
...
+  {
+    "name": "site_content_edit",
+    "displayName": "Site content management",
+    "description": "View, edit and configure the portal website"
+  },
+  {
+    "name": "team_roles_edit",
+    "displayName": "Team role management",
+    "description": "View and edit roles assigned to team members"
+  },
```
connorlbark, line 26: These roles are right now unenforced, right? Is there a followup PR to enforce these?

connorlbark (approved):
For some reason, the initial API restart didn't get the power_user account the power_user role, but after restarting it worked.

populate/src/main/resources/seed/iam/permissions.json
MatthewBemis: you're right- so far these existing permissions have only been enforced in the frontend for hiding/disabling functionality, but there was nothing on the backend actually checking for these permissions. I'll add a ticket to backfill enforcement in the backend

devonbush (approved):
Nice!  As mentioned in the other PR, let's go ahead and merge this without the annotation.  I want to spend a little more time making that more sophisticated in how it can pass the portal context through to the annotated method

MatthewBemis:
@connorlbark good eye! fixed in https://github.com/broadinstitute/juniper/pull/927/commits/97a43000dc5b12864b4e8b1879fc34691f5ee718

PR #929, opened 2024-06-07
[JN-1109] Fancy query builder

connorlbark:

Significantly upgrades the query builder:
- Per demo feedback, if the question has choices, then you can select them in the query builder. This required an overhaul to use `TypeDefinitions` instead of just the type
- You can switch between advanced / simple modes of editing (maybe in the future, we can remove the advanced editor)
- You can see the number of enrollees that this matches. I do feel like it would be cool that be able to click on the `10 matching enrollees` text and get a popup with a table of the enrollees that it matches, just to make sure it's matching the ones that you want, but I figured that was out of the scope of this.

image

Note that there are some questions that we still can't compare with :(. Anything with multiple checkboxes is an "object" type, which doesn't work, and any number ones doesn't work either. I have an idea of how that might work, so I might tackle that in member's choice.

- Restart admin api
- Try out the new query builder in the survey creation tool
- Ensure that, after you create the survey, it works as expected when you go to edit the survey.


core/src/main/java/bio/terra/pearl/core/model/search/SearchValueTypeDefinition.java
```diff
@@ -0,0 +1,22 @@
...
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+@EqualsAndHashCode
+public class SearchValueTypeDefinition {
+    private SearchValue.SearchValueType type;
+    @Builder.Default
+    private List<QuestionChoice> choices = new ArrayList<>();
+    private boolean isMultiValued;
+    private boolean hasOtherQuestion;
```
devonbush: how about renaming these to match the SurveyQuestionDefintion fields?
```suggestion
    private boolean allowMultiple;
    private boolean allowOtherDescription;
```

ui-admin/src/study/participants/participantList/search/AdvancedSearchModal.test.tsx
```diff
@@ -1,10 +1,21 @@
...
+  fireEvent,
+  getByText,
+  render,
+  screen
+} from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import AdvancedSearchModal from './AdvancedSearchModal'
 import { setupRouterTest } from '@juniper/ui-core'
 
+jest.mock('../../../../api/api', () => ({
```
devonbush: I don't think you need the ...
```suggestion
jest.mock('api/api', () => ({
```

devonbush (approved):
Looks good!  couple of small tweak suggestions, nothing blocking

ui-admin/src/study/participants/participantList/search/AdvancedSearchModal.test.tsx
connorlbark: I finally figured out how to change intellij to never import relative 😄 I have no idea why someone would prefer this ugly `../../` chaining... Apologies for this being in almost all of my PRs up until now!

ui-admin/src/search/SearchQueryBuilder.tsx
```diff
@@ -35,6 +52,121 @@ export const SearchQueryBuilder = ({
...
+            studyEnvContext={studyEnvContext}
+            onSearchExpressionChange={onSearchExpressionChange}
+            searchExpression={searchExpression}
+          />
+      }
+    </div>
+
+    <LoadingSpinner isLoading={isLoadingSearchResults}>
+      <span>
+        {searchResults.length} matching enrollee{searchResults.length !== 1 ? 's' : ''}
```
MatthewBemis: nit: check out the `pluralize` function, it can handle modifying the string plurality for you

ui-admin/src/study/surveys/FormOptionsModal.tsx
```diff
@@ -114,27 +114,15 @@ export const FormOptions = ({ studyEnvContext, initialWorkingForm, updateWorking
...
-                })
-              }
-              }/>
+              && <div className="my-2">
+                <LazySearchQueryBuilder
+                  studyEnvContext={studyEnvContext}
+                  onSearchExpressionChange={exp => updateWorkingForm({
+                    ...workingForm, eligibilityRule: exp
+                  })}
+                  searchExpression={(workingForm as Survey).eligibilityRule || ''}/>
```
MatthewBemis: nit: a recent PR cleaned up the type assertion for this, so this can be cleaned up slightly:

```suggestion
                  searchExpression={workingForm.eligibilityRule || ''}/>
```

MatthewBemis (approved):
looks good! Just a few small optional nits

ui-admin/src/search/SearchQueryBuilder.tsx
connorlbark: sweet!

PR #933, opened 2024-06-11
[JN-1099] Adding a custom annotation for permissions

devonbush:

This follows up the previous PR with affordances made for the service methods to access the objects retrieved during the authorization checks (e.g. the Portal or StudyEnvironment objects).  So, e.g.

```
@EnforcePortalStudyEnvPermission(permission = "tdr_export")
  public List<Dataset> listDatasetsForStudyEnvironment(PortalStudyEnvAuthContext authContext) {
    return dataRepoExportService.listDatasetsForStudyEnvironment(
        authContext.getStudyEnvironment().getId());
  }
```
here, the authContext is pased from the cotnroller with just the AdminUser and the shortcode parameters. The annotation then hydrates the authContext with the studyEnvironment as part of the auth checks. This definitely adds a new layer of automagic.  But I think it's worth it for stripping out a ton of boilerplate.  It let me notice a couple of bugs in the existing datarepoExtService where we weren't authing the dataset to the specified portal.   I'm open to lots of suggestions for how to make it clearer what does/doesn't get "magically" populated in the authContext.

The general expected auth pattern is now:

1. Controller method invokes AuthUtilService.requireAdminUser(request) to confirm the token on the request corresponds to a user in the database
2. Controller method creates an authContext object with the AdminUser object from requireAdminUser call, and any params
3. Controller method invokes an ExtService method, passsing the authContext object as the first argument
4. The ExtService method is Annotatated with @EnforcePortalPermission(permission = "[[permissionName]]")
5. That annotation confirms the user has the given permission for the specified portal (or is a superuser.  If so, the annotation will populate the authContext object with the Portal, PortalStudy, and/or StudyEnvironment objects retrieved during the authorization process, so they are available to the service method.

Note that we *could* put this annotation around the Controller methods, but I decided against that because
1. we don't want an authContext argument to our controller methods, since that would show up in our openApi docs
2. Putting auth on ExtServices means testing auth doesn't requires mocking a request, and I like that splitting out the extServices gives more mocking options for dependent services.


api-admin/src/main/java/bio/terra/pearl/api/admin/service/DataRepoExportExtService.java
```diff
@@ -27,12 +29,12 @@ public DataRepoExportExtService(
     this.studyEnvironmentService = studyEnvironmentService;
   }
 
+  @EnforcePortalPermission(permission = "tdr_export")
```
MatthewBemis: I really like the idea of using an annotation for this. It makes it super easy to spot which permissions are required, and as our roles and permissions become more widely used, it makes it much easier to spot when a permission check is missing.

Let's see how this one goes, but I do think it would be slick to have other auth annotations like `@EnforceStudyAuthed`

api-admin/src/main/java/bio/terra/pearl/api/admin/service/DataRepoExportExtService.java
```diff
@@ -27,12 +29,12 @@ public DataRepoExportExtService(
...
   public List<Dataset> listDatasetsForStudyEnvironment(
       String portalShortcode,
       String studyShortcode,
       EnvironmentName environmentName,
       AdminUser user) {
-    authUtilService.authUserToPortalWithPermission(user, portalShortcode, "tdr_export");
     authUtilService.authUserToStudy(user, portalShortcode, studyShortcode);
```
connorlbark: I do really like this syntax, but I feel like we often use these auth methods like:

```
Portal portal = authUtilService.authUserToPortal<WithPermission>(user,...)
.... use portal somehow ....
```

Because of that, I'm not 100% in love that we would then have two places where authentication could happen (direct authUtilService calls and annotations). If this is a problem, we could just go full annotations for authentication, which would then maybe mean a couple more db calls to grab the portal/study in the cases we need it. But maybe it's not a problem and it's ok to do in two places!

MatthewBemis (approved):
pretty sweet- let's roll with it!

connorlbark (approved):
approving because I could go either way on my concerns below!

api-admin/src/main/java/bio/terra/pearl/api/admin/service/DataRepoExportExtService.java
devonbush: that's a great point -- I'll hold off on merging this for now, and take a look at how an annotation might also populate an "PortalAuthContext" object that the method could access, because you're right that there are lots of cases where we want that Portal object, and it's very important that that portal object is the same one we authed to.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/DataRepoExportExtService.java
```diff
@@ -27,62 +26,37 @@ public DataRepoExportExtService(
...
-    authUtilService.authUserToStudy(user, portalShortcode, studyShortcode);
-
-    StudyEnvironment studyEnv =
-        studyEnvironmentService.findByStudy(studyShortcode, environmentName).get();
-
-    return dataRepoExportService.listDatasetsForStudyEnvironment(studyEnv.getId());
+  @EnforcePortalStudyEnvPermission(permission = "tdr_export")
+  public List<Dataset> listDatasetsForStudyEnvironment(PortalStudyEnvAuthContext authContext) {
+    return dataRepoExportService.listDatasetsForStudyEnvironment(
+        authContext.getStudyEnvironment().getId());
```
connorlbark, line 32: Nice - this is clean! I like the `authContext` at the ext service level a lot - breaking it down for service methods is clean (and I feel harder to mess up than our current flows).

connorlbark (approved):

MatthewBemis (approved):

PR #936, opened 2024-06-11
[JN-1129] Synthetic family data

connorlbark:

Adds the infrastructure for families to populate and adds a synthetic family to demo.

- repopulate demo
- check site settings and observe that family linkage is enabled
- execute 
   ```sql
    select f.shortcode as family_shortcode,
           e.shortcode,
           p.given_name,
           s.shortcode as study_shortcode,
           se.environment_name
    from family f
    inner join family_enrollee fe on f.id = fe.family_id
    inner join enrollee e on fe.enrollee_id = e.id
    inner join profile p on e.profile_id = p.id
    inner join study_environment se on e.study_environment_id = se.id
    inner join study s on se.study_id = s.id
   ```
   to see who is in what family - should just be 1 family, the salks


devonbush (changes requested):
as discussed, I think we need to support an enrollee being in multiple families, which unfortunately changes a lot of this code :(

core/src/main/java/bio/terra/pearl/core/service/participant/FamilyEnrolleeService.java
```diff
@@ -20,15 +22,23 @@ public FamilyEnrolleeService(FamilyEnrolleeDao familyEnrolleeDao,
...
+        bulkDelete(objs, info);
+    }
 
+    @Transactional
+    public void deleteByEnrolleeId(UUID id, DataAuditInfo info) {
+        List<FamilyEnrollee> objs = dao.findByEnrolleeId(id);
 
+        bulkDelete(objs, info);
+    }
 }
```
connorlbark: Is there a better way to do this in an audited fashion? I don't imagine it's performance sensitive but it was easy to forget to do the auditing (and even get it approved and merged in)

populate/src/main/resources/seed/portals/demo/participants/jtsalk.json
```diff
@@ -0,0 +1,20 @@
+{
+  "participantUser": {
+    "username": "jtsalk@test.com",
+    "lastLoginHoursAgo": 5
+  },
+  "profile": {
+    "givenName": "Jonathan",
```
devonbush: Can we give him a first name that doesn't share 3 letters with Jonas, and an email that's a bit more visually distinguishable? I typically log in with chrome autocompleting the username, and I could easily see myself mixing up these logins

populate/src/main/java/bio/terra/pearl/populate/dto/participant/FamilyPopDto.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.List;
+
+@Getter @Setter @NoArgsConstructor
+public class FamilyPopDto extends Family {
+    private String probandShortcode;
+    private List<String> memberShortcodes;
+    private List<FamilyRelationship> relations;
+
+    public static record FamilyRelationship(String shortcode, String targetShortcode, String relationship) {
```
devonbush:
```suggestion
    public static record FamilyRelationshipPopDto(String shortcode, String targetShortcode, String relationship) {
```

core/src/main/java/bio/terra/pearl/core/service/participant/FamilyEnrolleeService.java
devonbush: confirming what we just talked about, these operations did not technically need to be audited, as they only occur in the context of enrollee deletion or study repopulation

core/src/main/java/bio/terra/pearl/core/service/participant/FamilyEnrolleeService.java
```diff
@@ -20,15 +22,23 @@ public FamilyEnrolleeService(FamilyEnrolleeDao familyEnrolleeDao,
...
+        List<FamilyEnrollee> objs = dao.findByFamilyId(familyId);
+
+        bulkDelete(objs, info);
     }
 
     @Transactional
-    public void deleteByEnrolleeId(UUID enrolleeId) {
-        dao.deleteByEnrolleeId(enrolleeId);
-    }
+    public void deleteByStudyEnvironmentId(UUID studyEnvironmentId, DataAuditInfo info) {
```
devonbush: this should definitely not be audited--this only happens in the context of study deletion

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeRelationService.java
```diff
@@ -127,4 +128,19 @@ public List<EnrolleeRelation> filterValid(List<EnrolleeRelation> enrolleeRelatio
...
+    }
+
+    @Transactional
+    public void deleteByFamilyId(UUID id, DataAuditInfo info) {
+        List<EnrolleeRelation> relations = findRelationsForFamily(id);
+        bulkDelete(relations, info);
+    }
+
+    @Transactional
+    public void deleteByStudyEnvironmentId(UUID studyEnvironmentId, DataAuditInfo info) {
```
devonbush: this should not be an audited operation -- only occurs in the case of study deletion

devonbush (changes requested):
looks good!  Just remove the auditing from the studyenvironment deletion operations, and then a couple suggestions to avoid family member typos going forward

populate/src/main/resources/seed/portals/demo/participants/jtsalk.json
connorlbark: Apparently he has another kid named Darrell! I'll go with that instead 😄

core/src/main/java/bio/terra/pearl/core/service/participant/FamilyEnrolleeService.java
connorlbark: Before this PR, I noticed that there was a delete happening at the study env level that just did:
```
        enrolleeRelationService.deleteByStudyEnvironmentId(studyEnvironmentId, null);
```
I think I slightly prefer having every delete be audit-able but passing null in the specific cases where we don't want it. That way we don't accidentally use an unaudited function without thinking about it. I could also see just simply labelling it with `deleteByStudyEnvironmentIdNotAudited` - verbose, but gets the message across. What do you think?

devonbush: hmmm... I think I like that the deleteByStudyEnvId doesn't take an audit context, since that's how it's communicating that we don't want the caller to audit that operation.   Agree that there's some danger of copy-pasting that method and creating a new one that should be audited, but isn't.  But I think that's better than getting in the habit of passing `null` to audit infos.  

Awhile ago, I toyed with the idea of having separate services for the "illegal/unaudited" operations.  So e.g. there would be a separate "EnrolleeRelationUnauditedService" that would contain the deleteByStudyEnvId method.  But I couldn't get the ergonomics to work out -- downstream services ended up needing to include both the unaudited and audited versions, and it was a dependency mess.  But something like that might be worth revisiting

core/src/main/java/bio/terra/pearl/core/service/study/StudyEnvironmentService.java
```diff
@@ -110,6 +119,9 @@ public StudyEnvironment create(StudyEnvironment studyEnv) {
     @Override
     public void delete(UUID studyEnvironmentId, Set<CascadeProperty> cascade) {
         StudyEnvironment studyEnv = dao.find(studyEnvironmentId).get();
+        enrolleeRelationService.deleteByStudyEnvironmentId(studyEnvironmentId, null);
```
devonbush: as discussed, I don't like passing in null here

devonbush (changes requested):

core/src/main/java/bio/terra/pearl/core/service/participant/FamilyEnrolleeService.java
connorlbark: per discussion, i refactored all remaining deletebystudyenv methods to not take in dataauditinfo

core/src/main/java/bio/terra/pearl/core/dao/participant/FamilyDao.java
```diff
@@ -25,10 +25,19 @@ public Optional<Family> findOneByShortcode(String shortcode) {
...
+    }
+
     public List<Family> findByEnrolleeId(UUID enrolleeId) {
         return jdbi.withHandle(handle -> handle.createQuery("SELECT family.* FROM family family INNER JOIN family_enrollee family_enrollee ON family_enrollee.family_id = family.id WHERE family_enrollee.enrollee_id = :enrolleeId")
                 .bind("enrolleeId", enrolleeId)
                 .mapToBean(Family.class)
                 .list());
     }
+
+    // WARNING: This method is not audited
```
MatthewBemis: can you add an explanation for why? I know why based on the context in this PR, but would be nice to see while perusing code

MatthewBemis (approved):

devonbush (approved):

PR #938, opened 2024-06-12
[JN-1112] link enrollee to multiple families via join table

connorlbark:

Per discussion, adds join table and replaces `familyId` with a list of families on the enrollee.

- make sure models look good

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/participant/FamilyDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/participant/FamilyDao.java
@@ -26,7 +26,7 @@ public class FamilyDao extends BaseMutableJdbiDao<Family> {
     }
 
     public List<Family> findByEnrolleeId(UUID enrolleeId) {
-        return jdbi.withHandle(handle -> handle.createQuery("SELECT f.* FROM family f INNER JOIN family_members fm ON fm.family_id = f.id WHERE fm.enrollee_id = :enrolleeId")
+        return jdbi.withHandle(handle -> handle.createQuery("SELECT f.* FROM family f INNER JOIN family_member fm ON fm.family_id = f.id WHERE fm.enrollee_id = :enrolleeId")
                 .bind("enrolleeId", enrolleeId)
                 .mapToBean(Family.class)
                 .list());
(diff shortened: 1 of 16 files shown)


core/src/main/java/bio/terra/pearl/core/dao/participant/FamilyDao.java
```diff
@@ -22,4 +24,11 @@ protected Class<Family> getClazz() {
     public Optional<Family> findOneByShortcode(String shortcode) {
         return findByProperty("shortcode", shortcode);
     }
+
+    public List<Family> findByEnrolleeId(UUID enrolleeId) {
+        return jdbi.withHandle(handle -> handle.createQuery("SELECT f.* FROM family f INNER JOIN family_member fm ON fm.family_id = f.id WHERE fm.enrollee_id = :enrolleeId")
```
devonbush: `f` and `fm` aren't great distinguishers.  I would either not abbreviate, or just use "family" and "member"

core/src/main/resources/db/changelog/changesets/2024_06_12_family_linkage_modified.yaml
```diff
@@ -0,0 +1,26 @@
+databaseChangeLog:
+  - changeSet:
+      id: "multiple_family_linkage"
+      author: connorlbark
+      changes:
+        - dropColumn:
+            tableName: enrollee
+            columnName: family_id
+        - createTable:
+            tableName: family_member
```
devonbush: Our convention is to have join tables be named as boringly as possible.  let's call this family_enrollee, or enrollee_family

devonbush (changes requested):
Structure looks solid, just rename the table to match our boring join convention

core/src/main/resources/db/changelog/changesets/2024_06_12_family_linkage_modified.yaml
connorlbark: I chose `FamilyMember` because `FamilyEnrollee` sounds like it is `a family that is signed up as an enrollee` and `EnrolleeFamily` sounds like it is the actual `Family` object, and I was torn what to do. I see your point about following convention, though, so I don't mind following it

devonbush (approved):

MatthewBemis (approved):

PR #939, opened 2024-06-12
[JN-1133] fix repopulate fk error

MatthewBemis:

Fixes a populate bug

* Register a user in any portal
* While logged in, join the mailing list
* Confirm that you can repopulate that portal without running into any FK constraint errors

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/portal/PortalEnvironmentService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/portal/PortalEnvironmentService.java
@@ -137,11 +137,11 @@ public class PortalEnvironmentService extends CrudService<PortalEnvironment, Por
         portalParticipantUserService.deleteByPortalEnvironmentId(id);
         // clean up any preregistration responses not associated with a user
         preregistrationResponseDao.deleteByPortalEnvironmentId(id);
+        mailingListContactService.deleteByPortalEnvId(id);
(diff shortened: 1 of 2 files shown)


populate/src/main/resources/seed/portals/hearthive/portal.json
```diff
@@ -56,7 +56,6 @@
       "password" : "broad_institute",
       "acceptingRegistration" : true,
       "initialized" : true,
-      "participantHostname" : "thehearthive.org",
```
MatthewBemis, line 59: unrelated, but I don't think this should be here- I almost registered a user in hearthive prod because my local UI pointed me there 🙃

core/src/main/java/bio/terra/pearl/core/service/portal/PortalEnvironmentService.java
```diff
@@ -137,11 +137,11 @@ public void delete(UUID id, Set<CascadeProperty> cascades) {
         portalParticipantUserService.deleteByPortalEnvironmentId(id);
         // clean up any preregistration responses not associated with a user
         preregistrationResponseDao.deleteByPortalEnvironmentId(id);
+        mailingListContactService.deleteByPortalEnvId(id);
```
MatthewBemis, line 140: mailing list records can reference participant user IDs, so the mailing list records need to be deleted first. we never really ran into this before because it's rare that a participant is signed in and then joins the mailing list

devonbush (approved):

connorlbark (approved):

PR #940, opened 2024-06-12
[JN-1132]  fix i18n save of SiteContents

devonbush:

Previously, we only fetched one language at a time for display in the admin tool.  This caused all sorts of fun loading and saving issues.  This changes the loading behavior to fetch all languages on initial load, which makes the save a simpler operation.

Also, the SiteContentEditor no longer relies on the portalEnvContext having a siteContent loaded on it -- instead a new `getCurrent` siteContent endpoint is added so that the component can just say "get me the siteContent for this environment"

1. repopulate demo
2. redeploy ApiAdminApp
3. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/siteContent
4. make changes to the english content
5. make changes to the spanish content
6. go back to english view, confirm your changes still appear
7. save
8. confirm english and spanish changes all still appear
9. use the history menu to go back to a prior version
10. confirm the prior version renders correctly


connorlbark (approved):
woo! this has been bothering me for a while, works flawlessly now.

MatthewBemis (approved):

PR #941, opened 2024-06-13
[JN-1132] add language content to siteContent editor

devonbush:
Adds the ability to clone the default language LocalSiteContent to a new language.  If the user selects a language that doesn't have content, a 'clone' button appears.

image

This also adds some safety to the participant side -- if you select a language that has no local site content, a warning appears and you get an option to reload with the default (previously, you were stuck with a permanent JS error and there was no way to unset the language without going into Chrome local storage).

image

1. repopulate ourhealth
2. add a new supported language 
```
insert into portal_environment_language values (gen_random_uuid() ,now(), now(), 'de', 'German', (select id from portal_environment where portal_id = (select id from portal where shortcode = 'ourhealth') and environment_name = 'sandbox'));
```
3. go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/siteContent
4. select the German language
5. press the "clone from default" button
6. make some changes to the German site
7. save
8. confirm the German

### DIFF
--- a/ui-participant/src/App.tsx
+++ b/ui-participant/src/App.tsx
@@ -49,7 +49,24 @@ function App() {
   const [cookiesAcknowledged, setCookiesAcknowledged] = useCookiesAcknowledged()
   const { localContent, portal, portalEnv } = usePortalEnv()
 
+  useEffect(() => {
+    const isCompatible = isBrowserCompatible()
+    if (!isCompatible) {
+      alert('Your browser does not support this page. ' +
+        'Please use the latest version of Chrome, Safari, Firefox, Edge, or Android')
+    }
+  }, [])
+
+
   const brandConfig: BrandConfiguration = {}
+  if (!localContent) {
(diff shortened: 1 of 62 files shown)


ui-participant/src/App.tsx
```diff
@@ -49,7 +49,24 @@ function App() {
...
+        'Please use the latest version of Chrome, Safari, Firefox, Edge, or Android')
+    }
+  }, [])
+
+
   const brandConfig: BrandConfiguration = {}
+  if (!localContent) {
+    return <div className="alert alert-warning">
+      No content has been configured for this language.
+      <button onClick={
```
MatthewBemis: Style nit:

```suggestion
      <button className='btn btn-outline-secondary ms-2' onClick={
```

MatthewBemis (approved):
works great! Thanks for adding all of these i18n site editor improvements

connorlbark (approved):

PR #942, opened 2024-06-14
[JN-1145] make search expressions even cooler

connorlbark:

If you all feel it would be easier to split this up into different PRs, I can do that - I just figured it might be easier to just lump it all together since they're all so related (& a lot of the code changes are tests/boilerplate/regenerated code).

Adds a laundry list of new features:
- Adds family term - and also support for arrays, since families are the first 1-many type supported in expressions
- You can not with !: `!({answer.basics.countryOfOrigin} = 'us' or {answer.basics.countryOfOrigin} = 'ca')`
- Function parsing, with a few basic useful ones (`lower`, `trim`, `min`, and `max`). Functions take in terms and are themselves terms, so you can nest them as much as you like!
    - Examples: `max({answer.s1.q1}, {answer.s2.q2}, 4) = 4` `trim(lower({enrollee.name})) contains 'jonas'`
    - I do not see an obvious way to implement this with react-querybuilder, so if an expression contains a function (or any parsing errors), it will force you to use the advanced editor
    - Right now, functions cannot take expressions (`{enrollee.shortcode} = 'asdf'`) as an argument - only terms (`'asdf'` or `{enrollee.shortcode}`). Potentially we could implement functions that take only expressions as arguments... but in my very limited antlr knowledge it seemed complex™ to support both expressions and terms intermixed as arguments. 99% of functions we care about just take terms, so I think it's OK.
- ~Using `lower`, basic participant search is now case insensitive!~ With devon's PR, this has been reverted
- More type intelligence in both frontend and backend parsing (e.g., if you attempt to use a function on the wrong type, it will error out with a useful message)
- Removal of integer type (we only really need doubles)

- Observe tests
- Try out new features (!, functions) in the participant list
- Try doing basic keyword searching but _case insensitive_ 
- Cause errors and try to break the search query builder - it has some type/facet intelligence and should be more robust than before
- Make sure it blocks basic view when you use functions as expected


core/src/main/java/bio/terra/pearl/core/dao/portal/PortalDao.java
```diff
@@ -125,4 +125,15 @@ public Optional<Portal> findByPortalEnvironmentId(UUID portalEnvironmentId) {
                         .mapTo(Portal.class)
                         .findOne());
     }
+
+    public Optional<Portal> findByStudyEnvironmentId(UUID studyEnvId) {
+        return jdbi.withHandle(handle ->
+                handle.createQuery("select * from " + tableName + " p " +
+                                " inner join portal_environment pe on pe.portal_id = p.id " +
+                                " inner join study_environment se on se.portal_environment_id = pe.id " +
```
devonbush: huh?  study_environment doesn't have a portal_environment_id column

core/src/main/java/bio/terra/pearl/core/service/search/terms/functions/LowerFunction.java
```diff
@@ -0,0 +1,66 @@
...
+import bio.terra.pearl.core.model.search.SearchValueTypeDefinition;
+import bio.terra.pearl.core.service.search.EnrolleeSearchContext;
+import bio.terra.pearl.core.service.search.sql.EnrolleeSearchQueryBuilder;
+import bio.terra.pearl.core.service.search.terms.SearchTerm;
+import bio.terra.pearl.core.service.search.terms.SearchValue;
+import org.jooq.Condition;
+
+import java.util.List;
+import java.util.Optional;
+
```
devonbush, line 12: We need to have some docs here -- either an example or a link to a zendesk page

core/src/main/java/bio/terra/pearl/core/service/search/terms/functions/MaxFunction.java
```diff
@@ -0,0 +1,32 @@
+package bio.terra.pearl.core.service.search.terms.functions;
+
+import bio.terra.pearl.core.model.search.SearchValueTypeDefinition;
+import bio.terra.pearl.core.service.search.terms.SearchTerm;
+import bio.terra.pearl.core.service.search.terms.SearchValue;
+
+import java.util.List;
+
```
devonbush, line 8: same, docs or zendesk link

core/src/main/java/bio/terra/pearl/core/service/search/terms/functions/VarArgFunction.java
```diff
@@ -0,0 +1,57 @@
...
+
+import bio.terra.pearl.core.service.search.EnrolleeSearchContext;
+import bio.terra.pearl.core.service.search.sql.EnrolleeSearchQueryBuilder;
+import bio.terra.pearl.core.service.search.terms.SearchTerm;
+import bio.terra.pearl.core.service.search.terms.SearchValue;
+import org.jooq.Condition;
+
+import java.util.List;
+import java.util.Optional;
+
```
devonbush, line 11: could use a class comment

devonbush (approved):
code looks good, could use docs on the functions.  I like the extra handling for integers/string comparisons

core/src/main/java/bio/terra/pearl/core/dao/portal/PortalDao.java
connorlbark: Oops 🤦 I don't remember writing this and it was never used

core/src/main/java/bio/terra/pearl/core/service/search/terms/functions/LowerFunction.java
connorlbark: I'm working on zendesk docs right now! I'll also add an example but we will have that to go from as well.

MatthewBemis (approved):

PR #943, opened 2024-06-17
[JN-1134] add env language

devonbush:

This enables editing of the PortalLanguages from the admin tool.  This copies a fair bit of logic from the AnswerMapping table editor, and moves in the direction of genericizing some of the functionality.  I think the next time we do an editable list will be when we should really focus on pulling out general purpose "editable list" hooks

image

This also adds portal languages to the publishing flow

image

1. redeploy apiAdminApp
2. repopulate demo
3. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/settings
4. add/remove languages
5. save languages
6. go to https://sandbox.ourhealth.localhost:3001/  -- confirm the updated list appears
7. go to https://localhost:3000/demo/studies/heartdemo/publishing
8. choose "copy" from sandbox to irb
9. select one or more of the language changes to copy
10. press "Copy Changes"
11. confirm IRB environment has updated language list

### DIFF
--- a/api-admin/src/test/java/bio/terra/pearl/api/admin/service/siteContent/SiteContentExtServiceTests.java
+++ b/api-admin/src/test/java/bio/terra/pearl/api/admin/service/siteContent/SiteContentExtServiceTests.java
@@ -12,7 +12,7 @@ import bio.terra.pearl.core.model.portal.PortalEnvironment;
 import bio.terra.pearl.core.model.portal.PortalEnvironmentLanguage;
 import bio.terra.pearl.core.model.site.LocalizedSiteContent;
 import bio.terra.pearl.core.model.site.SiteContent;
-import bio.terra.pearl.core.service.portal.PortalLanguageService;
+import bio.terra.pearl.core.service.portal.PortalEnvironmentLanguageService;
 import bio.terra.pearl.core.service.site.SiteContentService;
 import java.util.List;
 import org.junit.jupiter.api.Test;
@@ -25,20 +25,20 @@ public class SiteContentExtServiceTests extends BaseSpringBootTest {
   @Autowired private PortalEnvironmentFactory portalEnvironmentFactory;
   @Autowired private SiteContentFactory siteContentFactory;
   @Autowired private SiteContentService siteContentService;
-  @Autowired private PortalLanguageService portalLanguageService;
+  @Autowired private PortalEnvironmentLanguageService portalEnvironmentLanguageService;
 
   @Test
   @Transactional
   public void testGetLoadAllLanguages(TestInfo testInfo) {
     PortalEnvironment portalEnv =
         portalEnvironmentFactory.buildPersisted(getTestName(testInfo), EnvironmentName.sandbox);
-    portalLanguageService.create(
+    portalEnvironmentLanguageService.create(
         PortalEnvironmentLanguage.builder()
             .languageCode("es")
             .languageName("Spanish")
             .portalEnvironmentId(portalEnv.getId())
             .build());
-    portalLanguageService.create(
+    portalEnvironmentLanguageService.create(
         PortalEnvironmentLanguage.builder()
             .languageCode("en")
             .languageName("English")
--- a/ui-admin/src/portal/PortalEnvConfigView.tsx
+++ b/ui-admin/src/portal/PortalEnvConfigView.tsx
@@ -12,7 +12,10 @@ import useUpdateEffect from '../util/useUpdateEffect'
 import useReactSingleSelect from '../util/react-select-utils'
 import { PortalEnvironmentLanguage } from '@juniper/ui-core'
 import Select from 'react-select'
-import { usePortalLanguage } from './usePortalLanguage'
+import { usePortalLanguage } from './languages/usePortalLanguage'
+import PortalEnvLanguageEditor from './languages/PortalEnvLanguageEditor'
+import _cloneDeep from 'lodash/cloneDeep'
+import InfoPopup from '../components/forms/InfoPopup'
 
 
 type PortalEnvConfigViewProps = {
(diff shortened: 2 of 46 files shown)


ui-admin/src/portal/PortalEnvConfigView.tsx
```diff
@@ -58,60 +73,80 @@ const PortalEnvConfigView = ({ portalContext, portalEnv }: PortalEnvConfigViewPr
...
+        {isLoading && <LoadingSpinner/>}
+        {!isLoading && 'Save website config'}
+      </Button>
+    </form>
+    <form className="mt-5">
+      <div>
+        <h4 className="h5">
+          Supported languages <InfoPopup content={'This list determines which languages appear in the dropdown' +
+          'of the participant view.  It is up to you to create the content for that language.'}/>
+        </h4>
```
connorlbark, line 138: For some reason, clicking this tooltip refreshes the page and you can't read it :'(

ui-admin/src/portal/languages/PortalEnvLanguageEditor.tsx
```diff
@@ -0,0 +1,230 @@
...
+    [items, newItem])
+
+  const table = useReactTable<PortalEnvironmentLanguageRow>({
+    data,
+    columns,
+    getCoreRowModel: getCoreRowModel()
+  })
+
+  return <div className='px-3 pt-1'>
+    {basicTableLayout(table, { tdClass: 'col-1 ' })}
```
connorlbark, line 207: Love the answer mapping-style table becoming more common!

From a UX perspective, I feel like the whole table being viewable at the width of the screen dominates the page visually when you scroll down. I wonder if it would be better to have the table only show up if you go into an edit mode on the languages, otherwise it has a more condensed view (and the save languages button disabled or not visible). But I feel like the whole page needs UX love, so I'm cool either way

ui-admin/src/portal/publish/PortalEnvDiffView.tsx
```diff
@@ -101,6 +108,8 @@ export default function PortalEnvDiffView(
     })
   }
 
+  // @ts-ignore
+  // @ts-ignore
```
connorlbark: nit: accidental?

ui-admin/src/portal/languages/PortalEnvLanguageEditor.tsx
```diff
@@ -0,0 +1,230 @@
...
+          return row.original.isEditing && <input
+            type="text"
+            aria-label={'Language code'}
+            onChange={e => setNewItem({
+              ...row.original,
+              languageName: e.target.value,
+              isEditing: true
+            })}
+            value={row.original.languageCode}
+          />
```
connorlbark: Nit: needs styling.

Also, editing this didn't work for me; I think it should set `languageCode` instead of `languageName`.

connorlbark (changes requested):
Ran into some bugs + some UX nits, but otherwise excited to have more UI-editability on this!!

ui-admin/src/portal/publish/PortalEnvDiffView.tsx
devonbush: good catch

ui-admin/src/portal/languages/PortalEnvLanguageEditor.tsx
devonbush: made a change to make the table half-width on big screens so it doesn't dominate as much -- good call.   Agree we can revisit the whole page at a later time

ui-admin/src/portal/PortalEnvConfigView.tsx
devonbush: ah, another button not of type "button" problem -- good catch

api-admin/src/test/java/bio/terra/pearl/api/admin/service/siteContent/SiteContentExtServiceTests.java
```diff
@@ -25,20 +25,20 @@ public class SiteContentExtServiceTests extends BaseSpringBootTest {
   @Autowired private PortalEnvironmentFactory portalEnvironmentFactory;
   @Autowired private SiteContentFactory siteContentFactory;
   @Autowired private SiteContentService siteContentService;
-  @Autowired private PortalLanguageService portalLanguageService;
+  @Autowired private PortalEnvironmentLanguageService portalEnvironmentLanguageService;
```
MatthewBemis, line 28: Thanks for renaming this and the DAO

ui-admin/src/portal/languages/PortalEnvLanguageEditor.tsx
```diff
@@ -0,0 +1,230 @@
...
+          return row.original.isEditing && <input
+            type="text"
+            aria-label={'Language code'}
+            onChange={e => setNewItem({
+              ...row.original,
+              languageCode: e.target.value,
+              isEditing: true
+            })}
+            value={row.original.languageCode}
+          />
```
MatthewBemis: this `input` doesn't behave particularly well for me. I type one character, and then I lose focus and have to click back into the field to add another character.

MatthewBemis: Also can you switch this to use `TextInput` instead, for styling consistency?

```
<TextInput
  type="text"
  aria-label={'Language code'}
  onChange={value => setNewItem({
    ...row.original,
    languageCode: value,
    isEditing: true
  })}
  value={row.original.languageCode}
/>
```

MatthewBemis (approved):
Looks good! just a few small comments

connorlbark (approved):

ui-admin/src/portal/languages/PortalEnvLanguageEditor.tsx
devonbush: good catch, was a problem with how I had the columns useMemo dependencies.  Also added a limit of 2 chars for the field

PR #944, opened 2024-06-17
[JN-1135] adding SandboxOnly annotation

devonbush:

Adds an `@SandboxOnly` annotation for methods that should be prevented from running in non-sandbox environments. (which forces users to use the publishing workflow for some changes).

1.  I don't think there's a way to test this via the UI (at least not until we add this to other methods), so just take a look at the unit tests and confirm they make sense

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/auth/BaseEnforcePermissionAspect.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/auth/BaseEnforcePermissionAspect.java
@@ -1,14 +1,17 @@
 package bio.terra.pearl.api.admin.service.auth;
 
+import bio.terra.pearl.api.admin.service.auth.context.PortalAuthContext;
 import java.lang.annotation.Annotation;
 import org.apache.commons.lang3.NotImplementedException;
+import org.aspectj.lang.JoinPoint;
 import org.aspectj.lang.ProceedingJoinPoint;
 
 /** base class for annotations enforcing permissions around methods. */
 public abstract class BaseEnforcePermissionAspect<
     T extends PortalAuthContext, A extends Annotation> {
+  private static final String ANNOTATION_NAME = "EnforcePermission";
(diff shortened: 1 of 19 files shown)


MatthewBemis (approved):
very nice. lgtm!

api-admin/src/main/java/bio/terra/pearl/api/admin/service/auth/BaseEnforcePermissionAspect.java
```diff
@@ -1,14 +1,17 @@
...
+import org.aspectj.lang.JoinPoint;
 import org.aspectj.lang.ProceedingJoinPoint;
 
 /** base class for annotations enforcing permissions around methods. */
 public abstract class BaseEnforcePermissionAspect<
     T extends PortalAuthContext, A extends Annotation> {
+  private static final String ANNOTATION_NAME = "EnforcePermission";
   private static final String AUTH_CONTEXT_USAGE_ERROR =
-      "EnforcePortalPermission annotation must be used on a method whose first argument is a %s";
+      "%s annotation must be used on a method whose first argument is a %s";
```
connorlbark, line 14: Would it be possible to have `@SandboxOnly` and also `@EnforcePortalPermission` at the same time? Since we're using argument types we know (`<something>AuthContext`), maybe these could search for that type instead of only being in the first argument for even more cool auto-magic composability.

connorlbark (approved):
Works great! Just an optional idea for even more cool composability 😄

api-admin/src/main/java/bio/terra/pearl/api/admin/service/auth/BaseEnforcePermissionAspect.java
devonbush: yes!  these are designed to be composable.  The idea is that there will be only one authContext argument (the first arg), and that can be either a PortalAuthContext or an EnvironmentAwareAuthContext, or *both*, since EnvironmentAwareAuthContext is just an interface

PR #945, opened 2024-06-18
[JN-1136] moving publishing logic into entity services

devonbush:

Starts the process of moving diff/publishing logic from a centralized publishing service into being owned by individual entity services.  The main benefit is that by providing the PublishingSupport interface, it gives a clearer roadmap for developing a new config that will need to get published.  Would love a better name for it than `PublishingSupport` -- that doesn't feel quite right.

1. no change in application behavior, so just run unit tests


core/src/main/java/bio/terra/pearl/core/service/publishing/PublishingSupport.java
```diff
@@ -0,0 +1,65 @@
...
+import bio.terra.pearl.core.model.Versioned;
+import bio.terra.pearl.core.model.portal.PortalEnvironment;
+import bio.terra.pearl.core.model.publishing.*;
+import bio.terra.pearl.core.model.study.StudyEnvironment;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+public interface PublishingSupport {
```
connorlbark: maybe simply `Publishable`? I feel like interfaces tell you what a thing is `able` to do.

core/src/main/java/bio/terra/pearl/core/service/publishing/PublishingSupport.java
```diff
@@ -0,0 +1,65 @@
...
+    public static final List<String> CONFIG_IGNORE_PROPS = List.of("id", "createdAt", "lastUpdatedAt", "class",
+            "studyEnvironmentId", "portalEnvironmentId", "emailTemplateId", "emailTemplate",
+            "consentFormId", "consentForm", "surveyId", "survey", "versionedEntity", "trigger");
+    public void loadForDiffing(PortalEnvironment portalEnv);
+    public void loadForDiffing(StudyEnvironment studyEnv);
+    public void updateDiff(PortalEnvironment sourceEnv, PortalEnvironment destEnv, PortalEnvironmentChange change);
+    public void updateDiff(StudyEnvironment sourceEnv, StudyEnvironment destEnv, StudyEnvironmentChange change);
+    public void applyDiff(PortalEnvironment sourceEnv, PortalEnvironment destEnv, PortalEnvironmentChange change);
+    public void applyDiff( StudyEnvironmentChange change, StudyEnvironment destEnv, PortalEnvironment destPortalEnv);
```
connorlbark: Just for my own understanding: why are there separate portalenvironment and studyenvironment methods? It feels like some things are publishable only at a portal level and some only at a study level, but it seems like that isn't the case for triggers, which update for both?

connorlbark (approved):

core/src/main/java/bio/terra/pearl/core/service/publishing/PublishingSupport.java
devonbush: that's exactly right.  This was the only way I could think of structuring it so that the entity services didn't have to have the logic for iterating across studies within a portal.

MatthewBemis (approved):

PR #946, opened 2024-06-18
[JN-1126] auto-trimming answer mappings

devonbush:

auto-trim whitespace for STRING_TO_STRING map types.  I think that makes sense as the default, if we ever need something different, we can add a STRING_TO_STRING_RAW or similar.

1.  redeploy apiParticipantApp
2. go to sandbox.ourhealth.localhost:3001
3. log in as "consented@test.com"
4. fill in your name on the basics survey, but with leading/trailing whitespace
5. run `select given_name, family_name from profile;` in the database, and confirm the whitespace got trimmed


connorlbark (approved):
nice!

MatthewBemis (approved):

PR #949, opened 2024-06-20
[JN-1116] Add survey response editing permission

MatthewBemis:

Puts survey response editing behind an admin permission- currently given to all study admins, but we could eventually break it out into another role in the future.

* Redeploy the admin API app to populate the new permission
* Log in as `staff@heartdemo.org` and confirm that you can still edit responses for users: https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants/HDINVI

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/SurveyResponseController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/SurveyResponseController.java
@@ -2,6 +2,7 @@ package bio.terra.pearl.api.admin.controller.forms;
 
 import bio.terra.pearl.api.admin.api.SurveyResponseApi;
 import bio.terra.pearl.api.admin.service.auth.AuthUtilService;
+import bio.terra.pearl.api.admin.service.auth.context.PortalStudyEnvAuthContext;
 import bio.terra.pearl.api.admin.service.forms.SurveyResponseExtService;
 import bio.terra.pearl.core.model.EnvironmentName;
 import bio.terra.pearl.core.model.admin.AdminUser;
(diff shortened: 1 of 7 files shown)


ui-admin/src/study/participants/survey/SurveyResponseView.tsx
```diff
@@ -100,7 +100,10 @@ export function RawEnrolleeSurveyView({
               <DropdownButton
                 onClick={() => setIsEditing(true)}
                 icon={faPencil}
-                disabled={!configSurvey.survey.allowAdminEdit}
+                disabled={
+                  !configSurvey.survey.allowAdminEdit ||
+                    !userHasPermission(user, studyEnvContext.portal.id, 'survey_response_edit')
+                }
```
connorlbark: ux nit: should there be a tooltip explaining that you do not have permissions? i could see being confused when you can't do something that you can see. or just not render the dropdown to edit.

connorlbark (approved):

ui-admin/src/study/participants/survey/SurveyResponseView.tsx
MatthewBemis: that's a good point- I kinda went back and forth on whether or not to display it. I don't think we're consistent about this in our UI, but I wonder if we want to establish some patterns like:

* If the user can't perform the action and there's nothing they can do on their own to enable the action, don't display it at all (i.e. they don't have the right permission or role to do the action)
* If the user can't perform the action but they can do something to enable it, display it with a tooltip telling them what they can do. An example would be fixing an error in a survey question before being able to click the 'Save' button

So in this case, I wouldn't render the option here at all.

I do think there are cases where the user might want to know that the action is possible so they can ask someone on their team to grant them that action, but I'm not sure how important/useful that is

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/SurveyResponseController.java
```diff
@@ -44,9 +45,17 @@ public ResponseEntity<Object> update(
...
     SurveyResponse responseDto = objectMapper.convertValue(body, SurveyResponse.class);
+    PortalStudyEnvAuthContext authContext =
+        PortalStudyEnvAuthContext.of(user, portalShortcode, studyShortcode, environmentName);
     HubResponse hubResponse =
         surveyResponseExtService.updateResponse(
-            user, portalShortcode, environmentName, responseDto, enrolleeShortcode, taskId);
+            authContext,
+            user,
+            portalShortcode,
+            environmentName,
```
devonbush: we don't need to pass these in -- they should be read from the authContext

ui-admin/src/study/participants/survey/SurveyResponseView.tsx
devonbush: I'd recommend (not necessarily in this PR) we update our Button and DropdownButton components to take a "forbidden" boolean to be true if the user doesn't have permission to click the button.  then we can adopt a consistent pattern of either showing/hiding forbidden buttons.

devonbush (approved):
nice, 100% agree that this was worth a separate permission

ui-admin/src/study/participants/survey/SurveyResponseView.tsx
MatthewBemis: 👍 I'll add a ticket for this

PR #952, opened 2024-06-20
[JN-1124] [part 1] readonly family view

connorlbark:

Part 1 / 2 for this ticket. To start, I just wanted to establish a readonly view / router / etc. to the family and then focus on making it editable in a second PR.

- Restart admin
- Repop demo
- Go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/families/F_SALKFM
- Click around, ensure it looks good


ui-admin/src/study/families/FamilyStructureTable.tsx
```diff
@@ -0,0 +1,261 @@
...
+type FamilyMemberWithSubrows = {
+  member: Enrollee,
+  relationToParentRow?: EnrolleeRelation
+  subrows: FamilyMemberWithSubrows[]
+}
+
+/**
+ * Renders a table of family members grouped by their relationships to each other.
+ */
+export const FamilyStructureTable = (
```
MatthewBemis, line 46: Getting a console warning from this component: ```Warning: Each child in a list should have a unique "key" prop. Check the render method of `FamilyStructureTable`.```

ui-admin/src/study/families/FamilyView.tsx
```diff
@@ -0,0 +1,110 @@
...
+        </h4>
+      </div>
+    </div>
+    <div className="row mt-2">
+      <div className="col-12">
+        <div className="d-flex">
+          <div style={navDivStyle}>
+            <ul className="list-unstyled">
+              <li style={navListItemStyle} className="ps-3">
+                <NavLink to="." className={getLinkCssClasses}>Overview</NavLink>
```
MatthewBemis:
```suggestion
                <NavLink end to="." className={getLinkCssClasses}>Overview</NavLink>
```
style nit: this will make it so the Overview item isn't always bolded, even when another section is selected

ui-admin/src/study/families/FamilyStructureTable.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+  useEffect(() => {
+    table.toggleAllRowsExpanded(true)
+  }, [])
+
+  return <div>
+    {basicTableLayout(table, {
+      customRowHeader,
+      tableClass: 'table table-striped m-0'
+    })}
```
MatthewBemis, line 152: sorting a column causes the relationship groupings to get incorrectly shuffled around:

Screenshot 2024-06-25 at 1 32 36 PM

I assume that's not going to be a super trivial fix, and I wonder if it makes sense to just disable sorting for this table anyway?

ui-admin/src/study/families/FamilyRouter.tsx
```diff
@@ -0,0 +1,39 @@
...
+  userHasPermission,
+  useUser
+} from 'user/UserProvider'
+
+/** routes to family views as appropriate */
+export default function FamilyRouter({ studyEnvContext }: {studyEnvContext: StudyEnvContextT}) {
+  const { user } = useUser()
+
+  if (!userHasPermission(user, studyEnvContext.portal.id, 'prototype')) {
+    return <></>
```
MatthewBemis: I might lean towards at least rendering some text like "You do not have permission to access this prototype feature.". A blank page usually makes me think the app is broken.

Just thinking out loud though, since this might not be an issue depending on how we roll all of this out.

MatthewBemis (approved):
looks great! just had a few small notes

ui-admin/src/study/families/FamilyView.tsx
connorlbark: 😮  That has been bothering me so much!!!!

ui-admin/src/study/families/FamilyStructureTable.tsx
connorlbark: TIL: You can add keys to fragments by doing `<Fragment key={...}></Fragment>` instead of `<></>`

api-admin/src/main/java/bio/terra/pearl/api/admin/service/family/FamilyExtService.java
```diff
@@ -0,0 +1,29 @@
...
+    this.familyService = familyService;
+  }
+
+  @EnforcePortalStudyEnvPermission(permission = "study_admin")
+  public Family find(PortalStudyEnvAuthContext authContext, String familyShortcode) {
+    StudyEnvironment studyEnvironment = authContext.getStudyEnvironment();
+
+    return familyService
+        .findOneByShortcode(familyShortcode)
+        .filter(f -> f.getStudyEnvironmentId().equals(studyEnvironment.getId()))
```
devonbush: this is a pretty fundamental operation -- let's add a findOneByShortcodeAndEnvironment(shortcode, envId) method to the core FamilyService

core/src/main/java/bio/terra/pearl/core/service/participant/FamilyService.java
```diff
@@ -40,6 +52,19 @@ public Optional<Family> findOneByShortcode(String shortcode) {
         return dao.findOneByShortcode(shortcode);
     }
 
+    public Family loadForAdminView(Family family) {
+        family.setMembers(enrolleeDao.findAllByFamilyId(family.getId()).stream().map(enrollee -> {
+            enrollee.setProfile(profileDao.find(enrollee.getProfileId()).orElse(null));
```
devonbush: I suspect this "Load some enrollees and attach their profiles" is a common enough operation that it's worth making an EnrolleeService  "attachProfiles" method.   That method should use the findAllPreserveOrder dao method to avoid making a separate DB call for each enrollee.

List<Profile> profiles = profileDao.findAllPreserveOrder(enrollees.stream().map(Enrollee::getProfileId).toList());

core/src/main/java/bio/terra/pearl/core/service/participant/FamilyService.java
```diff
@@ -40,6 +52,19 @@ public Optional<Family> findOneByShortcode(String shortcode) {
...
+            return enrollee;
+        }).toList());
+        family.setRelations(enrolleeRelationDao.findRelationsForFamily(family.getId()));
+        family.setProband(enrolleeDao.find(family.getProbandEnrolleeId()).map(e -> {
+            e.setProfile(profileDao.find(e.getProfileId()).orElse(null));
```
devonbush: e is for error.  make this `enrollee` or `proband`

devonbush: also this is another case where `attachProfiles` would be useful

ui-admin/src/study/families/FamilyRelations.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+
+  const table = useReactTable({
+    data: family.relations || [],
+    columns,
+    getCoreRowModel: getCoreRowModel(),
+    getSortedRowModel: getSortedRowModel()
+  })
+  return <div>
+    {basicTableLayout(table)}
```
devonbush: the display of the table is a bit wonky due to the select sizing.  let's update BasicTableConfig to take a `trClass` and then you can pass `align-middle` to get everything centered

ui-admin/src/study/families/FamilyRouter.tsx
```diff
@@ -0,0 +1,40 @@
...
+import {
+  userHasPermission,
+  useUser
+} from 'user/UserProvider'
+
+/** routes to family views as appropriate */
+export default function FamilyRouter({ studyEnvContext }: {studyEnvContext: StudyEnvContextT}) {
+  const { user } = useUser()
+
+  if (!userHasPermission(user, studyEnvContext.portal.id, 'prototype')) {
```
devonbush, line 20: just to confirm -- the thought here is that there won't be a need for a separate "family_view" permission, and so once this feature is finalized, there won't be a permission check here at all?

ui-admin/src/study/families/FamilyRouter.tsx
```diff
@@ -0,0 +1,40 @@
...
+    return <p>You do not have the appropriate permissions to access this prototype feature.</p>
+  }
+
+  return <>
+    <NavBreadcrumb value={studyEnvContext.currentEnvPath}>
+      <Link to={`${studyEnvContext.currentEnvPath}/participants`} className="me-1">
+          participants</Link>
```
devonbush: Give some indication on the breadcrumbs that were in a family path
```suggestion
          participants</Link> family
```

devonbush (approved):
looks good for v1! Feel free to address comments in this or a follow-on PR

ui-admin/src/study/families/FamilyRouter.tsx
connorlbark: Yes - I am totally open to however we want these perms to work, I assumed something like this wouldn't be locked down since, e.g., viewing participants is not locked down right now

core/src/main/java/bio/terra/pearl/core/service/participant/FamilyService.java
connorlbark: The only problem is that the EnrolleeService uses the FamilyService, so I have to use the direct dao methods here to avoid a circular dependency 😢

ui-admin/src/study/families/FamilyRelations.tsx
connorlbark: Hmm, I'm not sure what you mean that it looks wonky. On my screen, everything is already centered and `align-middle` has no effect:
image

ui-admin/src/study/families/FamilyRouter.tsx
connorlbark: I made it `families` and in the PR where I will incorporate families into the participant list, I'll make sure that when you use the link it will pre-group by family

PR #955, opened 2024-06-21
[JN-1114] Add justification flow to response editing

MatthewBemis:

* Updates the survey response editing flow to require a justification. We ask for the justification up front because autosaves will happen automatically in the background. The justification will be used for the duration of their edit session. A justification is _not_ required on study-staff forms, only for participant forms.
* Updates the `AdminUserProvider` to also fetch the list of superusers if the logged-in user is a superuser. This allows us to display the usernames of superusers throughout the admin UI instead of just displaying `superuser (4a6f...)`
* Updates the audit table to show _which_ admin made the changes to the user

Note: in the video below, some entries show `Admin (undefined)`. That's a display bug that I've fixed, but I didn't feel like recording the video again 🙂 

https://github.com/broadinstitute/juniper/assets/7257391/4f58239f-c2c5-4891-8151-851cc3bfe814

* Rebuild ui-core and restart admin UI and admin API
* Log in and start editing a user's existing responses
* Confirm that you see your edits and justification in the audit history


core/src/main/java/bio/terra/pearl/core/model/audit/ResponsibleEntity.java
```diff
@@ -34,6 +35,12 @@ public ResponsibleEntity(String systemProcess) {
         }
         this.systemProcess = systemProcess;
     }
+    public ResponsibleEntity(PortalParticipantUser ppUser) {
+        if (ppUser == null) {
+            throw new IllegalArgumentException("portal participant user cannot be null");
+        }
+        this.participantUser = ppUser.getParticipantUser();
+    }
```
MatthewBemis, line 43: I'm not 100% sure that doing this is correct, but I saw a few places where we take a PortalParticipantUser and set the responsibleEntity to the underlying ParticipantUser

core/src/main/java/bio/terra/pearl/core/model/audit/TraceableEntity.java
```diff
@@ -0,0 +1,29 @@
...
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class TraceableEntity extends BaseEntity {
+    private UUID responsibleUserId; // id of the user making the change, if it was a participant
+    private UUID responsibleAdminUserId; // id of the user making the change, if it was an admin
+    private String systemProcess; // if the change was the result of an automatic process, store Class+method here
+    private Boolean anonymousUser; // if the change was made by an anonymous user
+    private String justification; // if an admin changes a participant's data, a justification is needed
```
MatthewBemis, line 20: I wanted to add some logic in here to enforce that a justification was always provided when a responsibleAdminUserId was provided, but unfortunately we also use this for tracking mailing list contact uploads from the admin tool- which don't need a justification

core/src/test/java/bio/terra/pearl/core/service/survey/SurveyResponseServiceTests.java
```diff
@@ -138,20 +143,32 @@ public void testSurveyResponseUpdateAnswers(TestInfo testInfo) {
 
         List<DataChangeRecord> changeRecords = dataChangeRecordService.findByEnrollee(savedResponse.getEnrolleeId());
         assertThat(changeRecords.size(), equalTo(1));
-        assertThat(changeRecords.get(0), samePropertyValuesAs(DataChangeRecord.builder()
```
MatthewBemis, line 141: `samePropertyValuesAs` was erroring on nulls even if I excluded the null fields from being checked. What's odd is that there were already null properties being checked in the existing test, so I'm not sure what's going on here. I spent some time in the debugger trying to figure out what was going on and was completely stumped- everything was identical

MatthewBemis:
per conversation with Connor, throwing this back into draft to handle the audit info stuff a little more clearly

ui-admin/src/study/participants/DataChangeRecords.tsx
```diff
@@ -79,7 +81,9 @@ export default function DataChangeRecords({ enrollee, studyEnvContext }:
     {
       header: 'Source',
       cell: ({ row }) => (
-        row.original.responsibleUserId ? 'Participant' : 'Admin'
+        row.original.responsibleUserId ? 'Participant' :
+          row.original.responsibleAdminUserId &&
+            `Admin (${(users.find(u => u.id === row.original.responsibleAdminUserId)?.username)})`
```
connorlbark, line 86: NICE! this is way easier and simpler than I thought it would be, you'll get a gold star from Parker for sure 😄

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -255,31 +258,39 @@ public List<Answer> createOrUpdateAnswers(List<Answer> answers, SurveyResponse r
         List<Answer> updatedAnswers = answers.stream().map(answer -> {
             Answer existing = existingAnswerMap.get(answer.getQuestionStableId());
             if (existing != null) {
-                return updateAnswer(existing, answer, response, survey, ppUser, changeRecords);
+                return updateAnswer(existing, answer, response, survey, ppUser, changeRecords, operator);
             }
-            return createAnswer(answer, response, survey, ppUser);
+            return createAnswer(answer, response.getSurveyResponse(), survey, ppUser);
```
connorlbark: Should we be creating audit records on create as well? - I could see it being a not uncommon flow for admins to fill out surveys for users on their behalf over the phone, so having a record of that seems important.

core/src/main/java/bio/terra/pearl/core/model/survey/SurveyResponseDto.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class SurveyResponseDto {
```
devonbush: since this is a container with other properties, and not just a DTO with convenience properties, maybe call this `ResponseWithJustification` or something like that.

If you are going to keep this name, consider extending SurveyResponse, instead of wrapping it

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/survey/SurveyResponseController.java
```diff
@@ -67,7 +67,7 @@ public ResponseEntity<Object> update(
       Object body) {
     ParticipantUser user = requestUtilService.requireUser(request);
     EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
-    SurveyResponse responseDto = objectMapper.convertValue(body, SurveyResponse.class);
+    SurveyResponseDto responseDto = objectMapper.convertValue(body, SurveyResponseDto.class);
```
devonbush: since we don't have any workflows where the participant has a justification, it would be nice if this deserialization didn't have to contemplate it.  that makes me lean more towards making SurveyResponseDto extend SurveyResponse, and renaming it to SurveyResponseWithJustification.

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
devonbush: we don't need a separate audit record for created answers -- the Answer object itself contains the time and author.  But we should update the createAnswer method to take the operator too

ui-admin/src/providers/AdminUserProvider.tsx
```diff
@@ -26,14 +27,24 @@ export const useAdminUserContext = () => {
...
 export default function AdminUserProvider({ portalShortcode, children }:
 {portalShortcode: string, children: React.ReactNode}) {
+  const { user } = useUser()
   const [userState, setUserState] = useState<AdminUserContextT>({ users: [], isLoading: true })
   useEffect(() => {
-    Api.fetchAdminUsersByPortal(portalShortcode).then(result => {
-      setUserState({
-        users: result,
-        isLoading: true
+    if (user?.superuser) {
```
devonbush, line 33: oooh, I like this change!

devonbush (changes requested):
Nice, love how seamless this is in the audit history.

Blockers are that `SurveyResponseService.createAnswer` should use the `operator`, and we should have a populate example to see the how justifications are saved/presented in the admin tool without having to manually edit a survey

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
connorlbark: I see - I guess my issue then is more that it doesn't show up in the data audit table, so it's weird to have to go to the DB for some changes but others show up like normal in the table. I was surprised when I made changes and they didn't show up in the table - I thought there was a bug until I did some investigation in the code, and I could see users feeling like the audit table wasn't working properly.

core/src/main/java/bio/terra/pearl/core/model/survey/SurveyResponseWithJustification.java
```diff
@@ -0,0 +1,15 @@
...
+import bio.terra.pearl.core.model.participant.Profile;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
```
devonbush, line 18: this could use a comment about how this is split from SurveyResponse since justifications aren't needed for participant flows

populate/src/main/resources/seed/portals/demo/studies/heartdemo/enrollees/jsalk.json
```diff
@@ -55,12 +55,23 @@
...
       "currentPageNo": 1,
       "complete": true,
       "submittedHoursAgo": 48
+    }, {
+      "surveyStableId": "hd_hd_basicInfo",
+      "surveyVersion": 1,
+      "answerPopDtos": [
+        {"questionStableId": "hd_hd_basic_streetAddress", "stringValue": "123 Walnut Street", "viewedLanguage": "es"}
+      ],
+      "justification":  "User entered the incorrect mailing address.",
```
devonbush, line 70: nice!

devonbush (approved):
populate works great -- thanks for adding it!

ui-admin/src/providers/AdminUserProvider.tsx
```diff
@@ -26,14 +27,24 @@ export const useAdminUserContext = () => {
...
+      Api.fetchAdminUsers().then(result => {
+        setUserState({
+          users: result,
+          isLoading: true
+        })
```
connorlbark, line 38: What happens if a superuser not in your portal makes a change that shows up in the audit table? Their username wouldn't show up right? Maybe in the audit table we could have the text default to `Juniper Admin` or something like that if it doesn't show up

connorlbark (approved):

ui-admin/src/providers/AdminUserProvider.tsx
MatthewBemis: If I'm understanding the question correctly, this shouldn't be an issue because superusers have access to all portals. There's a related case where if a non-superuser views this table, they won't have access to load the list of superusers, so those entries will display the text "superuser (first few digits of user id)"

PR #960, opened 2024-06-24
[JN-1122] Log exception message in processNotification

MatthewBemis:

Adds the exception message to the log message that we're seeing a bunch in the prod alerts channel.

I spent a little bit of time perusing prod logs but didn't find anything. I also didn't have any luck trying to reproduce the condition locally. What I really want to see is the error message- it's most likely coming from the database.

* Add an exception to line 86
* Try sending an email
* Verify that the exception message is logged


connorlbark (approved):

devonbush (approved):

devonbush:
We might want to remove this logging after we troubleshoot -- the reason we didn't log the exception before was that it felt like an exception likely to have PHI in it

PR #963, opened 2024-06-25
[JN-1150] include families on export

connorlbark:

Also, since it would be yet another copy of kitrequestformatter, refactors enrolleerelation/kitrequest formatters to have a parent `BeanListFormatter` type.

- ensure heartdemo synthetic data exports as expected


core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -179,7 +184,8 @@ protected EnrolleeExportData loadEnrolleeData(Enrollee enrollee) {
                 participantTaskService.findByEnrolleeId(enrollee.getId()),
                 surveyResponseService.findByEnrolleeId(enrollee.getId()),
                 kitRequestService.findByEnrollee(enrollee),
-                enrolleeRelationService.findByTargetEnrolleeIdWithEnrollees(enrollee.getId())
+                enrolleeRelationService.findByTargetEnrolleeIdWithEnrolleesAndFamily(enrollee.getId()),
```
devonbush: Gating these on the StudyEnvironmentConfig.enableFamilyLinkage will save a couple thousand DB queries for most studies.

devonbush (approved):
Looks great, thanks for pulling out the BeanList formatter.  One performance improvement suggestion -- OurHealth export currently takes ~20 seconds in prod for 1180 enrollees, so we should start thinking about export performance a little bit.

MatthewBemis (approved):

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
connorlbark: Refactored proxies to also only be grabbed when proxy enrollment is enabled + added tests for these.

PR #964, opened 2024-06-26
[JN-1160] adding profile editing to form preview

devonbush:

In order to test dynamic texts for A-T more easily, I added the ability to change the enrollee/proxy profile during form editing.  For now this is a simple sidebar, later we may need to make it into a full-fledged editor modal so that birthdate, sexAtBirth, and other profile fields can be customized to test logic.

image

to make room for this, I condensed the other preview options with liberal use of InfoPopups.

This also contains a bugfix for the language selector -- adding multiple languages would sometimes have timing bugs

1. repopulate demo
2. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms/surveys/hd_hd_lost_interest
3. go to the `Preview` tab
4. fill in values for profile given name or profile family name.  Observe the question gets updated with the values

### DIFF
--- a/ui-admin/src/forms/FormPreviewOptions.tsx
+++ b/ui-admin/src/forms/FormPreviewOptions.tsx
@@ -1,15 +1,17 @@
 import React, { useState } from 'react'
-import { PortalEnvironmentLanguage } from '@juniper/ui-core'
+import { PortalEnvironmentLanguage, Profile } from '@juniper/ui-core'
 import Select from 'react-select'
 import useReactSingleSelect from 'util/react-select-utils'
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faGlobe } from '@fortawesome/free-solid-svg-icons'
 import { usePortalLanguage } from 'portal/languages/usePortalLanguage'
+import InfoPopup from '../components/forms/InfoPopup'
+import { TextInput } from '../components/forms/TextInput'
 
 type FormPreviewOptions = {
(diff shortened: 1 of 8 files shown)


ui-admin/src/forms/FormPreviewOptions.tsx
```diff
@@ -84,10 +93,68 @@ export const FormPreviewOptions = (props: FormPreviewOptionsProps) => {
...
+            profile: {
+              ...value.profile,
+              familyName: text
+            }
+          })}
+          label={'Family name'}
+          unboldLabel={true}
+          value={value.profile?.familyName ?? ''} />
+        </div>
+        <div data-testid="proxyInfoFields">
```
connorlbark, line 126: Should we put this behind `config.proxyEnrollmentEnabled` check?

connorlbark (approved):
Sweet! works super well. If we add all of the fields of the profile, I would think we should maybe consider making them expandable/contractable, but with just given name/family name it looks/works great!

ui-admin/src/forms/FormPreviewOptions.tsx
devonbush: hah, I considered that, but the config was only accessible about 5 components up, so it wasn't worth the drilldown.  

Related, I think there's an engineering infrastructure ticket to make the PortalContext more explicit about what it includes (e.g. config + languages) and doesn't, so that other components can work with it more easily

MatthewBemis (approved):

PR #965, opened 2024-06-27
[JN-1124] [part 2] editable family

connorlbark:

Makes families robustly editable (able to add/remove members and add/remove relationships and members at the same time).

You can create relationships between _anyone_ - but if they aren't in the family, they will be added (and you will be warned). All operations also require justifiaction.

- Go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/families/F_SALKFM
- Try:
   - Adding new members via members table
   - Creating relations between: 2 people not in family, 1 person in and one person out of, and 2 people in family. Should add to the family if they aren't already.
   - Deleting members with relations (should delete relations)
   - Changing proband
- Ensure operations ^ end up in data audit table with justifications


ui-admin/src/study/families/FamilyStructureTable.tsx
```diff
@@ -72,21 +71,10 @@ export const FamilyStructureTable = (
...
     {
-      id: 'shortcode',
-      header: 'Shortcode',
-      accessorKey: 'member.shortcode',
+      id: 'enrollee',
+      header: 'Enrollee',
       cell: ({ row }) => {
-        return <Link to={`${studyEnvContext.currentEnvPath}/participants/${row.original.member.shortcode}`}>
-          {row.original.member.shortcode}
-        </Link>
```
connorlbark, line 81: the family structure table is.... wonky at best. I have opinions how to make it better & I'll come by in a followup PR to make it more robust/useful.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeRelationExtService.java
```diff
@@ -44,4 +50,68 @@ public List<EnrolleeRelation> findRelationsForTargetEnrollee(
             .orElseThrow(() -> new NotFoundException("Enrollee not found"));
     return enrolleeRelationService.findByTargetEnrolleeIdWithEnrolleesAndFamily(enrollee.getId());
   }
+
+  // todo
+  @EnforcePortalStudyEnvPermission(permission = "???")
```
connorlbark: I have no idea what to put on these methods (aside from `BASE`). I would like to make this a prototype_user only thing for now - but that's a role, not a permission. It could be cool to make a `can_use_beta_features` permission that this could sit under until our permissions are more mature and we can decide what to do with these.

core/src/main/java/bio/terra/pearl/core/service/search/sql/EnrolleeSearchQueryBuilder.java
```diff
@@ -68,9 +68,19 @@ public SelectQuery<Record> toQuery(DSLContext context, EnrolleeSearchOptions opt
...
+            var ordered = selectConditionStep.orderBy(field(sortField));
+            if (Objects.nonNull(opts.getLimit())) {
+                return ordered.limit(opts.getLimit()).getQuery();
+            }
+            return ordered.getQuery();
+        }
+
+        if (Objects.nonNull(opts.getLimit())) {
+            return selectConditionStep.limit(opts.getLimit()).getQuery();
         }
```
connorlbark, line 84: jooq is great, but because jooq seems to change types every time you call a function on the query, it's really hard to have multiple conditional branches on the same query without a lot of reused code. It's possible I'm missing some easy way to do this

api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeRelationExtService.java
devonbush: That's a great point, let's talk after standup.  I suspect what we want to do is replace the current `survey_response_edit` permission with a more general `participant_data_edit` permission.  And then this method will have that

How I'd assumed beta permissions would would is that you would annotate this feature with a specific permission that you create for the feature like "family_relation_edit", and then you'd add that permission to the "prototype_user" role.  That may be more complexity/trouble than these are worth though -- we might be better off as you said, adding a "beta_feature" permission.

devonbush (commented):
Still going through the code, but wanted to say that I tested out the functionality, and it seems great.  Love the UX tweaks made since demo.  Two thoughts:

1.  we're not using disks as save icons anywhere else -- I'd put a screenshot in the team slack and ask for thoughts.  I don't have a strong opinion, but if we're going to use save disk icons, we should do it everywhere.
2. I love the members & relations view.  I'd consider dropping the "Overview" for now, or making the members&relations view the primary view.  But of course I defer to the CMI folks who know a lot more about what they want to do here.

connorlbark:
got lots of feedback -gonna put this back into draft until I can work it all out.

edit: alright! should be good to go now:
- floppy disk -> save 
- after talking to whitney, we changed `survey_response_edit` to `participant_data_edit` and  added `participant_data_view` (not being used in all the places it should be)
- added lots of authentication tests with the new fancy stuff devon built
- made overview a beta page only viewable by superusers - it's not super useful right now, and might just be chopped off. erin brought up some good points about it, and i think it should probably stay long-term but it's not quite useful yet. I might dedicate a member's choice to making it more robust - I found some cool algorithms that will make the family table able to handle every family structure you throw at it, so stay tuned 😄

ui-admin/src/util/tableUtils.tsx
```diff
@@ -172,14 +173,18 @@ function SearchFilter<A>({
...
 ) {
   const sortDirection = options.sortable && header.column.getIsSorted()
   const ariaSort = options.sortable ?
     sortDirection ? (sortDirection === 'desc' ? 'descending' : 'ascending') : 'none' : undefined
 
   return <th key={header.id}
-    aria-sort={ariaSort}>
+    aria-sort={ariaSort}
+    style={{
+      width: options.useSize ? `${header.getSize()}px` : undefined
```
devonbush: setting column sizes in px seems like it will cause responsiveness problems.  Try using bootstrap col-* classes instead?  If that doesn't work and you really need to pass an explicit size parameter, have it be a string so it can be specified in px/em/% as needed

ui-admin/src/study/participants/enrolleeView/ParticipantTaskView.tsx
```diff
@@ -59,7 +62,9 @@ const ParticipantTaskView = ({ enrollee }: {enrollee: Enrollee}) => {
     <table className="table table-striped">
       <thead>
         <tr>
-          {table.getFlatHeaders().map(header => tableHeader(header, { sortable: true, filterable: false }))}
+          {table
+            .getFlatHeaders()
+            .map(header => tableHeader(header, { sortable: true, filterable: false, useSize: false }))}
```
devonbush: isn't false the default?

ui-admin/src/study/participants/enrolleeView/EnrolleeLink.tsx
```diff
@@ -11,6 +11,10 @@ export const EnrolleeLink = ({ studyEnvContext, enrollee }: {
   studyEnvContext: StudyEnvContextT,
   enrollee: Enrollee
 }) => {
+  if (!enrollee) {
```
devonbush: this seems odd since we have enrollee as a non-optional type.  Either change the type to optional or remove this check

core/src/main/java/bio/terra/pearl/core/service/participant/FamilyService.java
```diff
@@ -96,4 +104,74 @@ public List<Family> findByEnrolleeIdWithProband(UUID enrolleeId) {
...
+
+        Enrollee enrollee = enrolleeService.findByShortcodeAndStudyEnvId(enrolleeShortcode, studyEnvId)
+                .orElseThrow(() -> new NotFoundException("Enrollee not found"));
+
+        // ensure the enrollee is not already a member of the family
+        List<FamilyEnrollee> existingMembers = familyEnrolleeService.findByFamilyId(family.getId());
+        if (existingMembers.stream().anyMatch(fe -> fe.getEnrolleeId().equals(enrollee.getId()))) {
+            throw new IllegalArgumentException("Enrollee is already a member of the family");
+        }
+
```
devonbush: let's pull this into a helper `isEnrolleeInFamily` method

api-admin/src/main/java/bio/terra/pearl/api/admin/service/participant/ProfileExtService.java
```diff
@@ -21,8 +23,13 @@ public ProfileExtService(AuthUtilService authUtilService, ProfileService profile
   /**
    * Updates the profile on behalf of the enrollee; requires a justification for auditing purposes.
    */
+  @EnforcePortalStudyEnvPermission(permission = "participant_data_edit")
```
devonbush, line 26: I'd love it if you tried creating an @EnforcePortalEnrolleePermission (possibly in a follow-on PR).  This pattern of authing an admin to a specific enrollee is very common. 

You'll create a EnforcePortalEnrolleePermission interface, and a PortalEnrolleeAuthContext.  the latter will extend PortalStudyEnvAuthContext since we'll want to auth the user to the enrolleee's study environment too. Then create a EnforcePortalStudyEnvPermissionAspect class to implement the auth check and set the enrollee on the authContext

api-admin/src/test/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeRelationExtServiceTest.java
```diff
@@ -0,0 +1,164 @@
...
+class EnrolleeRelationExtServiceTest extends BaseSpringBootTest {
+
+  @Autowired private EnrolleeRelationExtService enrolleeRelationExtService;
+  @Autowired private StudyEnvironmentFactory studyEnvironmentFactory;
+  @Autowired private EnrolleeFactory enrolleeFactory;
+  @Autowired private AdminUserFactory adminUserFactory;
+  @Autowired private FamilyFactory familyFactory;
+
+  @Test
+  public void testAuthentication() {
```
devonbush, line 37: nice! thanks for giving the new auth test utils a spin!

core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeRelationDao.java
```diff
@@ -99,4 +99,18 @@ public void deleteByStudyEnvironmentId(UUID studyEnvironmentId) {
                         .execute()
         );
     }
+
+    public List<EnrolleeRelation> findAllFamilyRelationshipsByEnrolleeIdOrTargetIdAndFamilyId(UUID enrolleeId, UUID familyId) {
```
devonbush: how about `findAllRelationsByFamilyAndEitherEnrollee()` for something a bit shorter

devonbush (approved):
Sweet, looks good and worked well.  Lots of suggestions, no blockers since this functionality won't be used in prod for a bit, so no need to refine everything too much

populate/src/main/resources/seed/iam/permissions.json
```diff
@@ -30,8 +30,13 @@
     "description": "Allows user to export study data to Terra Data Repo"
   },
   {
-    "name": "survey_response_edit",
-    "displayName": "Survey response editing",
-    "description": "Allows user to edit participant survey responses"
+    "name": "participant_data_edit",
```
MatthewBemis, line 33: nice, i like that you made this more general

MatthewBemis (approved):
Looks good, just one testing note: I'm not seeing any options in the relation dropdown when creating a new relation. Am I using this correctly?

Screenshot 2024-07-02 at 7 34 06 AM

connorlbark:
> Looks good, just one testing note: I'm not seeing any options in the relation dropdown when creating a new relation. Am I using this correctly?

Yes! I got feedback in demo that this should be a set list of options that the user specifies somewhere - there is nowhere to specify it right now, so it's just creatable, but eventually it will have options.

ui-admin/src/util/tableUtils.tsx
connorlbark: Unfortunately, tanstack table is all-in on size being a number, but the col- idea is clever! I'll go with that.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/participant/ProfileExtService.java
connorlbark: Seems non-trivial, so I'll do this in a follow up PR!

PR #968, opened 2024-06-28
[JN-1159] rename relations in participant ui to proxyrelations

connorlbark:

Based on a few thoughts:
1. We shouldn't send family relations in the frontend because it's potential a PII leak & we (as of right now) have no need for it
2. Because of this, `relations` (in logindto/useuser) really is just `proxyRelations`, so I renamed it.
3. If we want to include family relations in the future, it would be less cumbersome to make it a separate logindto field: `familyRelations`. That way in the frontend we never have to do any awkward `relations.filter(r => r.type === 'FAMILY')` and we can keep `proxyRelations` and `familyRelations` totally separate.

- Ensure no regressions in proxy functionality
- Log in as jsalk and psalk and check network traffic to ensure no family relations sent in the logindto


api-participant/src/main/java/bio/terra/pearl/api/participant/service/CurrentUserService.java
```diff
@@ -101,7 +101,7 @@ public UserLoginDto loadFromUser(ParticipantUser user, String portalShortcode) {
 
     List<Enrollee> enrollees = loadEnrollees(ppUser);
 
-    List<EnrolleeRelation> relations = loadEnrolleeRelations(enrollees);
+    List<EnrolleeRelation> relations = loadProxyRelations(enrollees);
```
devonbush:
```suggestion
    List<EnrolleeRelation> proxyRelations = loadProxyRelations(enrollees);
```

devonbush (approved):
nice, glad we're filtering these out now.  One variable name suggestion

MatthewBemis (approved):

PR #969, opened 2024-06-28
[JN-1114] Answer edit history

MatthewBemis:

Adds a "View history" button to the survey response view. While most of this information is also displayed in the audit history table, it can be hard to piece together what's happening and you also can't view who originally answered the question.

This is intended to address https://github.com/broadinstitute/juniper/pull/955#discussion_r1651513466

Screenshot 2024-07-01 at 2 48 06 PM

Restart participant api, admin api, and admin UI
Repopulate demo
Make some answer edits as an admin or participant
Confirm that they show up in the response history view

_For a pre-populated test:_
View the basics survey response for Jonas Salk: https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants/HDSALK
You should see the edit history for their street address pre-populated:
Screenshot 2024-07-01 at 2 48 23 PM

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
@@ -302,7 +302,12 @@ public class SurveyResponseService extends ImmutableEntityService<SurveyResponse
 
     private Answer createAnswer(Answer answer, SurveyResponse response,
                                 Survey survey, PortalParticipantUser ppUser, ResponsibleEntity operator) {
-        answer.setCreatingEntity(operator);
+        if(operator.getParticipantUser() != null || operator.getAdminUser() != null) {
+            answer.setCreatingEntity(operator);
+        } else {
+            //Answers created by a system process (i.e. enrollee import) should be associated with the participant user
+            answer.setCreatingParticipantUserId(ppUser.getParticipantUserId());
+        }
         answer.setSurveyResponseId(response.getId());
         answer.setSurveyStableId(survey.getStableId());
         if (answer.getSurveyVersion() == 0) {
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
@@ -302,7 +302,7 @@ public class SurveyResponseService extends ImmutableEntityService<SurveyResponse
 
     private Answer createAnswer(Answer answer, SurveyResponse response,
                                 Survey survey, PortalParticipantUser ppUser, ResponsibleEntity operator) {
-        if (answer.getCreatingAdminUserId() != null) {
+        if (response.getCreatingAdminUserId() != null) {
             answer.setCreatingAdminUserId(operator.getAdminUser().getId());
         } else {
             // If an admin wasn't specified, we'll default to attributing this to the participant user.
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
@@ -303,7 +303,7 @@ public class SurveyResponseService extends ImmutableEntityService<SurveyResponse
     private Answer createAnswer(Answer answer, SurveyResponse response,
                                 Survey survey, PortalParticipantUser ppUser, ResponsibleEntity operator) {
         if (response.getCreatingAdminUserId() != null) {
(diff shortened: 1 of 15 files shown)


core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -302,14 +302,7 @@ public Answer updateAnswer(Answer existing, Answer updated, SurveyResponse respo
...
                                 Survey survey, PortalParticipantUser ppUser, ResponsibleEntity operator) {
-        if (answer.getCreatingAdminUserId() != null) {
-            answer.setCreatingAdminUserId(operator.getAdminUser().getId());
-        } else {
-            // If an admin wasn't specified, we'll default to attributing this to the participant user.
-            // It's also possible that a system process (such as import) could have created the response,
-            // but we still want to tie it to the participant user in that case.
-            answer.setCreatingParticipantUserId(ppUser.getParticipantUserId());
-        }
+        answer.setCreatingEntity(operator);
```
MatthewBemis: this should have been using the `operator` to begin with

ui-admin/src/study/participants/survey/SurveyFullDataView.tsx
```diff
@@ -105,19 +130,94 @@ export const ItemDisplay = ({
...
     </dt>
     <dl>
       <pre className="fw-bold">{displayValue}</pre>
     </dl>
   </>
 }
 
+/**
+ * Renders a dropdown with the edit history for a question response
+ */
```
devonbush: this feels like it belongs in a separate file

ui-admin/src/study/participants/survey/SurveyFullDataView.tsx
```diff
@@ -105,19 +130,94 @@ export const ItemDisplay = ({
...
+}
+
+/*
+ * Displays the original answer value and the entity responsible for answering it. If changes have
+ * been made to the answer, we backtrack through the change records to find the original answer.
+ */
+const getOriginalAnswer = (
+  question: Question | CalculatedValue, answer: Answer, changeRecords: DataChangeRecord[], users: AdminUser[]
+) => {
+  const originalChangeRecord = changeRecords.sort((a, b) => a.createdAt > b.createdAt ? 1 :
```
devonbush: let's use lodash sortBy here

ui-admin/src/study/participants/survey/SurveyFullDataView.tsx
```diff
@@ -105,19 +130,94 @@ export const ItemDisplay = ({
...
+    <FontAwesomeIcon icon={faArrowRight} className="mx-1"/>
+    <div className="bg-success-subtle fw-medium">{changeRecord.newValue}</div>
+  </div>
+}
+
+/*
+ * Displays the original answer value and the entity responsible for answering it. If changes have
+ * been made to the answer, we backtrack through the change records to find the original answer.
+ */
+const getOriginalAnswer = (
```
devonbush: I think of `gets` as information retrieval, rather than display.  So maybe call this `renderOriginalAnswer`

ui-admin/src/study/participants/survey/SurveyFullDataView.tsx
```diff
@@ -105,19 +130,94 @@ export const ItemDisplay = ({
...
+            </div>
+          </div>
+        </div>
+      )}
+      {getOriginalAnswer(question, answer, editHistory, users)}
+    </div>
+  </>
+}
+
+const getBeforeAndAfterAnswer = (changeRecord: DataChangeRecord) => {
```
devonbush: how about `renderChangeRecordSimple`?

devonbush (commented):
this is sweet!   I think you definitely chose the right data-fetching paradigm in grabbing the history at once.  

For the UX, I'd lean towards something that's less text-y, and also gives an at-a-glance indication of whether the answer has changed at all.  Maybe make the answer value itself clickable to show the dropdown instead of a separate "View History" label, and conditionally show either an info or history icon next to the answer value depending on whether or not there are change records for the answer

connorlbark (approved):
Nice!!! Works really well and it's a lot less code than I thought it would take to do this.

It would be cool to see the justification in-line, too, but we'd be missing the justification on create (since we aren't making a data change record for it) and it might get somewhat busy. I just fear the data audit table is becoming too unwieldy to be useful haha

devonbush (approved):
nice!  this will be a great feature to get demo feedback on. One more tweak you might consider is adding `text-muted` to the info/history icons so they spearate more from the answer.  Also maybe faChevronDown instead of faInfoCircle?

image

PR #971, opened 2024-06-30
[JN-1168] Fixing rule evaluation bug

devonbush:

This showed up as a bug when we were testing Hearthive outreach survey assignment.  Rule evaluations weren't working in the context of survey assignment.  The issue was that one of the context loading paths wasn't including the mailing address.  This is a quick fix.  On Monday, I want to strategize with Connor on the best way we can consolidate EnrolleeContext and EnrolleeSearchContexts to make them more robust to these sorts of bugs.

1. redeploy ApiAdminApp and repopulate demo
2. go to `https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants/HDBASC`
3. confirm the participant has been assigned the "Massachusetts survey" (but has not responded)
4. in the database, clear the tasks by running `delete from participant_task where target_stable_id = 'massachusettsSurvey';`
5. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms
6. Use the 'participant assignment' for the Massachusetts survey
7. click 'assign to sandbox participants'
8. confirm the survey is assigned to 2 participants (HDBASC and HDNEWB)


connorlbark (approved):
Hmm 🤔 Perhaps would could have a new service with a method like:
```
List<Enrollee> enrollees = ...;
List<EnrolleeContext> enrolleeContexts = ...;
List<EnrolleeSearchContext> searchCtx = enrolleeSearchContextService.load(enrollees); // does db fetching
List<EnrolleeSearchContext> searchCtx2 = enrolleeSearchContextService.load(enrolleeContexts); // does not do (or does minimal) db fetching
```
I'm also not opposed to just completely replacing EnrolleeSearchContext with EnrolleeContext. Or, perhaps more robustly, just having it take in `List<Enrollee>` and doing all the fetching, since (at least I think it should be assumed) that when you're using the evaluate functionality, it should be O(10) or O(100) at most. Additionally, each of these classes also has an overridable `evaluate(List<EnrolleeSearchContext>` which by default just runs through the list and evaluates one by one, but we could override that on ones that do db fetching to make them more efficient and only do one db fetch, so we wouldn't really lose much by making it just `List<Enrollee>` if we did that.

On a related meta note: Is there a reason we're using `evaluate` here as opposed to evaluating the expression on the enrolleesearchexpressiondao?

MatthewBemis (approved):

devonbush:
as discussed with Connor, we'll do a follow-on PR to consolidate EnrolleeSearch/Context

PR #972, opened 2024-06-30
[JN-1169] add BASE permission

devonbush:

This adds a "BASE" permission for operations that should be allowed for any signed-in user (e.g. reading semi-public config/forms).  This also enhances our test infrastructure so the we can assert that a given service has all public methods annotated with permissions.  This should help us continue our trend of avoiding mocks in service tests -- we can quickly assert a whole bunch of security stuff, and then focus the bulk of our test code on actual scenarios.

1. redeploy ApiAdminApp
2. in the admin tool, try creating, deleting, editing, and publishing surveys, and confirming stuff generally still works.


connorlbark (approved):

api-admin/src/main/java/bio/terra/pearl/api/admin/service/auth/AuthUtilService.java
```diff
@@ -25,6 +25,13 @@
 /** Utility service for common auth-related methods */
 @Service
 public class AuthUtilService {
+  /**
+   * 'BASE_PERMISSON' indicates the operation is permitted for any PortalAdminUser as they are
+   * authorized to access the given portal. It might include public-ish operations like viewing
+   * surveys, etc.
+   */
+  public static final String BASE_PERMISSON = "BASE";
```
MatthewBemis, line 33: For my own understanding: how does the idea of a "base" permission differ from a "read"/"view" permission? Is the idea that there will be other non-read actions that we'll want to capture with the base permission without having to enumerate every piece of functionally as a new permission, and so calling it "read" wouldn't be accurate?

api-admin/src/test/java/bio/terra/pearl/api/admin/service/forms/SurveyExtServiceTests.java
```diff
@@ -26,6 +38,60 @@ public class SurveyExtServiceTests extends BaseSpringBootTest {
   @Autowired private SurveyExtService surveyExtService;
   @Autowired private StudyEnvironmentSurveyService studyEnvironmentSurveyService;
   @Autowired private AdminUserFactory adminUserFactory;
+  @Autowired private PortalAdminUserFactory portalAdminUserFactory;
+  @Autowired private PortalFactory portalFactory;
+
+  @Test
+  public void assertAllMethods() {
+    AuthTestUtils.assertAllMethodsAnnotated(
```
MatthewBemis, line 46: 🎉

MatthewBemis (approved):
this is sweet! just have one question

api-admin/src/main/java/bio/terra/pearl/api/admin/service/auth/AuthUtilService.java
devonbush: I think there are reads that would not be included in the BASE permission.  For example, reading participant data shouldn't be in "BASE"--that should require a read_participant_data.  And there might be trivial writes that don't require a permission (e.g. updating your own user preferences)

MatthewBemis: Got it- makes sense! I appreciate the clarification

PR #973, opened 2024-07-01
[JN-1151] adding easy demo refresh button

devonbush:

adds an easy "Refresh demo" button to the populate home page

image

This also makes some UX tweaks to the populate portal page -- hiding the overwrite text in an infopopup and reducing oversized button.  (originally I added the reset demo button to that page, but realized it would be easier to find on the default page.)
image

1. go to https://localhost:3000/populate
2. press the button
3. confirm demo repopulates


MatthewBemis (approved):
This is such a great idea!

connorlbark (approved):
Really useful, I concur it's a great idea!!
Super minor, but for some reason, I have to click confirm twice before it actually repopulates demo. I also find the ergonomics of it not being its own tab slightly awkward, but maybe it could be expanded to be a `Quick` tab that also has buttons for ourhealth/hearthive/etc.? Either, way, it's dev only, so I'm down to ship as-is

devonbush:
good call on adding it to the menu.
image
I couldn't reproduce the double-click on the confirm issue.  Agree we don't need to solve that yet...

PR #979, opened 2024-07-03
[JN-1085] Report code cov to sonar [wip]

MatthewBemis:


devonbush (approved):
I see this is WIP, but generally looks good to merge and at least test

MatthewBemis:
@devonbush unfortunately it doesn't work yet- the Sonar step isn't able to find the test coverage reports. Likely needs a few more quick iterations before it's good to go. Sonar reports coverage for branches so we can test this without actually merging to `development`

PR #980, opened 2024-07-03
[JN-1175] family audit table

connorlbark:

Allows associating data change records with a family, and creates a view into changes related to the family.

Because these changes rely on the raw enrollee id, I created new columns `Enrollee` and `Target Enrollee` and hydrate them either with the members on the family or by fetching from the backend. This way it's easier to see who the change is related to.

That being said - it's not the greatest or most intuitive UI in the world. If you're not familiar with the internals of juniper, enrollee vs target enrollee might be confusing. But it feels like a poweruser thing that you would know what you're looking for when you go to this view, so I'm not sure it's worth spending lots of time refining it. IMO to make it better would require that the table was intelligent enough to have different views for each model (e.g., it shows a custom row for family enrollee creation, or deletion, or enrollee relation creation, etc.) which would be a very non-trivial amount of work.

- Go to `https://localhost:3000/demo/studies/heartdemo/env/sandbox/families/F_SALKFM`
- Do a bunch of operations on the family - change proband, add member, remove member, add relation, remove relation, etc.
- Verify that they show up in the audit table with the relevant enrollees attached

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/family/FamilyController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/family/FamilyController.java
@@ -110,4 +110,24 @@ public class FamilyController implements FamilyApi {
 
     return ResponseEntity.ok(family);
   }
+
+  @Override
+  public ResponseEntity<Object> listChangeRecords(
+      String portalShortcode,
+      String studyShortcode,
+      String envName,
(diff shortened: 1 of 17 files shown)


devonbush (approved):
nice!  but don't forget to include screenshots in your PR

MatthewBemis (approved):

PR #982, opened 2024-07-08
[JN-1123] add family view to participant list

connorlbark:

Allows grouping by family.

image
image

- Go to participant list
- Ensure no regressions in functionality
- Click `Family view`
- Click around, make sure everything is laid out sensibly
- Try searching - it should make the list of families smaller and only show the enrollees that match the search criteria
- Click into `F_SALKFM`
- Click `Families` on the navbar crumb and make sure it brings you to the family view
- Go to ourhealth and ensure that the family shortcode column doesn't show up


ui-admin/src/util/tableUtils.tsx
```diff
@@ -417,7 +418,7 @@ const defaultBasicTableConfig = {
 /** helper function for simple table layouts */
 export function basicTableLayout<T>(table: Table<T>, config: BasicTableConfig<T> = {}) {
   const { filterable } = { ...defaultBasicTableConfig, ...config }
-  return <table className={config.tableClass ? config.tableClass : 'table table-striped'}>
```
connorlbark, line 420: Might be controversial, but since these tables were becoming more complex, striping started to look kinda bad. And at the demo, nobody seemed to notice 😉.  Upon googling it, most people seem to recommend _not_ zebra striping, so I thought I'd open it to the council to see what people's thoughts are.

MatthewBemis: Hugely in favor of getting rid of the striped styling. I think it looks a lot cleaner like this!

MatthewBemis (approved):
Very nice, everything looks great!

ui-admin/src/study/families/FamilyLink.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * Links to a family, similar to the EnrolleeLink component.
+ */
+export const FamilyLink = ({
+  studyEnvContext,
+  family
+}: {
+  studyEnvContext: StudyEnvContextT,
+  family: Family
+}) => {
+  const path = `${studyEnvContext.currentEnvPath}/families/${family.shortcode}`
```
devonbush: let's pull this into a `familyPath` helper method in StudyEnvironmentRouter

ui-admin/src/study/families/FamilyLink.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const familyNames = getFamilyNames(family)
```
devonbush: To make it clearer that this isn't an array
```suggestion
  const familyNameString = getFamilyNameString(family)
```

ui-admin/src/study/families/FamilyRouter.tsx
```diff
@@ -23,7 +23,7 @@ export default function FamilyRouter({ studyEnvContext }: {studyEnvContext: Stud
 
   return <>
     <NavBreadcrumb value={studyEnvContext.currentEnvPath}>
-      <Link to={`${studyEnvContext.currentEnvPath}/participants`} className="me-1">
+      <Link to={`${studyEnvContext.currentEnvPath}/participants?groupByFamily=true`} className="me-1">
```
devonbush: you can use the `participantListPath` helper from StudyEnvironmentRouter here

ui-admin/src/study/families/FamilyView.tsx
```diff
@@ -55,13 +55,13 @@ export function LoadedFamilyView({ family, studyEnvContext, reloadFamily }:
 
   return <div className="ParticipantView mt-3 ps-4">
     <NavBreadcrumb value={family?.shortcode || ''}>
-      <Link to={`${currentEnvPath}/family/${family.shortcode}`}>
+      <Link to={`${currentEnvPath}/families/${family.shortcode}`}>
```
devonbush: use the familyPath helper here

ui-admin/src/study/participants/participantList/ParticipantList.tsx
```diff
@@ -1,62 +1,38 @@
...
+
+  const [searchParams, setSearchParams] = useSearchParams()
+
+  const groupByFamily = searchParams.get('groupByFamily') === 'true'
+  const setGroupByFamily = (groupByFamily: boolean) => {
+    setSearchParams(params => {
+      params.set('groupByFamily', groupByFamily.toString())
+      return params
+    })
+  }
```
devonbush: This is a pattern we now have in a couple different places.  I wonder if it's worth making a `useSingleSearchParam(paramName: string)` helper for this

ui-admin/src/study/participants/participantList/ParticipantList.tsx
```diff
@@ -65,147 +41,52 @@ function ParticipantList({ studyEnvContext }: {studyEnvContext: StudyEnvContextT
...
-      checked={table.getIsAllRowsSelected()} indeterminate={table.getIsSomeRowsSelected()}
-      onChange={table.getToggleAllRowsSelectedHandler()}/>,
-    cell: ({ row }) => (
-      <div className="px-1">
-        <IndeterminateCheckbox
-          checked={row.getIsSelected()} indeterminate={row.getIsSomeSelected()}
-          onChange={row.getToggleSelectedHandler()} disabled={!row.getCanSelect()}/>
-      </div>
+  const { isLoading } = useLoadingEffect(async () => {
+    const fullSearchExpression = concatSearchExpressions(
```
devonbush: instead of (or maybe in addition to) a `concatSearchExpressions` function, I'd lean towards making a "generateFullSearchExpression" function that take the searchExpression and familyLinkageEnabled as arguments

ui-admin/src/study/participants/participantList/ParticipantListTable.tsx
```diff
@@ -0,0 +1,228 @@
...
+import { Button } from 'components/forms/Button'
+import { FamilyLink } from 'study/families/FamilyLink'
+import { isEmpty } from 'lodash'
+import TableClientPagination from 'util/TablePagination'
+
+/**
+ * Participant table used by the participant list. Does not include searching functionality, just table
+ * rendering.
+ */
+function ParticipantListTable({
```
devonbush, line 46: Great move pulling this out into a separate file

ui-admin/src/study/participants/participantList/ParticipantListTableGroupedByFamily.tsx
```diff
@@ -0,0 +1,207 @@
...
+    header: 'Proband',
+    accessorKey: 'proband',
+    enableColumnFilter: false,
+    cell: ({ row }) => {
+      if (!row.original.proband) {
+        return <></>
+      }
+      return <EnrolleeLink studyEnvContext={studyEnvContext} enrollee={row.original.proband}/>
+    }
+  }, {
```
devonbush: this  `createAt` column occurs a lot in our code -- maybe pull it out into a `tableColumnUtils` file?

ui-admin/src/util/familyUtils.tsx
```diff
@@ -0,0 +1,23 @@
+import { Family } from '@juniper/ui-core'
+import { uniq } from 'lodash/fp'
+import { isEmpty } from 'lodash'
+
+/**
+ * Returns a string of family names, separated by commas and an 'and' before the last name.
+ * For example, a family with the members 'Alice Tester', 'Bob Reviewer', and 'Charlie Examiner'
+ * would return 'Tester, Reviewer and Examiner'.
+ */
+export const getFamilyNames = (family: Family) : string => {
```
devonbush: Normally, putting the data type of the function return in the name is redundant, but here I think it really helps clarify what's happening
```suggestion
export const getFamilyNameString = (family: Family) : string => {
```

ui-admin/src/util/tableUtils.tsx
devonbush: I agree it looks a bit nicer, but I find it harder to use -- my eyes have a harder time tracking across an individual row.  Maybe we pose the question to Erin to see if we can get a one-word response?  Alternatively, consider keeping the striping but getting rid of the borders around the stripes for a somewhat cleaner look:

image

ui-admin/src/study/participants/participantList/ParticipantListTableGroupedByFamily.tsx
```diff
@@ -0,0 +1,207 @@
...
+    header: '',
+    accessorKey: 'expanded',
+    enableColumnFilter: false,
+    enableColumnSort: false,
+    cell: ({ row }) => {
+      return <>
+        <button
+          className="btn btn-link m-0 p-0"
+          onClick={() => row.toggleExpanded()}>
+          {row.getIsExpanded()
```
devonbush, line 73: switch these to match the expand/collapse actions in the sidebar

ui-admin/src/study/participants/participantList/ParticipantListTable.tsx
```diff
@@ -0,0 +1,228 @@
...
+    },
+    filterFn: 'equals',
+    cell: info => info.getValue() ? <FontAwesomeIcon icon={faCheck}/> : ''
+  }], [study.shortcode, currentEnv.environmentName])
+
+  const familyLinkageEnabled = studyEnvContext.currentEnv.studyEnvironmentConfig.enableFamilyLinkage
+
+  const table = useReactTable({
+    data: participantList,
+    columns: useMemo(
```
devonbush: this double useMemo was really tricky to follow.  let's move this filtering into the initial column definition useMemo

devonbush (changes requested):
Overall really solid.  Lots of small stuff, none of which individually are deal-breakers, but there's enough of them that it's worth another round.  Also, we should definitely get Erin's quick take before we switch the table styling for the whole app

ui-admin/src/util/tableUtils.tsx
MatthewBemis: FWIW I bumped into Erin yesterday and showed off a few of our recent UX changes. The new table styling got a fairly indifferent 🤷. It doesn't seem like there are any well defined best practices around it

connorlbark: Agree on talking to erin - what prompted to me to do this now specifically is that striping looks (imo) objectively bad with the family view:
image
The reason for this is that the sub-participant-table has to be on its own row to be the whole width of the parent table (colspan = 5). I did some digging into bootstrap, and I don't think it's possible to modify the behavior of table-striped in any meaningful way. So I think the choices are: accept that it looks bad, turn off striping _just_ for the family view, turn off striping everywhere, or manually apply striping ourselves. If it's a dealbreaker, I wouldn't be opposed to doing manual striping - it would probably take all of about 10 minutes to get it working.

ui-admin/src/study/families/FamilyLink.tsx
connorlbark: Per discussion in standup, I think I have realized why `StudyEnvironmentRouter` methods are falling out of favor: if I have access to a studyEnvContext, which is most of the time, then I have to fully deconstruct it into its portal shortcode, study shortcode, etc. to use any of the routing methods, whereas studyEnvContext has a `currentEnvPath` that shortcuts all of that destructuring.

For the purposes of this PR, I don't mind going through and destructuring like that for all of them for consistency, but I do feel like we could do something like this:


```ts
type RouterContext = StudyEnvContextT | {portalShortcode: string, studyShortcode: string, envName: string}

const getCurrentEnvPath = (ctx: RouterContext) {
   ... do type checking etc to get current env ...
}

const familyRouter = (ctx: RouterContext) => {
    return `${getCurrentEnvPath(ctx)}/families`
}
```

So that we had the flexibilty to just go:

`familyRouter(studyEnvContext)` or fully deconstruct it, if we want.

ui-admin/src/study/participants/participantList/ParticipantList.tsx
connorlbark: good call, way more readable

ui-admin/src/study/participants/participantList/ParticipantListTableGroupedByFamily.tsx
connorlbark: 🤔 Hmm, they match on my machine; is it possible you didn't get the latest commits?

image

connorlbark:
@devonbush re-requesting feedback to get thoughts on table-striped approach (& review for other feedback items)

ui-admin/src/util/tableColumnUtils.tsx
```diff
@@ -0,0 +1,17 @@
+import { instantToDefaultString } from '@juniper/ui-core'
+import { ColumnDef } from '@tanstack/react-table'
+
+/**
+ * Creates a column definition for a "created at" column.
+ */
+export function createCreatedAtColumn<T>(): ColumnDef<T> {
```
devonbush: I think the extra create is a bit redundant here (and hard to read).  just like for our mocks we just call them, e.g., `mockSurvey` and not `createMockSurvey()`, I think `createdAtColumn` is fine. 
```suggestion
export function createdAtColumn<T>(): ColumnDef<T> {
```

devonbush (approved):
cool, we can have the striping off for awhile, and see how folks react

PR #983, opened 2024-07-08
[JN-1183] `include` search exp

connorlbark:

Needed for the group by family functionality, but it's so unrelated I figured it should be in its own PR. Allows you to fetch data in a search expression without filtering on it - for example, if the participant list needs family data but doesn't need any filtering, then it can do `<search filters> and include({family.shortcode})`.

The participant list also currently has a bug where it doesn't show `lastLogin` because participant user doesn't exist/isn't ever populated - we can solve that similarly by adding a `UserTerm` and tacking `include({user.lastLogin})` onto our participant list expressions.

Observe tests. If you want to try it out, swagger might be the easiest way.


core/src/main/antlr/bio/terra/pearl/core/antlr/CohortRule.g4
```diff
@@ -4,8 +4,7 @@ grammar CohortRule;
    See https://github.com/surveyjs/survey-library/blob/master/src/expressions/grammar.pegjs */
 
 // Parser rules
-expr: PAR_OPEN expr PAR_CLOSE | term OPERATOR term | expr AND expr | expr OR expr | NOT expr;
-term: NUMBER | STRING | VARIABLE | BOOLEAN | NULL | FUNCTION_NAME PAR_OPEN term (',' term)* PAR_CLOSE;
+expr: PAR_OPEN expr PAR_CLOSE | term OPERATOR term | expr AND expr | expr OR expr | NOT expr | INCLUDE PAR_OPEN term PAR_CLOSE;term: NUMBER | STRING | VARIABLE | BOOLEAN | NULL | FUNCTION_NAME PAR_OPEN term (',' term)* PAR_CLOSE;
```
devonbush: why did these get put on a single line?

devonbush (approved):
slick!

core/src/main/antlr/bio/terra/pearl/core/antlr/CohortRule.g4
connorlbark: I have no idea - I copied and pasted them so I guess copy/paste error. Good catch!

MatthewBemis (approved):

PR #984, opened 2024-07-08
[JN-1184] Enable editing of multi-language surveys

devonbush:

Adds a language selector to survey editor, which enables viewing and editing content for multi-language forms
image

Currently, text questions, html questions, and basic choice lists are supported for i18n editing.  This is not intended to provide exhaustive support for all i18n editing -- some cases, like editing the "none of the above" option text will need to be handled separately.  But in your testing, please note any cases like that that you find so we can ticket them appropriately

1. go to `https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms/surveys/hd_hd_mental`
2. edit some content in English and spanish
3. save the content
4. confirm the saved content reflects your changes


connorlbark (commented):
It seems like I am unable to edit the existing english on the basics form:

image

The other languages seem to work. I think it's because the english one is `default` in this survey instead of `en`. But maybe this is OK because future content will have `en` specified and won't use default - lmk what you think.

devonbush:
ah, good catch on the 'default' handling.  Instead of having "default" text, maybe we just default to the portal's default language?   @MatthewBemis  what do you think?

MatthewBemis:
@devonbush that generally sounds good to me- I think it's better to lean into our concept of a portal default anyway instead of relying on SurveyJS's logic. Although I'm not sure what would happen if you switched to a language that doesn't have all fields defined in the selected language. SurveyJS uses the default language as a fallback in that case. But maybe there's a way to override that in the surveyJS model

devonbush:
yeah, I think surveyjs currently just shows blanks if it can't find a text for the current language, and there is no 'default'.  I'll take a look if that can be changed.

devonbush:
Just updated our populated surveys to have "en" rather than "default"

connorlbark (approved):
Works great!
Per instructions: I only noticed that `Description` also needed i18n'ing

ui-admin/src/test-utils/mocking-utils.tsx
```diff
@@ -512,3 +529,33 @@ export const mockExpressionApis = () => {
...
+  languageCode: 'es',
+  languageName: 'Español',
+  id: '1'
+}
+
+/**
+ * renders the children in a PortalProvider context and simulating appropriate routes
+ * so that useStudyEnvParams hook works as expected. Hardcoded to sandbox for now
+ * */
+export const renderInPortalRouter = (portal: Portal, children: React.ReactNode, envName = 'sandbox') => {
```
MatthewBemis, line 548: nice!

MatthewBemis (approved):

PR #986, opened 2024-07-08
[JN-1119] adding support for i18n validations in ActivityImport

devonbush:

This adds some support for importing validation and required texts from Pepper.  We're assuming the logic will have to be hand-checked, but importing the translation text is a huge timesaver.

image

This also adds a superuser-only survey model import feature to the survey creation modal
image

1. Run 'PepperImportCliApp'
2. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms
3. click "add" under Research Surveys
4. in the "Import JSON" text box, paste the contents of the `pepper-import/out/medical-history.conf` file that was generated from your run above
5. click create (you may have to wait a while -- the survey is biiiig)
6. go to form preview
7. uncheck the "ignore validations" checkbox
8. go through the survey, observe the first question is required, and the required error text is internationalized

### DIFF
--- a/pepper-import/src/main/java/bio/terra/pearl/pepper/ActivityImporter.java
+++ b/pepper-import/src/main/java/bio/terra/pearl/pepper/ActivityImporter.java
@@ -161,12 +161,13 @@ public class ActivityImporter {
                     .name(pepperQuestionDef.getStableId())
                     .type(questionType)
                     .title(titleMap)
-                    .required(false)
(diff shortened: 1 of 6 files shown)


connorlbark (approved):
sweet! these imports are looking really sweet - it seems like we can automate more than I would have expected

MatthewBemis (approved):
very cool!

PR #992, opened 2024-07-10
[JN-1202] create enforceportalenrolleeperm

connorlbark:

Based on feedback from a PR last week - creates a new permission that also loads an enrollee and tries it out on the ProfileExtService

- Nothing to test, just observe tests


api-admin/src/main/java/bio/terra/pearl/api/admin/service/auth/EnforcePortalEnrolleePermissionAspect.java
```diff
@@ -0,0 +1,88 @@
...
+        studyEnvironmentService.verifyStudy(
+            authContext.getStudyShortcode(), authContext.getEnvironmentName());
+    authContext.setStudyEnvironment(studyEnv);
+
+    Enrollee enrollee =
+        enrolleeService
+            .findByShortcodeAndStudyEnvId(authContext.getEnrolleeShortcode(), studyEnv.getId())
+            .orElseThrow(() -> new NotFoundException("Enrollee not found"));
+
+    authContext.setEnrollee(enrollee);
```
connorlbark, line 66: lots of duplicated code with EnforcePortalStudyEnvPermissionAspect, but I'm not sure the best way to make it reusable (if we even want to do that)

devonbush: agreed -- I spent a while trying to dedupe stuff across the existing aspects, and couldn't figure it out because of the way these are based on annotations, which are interfaces, and therefore can't inherit from each other.

devonbush (changes requested):
You need to add a test that the enforcement code actually works, in addition to the annotation-detecting test

connorlbark:
> You need to add a test that the enforcement code actually works, in addition to the annotation-detecting test

Is there a particular place we test these, other than (in this case) on ProfileExtService? I couldn't find a place, if there is one, so I added tests to ProfileExtServiceTests

devonbush (approved):
nice!

MatthewBemis (approved):

PR #997, opened 2024-07-12
[JN-1119] Lots of AT survey polish

connorlbark:

Auto converts visibility expressions to surveyjs format!! I thought this would be difficult, but it ended up being really easy once copilot auto-generated the exact regex I wanted 😄. I figured it would be easier to test a bunch of new features in a batch, to save you all having to retake the same survey 15 times.

Also adds support for:
- email validation
- placeholder text
- picklist translations + yes/no question translations
- signature/comment questions
- date range validation
- regex validation
- other questions / details in checkbox list

Essentially, anything remaining _should_ be an edge case - this gets the surveys really about 90% there.

For ease of testing, you can unzip and use this atcp portal:
[atcp.zip](https://github.com/user-attachments/files/16249548/atcp.zip)
It has no surveys, but if you copy all of the `out/*.conf` files to the survey folder, all you need to do is populate atcp.

Note that in pepper, if a question doesn't have a title, then there will be no text above the question. With surveyjs, if you don't provide a title, the stableid will become the title. If you see questions with stableids hanging above them, that should be because they don't have a title in pepper.

- Run pepper import cli app
- Copy surveys to atcp portal and populate it
- Go to `Registration` survey
- Verify that if you switch the country, it will show up the appropriate options for the state/province
   - Note that these are only translated in english on the pepper side, so going to a different language isn't expected to work
- Go to contacting physician and genome study to verify that the checkboxes and dropdowns are translated properly; switch languages, etc.
- In genome study, verify that placeholders work for the dropdown questions, e.g. `Select gender...` and are translated
- Go to medical history and hit `No` and verify the only option is completion; also verify if you change the language Yes and No are translated
- Hit `Yes` and observe that more sections show up
- Go to the next page; hit `Other` on the `How was the diagnosis ...` question and see the essay box show up
- Observe that placeholders are working for text / numeric questions
- Go to question 10 (Cancer) and click `Yes`, observe that the next page has questions about cancer, go back and hit `No` and observe that the page is skipped
- Go to question 15 (Telangiectasia) and observe that clicking either prompts you to explain more
- Go to question 26 (Siblings), add a new one, turn on validation, and observe that date range validation works (pick one in the future, and one from a loooong time ago)
- Go to consent/assent and verify it asks for a signature
- Go to stay informed and verify that you can't submit it with an invalid email; change translations and verify the error is translated.


devonbush (approved):
This is great.  One thing you might want to consider is a custom ObjectMapper so that you don't need to serialize parameters that are set to nulls/default.  This will make the surveys shorter and a bit easier to work with.   E.g. every question has 
```
"validators" : [ ],
"required" : false
```

I believe this could be done by defining a new objectMapper (see ExtractionObjectMapperConfig for an example of defining a new mapper and then adding mixins), and then using mixins with @JsonGetter that returns null if the property is the default.
e.g.
```
@JsonGetter("required")
public Boolean isRequired() {
    return required ? true : null;
}
```

connorlbark:
> This is great. One thing you might want to consider is a custom ObjectMapper so that you don't need to serialize parameters that are set to nulls/default. This will make the surveys shorter and a bit easier to work with. E.g. every question has
> 
> ```
> "validators" : [ ],
> "required" : false
> ```
> 
> I believe this could be done by defining a new objectMapper (see ExtractionObjectMapperConfig for an example of defining a new mapper and then adding mixins), and then using mixins with @JsonGetter that returns null if the property is the default. e.g.
> 
> ```
> @JsonGetter("required")
> public Boolean isRequired() {
>     return required ? true : null;
> }
> ```

Great idea! I think I will do that in a follow-up PR since that seems non-trivial

MatthewBemis (approved):
Nice!

PR #1000, opened 2024-07-15
[JN-1191] Prototype split-view survey designer

MatthewBemis:

Adds a WIP prototype split-view designer. Unlike the "beta" site designer, this new survey designer _actually_ is a beta feature. It's missing a lot of functionality (particularly around easy navigation, panel editing, and adding new pages).

I tried to wire this in to using as many existing question designer components as possible.

The point of this PR is to:

1) Start building out the core components and UX of the new survey designer
2) Start getting something in front of people for feedback. I'm particularly interested in Whitney's feedback. For now, this requires superuser or prototype_tester permissions.

https://github.com/user-attachments/assets/e15c381f-29d3-4721-af23-39177ad5f35d

This does not have unit tests yet because I anticipate this will be changing quite a bit.

Rebuild UI core
Load up the form editor and try out the split designer


ui-admin/src/forms/designer/split/SplitFormElementDesigner.tsx
```diff
@@ -0,0 +1,102 @@
...
+    pages: [{ elements: [element] }],
+    questionTemplates: editedContent.questionTemplates
+  }
+  const surveyModel = surveyJSModelFromFormContent(surveyFromQuestion)
+
+  surveyModel.showInvisibleElements = true
+  surveyModel.showQuestionNumbers = false
+
+  // @ts-ignore
+  const elementType = element.type === 'panel' ? 'panel' : 'question'
```
connorlbark, line 47: I wonder how to make it easy in the UX to move questions into and out of panels. For example, I could see someone cleaning up ATs surveys spending a lot of time moving related questions into panels so that they're grouped together since the export exports without any concept of panels. Right now, in the UX, I'm not sure how we would make that functionality visible. Maybe a `Move Question` button that shows all the panels and questions and you can click where it should go?

I could also see them needing to move translations around - some way of doing "copy all translations for this label" and "paste all translations" could be nice - maybe next to every text field is a json button that converts the text field into a larger one with the translation object? that way you can copy/paste it and see all the translations for one question at a time without diving into the whole question's json?

Idk just some thoughts on things I have felt were missing while we're still early on 😄

ui-admin/src/forms/designer/split/PanelEditor.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return (
+    <>
+      <TextInput className={'mb-2'} label={'Title'} value={panel.title}
+        onChange={title => onChange({ ...panel, title })}/>
+      <div>
+        {panel.elements.map((element, elementIndex) => {
+          const questionLabelId = useId()
+
+          return <div key={elementIndex}>
+            <CollapsibleSectionButton
```
connorlbark, line 23: Maybe controversial, but since you're only confined to see one page at a time, I think I would personally prefer seeing these open by default - having some open and some collapsed when i opened the page confused me a little bit at first

connorlbark (approved):
Sweeeeet! An awesome first step towards a great survey editor. I'm kind of excited about it, I might throw in some of my own refinements as I'm working on AT survey import 😉

ui-admin/src/forms/designer/split/FormElementEditor.tsx
```diff
@@ -0,0 +1,96 @@
...
+ */
+export const FormElementEditor = ({
+  element, elementIndex, currentPageNo, editedContent, onChange, currentLanguage,
+  supportedLanguages
+}: {
+  element: FormElement, elementIndex: number, currentPageNo: number,
+  currentLanguage: PortalEnvironmentLanguage, supportedLanguages: PortalEnvironmentLanguage[],
+  editedContent: FormContent, onChange: (newContent: FormContent) => void
+}) => {
+  const [isStableIdLocked, setIsStableIdLocked] = useState(true)
```
devonbush, line 21: isStableIdLocked could use a comment

ui-admin/src/forms/designer/split/SplitFormElementDesigner.tsx
devonbush: yeah, related is a "copy question" button for all the times where you're changing the question text but giving the exact same multiple choice options

devonbush (approved):
Worked great in local testing.  For future improvements things I'd suggest:  

1.  look to hiding "description" "show other" and the other configurations behind a "more configuration" expander.  Especially "description" which is almost never used
2. move "required" to some sort of iconized style, or someway that it can be included in the main display without taking up as much vertical space.  
 3. Think about always showing at least one or two choices for multiple choice questions and having the expander be "show all"
 4. "Edit Question" is kind of redundant to have appear over every question/panel.  I'd replace it with a softer and smaller "question" or "panel" label.   Or maybe just an icon since the contents of the editor strongly tell you what it is

PR #1004, opened 2024-07-17
[JN-1205] fix repopulate demo

connorlbark:

This took an embarrassingly long time because apparently there was a comma in the foreign key's name. TIL that was even possible.

- Populate demo
- Do something to F_SALKFM (add a relation or whatev)
- Repopulate demo


devonbush (approved):
wow, TIL.  And yeah, I'm guessing that snuck by everyone since we're so used to reading JSON

MatthewBemis (approved):

PR #1006, opened 2024-07-18
[JN-1191] Split designer UX improvements

MatthewBemis:

Adds some more functionality related to panels (adding questions to panels, move questions within a panel, also fixes a bug with deleting panels)

Takes some of the great UX suggestions left on the previous PR to refine and simplify the view quite a bit. There are some other UX ideas that Connor and I have discussed regarding stealing a few things from the SurveyJS designer that seemed to work well and made things a bit more clear. Those are not included yet- this is really just the first batch of UX refinement before it's good enough to take it for a spin.

Before
Screenshot 2024-07-29 at 11 03 52 AM

After
Screenshot 2024-07-29 at 11 02 02 AM

Other functionality still to come:

* Better navigation
* i18n support
* Lots of tests
* Question template support
* Type improvements
* Calculated values
* Validators
* Better client-side validation for questions as they're being filled out by the admin

To avoid maintaining completely separate code paths for all components of the form designer, this has some minimal UX effects on the existing question designer:

Before
Screenshot 2024-07-29 at 11 00 11 AM

After
Screenshot 2024-07-29 at 11 00 24 AM

* Take the new survey editor for a spin. Try doing stuff with panels, make sure the new UX works as expected
* Also make sure that the existing survey designer experience isn't too dramatically altered


ui-admin/src/forms/designer/questions/BaseFields.tsx
```diff
@@ -44,24 +48,28 @@ export const BaseFields = (props: BaseFieldsProps) => {
...
+            rows={2}
+            value={i18nSurveyText(regularQuestion.description)}
+            onChange={value => {
+              onChange({
+                ...regularQuestion,
+                description: value
+              })
+            }}
+          />
+        </div>
```
connorlbark: Imo I think it would make sense to move this to the very bottom - it feels like this is a `misc` section that comes after more important things

ui-admin/src/forms/designer/split/FullQuestionDesigner.tsx
```diff
@@ -0,0 +1,159 @@
...
+          icon={faAsterisk}
+          aria-label={'Toggle required'}
+          onClick={() => {
+            onChange({
+              ...question,
+              isRequired: !question.isRequired
+            })
+          }}
+        />
+      </div>
```
connorlbark: Love this! My only confusion is that it's slightly unclear if blue or red means checked - I wonder if we did grey instead of blue it would be clearer? If there was some `optional` icon that would make our lives so much easier, but alas...

Maybe it's the presence of the asterisk at all that's confusing. What if it was just an empty box (faSquare) by default. If you clicked it, it filled the box with an asterisk?

ui-admin/src/forms/designer/questions/VisibilityFields.tsx
```diff
@@ -37,44 +39,47 @@ export const VisibilityFields = (props: VisibilityFieldsProps) => {
...
+              required={true}
+              value={question.visibleIf}
+              onChange={value => {
+                onChange({
+                  ...question,
+                  visibleIf: value
+                })
+              }}
+            />
+          </div>
```
connorlbark, line 80: I noticed that when I clicked `Conditionally show this question` the field shows up immediately being red but don't show up at the top as an error that blocks saving.

I would expect that it would show up red only if I put stuff in it and cleared it, or tried to save it without providing a value. I noticed this was a pattern across multiple flows - if I add a new question for example, a bunch of stuff shows up immediately red.

I don't mind the way it is now, but I wonder if it would be good to beef up our client side validation so that missing visibility expressions show up as errors that block saving, and none of these fields are red by default, only upon saving. Just something to consider 😄

ui-admin/src/forms/designer/split/FullQuestionDesigner.tsx
```diff
@@ -0,0 +1,159 @@
...
+              <TextFields
+                disabled={readOnly}
+                question={question}
+                onChange={onChange}
+              />
+            )
+          }
+          {
+            question.type === 'html' && <Textarea
+              disabled={readOnly}
```
connorlbark: I think this needs a bit of padding, it currently touches the area beneath it: 
```suggestion
            question.type === 'html' && <div className="mb-2"><Textarea
              disabled={readOnly}
```

connorlbark (approved):
Looks great! Loving how much easier things are already getting. I think it's officially time for me to stop using the JSON editor 😉. My comments are all just ideas that I wanna put out there, optional if you ever want to tackle them.

devonbush (approved):

ui-admin/src/forms/designer/questions/VisibilityFields.tsx
MatthewBemis: Agreed- the validation is super rough right now. I've added this to the list of "functionality still to come" in the PR description, which we can turn into backlog tickets.

ui-admin/src/forms/designer/split/FullQuestionDesigner.tsx
MatthewBemis: I just switched this to use text-muted when it's disabled, so there's a little bit more contrast in states. It's still not perfect, but I think we'll probably change this part of the UX once the prototype matures a bit

PR #1008, opened 2024-07-19
[JN-1214] adding variables to openapi params

devonbush:

Deletes one unused endpoint (SAM access), and then does a bunch of variable substitution to standardize and shorten params.  I've spot checked a number of places, but obviously haven't tested every route -- this was all Ctrl-R replacing though, so hopefully there aren't any errors.

Also a drive-by missing key fix in EnrolleeOverview.

1. Redeploy ApiAdminApp
2. spot check various admin tool functions


devonbush:
test failure is the SurveyServiceTest false positive Connor fixed in another branch

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -94,12 +64,12 @@ paths:
       tags: [ portal ]
       operationId: get
       parameters:
-        - { name: portalShortcode, in: path, required: true, schema: { type: string } }
```
connorlbark, line 97: I don't feel super strongly, but I was unfamiliar with the syntax so I was searching around, and it looks like you can define everything at the top of the file:
```
define: &background_color '#998088'
define: &text_color "#111122"

openapi: 3.0.3
...
```
Is there any reason we didn't do that and instead went for defining them as we went along?

connorlbark (approved):
clicked around for a bit and nothing seems to have regressed!

api-admin/src/main/resources/api/openapi.yml
MatthewBemis: I had a similar thought, I'd also be in favor of having these defined in a separate section- but am curious if there are benefits to doing it this way

MatthewBemis (approved):

api-admin/src/main/resources/api/openapi.yml
devonbush: works for me!

devonbush: bah, openapi doesn't support the `define` property

PR #1012, opened 2024-07-22
[JN-1217] triggered task completion

devonbush:

The immediate functionality provided by this is to enable triggers to update task statuses.  This will give HeartHive the ability to mark an outreach task as complete when a participant enrolls in the targeted study.

Long term, this lays the groundwork for making triggers much more general.  The next step would be to have a trigger action "assign survey".  this would enable us to get rid of StudyEnvironmentSurveys and have them just replaced by triggers on study enrollment.

The UX for this is currently bare-minimum CRUD.  I'm going to take an IOU on frontend tests, because I've got a redo of the Trigger list/editing screen in the works, so much of this will change.

image

image

1. check out the test in `TriggerActionServiceTests.testUpdateTaskStatus` and see that it makes sense
2. redeploy ApiAdminApp
3. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/notificationContent/triggers
4. confirm you can make a trigger with an Action of "Update task status"


core/src/main/java/bio/terra/pearl/core/model/notification/Trigger.java
```diff
@@ -29,19 +30,27 @@ public class Trigger extends BaseEntity implements VersionedEntityConfig {
...
+    private TriggerScope actionScope = TriggerScope.STUDY;
+    private TaskStatus statusToUpdateTo;
+    private String updateTaskTargetStableId; // will update any tasks with this stableId
+
+    // fields below are for triggerActionType of NOTIFICATION
     @Builder.Default
     private NotificationDeliveryType deliveryType = NotificationDeliveryType.EMAIL;
     private UUID emailTemplateId;
     private EmailTemplate emailTemplate;
     private String rule;
```
connorlbark, line 48: If we're going to have this `rule` here, and triggers are going to be more general to be the things that assign surveys, then what would be the distinction between the `eligibilityRule` on a survey and the `rule` here?

I ask this because for `required documents` tasks I don't think it makes sense to extend from surveys because we don't need surveyjs stuff, but we do want to get the assignment-related stuff, like `eligibilityRule` and `assignToExistingParticipants`, etc. This is making me wonder if it makes sense to abstract those things from surveys to something that only worries about task assignment - regardless of task type - which could be a Trigger. I think it would open a lot of doors and separate concerns better.

core/src/main/java/bio/terra/pearl/core/service/workflow/TriggerActionService.java
```diff
@@ -0,0 +1,84 @@
...
+    @Order(DispatcherOrder.ACTION)
+    public void handleEvent(EnrolleeEvent event) {
+        List<Trigger> applicableTriggers = triggerService
+                .findByStudyEnvironmentId(event.getEnrollee().getStudyEnvironmentId(), true)
+                // only EVENT triggers
+                .stream().filter(trigger  -> trigger.getTriggerType().equals(TriggerType.EVENT))
+                // that match the event type
+                .filter(trigger -> trigger.getEventType().eventClass.isInstance(event))
+                // that satisfy the trigger's rule
+                .filter(trigger -> EnrolleeRuleEvaluator.evaluateRule(trigger.getRule(), event.getEnrolleeContext()))
```
connorlbark, line 46: related to my other comment: is the plan to ever move this to a `search expression`?

ui-admin/src/study/notifications/TriggerBaseForm.tsx
```diff
@@ -0,0 +1,92 @@
...
+]
+
+const configTypeOptions: { label: string, value: TriggerType}[] = [
+  { label: 'Event', value: 'EVENT' },
+  { label: 'Task reminder', value: 'TASK_REMINDER' },
+  { label: 'Ad hoc', value: 'AD_HOC' }
+]
+
+/** configures the notification type and event/task type */
+export default function TriggerBaseForm({ trigger, setTrigger }:
```
connorlbark, line 32: i find the UX here very confusing, so I'm looking forward to the trigger list/editing rework!

connorlbark (approved):

core/src/main/java/bio/terra/pearl/core/model/notification/Trigger.java
devonbush: great question -- yes I suspect we would remove the "eligibilityRule" from Survey and have it just live on the Trigger, along with some of the other fields like `assignToExistingParticipants` (which would get renamed to something like `triggerForExistingParticipants`).

core/src/main/java/bio/terra/pearl/core/service/workflow/TriggerActionService.java
devonbush: oh, I hadn't even noticed this -- the trigger rules are using search expression syntax, so yes, they should use search expression evaluation.  (AFAIK no triggers have rules yet in prod)

ui-admin/src/study/notifications/TriggerBaseForm.tsx
devonbush: likewise!

MatthewBemis (approved):

PR #1021, opened 2024-07-25
[JN-1224] AT pre-enroll/registration flow

connorlbark:

Bunch of random changes for things I ran into:
- Get on site media controller automatically lowercases file input
   - (this is becuase the files get automatically lowercased on creation, so e.g. US_states.json becomes us_states.json. This is fine - but surveyjs doesn't have a `lower` function... so I either have to implement a lower function and apply it to the country code or lower case the input to site media - which is fine because everything is lowercased anyway)
- better null handling to search expressions
- `{isGovernedUser}` variable available in surveys, with the ability to set it on the admin form preview
- proxyProfile and profile as surveyjs vars were broken, so I fixed those
- Answer mappings weren't saved for prereg/pre-enroll on edit in admin tool, which is now fixed. Also, now calculated values are options for answer mappings.
- Supports different pepper survey types and general cleanup of AT import/export.
- Fixed issue where if you signed up for a country without states (e.g. `Guam`) then the participant profile should show undefined for stateProvince

You can test all of these out by importing this [this portal](https://drive.google.com/file/d/1qTFDp5xi0UXFeMMg9LucxP6Sn0gi_gIW/view?usp=drive_link) (which is in a new portal extracts folder) and playing around with it. You shouldn't use PepperImportCliApp, because these surveys have manual edits made to them.
- Testing site media stuff: fill out pre-enroll and try out different countries
- create proxy user and observe correct behavior
- observe that assents/consents are applied in the correct situations (consent < 7, assent 7 to 18, >=18 consent)
- observe that answer mappings are created appropriately 
- `{isGovernedUser}` and profile fixes can be tested by going to the `blood type` survey as a proxy and seeing that it correctly says your governed user's name.
   - go back as a normal/non-proxy user and the text should be different

### DIFF
--- a/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/SiteMediaController.java
+++ b/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/SiteMediaController.java
@@ -31,7 +31,9 @@ public class SiteMediaController implements SiteMediaApi {
         "https://hearthivedev.b2clogin.com", // HeartHive (prod)
         "https://hearthive.b2clogin.com", // HeartHive (demo)
         "https://gvascdev.b2clogin.com", // gVASC (demo)
-        "https://gvascprod.b2clogin.com" // gVASC (prod)
+        "https://gvascprod.b2clogin.com", // gVASC (prod)
+        "https://localhost:3000",
+        "https://sandbox.atcp.localhost:3001"
       },
       maxAge = 3600,
       methods = {RequestMethod.GET, RequestMethod.OPTIONS})
--- a/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/SiteMediaController.java
+++ b/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/SiteMediaController.java
@@ -32,7 +32,9 @@ public class SiteMediaController implements SiteMediaApi {
         "https://hearthivedev.b2clogin.com", // HeartHive (prod)
         "https://hearthive.b2clogin.com", // HeartHive (demo)
         "https://gvascdev.b2clogin.com", // gVASC (demo)
-        "https://gvascprod.b2clogin.com" // gVASC (prod)
+        "https://gvascprod.b2clogin.com", // gVASC (prod)
+        "https://localhost:3000",
+        "https://sandbox.atcp.localhost:3001"
       },
       maxAge = 3600,
       methods = {RequestMethod.GET, RequestMethod.OPTIONS})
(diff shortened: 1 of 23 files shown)


api-participant/src/main/java/bio/terra/pearl/api/participant/controller/SiteMediaController.java
```diff
@@ -31,7 +31,9 @@ public SiteMediaController(SiteMediaService siteMediaService) {
         "https://hearthivedev.b2clogin.com", // HeartHive (prod)
         "https://hearthive.b2clogin.com", // HeartHive (demo)
         "https://gvascdev.b2clogin.com", // gVASC (demo)
-        "https://gvascprod.b2clogin.com" // gVASC (prod)
+        "https://gvascprod.b2clogin.com", // gVASC (prod)
+        "https://localhost:3000",
+        "https://sandbox.atcp.localhost:3001"
```
connorlbark: Temp change to allow testing: must delete before merge. I'm not really sure the best way to enable local testing, since I was (for some reason that I don't honestly fully understand) getting a CORS error locally without this. I think we should figure out - in a separate PR - how we want to handle this. Since this works just fine without a `@CrossOrigin` annotation, Matt suggested a `get` and `getExternal` method, but I wonder if we can get away with just `origins = '*'`. I don't think there are any security implications of doing that since this is a fully public API anyway. The final option would be to implement a custom cors policy (🤢)

devonbush: if you were getting CORS errors for this on localhost, it's likely something is off in our vite proxy configs.  Try setting 'changeOrigin' to true for line 34 of `ui-participant/vite.config.ts`

core/src/main/java/bio/terra/pearl/core/service/survey/AnswerProcessingService.java
```diff
@@ -56,7 +50,11 @@ public void processAllAnswerMappings(
             return;
         }
         processProfileAnswerMappings(enrollee, answers, mappings, operator, auditInfo);
-        processProxyProfileAnswerMappings(enrollee, answers, mappings, ppUser, auditInfo);
+        // if the ppUser is not the same as the enrollee, we're in a proxy environment,
+        // so we should check proxy profile mappings. otherwise, ignore them.
+        if (!ppUser.getProfileId().equals(enrollee.getProfileId())) {
```
devonbush: Let's move this logic check to inside the processProxy... function, so that function owns proxy related logic

devonbush (changes requested):
looks great, appreciate all the small fixes.  Requesting changes so that we can figure out what's going on with the CORS stuff

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/SiteMediaController.java
connorlbark: Woo! Thanks to Devon, we figured it out. If we just strip the hostname in the survey: `"url": "/api/..."` then it works!

devonbush (approved):
nice

MatthewBemis (approved):

PR #1023, opened 2024-07-26
[JN-1199] improving publishing infrastructure

devonbush:

This adds portalId and environmentName to portalEnvironmentChangeRecord.  A basic table view for seeing the history is added, with a link from the main publishing page

image

image

image

There's no upgrade script for adding portalIds/envNames to past records -- I'll manually add a few in prod, but it's very non-trivial to deduce, and so wasn't worth the script.

Of note, this substantially improves our "renderInPortal" test helper, to make it very easy to run tests with given user permissions.  I'd like for us to move towards having the default be that frontend tests execute with non-superuser permissions.

1. redeploy ApiAdminApp
2. repopulate demo
3. go to https://localhost:3000/demo/studies/heartdemo/publishing
4. click "view history"
5. confirm the row appears, and a detail modal pops up on clicking "detail"


ui-admin/src/test-utils/mocking-utils.tsx
```diff
@@ -583,10 +602,20 @@ export const renderInPortalRouter = (portal: Portal, children: React.ReactNode,
   if (opts.mockWindowAlert) {
     jest.spyOn(window, 'alert').mockImplementation(jest.fn())
   }
+
+  if (opts.permissions) {
+    opts.user.portalPermissions = { [portal.id]: opts.permissions }
+  }
```
connorlbark, line 608: this is sweet!

connorlbark (approved):
love!

ui-admin/src/test-utils/mocking-utils.tsx
```diff
@@ -563,16 +567,31 @@ export const MOCK_SPANISH_LANGUAGE = {
...
+  adminUsers?: AdminUser[], // will be put in the AdminUserContext
+  user: AdminUser, // will be put in the UserContext
+  permissions?: string[] // convenience for setting permissions without setting a full user with portalPermissions
+}
+
+export const defaultRenderOpts = {
+  envName: 'sandbox',
+  mockWindowAlert: true,
+  adminUsers: [],
+  user: mockAdminUser(false),
```
MatthewBemis, line 582: This is a great change. Totally agree with testing as a non-superuser by default going forward!

populate/src/main/resources/seed/iam/roles.json
```diff
@@ -12,6 +12,14 @@
       "participant_data_view"
     ]
   },
+  {
+    "name": "publisher",
+    "displayName": "Study publisher",
+    "description": "publish changes to irb and live environments",
+    "permissionNames": [
+      "publish"
+    ]
```
MatthewBemis, line 21: I'm curious, why is this split out into a new role and not a part of `study_admin`? To me, study admin feels like the type of role that should be able to do _everything_ related to administering a study.

Or is the goal with this to treat this as a bit of a feature-gate since we're expanding this from superuser-only to non-superusers and we don't quite want to make the full jump yet?

MatthewBemis (approved):
Looks great. Just have one question related to the new publisher role.

populate/src/main/resources/seed/iam/roles.json
devonbush: exactly -- we're still not ready for, e.g., OurHealth to own their own publishing.  So we want this separate for now.  In a few months we'll fold this into study_admin

PR #1025, opened 2024-07-26
[JN-1221] Add SuperuserOnly annotation

MatthewBemis:

Adds an `@SuperuserOnly` auth annotation and shows usage of it in LoggingExtService

Restart AdminApiService
Log in as a superuser and confirm you can still load https://localhost:3000/logEvents
Log in as a regular study staff and go to the same page. Confirm that both: 1) the UI displays an error and 2) the API calls themselves return 403


api-admin/src/main/java/bio/terra/pearl/api/admin/service/auth/context/OperatorAuthContext.java
```diff
@@ -0,0 +1,17 @@
...
+
+@SuperBuilder
+@Getter
+@Setter
+public class OperatorAuthContext {
+  private AdminUser operator;
+
+  public static OperatorAuthContext of(AdminUser operator) {
+    return OperatorAuthContext.builder().operator(operator).build();
+  }
```
connorlbark, line 16: Should the other auth contexts extend this class? Or perhaps there could be an interface, `OperatorAwareAuthContext`. That way you can do

```
@SuperUserOnly
@EnforcePortalEnvPermission(...)
public static verySecureMethod(PortalEnvAuthContext authContext, ...) {...}
```

devonbush: I think it's safe to have them extend it.  The reason we had to do EnvironmentAware as an interface was to handle the multple permutations of portal/study/env contexts.  But I think all contexts will have an operator

devonbush (approved):
nice!

connorlbark (approved):

PR #1028, opened 2024-07-29
[JN-1199] fixing double-survey publishing

devonbush:

The critical part of this is the logic preventing multiple active versions of the same survey to be saved.  This will fix one of the bugs OurHealth had last week where two staff members saved the same survey at the same time.  Fun fact, this save lgoic revealed a random typo in a test written last year.

This also updates the UX for the survey history to make it easier to switch/edit past versions.  It now mimics the (more recent and less complex) flow of the Website editor.

image

1. redeploy ApiAdminApp
2. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms/surveys/hd_hd_socialHealth
3. in the ... dropdown, select "version history"
4. play around with editing/switching versions, confirm UX behaves as expected

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/study/StudyEnvironmentSurveyService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/study/StudyEnvironmentSurveyService.java
@@ -43,11 +43,30 @@ public class StudyEnvironmentSurveyService extends CrudService<StudyEnvironmentS
         return dao.update(ses);
     }
 
-    public Optional<StudyEnvironmentSurvey> findActiveBySurvey(UUID studyEnvId, String stableId) {
-        List<StudyEnvironmentSurvey> configs = dao.findActiveBySurvey(studyEnvId, stableId);
-        // we don't yet have robust support for having multiple surveys with the same stableId configured for an
-        // environment.  For now, just pick one
-        return configs.stream().findFirst();
+    @Transactional
+    @Override
+    public StudyEnvironmentSurvey create(StudyEnvironmentSurvey studyEnvSurvey) {
+        validateSurveyNotAlreadyActive(studyEnvSurvey);
+        return super.create(studyEnvSurvey);
+    }
+
+    @Transactional
+    @Override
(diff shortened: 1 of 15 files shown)


core/src/main/java/bio/terra/pearl/core/service/study/StudyEnvironmentSurveyService.java
```diff
@@ -43,11 +43,30 @@ public StudyEnvironmentSurvey deactivate(UUID id) {
...
+    public StudyEnvironmentSurvey update(StudyEnvironmentSurvey studyEnvSurvey) {
+        validateSurveyNotAlreadyActive(studyEnvSurvey);
+        return super.update(studyEnvSurvey);
+    }
+
+    public void validateSurveyNotAlreadyActive(StudyEnvironmentSurvey studyEnvSurvey) {
+        if (studyEnvSurvey.isActive() &&
+            dao.isSurveyActiveInEnv(studyEnvSurvey.getSurveyId(), studyEnvSurvey.getStudyEnvironmentId(), studyEnvSurvey.getId())
+        ) {
+            throw new IllegalArgumentException("Cannot save -- another version of the survey is already active, likely due to multiple saves overlapping.  Confirm no one else is working on the survey and then retry.");
```
connorlbark, line 64: nice! love a helpful error message

connorlbark (approved):
Worked great! I noticed a couple weird things, but they're unrelated to this PR, so I just wanted to document them. First, I tried going back to an old survey and changing the title and saving it. The title didn't update; it changed in surveyjs, but the language keys stayed the same. Second, I noticed when I went to an old version, it showed version 3 multiple times on the version page:
Screenshot 2024-07-30 at 9 49 12 AM
Super minor, but it confused me for half a second.

devonbush:
Good find on version 3 showing up twice -- that's because that's showing the respective launch times, and version 3 was launched 2 different times, and so it will show up twice.  It's just not apparent from the UX since all those launches are from the same date.  Not a bug, but maybe an opportunity for some improved UX down the road to make it clearer what we're showing

MatthewBemis (approved):

PR #1032, opened 2024-07-31
[JN-1237] Mixpanel for participant analytics

MatthewBemis:

Adds mixpanel events to participant UI

Future steps:

* Log events from EventService in the backend
* Deprecate/remove logging errors/webvitals to the DB (for now this PR logs to both the DB and Mixpanel)
* Enabling on the demo server in another PR in terra-helmfile

You both should have access to the Juniper Demo Mixpanel org: https://mixpanel.com/project/3368887/view/3875335/app/events

* Set two new env variables for your local participant api: `MIXPANEL_TOKEN=...` and `MIXPANEL_ENABLED=true`
* Restart participant API
* Grab the token from https://mixpanel.com/project/3368887/app/settings#account/projects (the token is public but want to show you were you can find it in case you're not familiar)
* Start doing things in the UI and confirm that the events show up in mixpanel (currently webvitals, JS errors, JS compatibility errors, language changes, form completions, profile edits, logins and logouts are captured)


api-participant/src/main/java/bio/terra/pearl/api/participant/controller/EventController.java
```diff
@@ -0,0 +1,31 @@
+package bio.terra.pearl.api.participant.controller;
+
+import bio.terra.pearl.api.participant.api.EventApi;
+import bio.terra.pearl.core.service.events.MixpanelService;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+
+@Controller
+public class EventController implements EventApi {
```
devonbush: let's consolidate this with LoggingController.  Also, let's look into whether we should drop/refactor our existing loggingUtils.tsx and use mixpanel to instrument JS errors and web vitals (I think we should)

api-participant/src/main/resources/application.yml
```diff
@@ -8,8 +8,6 @@ env:
     name: ${DATABASE_NAME:pearl}
     password: ${DATABASE_USER_PASSWORD:dbpwd}
     user: ${DATABASE_USER:dbuser}
-  sam:
-    basePath: ${SAM_ADDRESS:https://sam.dsde-dev.broadinstitute.org}
   dsm:
```
devonbush, line 11: excellent, we'll list this on our `migration from DSP` accomplishments checklist for the month

core/src/main/java/bio/terra/pearl/core/service/events/MixpanelService.java
```diff
@@ -0,0 +1,57 @@
...
+            JSONObject event = events.getJSONObject(i);
+            JSONObject updatedEvent = messageBuilder.event(
+                    null,
+                    event.getString("event"),
+                    event.getJSONObject("properties")
+                            .put("token", env.getProperty("mixpanel.token"))
+            );
+            log.info("Preparing Mixpanel event: {}", updatedEvent.toString());
+
+            delivery.addMessage(updatedEvent);
```
devonbush: pull this into a separate function?

devonbush (commented):
What you have looks solid.  I think the main thing to think about are how to consolidate this with all our current logging infrastructure.  I would lean towards logging all events to mixpanel, including our events from EventService and loggingUtils.

devonbush:
I'm not able to access that mixpanel page -- signing in with my broadinstitute.org account gives me an error.  (I can still see the other DSP projects in MixPanel though)

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/LoggingController.java
```diff
@@ -31,4 +36,26 @@ public ResponseEntity<String> log(Object body) {
       return ResponseEntity.unprocessableEntity().body(e.getMessage());
     }
   }
+
+  @Override
```
devonbush, line 44: add a comment about how this is different from `log` (i.e. this is for mixpanel)

ui-participant/src/Navbar.tsx
```diff
@@ -52,6 +53,8 @@ export default function Navbar(props: NavbarProps) {
   const navLinks = localContent.navbarItems
 
   async function updatePreferredLanguage(selectedLanguage: string) {
+    //track the language change
```
devonbush, line 56: nice!

MatthewBemis:
@devonbush hmmm did you get an email invite from Mixpanel earlier today? I think that might need to be accepted first before being able to access the new project

ui-participant/src/Navbar.tsx
```diff
@@ -52,6 +53,8 @@ export default function Navbar(props: NavbarProps) {
...
+    mixpanel.track('languageUpdated', { language: selectedLanguage })
```
devonbush: how about adding `source: 'navbar'` so we'll know where the language was changed?

ui-participant/src/Navbar.tsx
```diff
@@ -357,10 +361,16 @@ export const AccountOptionsDropdown = () => {
               {i18n('manageProfiles')}
             </button>
           </NavLink>}
-          <button className="dropdown-item" aria-label="change password" onClick={doChangePassword}>
+          <button className="dropdown-item" aria-label="change password" onClick={() => {
+            mixpanel.track('changePassword')
```
devonbush: same, think about adding a source property

devonbush (approved):
looks good!  A great start.  Might want to think about adding some source/scope to events tracked

api-participant/src/main/resources/application.yml
```diff
@@ -32,6 +30,9 @@ env:
     participantApi: ${PARTICIPANT_API_HOSTNAME:localhost:8081}
     adminUi: ${ADMIN_UI_HOSTNAME:localhost:3000}
     adminApi: ${ADMIN_API_HOSTNAME:localhost:8080}
+  mixpanel:
+    enabled: ${MIXPANEL_ENABLED:false}
+    token: ${MIXPANEL_TOKEN:}
```
connorlbark, line 35: Do we have a central place where we keep `things you need access to / to do when you onboard`? Not a big deal, but it could be nice to keep track that we added a new thing, and having formal instructions for getting the token could be nice.

ui-participant/src/Navbar.tsx
```diff
@@ -52,6 +53,8 @@ export default function Navbar(props: NavbarProps) {
   const navLinks = localContent.navbarItems
 
   async function updatePreferredLanguage(selectedLanguage: string) {
+    //track the language change
+    mixpanel.track('languageUpdated', { language: selectedLanguage, source: 'navbar' })
```
connorlbark, line 57: this seemed to be weird for me - if I changed from en -> es it didn't fire one, then es -> dev fired one for the es language, then dev -> es fired one for dev. I wonder if there's some weird component refresh thing going on? Perhaps it needs to be in a `useEffect`?

connorlbark (approved):
some slight weirdness on language change, but otherwise looks great! excited for the insights it can give us

api-participant/src/main/resources/application.yml
MatthewBemis: Good callout, this should definitely be added to the `./local-dev/render_environment_vars.sh` script that's part of the onboarding process

ui-participant/src/Navbar.tsx
MatthewBemis: I haven't been able to reproduce this :( I have seen related behavior in the past though, where selecting a language was similar out of date by one click. Have never been able to intentionally reproduce it though

PR #1033, opened 2024-08-01
[JN-1218] Quick protection against log event viewer OOM

MatthewBemis:

Caps the UI at requesting a max of 1000 log events.

Screenshot 2024-08-01 at 10 59 18 AM

Restart admin API
Confirm sure you can still load up to 1,000 events in the UI

### DIFF
--- a/api-admin/src/main/resources/api/openapi.yml
+++ b/api-admin/src/main/resources/api/openapi.yml
@@ -1908,6 +1908,7 @@ paths:
       parameters:
         - { name: eventTypes, in: query, required: false, schema: { type: string } }
         - { name: days, in: query, required: true, schema: { type: string } }
+        - { name: limit, in: query, required: true, schema: { type: integer } }
       responses:
         '200':
           description: logs
--- a/api-admin/src/main/resources/api/openapi.yml
+++ b/api-admin/src/main/resources/api/openapi.yml
@@ -1908,7 +1908,7 @@ paths:
       parameters:
         - { name: eventTypes, in: query, required: false, schema: { type: string } }
         - { name: days, in: query, required: true, schema: { type: string } }
-        - { name: limit, in: query, required: true, schema: { type: integer } }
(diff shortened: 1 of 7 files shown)


api-admin/src/main/resources/api/openapi.yml
```diff
@@ -1908,6 +1908,7 @@ paths:
       parameters:
         - { name: eventTypes, in: query, required: false, schema: { type: string } }
         - { name: days, in: query, required: true, schema: { type: string } }
+        - { name: limit, in: query, required: true, schema: { type: integer } }
```
devonbush: should this be required?  or should there be a default if none is specified?

core/src/main/java/bio/terra/pearl/core/dao/log/LogEventDao.java
```diff
@@ -37,11 +37,12 @@ public LogEvent create(LogEvent event) {
     );
   }
 
-  public List<LogEvent> listLogEvents(String days, List<LogEventType> eventTypes) {
+  public List<LogEvent> listLogEvents(String days, List<LogEventType> eventTypes, Integer limit) {
```
devonbush: this should take an `int` so that it can't be null -- I don't think the query will work if it is

ui-admin/src/health/LogEventViewer.tsx
```diff
@@ -125,6 +125,10 @@ export default function LogEventViewer() {
   return (
     <div className="px-4 py-2">
       {renderPageHeader('Log Events')}
+      <span className="text-muted fst-italic">
+        <FontAwesomeIcon icon={faInfoCircle} className="me-2"/>
+        This table is limited to 1000 log events. If you need to view more, query the database directly.
```
devonbush:
```suggestion
        This table is limited to 1000 log events. If you need to view more, contact support.
```

devonbush (approved):
good safeguarding.  Just some suggestions around using defaults and `int` to avoid null errors on the backend

connorlbark (approved):

PR #1036, opened 2024-08-03
[JN-1263]  Create snythetic participants via UI

devonbush:

This will let us generate synthetic participants in the OurHealth and Hearthive demo environments even after we repopulate them from prod configs.  It will also let us (and customers) generate synth participants in prod to more easily test configs.

image

By specifying a username, this will also let us create synthetic participants that we can log into (e.g. jsalk@test.com)

image

To make room in the UX, I've created a new "Actions" menu that consolidates this with the 'download' and 'email' options.  

1. redeploy ApiAdminApp
2. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants
3. click 'Actions'
4. click 'Add synthetic participant'
5. pick a type, then create
6. observe the participant created in the participant list


api-admin/src/main/java/bio/terra/pearl/api/admin/service/PopulateExtService.java
```diff
@@ -114,54 +128,65 @@ public SiteContent populateSiteContent(
...
+            authContext.getStudyShortcode(),
+            authContext.getEnvironmentName(),
+            new HashMap<>(),
+            false,
+            null);
     return enrolleePopulator.populate(config, overwrite);
   }
 
+  @EnforcePortalStudyEnvPermission(permission = "BASE")
+  public Enrollee populateEnrollee(
```
MatthewBemis, line 147: Should we add an extra check here to prevent this endpoint from being used in `live`? I agree that this should be usable in Sandbox and IRB (as you've done in the frontend code)

MatthewBemis (approved):
One question, otherwise looks great

denniscunningham (approved):

api-admin/src/main/java/bio/terra/pearl/api/admin/service/PopulateExtService.java
devonbush: good question.  I waffled on that, since I could see a case where someone wants to create a synthetic participant in live to test some published configs (and then will withdraw them).  And so it's disabled on the frontend but not the backend.  The thinking is that the backend API might be useful for developers, e.g. populating enrollees into a live local environment

PR #1037, opened 2024-08-05
[JN-1194] Admin emails

connorlbark:

Adds the ability to trigger admin emails. These are sent to everyone in the study staff - in the future, I could see this being configurable ('only send to these admins') so it doesn't spam everyone.

- (skip if team testing) delete `REDIRECT_ALL_EMAILS_TO` in your environment variables for participant api to test out what happens in the real world
- populate demo
- go to emails & notifications and find the admin notif on study enrollment one to see what an admin notification looks like
- try a test email and make sure it ends up in your inbox
- enroll a new user in demo that says no to the marketing emails question & ensure you do _not_ get an email (we're not emailing superusers)
- add multiple versions of your email as admins, e.g. `email+1` `email+2`
- enroll a new user in the demo portal with no to marketing emails and make sure that you get multiple emails and each person is in the to line
- make sure the email looks good & and the enrollee url link works
- (skip if team testing) re-add `REDIRECT_ALL_EMAILS_TO` and trigger an email and observe that it is only sent to your email, and it shows `Redirected from <email>, <email>`
- look at the relevant participant's timeline page

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/notification/email/AdminEmailService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/notification/email/AdminEmailService.java
@@ -31,9 +31,9 @@ public class AdminEmailService {
   private final StudyEnvironmentService studyEnvironmentService;
   private final StudyService studyService;
   private final PortalEnvironmentService portalEnvironmentService;
-  private EmailTemplateService emailTemplateService;
-  private ApplicationRoutingPaths routingPaths;
-  private SendgridClient sendgridClient;
+  private final EmailTemplateService emailTemplateService;
+  private final ApplicationRoutingPaths routingPaths;
+  private final SendgridClient sendgridClient;
 
   public AdminEmailService(EmailTemplateService emailTemplateService,
                            ApplicationRoutingPaths routingPaths,
--- a/core/src/main/java/bio/terra/pearl/core/service/notification/email/AdminEmailService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/notification/email/AdminEmailService.java
@@ -5,6 +5,7 @@ import bio.terra.pearl.core.model.notification.*;
 import bio.terra.pearl.core.model.participant.Enrollee;
 import bio.terra.pearl.core.model.portal.Portal;
 import bio.terra.pearl.core.model.portal.PortalEnvironment;
+import bio.terra.pearl.core.model.portal.PortalEnvironmentConfig;
 import bio.terra.pearl.core.model.study.Study;
 import bio.terra.pearl.core.model.study.StudyEnvironment;
 import bio.terra.pearl.core.service.admin.AdminUserService;
@@ -12,6 +13,7 @@ import bio.terra.pearl.core.service.exception.NotFoundException;
 import bio.terra.pearl.core.service.notification.NotificationContextInfo;
 import bio.terra.pearl.core.service.notification.NotificationService;
 import bio.terra.pearl.core.service.notification.substitutors.AdminEmailSubstitutor;
+import bio.terra.pearl.core.service.portal.PortalEnvironmentConfigService;
 import bio.terra.pearl.core.service.portal.PortalEnvironmentService;
 import bio.terra.pearl.core.service.portal.PortalService;
 import bio.terra.pearl.core.service.rule.EnrolleeContext;
@@ -42,13 +44,14 @@ public class AdminEmailService {
   private final AdminUserService adminUserService;
   private final PortalService portalService;
   private final NotificationService notificationService;
+  private final PortalEnvironmentConfigService portalEnvironmentConfigService;
 
   public AdminEmailService(EmailTemplateService emailTemplateService,
                            ApplicationRoutingPaths routingPaths,
                            SendgridClient sendgridClient,
                            StudyEnvironmentService studyEnvironmentService,
                            StudyService studyService,
-                           PortalEnvironmentService portalEnvironmentService, AdminUserService adminUserService, PortalService portalService, NotificationService notificationService) {
+                           PortalEnvironmentService portalEnvironmentService, AdminUserService adminUserService, PortalService portalService, NotificationService notificationService, PortalEnvironmentConfigService portalEnvironmentConfigService) {
     this.emailTemplateService = emailTemplateService;
     this.routingPaths = routingPaths;
     this.sendgridClient = sendgridClient;
@@ -58,6 +61,7 @@ public class AdminEmailService {
     this.adminUserService = adminUserService;
     this.portalService = portalService;
     this.notificationService = notificationService;
+    this.portalEnvironmentConfigService = portalEnvironmentConfigService;
   }
 
   @Async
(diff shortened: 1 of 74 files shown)


core/src/main/java/bio/terra/pearl/core/service/notification/email/AdminEmailService.java
```diff
@@ -4,45 +4,63 @@
...
 
   @Async
   public void sendWelcomeEmail(Portal portal, AdminUser adminUser) {
     NotificationContextInfo contextInfo = loadContextInfo(WELCOME_EMAIL_TEMPLATE_STABLEID, WELCOME_EMAIL_TEMPLATE_VERSION, portal);
-    sendEmail(contextInfo, adminUser);
+    sendEmail(contextInfo, adminUser, null);
   }
 
-  public void sendEmail(NotificationContextInfo contextInfo, AdminUser adminUser) {
+  public void sendEmail(NotificationContextInfo contextInfo, AdminUser adminUser, @Nullable EnrolleeContext enrolleeContext) {
```
devonbush: Our current practices disfavor `@Nullable` annotations.  Instead, use a comment to explain what will happen if it's null
```suggestion
  public void sendEmail(NotificationContextInfo contextInfo, AdminUser adminUser, EnrolleeContext enrolleeContext) {
```

core/src/main/java/bio/terra/pearl/core/service/workflow/TriggerActionService.java
```diff
@@ -50,12 +63,30 @@ public void handleEvent(EnrolleeEvent event) {
...
                         event.getPortalParticipantUser().getPortalEnvironmentId());
+            } else if (TriggerActionType.ADMIN_NOTIFICATION.equals(trigger.getActionType())) {
+                sendEmailToAllAdmins(trigger, event);
             } else if (TriggerActionType.TASK_STATUS_CHANGE.equals(trigger.getActionType())) {
                 updateTaskStatus(trigger, event);
             }
         }
     }
 
+    private void sendEmailToAllAdmins(Trigger trigger, EnrolleeEvent event) {
```
devonbush: Hmm... seeing this next to the method above suggests that we've got too much unrelated logic in the same class.  I'd move this entire method to the AdminEmailService.   TriggerActionService can own loading triggers and dispatching, and other services can handle operations

Not sure where the best place is for the updateStatus logic (maybe ParticipantTaskService?) but that's not essential to figure out in this PR

devonbush (changes requested):
Overall looks good, some minor service rearranging suggestions.  Tested locally, was able to add an admin email to consent events and it worked as expected.  

Requesting changes because this needs a populate example inside the demo study.  One idea might be to have it triggered on study enrollment of someone who opts out of marketing comms so it's easy to manually test but also doesn't spam us during populate.  (another possibility would be to trigger an admin email on kit sent or received events but for enrollees from a given state)

core/src/main/java/bio/terra/pearl/core/service/workflow/TriggerActionService.java
```diff
@@ -50,12 +63,30 @@ public void handleEvent(EnrolleeEvent event) {
...
+                .orElseThrow(() -> new IllegalStateException("Portal not found"));
+
+        EmailTemplate emailTemplate = emailTemplateService.find(trigger.getEmailTemplateId())
+                .orElseThrow(() -> new NotFoundException("Email template not found"));
+        emailTemplateService.attachLocalizedTemplates(emailTemplate);
+
+        NotificationContextInfo contextInfo = adminEmailService.loadContextInfoFromEnrollee(emailTemplate, portal, event.getEnrolleeContext().getEnrollee().getStudyEnvironmentId());
+
+        List<AdminUser> adminUsers = adminUserService.findAllWithRolesByPortal(portal.getId());
+        for (AdminUser adminUser : adminUsers) {
```
devonbush: SendGrid should support sending email that has multiple cc'ed recipients, and that will save us this loop (although might require refactoring some of the underlying methods to take a List). 

We definitely should do that, although it's fine if that's in a separate PR

devonbush (commented):
Also, we should make sure these emails show up in the timeline for the enrollee (they currently do not).  I think it'll be important for customers to be able to confirm if/when an admin alert was sent, and to who.    That also means you don't have to mock out sendgrid as much, you can just query the DB in tests to determine whether the email went out related to someone

core/src/main/java/bio/terra/pearl/core/service/notification/email/SendgridClient.java
```diff
@@ -38,8 +37,8 @@ public SendgridClient(Environment env, ApplicationRoutingPaths applicationRoutin
   public String sendEmail(Mail mail) {
     if (StringUtils.isEmpty(sendGridApiKey)) {
       // if there's no API key, (likely because we're in a CI environment), don't even attempt to send an email
-      log.info("Email send skipped: no sendgrid api provided");
-      throw new UnsupportedOperationException("Attempted to send email without sendgrid key");
+      log.warn("Email send skipped: no sendgrid api provided");
+      return null;
```
connorlbark: It felt weird that we were throwing a full exception on this and it made testing difficult - is a warning log and not saving the key the expected behavior here?

devonbush: good question -- it's definitely an error if this happens in prod.  So maybe have it only be an error if deploymentZone is prod, and otherwise have it be a warning

devonbush (commented):
Nice, the refactor to support multi-cc's looks like it's handling the complexity well.  Just need to add an admin email trigger to demo populate so this is easier to test

core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
```diff
@@ -158,7 +156,7 @@ public boolean shouldSendEmail(Trigger config,
                     enrolleeContext.getEnrollee().getShortcode(), config.getId(), config.getPortalEnvironmentId());
             return false;
         }
-        if (enrolleeContext.getProfile().isDoNotEmailSolicit() && config.getTaskType().equals(TaskType.OUTREACH)) {
+        if (enrolleeContext.getProfile().isDoNotEmailSolicit() && Objects.nonNull(config.getTaskType()) && config.getTaskType().equals(TaskType.OUTREACH)) {
```
connorlbark, line 159: Drive by fix: task type is not technically required, so this could throw null pointer errors.

populate/src/main/resources/seed/portals/demo/studies/heartdemo/emails/adminNotificationTest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "stableId": "hd_hd_adminKitReceived",
+  "name": "Kit Received",
```
devonbush: I don't think this name/stableId matches the content

devonbush (approved):
Awesome, everything works great locally.  Just one cleanup of name/stableId leftovers

populate/src/main/resources/seed/portals/demo/studies/heartdemo/emails/adminNotificationTest.json
connorlbark: oops

core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
MatthewBemis: Ah, thanks for catching this and fixing it!

MatthewBemis (approved):

PR #1038, opened 2024-08-05
[JN-1249] Configurable kit types

MatthewBemis:

Adds configurable kit types to study settings

Screenshot 2024-08-08 at 10 42 15 AM

* Restart admin api to repopulate the new kit types
* Confirm that you can add/remove kit types: https://localhost:3000/demo/studies/heartdemo/env/sandbox/settings
* Confirm that you can publish kit type changes made in sandbox to IRB and Live environments


api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtService.java
```diff
@@ -51,40 +55,86 @@ public StudyEnvironment update(PortalStudyEnvAuthContext authContext, StudyEnvir
...
+  protected void validateNoRemovalOfKitTypes(
+      List<String> updatedKitTypes, List<KitType> existingKitTypes) {
+    Set<String> updatedKitTypeSet = new HashSet<>(updatedKitTypes);
+    Set<String> existingKitTypeSet =
+        existingKitTypes.stream().map(KitType::getName).collect(Collectors.toSet());
+
+    if (!updatedKitTypeSet.containsAll(existingKitTypeSet)) {
+      throw new BadRequestException("You may not remove a kit type from a study environment");
+    }
+  }
```
connorlbark: Should we check to see if any kits of this type exist and only under those circumstances not allow deletion? I could see someone misclicking then not being able to remove it.

Also, I think this error should be an `IllegalArgumentException` - `BadRequest` triggers a 500 so the user doesn't see the error, so I was a bit confused at first. There's a few other usages of BadRequestException in this file that I'm not sure should trigger 500s also

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/StudyEnvironmentController.java
```diff
@@ -68,17 +68,47 @@ public ResponseEntity<Object> getKitTypes(
...
+  public ResponseEntity<Object> patchKitTypes(
+      String portalShortcode, String studyShortcode, String envName, Object body) {
+    AdminUser operator = requestService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+    PortalStudyEnvAuthContext authContext =
+        PortalStudyEnvAuthContext.of(operator, portalShortcode, studyShortcode, environmentName);
+
+    List<String> kitTypeNames =
+        objectMapper.convertValue(
+            body, objectMapper.getTypeFactory().constructCollectionType(List.class, String.class));
```
connorlbark: TIL about `constructCollectionType`, that's handy to know

connorlbark (approved):
sweet! worked great for me

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/StudyEnvironmentController.java
devonbush: I also hadn't seen this before, nifty.

But unless there's an advantage to doing it this way, let's stick with using `TypeReference`, which is how we do it in other controllers (see e.g. MailingListController:51.  (IMHO the syntax is a little easier to parse)

api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtService.java
devonbush: concur. with IllegalArgument instead of BadRequest

devonbush: you could either check that no kits of that type exist, or check that that no types that are configured for prod have been removed.  that way someone can experiment with types in sandbox

api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtService.java
```diff
@@ -51,40 +55,86 @@ public StudyEnvironment update(PortalStudyEnvAuthContext authContext, StudyEnvir
...
+        authContext.getStudyEnvironment().getId());
   }
 
-  public StudyEnvStats getStats(
-      AdminUser operator, String portalShortcode, String studyShortcode, EnvironmentName envName) {
-    StudyEnvironment studyEnv = authToStudyEnv(operator, portalShortcode, studyShortcode, envName);
-    return new StudyEnvStats(
-        enrolleeService.countByStudyEnvironmentId(studyEnv.getId()),
-        withdrawnEnrolleeService.countByStudyEnvironmentId(studyEnv.getId()));
+  @EnforcePortalStudyEnvPermission(permission = "study_settings_edit")
```
devonbush, line 74: Let's make this SandboxOnly, and add kitTypes to things that you can publish.  You should be able to model KitTypes as a ListChange.  Take a look at PortalPublishingService.applyChangesToLanguages and then you'll want to write a similar method in StudyPublishingService.
Updating StudyEnvDiff.tsx should also be straightforward (see line 207 of PortalEnvDiffView.tsx for the corresponding list diffing code you'll want to borrow)

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/StudyEnvironmentController.java
```diff
@@ -68,17 +68,47 @@ public ResponseEntity<Object> getKitTypes(
...
-        studyEnvExtService.getKitTypes(adminUser, portalShortcode, studyShortcode, environmentName);
+    PortalStudyEnvAuthContext authContext =
+        PortalStudyEnvAuthContext.of(adminUser, portalShortcode, studyShortcode, environmentName);
+
+    List<KitType> kitTypes = studyEnvExtService.getKitTypes(authContext);
+    return ResponseEntity.ok(kitTypes);
+  }
+
+  @Override
+  public ResponseEntity<Object> patchKitTypes(
```
devonbush: I don't think we're in the habit of using `patch` in method names.  I'd just as soon call this "updateKitTypes", but I could be persuaded otherwise

ui-admin/src/study/StudySettings.tsx
```diff
@@ -46,9 +48,33 @@ export function StudyEnvConfigView({ studyEnvContext, portalContext }:
...
+
+  const loadConfiguredKitTypes = async () => {
+    const selectedKitTypes = await Api.fetchKitTypes(studyEnvParams)
+    setSelectedKitTypes(selectedKitTypes.map(kt => ({ value: kt.name, label: kt.displayName })))
+  }
+
+  useEffect(() => {
+    loadAllowedKitTypes()
+    loadConfiguredKitTypes()
+  }, [studyEnvContext])
```
devonbush:
```suggestion
  }, [studyEnvContext.curretEnvPath])
```
making the dependency the path (string) rather than the object reference  is more concrete

devonbush (changes requested):
looks good -- some code suggestions.  Requesting changes to include this in publishing infrastructure.  Less because it's really important to be able to publish-control these types, and more just to keep the overall product experience simpler "everything is changed in sandbox, and then published"

api-admin/src/main/java/bio/terra/pearl/api/admin/service/study/StudyEnvironmentExtService.java
MatthewBemis: I thought about this more and I don't think there's actually a good reason to restrict kit type removal at all now that it's going through the publishing infrastructure.

There aren't any dangling FK constraints to worry about either as the actual kit requests reference the `kit_type` table, not the `study_environment_kit_type` table (which I think is right). Since the `kit_type`s are global and populated by the seed kitTypes.json file, those shouldn't be going anywhere

ui-admin/src/study/StudySettings.tsx
```diff
@@ -143,5 +177,27 @@ export function StudyEnvConfigView({ studyEnvContext, portalContext }:
       {isLoading && <LoadingSpinner/>}
       {!isLoading && 'Save study config'}
     </Button>
+
+    { user?.superuser &&
```
devonbush: Let's either just expose this outright, or maybe make a new permission "study_kit_type_edit", and put it under the "power_user" role.   

Minimizing the amount of stuff we gate behind superuser will minimize the amount of code changes we have to make later

devonbush (approved):
Works great locally.  One suggestion about changing the permission level so it's not superuser-gated

PR #1039, opened 2024-08-07
[JN-1234] handle dynamic panels effectively on answer save/export

connorlbark:

For questions within dynamic panels, "child" questions are created. These questions do not have an answer object associated with them; rather, they are extracted from their parent question at the time of export. E.g.,  the parent's answer value is parsed as json, then the value is extracted like so: 

if parent is "repeatable" (e.g., dynamic panel):
`parentAnswerValue[repeatIndex][childQuestionStableId]`.
if not repeatable (e.g., matrix) [not implemented, but just requires creating the child questions]:
`parentAnswerValue[childQuestionStableId]`

the number of repeats is calculated at time of export, so there could theoretically be as many repeats as you want. if there's no data, then it will have at least `[0]` show up, so you can introspect what the export will look like.

I'm debating refactoring `child`/`parent` language - perhaps `derived`/`source`? let me know what you guys think.

- create new enrollee
- fill out new 'languages' question at the bottom of the first page of the basics in demo
- note that you can't find these on the enrollee's response page - you can still see the full json object, but because this is mirroring the questions in surveyjs and these are kind of "fake questions" it would have been non-trivial to add here. It seems to me more important that they export nicely than that you can see them here, since you still have visibility into the data.
- find answers on data export preview 
- download & find answers in data export 
- go to data export in `live` and observe that the `hd_hd_basicInfo.languages.language[0]` / `proficiency[0]` show up even without data


core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
```diff
@@ -110,7 +152,17 @@ public static String unmarshalSurveyQuestionChoices(JsonNode question) {
...
+
+                String text;
+                if (textNode.isTextual()) {
+                    text = textNode.asText();
+                } else {
+                    // todo: make this grab default language
+                    text = textNode.get("en").asText();
+                }
+
+                choices.add(new QuestionChoice(choice.get("value").asText(), text));
```
connorlbark, line 165: Drive by fix: any questions that have text that was translated into multiple languages would have QuestionChoices with empty text strings, since we were trying to turn a json object into a string and it just doesn't do it

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
```diff
@@ -110,7 +153,20 @@ public static String unmarshalSurveyQuestionChoices(JsonNode question) {
...
+                    text = textNode.asText();
+                } else {
+                    if (textNode.has("en")) {
+                        text = textNode.get("en").asText();
+                    } else {
+                        text = textNode.get("value").asText();
+                    }
+                }
+
+                choices.add(new QuestionChoice(choice.get("value").asText(), text));
```
connorlbark, line 182: So, this is a drive by fix that might need more discussion. If you create a question with choices that have multi-language labels, our survey parsing infrastructure would create it with an empty string for the text, since `.asText()` will return an empty string if you call it on an object. This means in heartdemo, any multi-lang dropdown will actually show up empty on export.

The "correct" way to do this would be to create the question definitions based upon the default language. But the issue is that the default language is defined at the _portal environment_ level whereas surveys are at the _portal_ level.

I figured, survey question choices are only displayed in the admin API, so we should just try to match everything else and grab english. If that doesn't exist, grab the raw value. Do you guys think there's a better way to do this?

devonbush: Yeah, this default to English is great for now.  Probably the real fix is to store the full multi-language object, and then in the export options allow specifying the language to do human-friendly exports.  But that's for another day...

ui-core/src/types/forms.ts
```diff
@@ -227,6 +227,11 @@ export type HtmlQuestion = BaseQuestion & {
   html: I18nSurveyElement
 }
 
+export type PanelDynamicQuestion = TitledQuestion & {
```
devonbush: is this used anywhere?

core/src/main/resources/db/changelog/changesets/2024_08_12_add_parent_to_question_def.yaml
```diff
@@ -0,0 +1,19 @@
...
+            columns:
+              - column:
+                  name: parent_stable_id
+                  type: "text"
+                  constraints:
+                    nullable: true
+              - column:
+                  name: repeatable
+                  type: "boolean"
+                  constraints:
```
devonbush, line 17: we would never want this column to be null -- should be nullable: false.

core/src/main/java/bio/terra/pearl/core/service/export/formatters/item/AnswerItemFormatter.java
```diff
@@ -90,9 +111,25 @@ public AnswerItemFormatter(ExportOptions exportOptions, String moduleName, Surve
...
+        this.objectMapper = objectMapper;
     }
 
+    private static String buildBaseKey(String questionStableId, Integer repeatIndex, String parentStableId) {
+        String key = questionStableId;
+        if (Objects.nonNull(repeatIndex)) {
+            key += "[" + repeatIndex + "]";
+        }
+        if (StringUtils.isNotEmpty(parentStableId)) {
+            key = parentStableId + "." + key;
```
devonbush:
```suggestion
            key = parentStableId + ExportFormatUtils.COLUMN_NAME_DELIMITER + key;
```

core/src/main/java/bio/terra/pearl/core/service/export/formatters/item/AnswerItemFormatter.java
```diff
@@ -90,9 +111,25 @@ public AnswerItemFormatter(ExportOptions exportOptions, String moduleName, Surve
...
+        isChildQuestion = questionDef.getParentStableId() != null;
+        parentStableId = questionDef.getParentStableId();
         versionMap = new HashMap<>();
+        this.objectMapper = objectMapper;
     }
 
+    private static String buildBaseKey(String questionStableId, Integer repeatIndex, String parentStableId) {
+        String key = questionStableId;
+        if (Objects.nonNull(repeatIndex)) {
+            key += "[" + repeatIndex + "]";
```
devonbush: create a helper for this in ExportFormatUtils. something like 'formatIndex' that takes an int and returns "[%s]".formatted(int)

core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/SurveyFormatter.java
```diff
@@ -90,6 +170,13 @@ public String getColumnHeader(ItemFormatter itemFormatter, boolean isOtherDescri
             AnswerItemFormatter answerItemFormatter = (AnswerItemFormatter) itemFormatter;
             // for now, strip the prefixes to aid in readability.  Once we have multi-source surveys, we can revisit this.
             String cleanStableId = stripStudyAndSurveyPrefixes(answerItemFormatter.getQuestionStableId());
+            if (Objects.nonNull(answerItemFormatter.getParentStableId())) {
+                cleanStableId = stripStudyAndSurveyPrefixes(answerItemFormatter.getParentStableId()) + "." + cleanStableId;
```
devonbush:
```suggestion
                cleanStableId = stripStudyAndSurveyPrefixes(answerItemFormatter.getParentStableId()) + ExportFormatUtils.COLUMN_NAME_DELIMITER + cleanStableId;
```

core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/SurveyFormatter.java
```diff
@@ -90,6 +170,13 @@ public String getColumnHeader(ItemFormatter itemFormatter, boolean isOtherDescri
...
+            }
+            if (Objects.nonNull(answerItemFormatter.getRepeatIndex())) {
+                cleanStableId += "[" + answerItemFormatter.getRepeatIndex() + "]";
```
devonbush: use the helper method you created above.  In general, try to get a single source of truth for how these names are encoded

devonbush (approved):
Looks good.  I didn't test it locally since my local env is in the midst of Java 21, but the code looks solid.  A few cleanup suggestions

ui-core/src/types/forms.ts
connorlbark: good catch - artifact from experimenting with a different approach

MatthewBemis (approved):
Nice! Worked great locally.

Re: your question `I'm debating refactoring child/parent language - perhaps derived/source? let me know what you guys think.`

I don't feel too strongly either way, I think they both convey what's going on reasonably well. Up to you if you think it's worth the refactor!

PR #1050, opened 2024-08-15
[JN-1235] initial pepper translation script

connorlbark:

Creates new pepper translation script. Needs more work, but I got it to the point that it generally looks OK and imports synthetic participants from AT dev without errors. The format of the data itself is unlikely to actually work that well with SurveyJS, so that's what I'll refine next.

Download:
- [Portal](https://drive.google.com/file/d/1_-Falk3C4tp2rJmqVWPvDIMKNBtVOcik/view?usp=drive_link)
- [Participant data](https://drive.google.com/file/d/1f17uQk33Hu-BgOWymR5mT1C8AnA2-IIm/view?usp=drive_link)

1. Install `openpyxl` into your python3 installation of choice.
2. Extract the participant data & run this script:
   ```
   python3 ~/path/to/juniper/pepper-import/scripts/translate.py -DDj juniper_data_dict.xlsx -DDd pepper_data_dict.xlsx -I 
   pepper_data.tsv -O juniper_data.csv -T translation_overrides.csv --limit 50
   ```
   You'll get some warnings - they can be ignored.
3. Repopulate ATCP with the above portal
4. Delete the admin email configuration (or you'll end up like me...)
    image
    In the future, I might see if I can mute these on import, but it wasn't as easy as just setting `doNotEmail` so I decided to push it off.
6. Upload the generated `juniper_data.csv` file as a new participant import
7. Observe that all the imports succeeded
8. Click into them and make sure that everything looks OK. I didn't try to do much testing that the data is of the exactly correct format, just that it's 90% there. I'll do a follow up PR with minor changes to clean up any formatting that is off with how it works in juniper.


core/build.gradle
```diff
@@ -36,8 +36,8 @@ dependencies {
     implementation 'org.liquibase:liquibase-core:4.21.1'
     implementation 'org.yaml:snakeyaml:2.0'
     implementation 'jakarta.validation:jakarta.validation-api:3.0.2'
-    implementation 'com.fasterxml.jackson.core:jackson-databind:2.14.1'
-    implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.15.2'
+    implementation 'com.fasterxml.jackson.core:jackson-databind:2.17.2'
+    implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.17.2'
```
connorlbark, line 40: This seemed to be broken in my local config until I updated it 🤷

devonbush: I'm guessing one of these needed to be updated for security purposes, and then you were probably seeing weird behavior since they were out of sync

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
```diff
@@ -346,7 +346,7 @@ protected SurveyResponse importSurveyResponse(UUID portalId, SurveyFormatter for
 
             List<ParticipantTask> tasks = surveyTaskDispatcher.assign(assignDto, studyEnv.getId(),
                     new ResponsibleEntity(DataAuditInfo.systemProcessName(getClass(), "handleSurveyPublished.assignToExistingEnrollees")));
-            relatedTask = tasks.get(0);
+            relatedTask = tasks.stream().filter(task -> task.getEnrolleeId().equals(enrollee.getId())).findFirst().orElse(null);
```
devonbush: Instead of doing this, I think you want to change line 343 above from `null` to `List.of(enrollee.getId())` -- that will limit the assignment to the current enrollee

core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/SurveyFormatter.java
```diff
@@ -214,7 +214,13 @@ public String getColumnSubHeader(ItemFormatter itemFormatter, boolean isOtherDes
...
         // if there are >= 3 underscores, filter out everything before the third underscore
         int thirdUnderscoreIndex = StringUtils.ordinalIndexOf(stableId, "_", 3);
-        if (thirdUnderscoreIndex < 0) {
+
+        // warning: hacky!
+        // the `stableId.toUpperCase()` check is a temporary fix to allow AT's questions
+        // to not be affected by the arcane study stripping logic since many of its questions
+        // feature many underscores. no other study has all caps stableIds, so it's safe(ish)
+        // to do.
+        if (thirdUnderscoreIndex < 0 || stableId.toUpperCase().equals(stableId)) {
```
devonbush: Another slightly less hacky way we could do this is changing the logic to check for the regex `^[a-z]{2}_[a-z]{2}_[a-zA-Z]+_` for detecting prefixes to be stripped.  this would catch that the pattern was always, eg, `oh_oh_medHx_someQuestion`. and it wouldn't strip A-T since the first two strings are lowercase

devonbush (approved):

MatthewBemis (approved):

PR #1052, opened 2024-08-17
[JN-1298] Ability to add roles to users via UX

devonbush:

This has the main purpose of allowing users to specify roles when creating new users.  However, in order to get there, we needed to start auditing admin user stuff.  And updating a bunch of admin services/controllers to match our latest patterns.  

We'll need to put some thought into role visibility, and how we let users add/remove roles.  Right now, the main concern is not letting users add roles they don't already have, since we'll be using roles to hide features we don't want to expose to certain customers yet.  But they can still see all the roles since we don't mind (and in fact welcome) questions about them.  Later, we might want to add logic hide roles that users/portals don't have access to.

Setting roles for new users
image

Updating roles of an existing user can be done from either the edit button on the list page, or an "edit roles" button in the detail view.  Note that you can only access the update function from the Portal-specific users list.  The superuser users list (accessed from the "superuser functions" nav list) does not support this function yet since it's operating at a different scope.

image

image

image

1.  redeploy ApiAdminApp
2.  go to  `https://localhost:3000/demo/studies/heartdemo/users`
3.  click the 'edit' icon next to one of the users
4.  add/remove roles and click update
5. confirm roles are saved
6.  logout and log back in as heartdemostaff@gmail.com
7.   go to https://localhost:3000/demo/studies/heartdemo/users
8.  Edit heartdemostaff@gmail.com
9. Try to add the 'publish' role to yourself.  It should error -- you can't add roles you don't have
10. Generally click around the user admin UX and do stuff (add/delete users, etc...) -- this PR changed a lot of the underlying mechanics


devonbush:
This still needs some more tests, but marking as ready for review since it's a long one

ui-admin/src/user/EditUserModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+          <div>
+            <RoleSelector roles={roles} selectedRoleNames={updateParams.roleNames} setSelectedRoleNames={roleNames =>
+              setUpdateParams({ ...updateParams, roleNames })}/>
+          </div>
+        </div>
+      </form>
+    </Modal.Body>
+    <Modal.Footer>
+      <LoadingSpinner isLoading={isLoading}>
+        <button className="btn btn-primary" onClick={updateUser}>Update</button>
```
MatthewBemis: nit: I might use `Save` instead of `Update` here to communicate that it's immediately applied.

our UX is inconsistent right now, but I think we've got two flavors of applying changes:

A) "Save": immediately applies the changes
B) "Update": applies the changes to the component state, requiring a Save later on to actually persist

ui-admin/src/user/EditUserModal.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useState } from 'react'
```
MatthewBemis: nit: rename file to `EditPortalUserModa.tsx` to match component name?

ui-admin/src/user/CreateUserModal.tsx
```diff
@@ -62,11 +79,17 @@ const CreateUserModal = ({ onDismiss, portals, userCreated }:
...
-              Portal
-              <Select options={portalOpts} value={selectedPortalOpt}
-                onChange={opt => setPortalShortcode(opt?.value ?? null)}/>
-            </label>
+            <div>
+              <label className="w-100 mb-3">
+                Portal
+                <Select options={portalOpts} value={selectedPortalOpt}
+                  onChange={opt => setPortalShortcode(opt?.value ?? null)}/>
+              </label>
```
MatthewBemis: Not introduced in this PR, but I think we should hide this `Select` when you're adding a user from within the context of a portal

populate/src/main/java/bio/terra/pearl/populate/service/AdminUserPopulator.java
```diff
@@ -15,6 +16,8 @@
 import java.util.Optional;
 import org.springframework.stereotype.Service;
 
+import javax.xml.crypto.Data;
```
MatthewBemis:
```suggestion
```

looks like an accidental import

core/src/main/java/bio/terra/pearl/core/model/audit/ParticipantDataChange.java
```diff
@@ -18,7 +18,7 @@
 @Setter
 @SuperBuilder
 @NoArgsConstructor
-public class DataChangeRecord extends BaseEntity {
+public class ParticipantDataChange extends BaseEntity implements DataChange {
```
MatthewBemis, line 21: nice rename

MatthewBemis (commented):
Generally looks and works great, just a few suggestions

populate/src/main/java/bio/terra/pearl/populate/service/AdminUserPopulator.java
devonbush: so much for my crytpo scam

ui-admin/src/user/CreateUserModal.tsx
devonbush: I kept the select, but made it disabled (it only has one option in it anyway if you're within a portal context).  That way you still get the visual confirmation you're adding them to a specific portal, but it's clear you don't have a choice

ui-admin/src/user/EditPortalUserModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+    setUpdateParams({
+      username: loadedUser.username,
+      roleNames: loadedUser.portalAdminUsers![0].roles.map(role => role.name)
+    })
+    setRoles(allRoles)
+  })
+
+  const updateUser = async () => {
+    doApiLoad(async () => {
+      await Api.updatePortalUser(portalShortcode, userId, updateParams.roleNames)
```
MatthewBemis, line 42: Since this patch payload includes the entire list of roles that should be persisted, adding/removing one role from another user requires the admin to have _all_ of the final roles.

For example, I can't remove `prototype_access` from a user if they also have `study_publisher` and I'm not also a `study_publisher` myself

We probably want the backend to sort out the diff and only reject the call if they're actually modifying a role that they don't have access to.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/admin/AdminUserExtService.java
```diff
@@ -71,39 +82,126 @@ public AdminUser get(UUID id, String portalShortcode, AdminUser operator) {
...
+  public AdminUser createAdminUser(
+      PortalAuthContext authContext, String username, List<String> roleNames) {
+    authorizeRoles(authContext, roleNames);
+    AdminUser newUser = AdminUser.builder().username(username).superuser(false).build();
+    newUser
+        .getPortalAdminUsers()
+        .add(PortalAdminUser.builder().portalId(authContext.getPortal().getId()).build());
+    DataAuditInfo auditInfo =
+        DataAuditInfo.builder().responsibleAdminUserId(authContext.getOperator().getId()).build();
+    AdminUser adminUser = adminUserService.create(newUser, auditInfo);
```
MatthewBemis: We probably want to check for conflicts in the underlying create here, right now this throws a SQL exception

api-admin/src/main/java/bio/terra/pearl/api/admin/service/admin/AdminUserExtService.java
```diff
@@ -71,39 +82,126 @@ public AdminUser get(UUID id, String portalShortcode, AdminUser operator) {
...
+                    new NotFoundException(
+                        "No admin user with id %s in portal %s"
+                            .formatted(
+                                adminUserId.toString(), authContext.getPortal().getShortcode())));
+    portalAdminUserService.delete(paUser.getId(), auditInfo);
+
+    /** if there are no portal users left for the admin, delete the adminUser too */
+    List<PortalAdminUser> portalAdminUsers = portalAdminUserService.findByAdminUser(adminUserId);
+    if (portalAdminUsers.size() == 0) {
+      adminUserService.delete(adminUserId, auditInfo);
```
MatthewBemis: I think we need to handle FKs that still reference this user, i.e. data change records

MatthewBemis (commented):
Looks good! Just noted a few things that I found during testing

devonbush:
thanks for the great feedback, Matt

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/admin/AdminUserController.java
```diff
@@ -38,48 +44,88 @@ public AdminUserController(
...
+    return ResponseEntity.noContent().build();
+  }
+
+  @Override
+  public ResponseEntity<Void> deleteInPortal(String portalShortcode, UUID adminUserId) {
+    AdminUser operator = authUtilService.requireAdminUser(request);
+    adminUserExtService.deleteInPortal(
+        PortalAuthContext.of(operator, portalShortcode), adminUserId);
+    return ResponseEntity.noContent().build();
+  }
```
connorlbark, line 122: I'm having trouble removing an admin user. It's giving a foreign key constraint error:
```
org.jdbi.v3.core.statement.UnableToExecuteStatementException: org.postgresql.util.PSQLException: ERROR: update or delete on table "admin_user" violates foreign key constraint "fk_portal_env_change_record_admin_user" on table "portal_environment_change_record"
  Detail: Key (id)=(f2513040-5b96-4ade-9d79-00c8e85f0b51) is still referenced from table "portal_environment_change_record". [statement:"delete from admin_user where id = :id;", arguments:{positional:{}, named:{id:f2513040-5b96-4ade-9d79-00c8e85f0b51}, finder:[]}]
```

core/src/testFixtures/java/bio/terra/pearl/core/factory/admin/PortalAdminUserFactory.java
```diff
@@ -36,18 +42,65 @@ public PortalAdminUser.PortalAdminUserBuilder builderWithDependencies(String tes
...
+//        AdminUser adminUser = adminUserFactory.buildPersisted(testName);
+//        List<PortalAdminUser> portalUsers = new ArrayList<>();
+//        DataAuditInfo auditInfo = DataAuditInfo.builder().systemProcess(testName).build();
+//        portalUsers.add(portalAdminUserService.create(PortalAdminUser.builder()
+//                .adminUserId(adminUser.getId())
+//                .portalId(portal.getId())
+//                .build(), auditInfo));
+//        portalAdminUserRoleService.setRoles(portalUsers.get(0).getId(), roleNames, auditInfo);
+//        return new AdminUserBundle(adminUser, portalUsers);
+//    }
```
connorlbark: nit: should delete?

populate/src/main/java/bio/terra/pearl/populate/service/BaseSeedPopulator.java
```diff
@@ -30,7 +30,7 @@ public class BaseSeedPopulator {
                     "adminUsers/connor.json", "adminUsers/egwozdz.json",
                     "adminUsers/mflinn.json", "adminUsers/nwatts.json",
                     "adminUsers/mbemis.json", "adminUsers/cunningh.json",
-                    "adminUsers/andrew.json", "adminUsers/pegah.json", "adminUsers/sampath.json");
```
connorlbark, line 33: 😢 should we delete the other unused admin accounts while we're here?

connorlbark (changes requested):
Works great! Once the deletion foreign key thing is good, then 👍

api-admin/src/main/java/bio/terra/pearl/api/admin/service/admin/AdminUserExtService.java
```diff
@@ -194,8 +194,8 @@ public void delete(OperatorAuthContext authContext, UUID adminUserId) {
   }
 
   /**
-   * removes the user from the specified portal, if the user is in no other portals, also deletes
-   * the user
```
connorlbark: nice! glad we're keep audit history

connorlbark (approved):

MatthewBemis (approved):

PR #1054, opened 2024-08-21
[JN-1287] Grouped navbars

connorlbark:

Also, disentangles html pages and navbar items so that they can exist separately.

Admin side adds a new section to the top of editing the page, much like the footer editor:

Screenshot 2024-08-26 at 2 55 58 PM

Screenshot 2024-08-26 at 2 55 40 PM

Changes to participant side are simple:

image

Screenshot 2024-08-27 at 4 00 26 PM

Screenshot 2024-08-27 at 4 00 29 PM

- Checkout _development_ and restart admin api
- Populate demo / ourhealth / etc
- Checkout this branch and restart admin api / participant api
- Observe that the migration script worked and you can click through all the pages without issues
- Populate demo again
- Go back to participant experience to see the new dropdown item in action
- Go to website editor
- Observe that there's a new navbar section at the top (like the footer)
- Add some new items, remove, delete, etc. and ensure it works as expected
- Ensure that normal website editing functionality has not regressed
- Extract & reimport demo to test extract/populate changes


api-admin/src/main/java/bio/terra/pearl/api/admin/service/siteContent/SiteContentExtService.java
```diff
@@ -76,7 +76,14 @@ public SiteContent create(
...
     siteContent.setStableId(stableId);
-    return siteContentService.createNewVersion(siteContent);
+    SiteContent newSiteContent = siteContentService.createNewVersion(siteContent);
+
+    return loadSiteContent(
+            newSiteContent.getId(),
+            portalEnvironmentService
+                .findOne(portal.getShortcode(), EnvironmentName.sandbox)
+                .orElseThrow(() -> new IllegalStateException("Portal environment not found")))
+        .orElseThrow(() -> new IllegalStateException("Failed to load created site content"));
```
MatthewBemis: I haven't read through all of the code yet, but to speed up my understanding, I'm curious why this extra load is needed now

MatthewBemis (commented):
I want to do another more thorough sweep through the code since there are a lot of changes here, but initial impressions are that the approach is good and it works well. Just a couple of testing notes/thoughts so far:

1) Should the navbar have a JSON editor like the other section types?
2) I can't seem to get the mailing list option to work in the Navbar, I get the error `bootstrap.js?v=bfcc0032:2177 Uncaught TypeError: Cannot read properties of undefined (reading 'backdrop')` (adding a navbar option also breaks the e2e test for mailing lists)

api-admin/src/main/java/bio/terra/pearl/api/admin/service/siteContent/SiteContentExtService.java
connorlbark: Ah, I was able to fix it, so I reverted this back to the same as before.

That being said, since on 'create' it goes through a separate code path for attaching all of the children, it was really easy to get out of date and out of sync with the main `loadSiteContent` code path. I'm not sure maintaining both code paths is ideal, so I'm slightly in favor of doing it like I had done - creating the object, then calling "loadSiteContent". But since both  paths are working now I don't feel strongly.

connorlbark:
> 1. Should the navbar have a JSON editor like the other section types?

I tried this out and it's much much hairier than page sections (lots of child objects, lots of random IDs, etc. that we would have to clean out)  - plus, I don't think it would surface any functionality that wouldn't be possible without the JSON editor, so I wanna lean towards not allowing it as a first step towards a more guided admin experience.

> 2. I can't seem to get the mailing list option to work in the Navbar, I get the error `bootstrap.js?v=bfcc0032:2177 Uncaught TypeError: Cannot read properties of undefined (reading 'backdrop')` (adding a navbar option also breaks the e2e test for mailing lists)

Good catch! This was an existing issue, but it was an easy fix

core/src/main/resources/db/changelog/changesets/2024_08_20_nested_navbars.yaml
```diff
@@ -0,0 +1,26 @@
+databaseChangeLog:
+  - changeSet:
+      id: "nested_navbars"
+      author: connorlbark
+      changes:
+        - addColumn:
+            tableName: navbar_item
+            columns:
+              - column:
+                  name: internal_path
```
devonbush, line 10: nice

core/src/main/java/bio/terra/pearl/core/dao/site/SiteContentDao.java
```diff
@@ -63,18 +64,30 @@ public void attachChildContent(SiteContent siteContent, String language) {
             List<HtmlPage> htmlPages = htmlPageDao.findByLocalSite(localSite.getId());
             List<HtmlSection> htmlSections = htmlSectionDao.findByLocalizedSite(localSite.getId());
             navbarItems.forEach(item -> {
-                item.setHtmlPage(htmlPages.stream().filter(page -> page.getId().equals(item.getHtmlPageId()))
-                        .findFirst().orElse(null));
+                // get the children of this item, if they exist
```
devonbush:
```suggestion
                // attach the children of this item, if they exist
```

devonbush (changes requested):
this is good -- the page paths need to be sanitized.  blocking concern is that I was able to create a page with path "blah blah $#^%@". They could also use some help text on the modal.

I found the navabar editing UX confusing.  I added a new page, and it looked like I was editing the navbar.  As we discussed, let's create an explicit "edit navbar" button

connorlbark:
> I found the navabar editing UX confusing. I added a new page, and it looked like I was editing the navbar. As we discussed, let's create an explicit "edit navbar" button

I deleted the buttons above the section that were mostly UX clutter to make it look more consistent and made the default page have some content - I feel that this drags the user's eyes down to the actual content of the page & hopefully alleviates the confusing UX. Let me know what you think.

image

devonbush (approved):

MatthewBemis (approved):
Looks great, excited to have this!

Also, sorry for the slow turnaround time on this one 🦥

PR #1055, opened 2024-08-27
[JN-1293] Fix empty form bug in SplitFormDesigner

MatthewBemis:

Fixes a bug that prevented the split form designer from being used with an empty form.

Also adds Insert new question / Insert new panel buttons to the 0 index slot, allowing you to insert an initial question.

* Create a new form
* Confirm that you can use the split view designer to add a first question


connorlbark (approved):
sweet!

devonbush (approved):

PR #1057, opened 2024-08-28
[JN-1288] Confirm sensitive changes when publishing

MatthewBemis:

Adds an extra check during publish to prompt the user to double-check portal/study config changes. This should help us avoid accidental password protections, etc

Some unrelated changes: this PR also fixes a bunch of low hanging fruit re: unit test log spam

Screenshot 2024-08-28 at 3 28 01 PM

Modify some study configs for sandbox (password, acceptingEnrollment, etc)
Try and publish them, you should see an extra confirmation modal

### DIFF
--- a/ui-admin/src/portal/publish/PortalEnvDiffView.tsx
+++ b/ui-admin/src/portal/publish/PortalEnvDiffView.tsx
@@ -20,7 +20,7 @@ import { userHasPermission, useUser } from 'user/UserProvider'
 import { Button } from 'components/forms/Button'
 import { isEmpty } from 'lodash'
 import Modal from 'react-bootstrap/Modal'
-import LoadingSpinner from '../../util/LoadingSpinner'
+import LoadingSpinner from 'util/LoadingSpinner'
 
 export const emptyChangeSet: PortalEnvironmentChange = {
   siteContentChange: { changed: false },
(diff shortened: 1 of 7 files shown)


ui-admin/src/portal/publish/PortalEnvDiffView.tsx
```diff
@@ -230,7 +234,14 @@ export default function PortalEnvDiffView(
     </div>
     <div className="d-flex justify-content-center mt-2 pb-5">
       { userHasPermission(user, portal.id, 'publish') && <>
-        <Button variant="primary" onClick={() => applyChanges(selectedChanges)}>
+        <Button variant="primary" onClick={() => {
+          if (isEmpty(selectedChanges.configChanges) &&
```
devonbush, line 239: why do we need this first isEmpty check?

ui-admin/src/portal/publish/PortalEnvDiffView.tsx
```diff
@@ -230,7 +234,14 @@ export default function PortalEnvDiffView(
...
+              isEmpty(selectedChanges.studyEnvChanges.filter(studyChange => !isEmpty(studyChange.configChanges)))) {
```
devonbush: we might want to exclude `enableFamilyLinkage` as a change that triggers the confirmation, since that's a setting that we expect to match across environments

devonbush (approved):
Looks good, a couple questions

ui-admin/src/portal/publish/PortalEnvDiffView.tsx
MatthewBemis: the first isEmpty checks for portal environment config changes, the second checks for any study environment config changes

there are a lot of isEmpty calls here as-written, so I just rewrote it slightly to hopefully be a bit more readable

ui-admin/src/portal/publish/PortalEnvDiffView.tsx
MatthewBemis: would we still want this safeguard in place to prevent people from accidentally toggling it in sandbox and inadvertently propagating to irb or live?

connorlbark (approved):
Nice! I would honestly even vote for showing the modal every time you publish.

PR #1059, opened 2024-08-28
[JN-1213] fix kit request export

connorlbark:

CSV/TSV formats and approachs to escaping seems like a total wild west. Apache Commons CSV uses RFC4180 - which honestly sucks ("""B""" == "B"???)  but it seems like the most portable and common format... 

- export demo
- make sure kit requests are quoted properly (open it in some csv readers)
- do a data import on your export 
- see that kits import as expected (e.g. jonas salk and basic done both have kits)

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/export/TsvExporter.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/export/TsvExporter.java
@@ -1,16 +1,17 @@
 package bio.terra.pearl.core.service.export;
 
+import bio.terra.pearl.core.service.export.formatters.module.ModuleFormatter;
+import com.opencsv.CSVWriterBuilder;
+import com.opencsv.ICSVWriter;
+
 import java.io.IOException;
 import java.io.OutputStream;
-import java.io.PrintWriter;
+import java.io.OutputStreamWriter;
 import java.util.List;
 import java.util.Map;
 
-import bio.terra.pearl.core.service.export.formatters.item.ItemFormatter;
-import bio.terra.pearl.core.service.export.formatters.module.ModuleFormatter;
-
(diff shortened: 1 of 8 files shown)


core/src/main/java/bio/terra/pearl/core/service/export/TsvExporter.java
```diff
@@ -1,16 +1,17 @@
...
+import java.io.OutputStreamWriter;
 import java.util.List;
 import java.util.Map;
 
-import bio.terra.pearl.core.service.export.formatters.item.ItemFormatter;
-import bio.terra.pearl.core.service.export.formatters.module.ModuleFormatter;
-
 public class TsvExporter extends BaseExporter {
-    public static final String DELIMITER = "\t";
+    public static final Character DELIMITER = '\t';
```
MatthewBemis: it doesn't look like this value is used anymore

MatthewBemis (approved):
looks good!

core/src/main/java/bio/terra/pearl/core/service/export/TsvExporter.java
```diff
@@ -21,42 +22,22 @@ public TsvExporter(List<ModuleFormatter> moduleExportInfos, List<Map<String, Str
      * can be supported
      */
     @Override
-    public void export(OutputStream os) {
-        PrintWriter printWriter = new PrintWriter(os);
+    public void export(OutputStream os) throws IOException {
```
devonbush: I don't think this method should throw a checked exception for IOException -- there's nothing the caller can do.  Let's handle the exception here, and turn it into an IOInternalException

devonbush (approved):
Looks good, just change the exception handling

PR #1061, opened 2024-09-03
[JN-1259] Assign kits to participants

MatthewBemis:

* Adds "kit origin" to kit requests: this will be either "ASSIGNED" for in-person kits, or "SHIPPED", for kits that come from DSM.
* Adds a new kit request status for "COLLECTED" and tracks which admin user collected the kit
* Simplifies the participant UX quite a bit. Instead of displaying a bunch of steps to the user, the content shown reflects the current state of the kit and explains the actions that can be taken in that moment
* Adds two modes to the admin kit scanner: "assign kit" and "collect kit"
* Adds auth annotations to KitRequestExtService

Testing Notes

Note: see updated screenshots and testing instructions on https://github.com/broadinstitute/juniper/pull/1091


core/src/main/java/bio/terra/pearl/core/model/kit/KitOriginType.java
```diff
@@ -0,0 +1,6 @@
+package bio.terra.pearl.core.model.kit;
+
+public enum KitOriginType {
+    SHIPPED, //kit was shipped to the participant by GP
```
MatthewBemis: Wondering if "SHIPPED" should just be "DSM" instead to provide more specificity in the future when we may have other kit providers integrated

ui-admin/src/study/kits/kitcollection/KitScanner.tsx
```diff
@@ -0,0 +1,299 @@
...
+  const assignKit = async () => {
+    if (!enrollee || !kitBarcode) { return }
+    doApiLoad(async () => {
+      await Api.createKitRequest(
+        studyEnvContext.portal.shortcode,
+        studyEnvContext.study.shortcode,
+        studyEnvContext.currentEnv.environmentName,
+          enrollee!.shortcode,
+          {
+            kitType: 'SALIVA',
```
MatthewBemis: I'm hoping there's actually a way to determine the kit type from the barcode. I'd rather not have the admin select it here to prevent accidental mismatch because the admin isn't the source of truth for this stuff

core/src/main/java/bio/terra/pearl/core/model/kit/KitOriginType.java
```diff
@@ -0,0 +1,6 @@
...
+    ASSIGNED //kit was assigned to the participant in person by study staff
```
devonbush: IN_PERSON

core/src/main/java/bio/terra/pearl/core/model/kit/KitOriginType.java
devonbush: "MAILED"?

core/src/main/java/bio/terra/pearl/core/model/kit/KitRequestStatus.java
```diff
@@ -17,6 +17,7 @@ public enum KitRequestStatus {
     CREATED,
     QUEUED,
     SENT,
+    COLLECTED, // for a kit that has been handed back to study staff in person. staff will mail it back to GP
```
devonbush: COLLECTED_BY_STAFF?

core/src/main/java/bio/terra/pearl/core/model/kit/KitOriginType.java
connorlbark: It feels like "MAILED"/"IN_PERSON" is separate from, perhaps, `KitProvider` which could be `GP` (since we're moving away from DSM) `External` etc.

core/src/main/java/bio/terra/pearl/core/model/kit/KitRequestStatus.java
connorlbark: If we're keeping track of the fact that it's sent via mail vs in person elsewhere, what is the distinction between this and `RECEIVED`? I would personally lean towards only have one option that denotes "We got it back from the participant" so that future if/switch statements are less complex/error prone, but I could go either way.

ui-participant/src/hub/kit/KitInstructions.tsx
```diff
@@ -1,162 +1,146 @@
...
-                    <Link to='/hub'> participant dashboard.</Link>
-                  </div>
-                  <div className="d-flex justify-content-center">
-                    <Link to={'/hub'} className="btn rounded-pill ps-4 pe-4 fw-bold btn-primary">
-                      Return to Dashboard
-                    </Link>
-                  </div>
-                </KitCollectionStepWrapper>
-              }
+  if (!enrollee.consented) {
```
connorlbark: Minor nit: I feel like this component might be a bit more readable if each of these branches was its own component

core/src/main/java/bio/terra/pearl/core/model/kit/KitRequestStatus.java
MatthewBemis: `RECEIVED` currently means "received by GP"

`COLLECTED` represents a new intermediate step, eventually a kit in this state would progress to `RECEIVED` once it's sent back to GP by study staff

the current statuses are a bit confusing especially now that the kit lifecycle is becoming more complex, I wonder if it would be worthwhile to migrate/refactor the existing statuses to make them more descriptive

some additional comments on here would be useful too

core/src/main/resources/db/changelog/changesets/2024_08_28_in_person_kits.yaml
```diff
@@ -0,0 +1,19 @@
+databaseChangeLog:
+  - changeSet:
+      id: "in_person_kits"
+      author: mbemis
+      changes:
+        - addColumn:
+            tableName: kit_request
+            columns:
+              - column: { name: kit_origin_type, type: text, defaultValue: 'SHIPPED', constraints: { nullable: false } }
```
devonbush: maybe `distribution_method`?  origin type sounds more like it would be the provider of the kit (e.g GP or some 3rd party)

devonbush (approved):
code looks solid -- one naming suggestion

devonbush (approved):
I'll make one final appeal for changing the column name to `distribution_method`, but otherwise looks good

ui-admin/src/study/kits/kitcollection/KitScanner.tsx
```diff
@@ -0,0 +1,303 @@
...
+          setEnrolleeCodeError(undefined)
+          loadEnrollee(enrolleeShortcodeOverride!)
+        }}>
+          <FontAwesomeIcon icon={faSearch}/>
+        </Button>
+      </div> }
+      { enrolleeCodeError &&
+        <div className="text-danger">{enrolleeCodeError}</div>
+      }
+    </KitCollectionStepWrapper>
```
connorlbark, line 223: There's a component called `EnrolleeSearchbar` that uses the search api for doing this (with preview dropdown results, ability to search by name, etc.). It might be worth using here, but this works great too.

api-admin/src/main/java/bio/terra/pearl/api/admin/service/kit/KitExtService.java
```diff
@@ -81,41 +55,74 @@ public void addException(Exception exception) {
...
+  public Collection<KitRequestDto> getKitRequests(PortalEnrolleeAuthContext authContext) {
+    return kitRequestService.findByEnrollee(authContext.getEnrollee());
   }
 
-  public void refreshKitStatuses(
-      AdminUser adminUser,
-      String portalShortcode,
-      String studyShortcode,
-      EnvironmentName environmentName)
+  @EnforcePortalStudyEnvPermission(permission = "BASE")
```
connorlbark, line 124: I don't think for this PR, but should we create a new perm for this? `kit_request_view` and `kit_request_edit`?

ui-admin/src/study/kits/kitcollection/KitScanner.tsx
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  //TODO: JN-1295, while under development, we'll only allow superusers to access this page
+  if (!user?.superuser) {
+    return <div className="m-2">
+      <div className="alert alert-danger" role="alert">
+          You do not have permission to access this page.
+      </div>
+    </div>
+  }
```
connorlbark, line 141: There's now a `<RequireUserPermission superuser noPermsMessage={<></>}>` component you could use for this

ui-admin/src/study/kits/kitcollection/KitScanner.tsx
```diff
@@ -0,0 +1,303 @@
...
+  return <div className='m-2' style={{ maxWidth: '450px' }}>
+    {renderPageHeader('Scan a kit')}
+
+    <div className={'text-muted mb-1'}>
+      To assign or collect an in-person kit, follow the steps below. Please ensure
+      that all information is correct before submitting.
+    </div>
+
+    { user.superuser && <Checkbox
+      label={'Enable manual override mode'}
```
connorlbark, line 152: I think a tooltip here would be useful just to explain what this is

connorlbark (approved):
very clean experience! I love it

PR #1064, opened 2024-09-04
[JN-1310] Fix participant profile page for non-US users

connorlbark:

Addresses a hearthive prod error that was triggered by someone editing their non-US profile from their profile page.

- Redeploy participant api
- Go to profile and attempt to put in `415 IMPROVABLE St` in the US
- Observe that it suggests a new address
- Change the addrress to non-US
- Observe that it saves no issues (no validation is ran)

### DIFF
--- a/ui-participant/src/participant/EditParticipantProfileModals.tsx
+++ b/ui-participant/src/participant/EditParticipantProfileModals.tsx
@@ -14,7 +14,6 @@ import {
 import ThemedModal from '../components/ThemedModal'
 import Modal from 'react-bootstrap/Modal'
 import { isNil } from 'lodash'
-import { ApiErrorResponse } from '@juniper/ui-admin/src/api/api-utils'
 
 
 // skeleton for all profile edit modals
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):

ui-participant/src/participant/EditParticipantProfileModals.tsx
```diff
@@ -359,6 +361,24 @@ export function EditMailingAddressModal(props: EditModalProps) {
     }
   }
 
+  const tryValidateAndSave = async () => {
+    // only validate US addresses
+    if (mailingAddress.country === 'US') {
+      try {
+        await validateAndSave()
+      } catch (e) {
+        await Api.log({
```
devonbush: no reason to `await` this call -- generally it's a bad idea to `await` logging

devonbush: also, would this `validateAndSave()` ever fail for client-side reasons?  I'm wondering if we need this extra log at all, or whether the server-side logging is sufficient

devonbush (approved):

ui-participant/src/participant/EditParticipantProfileModals.tsx
connorlbark: You're probably right - I can't quite remember why I did this Api.log so I'll remove it

PR #1066, opened 2024-09-04
[JN-1281] Switch email update retry to exponential backoff

MatthewBemis:

This updates our existing query retry logic to an exponential backoff that occurs after 4, 8, 16, and 32 seconds. Previously it was a fixed backoff that would retry every 2 seconds, a maximum of 3 times

Confirm that emails are still sent

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
@@ -19,7 +19,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.text.StringSubstitutor;
 import org.springframework.retry.backoff.ExponentialBackOffPolicy;
-import org.springframework.retry.backoff.FixedBackOffPolicy;
 import org.springframework.retry.support.RetryTemplate;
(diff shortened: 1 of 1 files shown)


core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
```diff
@@ -80,8 +80,9 @@ public void processNotification(Notification notification, Trigger config, Enrol
             // the notification might have been saved, but in a transaction not-yet completed (if, for example,
             // study enrollment transaction is taking a long time). So retry the update if it fails
             RetryTemplate retryTemplate = RetryTemplate.defaultInstance();
-            FixedBackOffPolicy backOffPolicy = new FixedBackOffPolicy();
-            backOffPolicy.setBackOffPeriod(2000);  // this will retry once every two seconds for 3 tries
+            ExponentialBackOffPolicy backOffPolicy = new ExponentialBackOffPolicy();
+            backOffPolicy.setInitialInterval(2000);
```
devonbush: this could use a comment explaining what will happen.  it will retry after intervals of 2, 4, 8, 16, and 32 seconds?  I might also make the initial interval longer -- maybe 5 seconds (or 8 if we want powers of 2)

devonbush (approved):
thanks for investigating this!

connorlbark (approved):

PR #1068, opened 2024-09-05
[JN-1315] Fix "No matching state found in storage" for double login

MatthewBemis:

When a user initiates the login flow, our oauth library adds a `state` value to local storage. This value must match the `state` query parameter on the b2c login page. Once the login is complete, the value is removed from local storage.

If the user hits the back button, they'll return to the login page with the same `state` query param as their original login, but the value doesn't exist in local storage anymore, so when the login flow completes, the "No matching state found in storage" error is thrown.

Now, we'll honor the pre-existing login session, which was already the case, but we just won't redirect them to the error page and log as `ERROR`

On demo, or locally running `development` branch, confirm the original issue:

Log in as an existing participant using b2c
Hit the back button from the dashboard
Log in again
Confirm you get the no matching state found in storage error

Confirm the fix:

While running this branch, repeat the above steps and confirm that you're redirected to the hub (since you're already logged in), and confirm that we logged the error as INFO: https://localhost:3000/logEvents

### DIFF
--- a/ui-participant/src/login/RedirectFromOAuth.tsx
+++ b/ui-participant/src/login/RedirectFromOAuth.tsx
@@ -14,7 +14,7 @@ import {
 import { enrollCurrentUserInStudy } from 'util/enrolleeUtils'
 import { PageLoadingIndicator } from 'util/LoadingSpinner'
 import { filterUnjoinableStudies } from 'Navbar'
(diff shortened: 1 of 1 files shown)


devonbush (approved):
great investigation!

connorlbark (approved):
niiiiiiiice

PR #1070, opened 2024-09-05
[JN-1313] Add participant user shortcodes

MatthewBemis:

Adds participant user shortcodes to allow enrollees to be tied together across studies w/o using PII

* Check out `development` and repopulate OurHealth
* Restart AdminApiApp and ParticipantApiApp with this branch checked out 
* Log in as an admin and verify that account.shortcode is now visible, but it's blank for everyone: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/export/dataBrowser
* Enroll a new user in OurHealth sandbox
* Confirm that new user has an account.shortcode generated for them, but everyone else is still blank: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/export/dataBrowser
* Run the `CREATE_PARTICIPANT_SHORTCODES` command: https://localhost:3000/populate/runCommand and confirm you get back a bunch of UUIDs (one for each participant user that was migrated)
* Go back to data export preview and verify that account.shortcode is now populated for everyone: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/export/dataBrowser

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
@@ -284,7 +284,15 @@ public abstract class BaseJdbiDao<T extends BaseEntity> {
     }
 
 
-    protected List<T> findAllByProperty(String columnName, Object columnValue) {
+    public List<T> findAllByProperty(String columnName, Object columnValue) {
+        if(columnValue == null) {
+            return jdbi.withHandle(handle ->
+                    handle.createQuery("select * from " + tableName + " where " + columnName + " is null;")
+                            .mapTo(clazz)
+                            .list()
+            );
+        }
+
         return jdbi.withHandle(handle ->
                 handle.createQuery("select * from " + tableName + " where " + columnName + " = :columnValue;")
                         .bind("columnValue", columnValue)
--- a/core/src/main/java/bio/terra/pearl/core/dao/participant/ParticipantUserDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/participant/ParticipantUserDao.java
@@ -3,6 +3,8 @@ package bio.terra.pearl.core.dao.participant;
 import bio.terra.pearl.core.dao.BaseMutableJdbiDao;
 import bio.terra.pearl.core.model.EnvironmentName;
 import bio.terra.pearl.core.model.participant.ParticipantUser;
+
+import java.util.List;
 import java.util.Optional;
 import org.jdbi.v3.core.Jdbi;
 import org.springframework.stereotype.Component;
@@ -18,6 +20,16 @@ public class ParticipantUserDao extends BaseMutableJdbiDao<ParticipantUser> {
 
    public Optional<ParticipantUser> findOne(String username, EnvironmentName environmentName) {
       return findByTwoProperties("username", username, "environment_name", environmentName);
+   }
+
+   public Optional<ParticipantUser> findOneByShortcode(String shortcode) {
+      return findByProperty("shortcode", shortcode);
+   }
+
+   public List<ParticipantUser> findAllWithMissingShortcode() {
+      return jdbi.withHandle(handle -> handle.createQuery("SELECT * FROM participant_user WHERE shortcode IS NULL")
+            .mapTo(ParticipantUser.class)
+            .list());
    }
 
    public Optional<ParticipantUser> findByToken(String token) {
--- a/core/src/main/java/bio/terra/pearl/core/dao/participant/ParticipantUserDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/participant/ParticipantUserDao.java
@@ -26,12 +26,6 @@ public class ParticipantUserDao extends BaseMutableJdbiDao<ParticipantUser> {
       return findByProperty("shortcode", shortcode);
    }
 
-   public List<ParticipantUser> findAllWithMissingShortcode() {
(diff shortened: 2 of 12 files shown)


core/src/main/java/bio/terra/pearl/core/service/participant/ParticipantUserService.java
```diff
@@ -44,4 +59,21 @@ public Optional<ParticipantUser> findOne(String username, EnvironmentName enviro
...
+
+    // creates an "ACC"-prefixed shortcode for each participant user that's missing one
+    public List<UUID> createMissingShortcodes() {
+        List<ParticipantUser> users = dao.findAllWithMissingShortcode();
+
+        users.forEach(participantUser -> {
+            String shortcode = shortcodeService.generateShortcode("ACC", dao::findOneByShortcode);
+            participantUser.setShortcode(shortcode);
+            dao.update(participantUser);
+        });
```
MatthewBemis, line 76: this will be a bunch of queries...but also this is one-off throwaway code. happy to switch to a batch query though that might be a bit trickier with the generateShortcode stuff

core/src/main/java/bio/terra/pearl/core/dao/participant/ParticipantUserDao.java
```diff
@@ -20,6 +22,16 @@ public Optional<ParticipantUser> findOne(String username, EnvironmentName enviro
       return findByTwoProperties("username", username, "environment_name", environmentName);
    }
 
+   public Optional<ParticipantUser> findOneByShortcode(String shortcode) {
+      return findByProperty("shortcode", shortcode);
+   }
+
+   public List<ParticipantUser> findAllWithMissingShortcode() {
```
devonbush: Can you just update `BaseJdbiDao.findAllByProperty` so that it checks if `columnValue` is null, and if so, uses a different query string?  (and findByProperty too while you're at it) -- those methods not behaving as expected for `null` has been a gotcha in our codebase for awhile

core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/ParticipantUserFormatter.java
```diff
@@ -11,7 +11,7 @@
 
 public class ParticipantUserFormatter extends BeanModuleFormatter<ParticipantUser> {
     public static final String PARTICIPANT_USER_MODULE_NAME = "account";
-    public static final List<String> INCLUDED_PROPERTIES = List.of("username", "createdAt");
+    public static final List<String> INCLUDED_PROPERTIES = List.of("username", "createdAt", "shortcode");
```
devonbush: let's list shortcode first

core/src/main/java/bio/terra/pearl/core/service/participant/ParticipantUserService.java
devonbush: yeah, it should be in the low thousands, which as you say, isn't great, but fine for a one-time thing (and probably a few thousand queries fewer than happen during data export 😛 ).  I would just add a comment that we expect to delete this method in a follow-up PR where we add the non-nullable constraint to shortcode

devonbush (approved):
Looks good -- I'm pleased at how little code you needed to get this working.  One suggestion about updating BaseJdbiDao that's been on my hit list for a while

core/src/main/java/bio/terra/pearl/core/service/participant/ParticipantUserService.java
MatthewBemis: Good idea, just filed https://broadworkbench.atlassian.net/browse/JN-1318 and referenced it in the code comment

core/src/main/java/bio/terra/pearl/core/dao/participant/ParticipantUserDao.java
MatthewBemis: That's a great idea, done in https://github.com/broadinstitute/juniper/pull/1070/commits/0364abb5a314b84dc8e41c7b4c8d59524623fe8e

core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
```diff
@@ -284,7 +284,15 @@ public Optional<T> findByProperty(String columnName, Object columnValue) {
...
+        if(columnValue == null) {
+            return jdbi.withHandle(handle ->
+                    handle.createQuery("select * from " + tableName + " where " + columnName + " is null;")
+                            .mapTo(clazz)
+                            .list()
+            );
+        }
+
         return jdbi.withHandle(handle ->
                 handle.createQuery("select * from " + tableName + " where " + columnName + " = :columnValue;")
```
devonbush, line 297: a bit cleaner...
```suggestion
        String queryString;
        if(columnValue == null) {
            queryString = "select * from %s where %s is null;".formatted(tableName, columnName);
        } else {
            queryString = "select * from %s where %s = :columnValue;".formatted(tableName, columnName);
        }
        return jdbi.withHandle(handle ->
                handle.createQuery(queryString)
```

devonbush (approved):
nice -- one code tidying suggestion

core/src/main/java/bio/terra/pearl/core/dao/BaseJdbiDao.java
MatthewBemis: I like this change from a code simplicity perspective, but jdbi throws an exception due to the fact that the bind parameter is unused in the null case, so I'm going to revert it

connorlbark (approved):
nice, i can do a followup PR to make it searchable as well once this is merged

PR #1074, opened 2024-09-09
[JN-1330] Organize site settings

connorlbark:

Organizes site settings into distinct pages

image

- _Should_ be changes to visuals only, so check for regressions in settings functionality


MatthewBemis (approved):
This is awesome! And I love the fact that everything is url routable

ui-admin/src/study/settings/study/KitSettings.tsx
```diff
@@ -0,0 +1,135 @@
...
+    config: StudyEnvironmentConfig,
+    updateConfig: (key: keyof StudyEnvironmentConfig, value: unknown) => void
+  }) => {
+  const { user } = useUser()
+
+  const studyEnvParams: StudyEnvParams = {
+    portalShortcode: studyEnvContext.portal.shortcode,
+    studyShortcode: studyEnvContext.study.shortcode,
+    envName: studyEnvContext.currentEnv.environmentName
+  }
```
devonbush:
```suggestion
  const studyEnvParams: StudyEnvParams = paramsFromContext(studyEnvContext)
```

ui-admin/src/study/settings/study/KitSettings.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  const loadConfiguredKitTypes = async () => {
+    const selectedKitTypes = await Api.fetchKitTypes(studyEnvParams)
+    setSelectedKitTypes(selectedKitTypes.map(kt => ({ value: kt.name, label: kt.displayName })))
+  }
+
+  useEffect(() => {
+    loadAllowedKitTypes()
+    loadConfiguredKitTypes()
+  }, [studyEnvContext.currentEnvPath])
```
devonbush: I don't think it's worth the complexity to track separate loading states for allowed and configured kit types--opens the possibility for more edge cases.  Also, it's clearer if the state more-or-less reflects what was fetched from the server, and the rendering methods handle transforming it into options

```suggestion
  const {isLoading} = useLoadingEffect(() => {
     const [selectedKitTypes, allowedKitTypes] = await Promise.all([
       Api.fetchKitTypes(studyEnvParams),
       Api.fetchAllowedKitTypes(studyEnvParams)
       ])
       setKitTypes(selectedKitTypes)
       setAllowedKitTypes(allowedKitTypes)
  }, [studyEnvContext.currentEnvPath])
```

ui-admin/src/study/settings/study/KitSettings.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+
+  const [isLoadingKitTypes, setIsLoadingKitTypes] = useState(false)
+  const [kitTypeOptions, setKitTypeOptions] = useState<{ value: string, label: string }[]>([])
+  const [selectedKitTypes, setSelectedKitTypes] = useState<{ value: string, label: string }[]>([])
```
devonbush: generally it's our practice to have state hooks at the top of the component -- similar to having the class members at the top of classes in Java

ui-admin/src/study/settings/study/KitSettings.tsx
```diff
@@ -0,0 +1,135 @@
...
+        studyShortcode: study.shortcode,
+        envName: studyEnv.environmentName
+      }, selectedKitTypes.map(k => k.value))
+      Store.addNotification(successNotification('Kit types saved'))
+      portalContext.reloadPortal(portalContext.portal!.shortcode)
+    }, {  setIsLoading: setIsLoadingKitTypes })
+  }
+
+  return <div>
+    {user?.superuser &&
```
devonbush: it could be convenient to create a `<RequireUserPermission>` component that takes role or superuser args, and conditionally renders its content.  Then that component could handle the useUser() hook

devonbush (approved):
Great UX -- a couple of code suggestions that are from code that pre-dated this change.

ui-admin/src/study/settings/study/KitSettings.tsx
connorlbark: great idea, re-requested feedback since this was a non-trivial change

ui-admin/src/util/RequireUserPermission.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import {
+  userHasPermission,
+  useUser
+} from 'user/UserProvider'
+import { Portal } from '@juniper/ui-core'
+
```
devonbush, line 7:
```suggestion
/** 
  Example use:
   <RequireUserPermission superuser> things </RequireSuperuserPermission>
   
   <RequireUserPermission permission="participant_data_view">stuff</RequireUserPermission>
*/

```

ui-admin/src/util/RequireUserPermission.tsx
```diff
@@ -0,0 +1,55 @@
...
+type RequireSuperuser = BaseRequireUserPermissionProps & {
+  superuser: true
+}
+
+const isRequireSuperuser = (props: RequireUserPermissionProps): props is RequireSuperuser => {
+  return (props as RequireSuperuser).superuser
+}
+
+type RequirePermissions = BaseRequireUserPermissionProps &  {
+  portal: Portal,
```
devonbush, line 32: how about making the portal argument optional, and using 
`const { portal } = useContext(PortalContext) as PortalContextT` to get the portal by default?

ui-admin/src/util/RequireUserPermission.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  if (hasPermission()) {
+    return <>
+      {props.children}
+    </>
+  }
+
+  return <>
+    {props.noPermissionMessage || 'You do not have permission to view this content'}
```
devonbush: I might default this to returning `null` if the user doesn't have permission.  It seems like that's the more common thing we want

ui-admin/src/study/settings/study/KitSettings.tsx
```diff
@@ -0,0 +1,123 @@
...
+        envName: studyEnv.environmentName
+      }, selectedKitTypes.map(k => k.value))
+      Store.addNotification(successNotification('Kit types saved'))
+      portalContext.reloadPortal(portalContext.portal!.shortcode)
+    }, { setIsLoading: setIsUpdatingKitTypes })
+  }
+
+  const isLoading = isLoadingKitTypes || isUpdatingKitTypes
+
+  return <RequireUserPermission superuser>
```
devonbush, line 74: super slick

ui-admin/src/util/RequireUserPermission.tsx
```diff
@@ -0,0 +1,55 @@
...
+      return user?.superuser
+    }
+    if (isRequirePermissions(props)) {
+      return props.perms.every(perm => userHasPermission(user, props.portal.id, perm))
+    }
+    return false
+  }
+
+  if (hasPermission()) {
+    return <>
```
devonbush: do you need to wrap this in a fragment?  can you just `return props.children`?

devonbush (approved):
nice, that permissions component will come in handy!  some codesmithing suggestions...

ui-admin/src/util/RequireUserPermission.tsx
devonbush: atually, nevermind -- don't do that -- we need to use this component outside of portal contexts sometimes for showing the superuser sidebar.  Although, that might be an argument that RequireSuperuser and RequirePortalPermission should be separate components

ui-admin/src/util/RequireUserPermission.tsx
connorlbark: surprisingly, yes. there might be a type other than `React.ReactNode` that plays nicer, but it technically isn't an element, so it will complain that `string | null | <etc...> is not a valid react element` unless it's mounted in JSX. So the fragment is necessary.

PR #1083, opened 2024-09-11
[JN-1323] terraform for getting admin dev exposed w/ load balancer

connorlbark:


terraform/gcp/cluster.tf
```diff
@@ -0,0 +1,48 @@
+resource "google_compute_network" "juniper_dev_compute_network" {
```
devonbush: what's the scope on these names?  since dev and prod will be in separate GCP folders/projects, can we just call the network juniper_compute_network?

terraform/gcp/cluster.tf
```diff
@@ -0,0 +1,48 @@
...
+  name = "juniper-dev-network"
```
devonbush: same question about scope

terraform/gcp/cluster.tf
```diff
@@ -0,0 +1,48 @@
...
+
+  auto_create_subnetworks  = false
+  enable_ula_internal_ipv6 = true
+}
+
+  resource "google_compute_subnetwork" "juniper_dev_subnetwork" {
```
devonbush: same question about scope

terraform/gcp/cluster.tf
```diff
@@ -0,0 +1,48 @@
...
+    ip_cidr_range = "192.168.0.0/24"
+  }
+
+  secondary_ip_range {
+    range_name    = "pod-ranges"
+    ip_cidr_range = "192.168.1.0/24"
+  }
+}
+
+resource "google_container_cluster" "juniper_dev_cluster" {
```
devonbush: same question about scope/name

terraform/gcp/postgres.tf
```diff
@@ -75,3 +78,46 @@ resource "google_service_account_iam_binding" "workload_identity_binding" {
   ]
 }
 
+
+# create db secrets
```
devonbush:
```suggestion
# create db secrets in google secret manager
```

terraform/gcp/k8s/admin-deployment.yml
```diff
@@ -50,23 +52,24 @@ spec:
...
+          image: us-central1-docker.pkg.dev/broad-ddp-dev/juniper/juniper-admin:1.3.75
           env:
+            - name: SPRING_PROFILES_ACTIVE
+              value: gcp
+            - name: DB_NAME_SECRET_ID
+              value: "d2p-db-name"
+            - name: DB_PASSWORD_SECRET_ID
+              value: "d2p-db-password"
+            - name: DB_USER_SECRET_ID
+              value: "d2p-db-user"
```
devonbush, line 67: it feels like it might be one extra layer of indirection to store the names of the secrets as environment variables.  But I guess it could come in handy for repointing to a different DB

devonbush (approved):
nice, congrats on getting it working!

terraform/gcp/k8s/admin-deployment.yml
connorlbark: i figured if somehow we need to have different names for dev/prod it would be nice, plus I wanted to avoid hardcoding any implementation-specific details like the names of things if possible in our `application.yml`, but happy to go either way with this

MatthewBemis (approved):

PR #1085, opened 2024-09-12
[JN-1335] Open split survey editor for all

MatthewBemis:

Opens up the split survey editor for all users.

Log in as a non-superuser and make sure you can access the split designer


connorlbark (approved):

devonbush (approved):
I'd lean towards putting the classic designer under the "prototype_tester" flag.  That way we can tell our users "there's a new, improved survey designer" and get feedback on it quickly.   Then if someone hates it or find something they can't do, we can re-enable the classic view for them

MatthewBemis:
@devonbush I think that's a great idea, I'll make that change

PR #1088, opened 2024-09-13
[JN-1324] terraform - use variables and recreate fully with remote state backend

connorlbark:

This is a lot of changes (add variables, SSL, juniper admin image being read from separate project), but it is FULLY up to date with what's running in GCP and the state is all stored in a remote backend. It also has the terraform for our eng infra project (which right now is just the artifact registry and terraform remote storage bucket).

This means that any and all future GCP changes have to go through terraform, but any of us will be able to do it and there should be locking so it's safe to do plan/apply/etc from our local machines without recourse.

There are some DNS changes which, becuase I tore the whole thing down and recreated it, will take 1-2 days to repropagate, so our admin app won't be fully running again in GCP until later this week. That'll be a separate PR. I was able to get the application running, downloading the image, logging errors, and connecting to the database - so it's working as much as it was before, just without external access.

Also, this creates a dual stack IPV4 and IPV6 cluster, which was recommended online. Currently, our DNS is just using IPV4, but it absolutely could be using IPV6. I _think_ this is the more futureproof option and generally recommended online, but if it proves to be annoying I'll change it before we create the prod cluster.

Next up is figuring out how we want to manage K8s manifests (helm, kustomize, etc.).

- if you haven't already, download `gcloud` and [get authenticated](https://cloud.google.com/docs/authentication/gcloud#local)
- `cd terraform/gcp`
- run terraform init (it might ask for workspace; use dev)
- switch to the `dev` workspace
- make sure you see dev and prod when you do `terraform workspace list`
- switch to dev: `terraform workspace select dev`
- (note that this means if you touch prod, you must both change workspaces *AND* use the prod.tfvars file)
- run `terraform plan -var-file="envs/dev.tfvars"` and ensure it says nothing should happen
- look at the files & make sure everything looks good (no glaring security issues, everything makes sense)


terraform/gcp/admin_dns.tf
```diff
@@ -0,0 +1,69 @@
...
+  name = "admin-ip"
+}
+
+resource "google_dns_managed_zone" "juniper_admin_dns_zone" {
+  description = "juniper-admin-dns-zone"
+  dns_name    =  "${var.admin_url}."
+  dnssec_config {
+    kind          = "dns#managedZoneDnsSecConfig"
+    non_existence = "nsec3"
+    state         = "off"
```
connorlbark, line 11: this is off because apparently you can't change to a new dns zone with dnssec on. it takes 24 hours to propagate - then I can switch domain to this dns zone - then in 24 MORE hours I can turn this back on. That's when the "fix dns" PR will come back online and we'll be able to go to `https://www.juniper-cmi.dev` once again.

terraform/gcp/db.tf
```diff
@@ -0,0 +1,36 @@
+# Create database
+resource "google_sql_database_instance" "d2p" {
+  name = "d2p"
+  database_version = "POSTGRES_15"
```
devonbush: I think we can go ahead and bump to 16. 
 https://cloud.google.com/blog/products/databases/postgresql-16-now-available-in-cloud-sql

devonbush (approved):
looks good, just one suggestion that we might as well go up to postgres 16

terraform/gcp/db.tf
connorlbark: good catch!

MatthewBemis (approved):

PR #1090, opened 2024-09-14
[JN-1244] Split designer navigation and UX upgrades

MatthewBemis:

* Renames to just "Designer" and marks the old designer with a "legacy" label (legacy designer was hidden by default in https://github.com/broadinstitute/juniper/pull/1085)
* Adds improved site navigation, including a table of contents that navigates you directly to the selected element. First/last page buttons also added to the pagination widget.
* Refined the styling of the editors. Previously didn't spend much time on this, but now we've committed to pushing forward on this so it warranted some cleanup
* Minimized vertical space usage as much as possible. Label padding, favoring info popups instead of descriptions, and moving additional question options (visibility, advanced, choices, input options, etc.) into a tabbed view
* Added ability to create and delete pages
* Added ability to clone existing questions
* Slightly improved panel UX (this area still needs a bit of attention, for a later PR)
* Also included a few minor bug fixes
* Full i18n support

Major remaining feature gaps before we have feature parity with legacy designer:

* Support question templates

Before:
Screenshot 2024-09-13 at 9 34 56 PM

After:

https://github.com/user-attachments/assets/bad193c0-23bc-43e2-9095-a963afa2c568

* Try out the new functionality described above
* Ensure that there are no functional regressions in the legacy designer


ui-admin/src/forms/FormContentEditor.tsx
```diff
@@ -62,15 +62,14 @@ export const FormContentEditor = (props: FormContentEditorProps) => {
...
-        className="mb-1"
         mountOnEnter
         unmountOnExit
         onSelect={setActiveTab}
       >
         {userHasPermission(user, portal.id, 'prototype_tester') &&< Tab
           disabled={activeTab !== 'designer' && !tabsEnabled}
           eventKey="designer"
-          title="Designer"
+          title={<>Designer<span className='badge bg-primary fw-light ms-2'>LEGACY</span></>}
```
devonbush: maybe move this tab all the way over to the right so it's less likely we'll get confused and click the wrong designer

devonbush (changes requested):
Looks great.  Big improvement!

One blocker: Html questions don't quite work -- the html text field doesn't get populated with the value from the form, so you can't edit the existing value, only start typing a new value

One non-blocker:
Focusing on items in the table of contents correctly updates the url, but going to a url doesn't scroll to the item.  e.g. going to `https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms/surveys/hd_hd_basicInfo?selectedElementPath=pages%5B0%5D.elements%5B6%5D` selects the item, but doesn't bring it into view

ui-admin/src/forms/designer/split/SplitFormDesigner.tsx
```diff
@@ -16,81 +22,78 @@ export const SplitFormDesigner = ({ content, onChange, currentLanguage, supporte
...
-            element={content.pages[currentPageNo].elements[elementIndex]}
-            currentLanguage={currentLanguage} supportedLanguages={supportedLanguages}
-            onChange={onChange}/>
-          <div className="d-flex">
-            {renderNewElementButton(content, onChange, elementIndex, currentPageNo, 'question')}
-            {renderNewElementButton(content, onChange, elementIndex, currentPageNo, 'panel')}
+  return <div className="container-fluid overflow-scroll">
+    <div className="row w-100 mx-0">
+      <div style={{ overflowY: 'scroll' }}
+        className={classNames('px-0 border-start border-end bg-white', hideTableOfContents ? 'd-none' : 'col-3')}>
```
connorlbark:
```suggestion
        className={classNames('px-0 bg-white', hideTableOfContents ? 'd-none' : 'col-3')}>
```
UX nit: I think it looks cleaner without borders

image

ui-admin/src/forms/designer/split/SplitFormDesigner.tsx
```diff
@@ -16,81 +22,78 @@ export const SplitFormDesigner = ({ content, onChange, currentLanguage, supporte
...
+        { !hideTableOfContents &&
+            <FormTableOfContents
+              formContent={content}
+              selectedElementPath={selectedElementPath}
+              onSelectElement={setSelectedElementPath}
+            />
+        }
+      </div>
+      <div className={classNames('col', hideTableOfContents ? 'col-12' : 'col-9')}>
+        <div className="d-flex justify-content-between border border-top-0 rounded-bottom-3 p-2 bg-light">
```
connorlbark:
```suggestion
        <div className="d-flex justify-content-between border mt-2 border-0 rounded-3 p-2 bg-light">
```
Another UX nit: Total personal preference, but I think it looks slightly cleaner matching the bottom
image

ui-admin/src/forms/designer/split/SplitFormDesigner.tsx
```diff
@@ -16,81 +22,78 @@ export const SplitFormDesigner = ({ content, onChange, currentLanguage, supporte
...
+  return <div className="container-fluid overflow-scroll">
+    <div className="row w-100 mx-0">
+      <div style={{ overflowY: 'scroll' }}
+        className={classNames('px-0 border-start border-end bg-white', hideTableOfContents ? 'd-none' : 'col-3')}>
+        { !hideTableOfContents &&
+            <FormTableOfContents
+              formContent={content}
+              selectedElementPath={selectedElementPath}
+              onSelectElement={setSelectedElementPath}
+            />
```
connorlbark, line 54: This might be a controversial opinion, but I find it really hard to map what you're seeing on the right to what's on the left. I think this is an awesome V1 and the functionality is great, but I would like to posit an alternative design idea: a `Go To Question` button.

image

This go to question button opens a modal with the table of contents (as-is). If you click on a question in the modal, it takes you straight there and scrolls you to it. I think I would personally find that more intuitive and save a ton of screen real estate. I say this because I exclusively use the table of contents for a "bring me straight to this question" purpose, but I could be missing other uses people have.

connorlbark (changes requested):
I love the new functionality! Because I'm excited about it, I have a bunch of totally optional opinionated UX ideas and one blocking bug.

Switching languages seems to not be working properly. It almost works, but doesn't update previews and doesn't update the existing text in the box

https://github.com/user-attachments/assets/c9d247fd-9cfc-4620-b704-74a25bba6f95

ui-admin/src/forms/designer/split/SplitFormDesigner.tsx
MatthewBemis: That's an interesting idea- let's keep it in mind as we start to roll this out and gather feedback. I agree that the ToC takes up quite a bit of real estate as-is. Hopefully the ability to hide (with that preference remembered in local storage) can help get it out of the way for when it isn't needed.

We could also roll out both options and let them live side by side for a bit, and then track with mixpanel to see what people are doing.

I think there's also some room to improve the current ToC in the future by adding a "search for stable id", with behavior similar to what you're suggesting.

devonbush (approved):

connorlbark (approved):

PR #1093, opened 2024-09-17
JN-1331 include jamf computers in vanta sync

zyme:

We need to keep the list of computers that are in scope for our audit in sync, similar to how we synchronize other vanta integrations.  This PR adds support for computers, assuming that vanta is using the jamf integration.  Associating a computer to use requires hitting a new vanta endpoint (`/people`), grabbing the internal `id` of the person, and then filtering the list of jamf computers to find which computer is owned by which person.

Along the way, a number of other improvements were made after observing some interesting behaviors in prod.  But before we get there, a bit of background...

This job is kicked off as a cloud run function via a cloud scheduler event that sends in a pubsub message that the job responds to.  Because this job takes over 10m to run, the pubsub message is never acked within the maximum pubsub ack timeout window (which is 5m).  As a result, pubsub retries were triggered, causing more instances of the job to run, causing more vanta rate limits pauses, which in turn caused yet more delays in processing and yet more pubsub retries. 

The "fix" for the storm of jobs is to keep track of pubsub ids received in a list, and then have the job consult this list and skip processing if it sees a pubsub message id that it's processed before.  This simple list is a file in GCP storage.  It's not transactional and if you were to run a bunch of jobs at once, the file data would probably be overwritten.  But for a job that runs twice a day and is not mission critical, this risk is acceptable.  Nonetheless, the log file includes some information about the contents of the file so that if someone does need to troubleshoot it, there will be a trail of breadcrumbs to follow.

The bucket to use to store the previously processed pubsub message ids is now part of the config secret.

This storm of jobs exposed another problem with the job: that the API call to change scope was itself not handling rate limits, leading to job failure instead of pause-and-retry.  This has been fixed.

Watch the logs for the job, confirm that there's not a storm of jobs.  Look in vanta to confirm that we don't have a list of a bajillion unmonitored computers.


compliance/src/main/java/bio/terra/pearl/compliance/SyncVantaUsers.java
```diff
@@ -170,17 +241,22 @@ private <T extends VantaObject> Collection<T> collectAllData(WebClient wc, Param
         return elements.get();
     }
 
-    private WebClient getWebClientForIntegration(String accessToken, String integrationId, String resourceKind) {
-        WebClient.Builder webClientBuilder = WebClient.builder().defaultHeaders(h -> {
+    private WebClient.Builder getWebClientForUrl(String accessToken, String url) {
```
zyme, line 244: Factored out a separate method so that the `patch` request can use the same kind of webClient as the `get` requests that list information about computers, git users, etc.

devonbush (approved):
looks good!

MatthewBemis (approved):

PR #1100, opened 2024-09-19
[JN-1346] Clear unsaved changes when switching env, add margins

MatthewBemis:

From team testing this week, tweaks some margins and clears unsaved state when switching environment

Before
Screenshot 2024-09-18 at 10 31 47 PM

After
Screenshot 2024-09-18 at 10 31 33 PM

Go to https://localhost:3000/demo/studies/heartdemo/env/live/settings
Confirm that editing settings and switching envs without saving clears the unsaved state

### DIFF
--- a/ui-admin/src/study/settings/SettingsView.tsx
+++ b/ui-admin/src/study/settings/SettingsView.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import Api, {
   PortalEnvironment,
   PortalEnvironmentConfig,
@@ -75,7 +75,6 @@ export function LoadedSettingsView(
     }, { setIsLoading: setIsLoadingPortalConfig })
   }
 
-
   const updateStudyConfig = (field: keyof StudyEnvironmentConfig, val: unknown) => {
     setStudyConfig(old => {
       return { ...old, [field]: val }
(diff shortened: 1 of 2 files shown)


ui-admin/src/study/settings/SettingsView.tsx
```diff
@@ -93,6 +92,13 @@ export function LoadedSettingsView(
     }, { setIsLoading: setIsLoadingStudyConfig })
   }
 
+  useEffect(() => {
+    setPortalConfig(portalEnv.portalEnvironmentConfig)
+    setHasPortalConfigChanged(false)
+    setStudyConfig(studyEnvContext.currentEnv.studyEnvironmentConfig)
+    setHasStudyConfigChanged(false)
+  }, [studyEnvContext.currentEnv.environmentName])
```
connorlbark: very nice! I think we need to do the same for `kits` and `languages` in their respective settings pages.

connorlbark (approved):

ui-admin/src/study/settings/SettingsView.tsx
devonbush: An easier way to do this is to add a `key` to the place where the component is invoked.  E.g. edit line 127 of StudyEnvironmentRouter to add `key={studyEnvContext.studyEnvPath}`. That tells react to destroy and recreate the whole component if the key changes.  Then you don't have to  worry about manually resetting every bit of state.  Alternatively, if you don't want StudyEnvRouter to have to `know` when to reset the component, you can add a wrapper component around LoadedSettingsView with the key

devonbush (approved):
looks good, one suggestion for easier state cleanup

PR #1102, opened 2024-09-19
[JN-1355] Get admin & participant APIs fully running in GCP

connorlbark:

I believe this is _everything_ for juniper dev, including apache proxy, sendgrid, etc.

- check the terraform is up to date
- sanity check the apache proxy config - I had to do a decent amount of changes and I don't really understand the config that well so I could have done something wrong


terraform/gcp/k8s/site-config.yml
```diff
@@ -0,0 +1,158 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: d2p-site-configmap
+data:
```
devonbush, line 5: Do I remember right that there's only one proxy per that serves both admin and participant?

terraform/gcp/main.tf
```diff
@@ -27,13 +27,13 @@ provider "google" {
 
 data "google_client_config" "provider" {}
```
devonbush, line 29: why is this commented out?

devonbush (approved):
looks good -- I think we're at a point where we should start treating this more like real code that's likely to last awhile.  So please add comments next to anything commented out as to why (or remove it)

terraform/gcp/k8s/site-config.yml
connorlbark: No, each pod has its own proxy. This means the proxies are also load balanced alongside the application. Right now we're not doing any load balancing, though, but if we scaled up the replicas we would have one proxy per admin pod, one proxy per participant.

terraform/gcp/main.tf
connorlbark: Oops - forgot to clean out commented code. Originally, I was creating the kubernetes service account in terraform, but this meant that in the same module terraform was creating a k8s cluster and also connecting to it and creating things on it. This is generally not recommended to do because terraform can drift really easily and get confused. I didn't know this until I started running into issues connecting to the cluster. But since the service account is just a standard k8s manifest, I just moved it to a normal yaml file that we create manually (like all other k8s manifests) instead of creating it here.

terraform/gcp/k8s/demo-cert.yml
```diff
@@ -0,0 +1,9 @@
+apiVersion: networking.gke.io/v1
+kind: ManagedCertificate
+metadata:
+  name: demo-cert
+spec:
+  domains:
+    - sandbox.demo.juniper-cmi.dev
+    - irb.demo.juniper-cmi.dev
+    - demo.juniper-cmi.dev
```
connorlbark, line 9: Moved these to their own cert object because I learned that managed certificates can only be _created_ and _deleted_ (not updated). This means that, if we wanted to add a new domain, there would be about 40 minutes of downtime if we created all of the certificates for juniper-cmi in one file as we would have to delete it then wait for the new one to provision. Thus, I think it's best to create a separate certificats for _each_ portal, that way that can be created/deleted and only effect that portal.

MatthewBemis (approved):

PR #1104, opened 2024-09-20
[JN-1360] fix proxy profile mappings

connorlbark:

Code shuffles broke the underlying logic for proxy profile mappings. This is definitely my fault because the code was really unclear that the `ppUser` was supposed to be the `operator` ppuser. There definitely could be clearer code here. For now, this is just a fix + test to get it working again.

- Extract CMI template portal from dev
- Create a proxy user
- Fill out the consent & make sure the proxy's name gets populated in their profile

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
@@ -174,6 +174,8 @@ public class SurveyParseUtils {
                 } else {
                     if (textNode.has("en")) {
                         text = textNode.get("en").asText();
+                    } else if (textNode.has("default")) {
+                        text = textNode.get("default").asText();
                     } else {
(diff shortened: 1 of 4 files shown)


core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
```diff
@@ -174,6 +174,8 @@ public static String unmarshalSurveyQuestionChoices(JsonNode question) {
                 } else {
                     if (textNode.has("en")) {
                         text = textNode.get("en").asText();
+                    } else if (textNode.has("default")) {
+                        text = textNode.get("default").asText();
                     } else {
```
connorlbark, line 179: drive by fix, cmi has "default" instead of "en" in some choices

devonbush (approved):

MatthewBemis (approved):

PR #1107, opened 2024-09-24
[JN-1365] export integration model + populate

devonbush:

Adds basic models and services for the export integration that will power regular airtable exports. No user-facing functionality yet.

1. redeploy apiAdminApp
2. run `./scripts/populate/populate_portal.sh demo`
3. in sql terminal, run `select * from export_integration`, confirm an integration populated with type AIRTABLE


core/src/main/java/bio/terra/pearl/core/model/export/ExportIntegration.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class ExportIntegration extends BaseEntity {
+    private UUID studyEnvironmentId;
+    private boolean enabled = true;
+    private ExportDestinationType destinationType;
+    private String destinationProps;
```
connorlbark, line 19: I don't see any usages of this; what does this do?

core/src/main/java/bio/terra/pearl/core/service/export/ExportOptionsParsed.java
```diff
@@ -0,0 +1,15 @@
...
+
+import bio.terra.pearl.core.service.search.EnrolleeSearchExpression;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@Setter @SuperBuilder
+public class ExportOptionsParsed extends ExportOptions {
+    private final EnrolleeSearchExpression searchExpression;
```
connorlbark, line 11: is there any reason not to add this to `ExportOptions` and just set the expression null before serializing? IMO I think that would be clearer. or maybe non-serialized export options could be `ExportOptionsUIDto` and this is the normal export options. the `Parsed` would throw me off, I think. also, I think maybe this should be called `filterExpression`  so it's clearer it's related to the `filter` string.

core/src/main/java/bio/terra/pearl/core/model/export/ExportIntegration.java
```diff
@@ -0,0 +1,21 @@
...
+    private String exportOptions;
```
connorlbark, line 20: would we want to use export options in other places later? I almost kinda feel like it might be worth making a full db object, and we could do other fun stuff (save user's preferred export options? or, users can create export presets for different aspects of their workflow, e.g. "Kit-Only Export", "Proxy Export" etc. so they can optimize their workflwos). The simplicity is nice, but it's got me thinking all the cool stuff we can do 😄 .

connorlbark (approved):
worked great! just some code ideas 😄

core/src/main/java/bio/terra/pearl/core/model/export/ExportIntegration.java
devonbush: nothing yet -- those are the props that will eventually tell it where to import to AirTable--e.g. which project and subtable to import to

core/src/main/java/bio/terra/pearl/core/model/export/ExportIntegration.java
devonbush: yeah, good ideas. I'll keep it as a string for now, and then we can make it a full table once the export options crystallize a bit more (we've still got to add support for including/excluding individual survey questions)

core/src/main/java/bio/terra/pearl/core/service/export/ExportOptionsParsed.java
devonbush: if we add it to ExportOptions, Jackson will still throw a fit about not knowing how to serialize it, even if it's null.  
and I like having the type indicate whether or not you're getting the parsed expression, because it makes function signatures clearer.   Good points on filter rename, I'll do that in the follow-on PR

devonbush: And as far as DTO -- our convention is that model names without suffixes are what's stored in the DB, and then suffixes ('parsed', 'dto') indicate classes with other purposes.   So ExportOptionsParsed tells you that that's *not* the version stored in the DB, while ExportOptions is.  That said, I'm 100% open to better suffixes than "Parsed".  Maybe "WithExpression"?

connorlbark: I think I like "WithExpression" more than "Parsed", I could see making a custom deserializer/serializer for expressions because they can be converted into and from strings really easily, but def outside of the scope of this PR.

MatthewBemis (approved):

devonbush:
Thanks for the reviews!  I'm going to hold off on merging this -- to avoid unnecessary double-migration scripts, this will get folded into a later PR with the export_options given its own table

PR #1109, opened 2024-09-24
[JN-1368] Make GKE cluster private

connorlbark:

GKE cluster & db are now on a fully private network.

The only endpoints that are exposed are the Ingress for the application and the kubernetes endpoint (internal node communication and db communication happens locally within the private network). The kubernetes endpoint is only accessible from broad IPs. This means you have to be in office or using the VPN to use `kubectl`, like in azure, so it should be quite secure.

~Currently, I don't think there's a way to directly connect to the database. I suspect we'll need something like we have for azure (a script that spins up a VM and connects to it via SSH so we can do things on the private network).~ Apparently you can do this in the UI!

- make sure we conform to [DSP's Secure GKE instance guide](https://dsp-security.broadinstitute.org/cloud-security/google-cloud-platform/gke) (I believe we do, most of these settings are default applied).
   - we still have the web dashboard enabled, but I want to keep that until we're closer to being finished because it makes my life so much easier for debugging 
- do `gcloud container clusters get-credentials juniper-cluster --region=us-central1 --project=broad-juniper-dev`
- make sure `kubectl config get-contexts` is connected to the broad-juniper-dev instance
- run `kubectl get ns` on your home network (or a hotspot or something) and make sure it fails
- run `kubectl get ns` in office or with the VPN and make sure it works


terraform/gcp/cluster.tf
```diff
@@ -32,6 +12,26 @@ resource "google_container_cluster" "juniper_cluster" {
     stack_type                    = "IPV4_IPV6"
   }
 
+  cluster_autoscaling {
```
devonbush, line 15: how does this autoscaling property interface with autopilot?

devonbush (approved):
Nice, one question.  Also, I believe we can still connect to postgres directly via cloud sql studio https://console.cloud.google.com/sql/instances/d2p/studio?authuser=0&project=broad-juniper-dev.  it's not quite as friendly as command-line, but will work just fine for support in the short-medium term

terraform/gcp/cluster.tf
connorlbark: The terraform docs / naming conventions are really bad here. [Here](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#auto_provisioning_defaults) you can see "A subset of fields also apply to GKE Autopilot clusters". Normally, you'd use `node_config` to change the service account, but unfortunately, that does not work with autopilot, so you have to use this arcane nested structure instead.

You can see it's still autopilot:
image
And it's using a custom service account:
image

connorlbark:
> Nice, one question. Also, I believe we can still connect to postgres directly via cloud sql studio https://console.cloud.google.com/sql/instances/d2p/studio?authuser=0&project=broad-juniper-dev. it's not quite as friendly as command-line, but will work just fine for support in the short-medium term

wow! i had no idea this worked with private networks. I guess it makes sense because google service networking is peered with our private network.

MatthewBemis (approved):
Nice! I confirmed that I was only able to access it on the Broad network 🎉

PR #1113, opened 2024-09-27
[JN-1383] [member's choice] json syntax highlighting

connorlbark:

Adds fancy syntax highlighting/code editor to the survey json editor. 

I evaluated a few options. The "simple" approach is to render a textbox and overlap a `<pre>` with syntax highlighting. But with json files >6000 lines long, like we have, this approach took about 1.5 seconds to render any changes in the textbox.

CodeMirror, on the other hand, despite being a relatively heavyweight solution, is much more cleverly optimized and is able to work with very large json files just fine. The downside being that we're no longer working with a regular textbox, but rather a complex series of divs and an emulated cursor.

Because of this, it doesn't seem easy to test and the current tests are failing. I decided to punt this off to work on other member's choice projects. In my admittedly quick evaulation, I think it might be difficult to get the tests working with the new editor - open to ideas here (putting a good amount of effort into the tests or significantly simplifying them).

image

Also, I ran into weird component refresh issues using it in other places in the app. I figured here was the most important place, and if we like it, we can put it elsewhere in the future. We could also use the markdown module in the website editor, as well.

- `npm install`
- try it out - make sure it works with no weirdness


ui-admin/src/util/json/JsonEditor.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import CodeMirror from '@uiw/react-codemirror'
+import { json } from '@codemirror/lang-json'
+import { githubLight } from '@uiw/codemirror-theme-github'
```
connorlbark, line 4: there are [other themes](https://uiwjs.github.io/react-codemirror/#/theme/home) if anyone has any preferences

devonbush (approved):
agree that this seems like the right approach -- might as well have a full-fledged editor given how deeply we work with json in the admin tool.  (And as long as we can lazy-load it as you've done)

ui-admin/src/util/json/JsonEditor.tsx
MatthewBemis: Any of the light ones look good imo- so what you've chosen works for me!

MatthewBemis (approved):
🎉 woohoo! works great

PR #1114, opened 2024-09-27
[JN-1318] Participant kits page [member's choice]

MatthewBemis:

Adds a new kits page to the participant app. Ran the UX by Whitney and Parker and both gave it a 👍 

This update was motivated by a few things:

1) The existing participant entrypoint into the in-person kits page felt very clunky. They would select "Sample Kits" from their account dropdown and then it would drop them right into the In-person kit instructions page. It felt like it should drop them into something a little more general, where they can get all of the context re: kits

2) Study staff will be able to enable/disable in-person kits at the study-level (not implemented yet- but will live in the same place where they can configure the kit types). When not enabled, we'll hide the text/button for in-person kits. It's going to be irrelevant to studies the _vast_ majority of the time, so there's no reason to keep it visible

3) There was no way for a participant to view information about their kits after they mailed them in. Most CMI studies will have multiple kits sent to users (blood and saliva), and certain studies will have recurring kits every X months. Having the full history visible to the participant felt valuable

screencapture-sandbox-ourhealth-localhost-3001-hub-kits-2024-09-27-18_20_04

screencapture-sandbox-ourhealth-localhost-3001-hub-kits-2024-09-28-19_03_56

Mobile
screencapture-sandbox-ourhealth-localhost-3001-hub-kits-2024-09-28-10_07_37


ui-core/src/participant/ParticipantNavbar.tsx
```diff
@@ -398,6 +398,11 @@ export const AccountOptionsDropdown = (
               {i18n('manageProfiles')}
             </button>
           </NavLink>}
+          <NavLink to="/hub/kits">
+            <button className="dropdown-item" aria-label="sample kits">
+              Sample Kits
+            </button>
+          </NavLink>
```
connorlbark: Should this be hidden behind study configuration settings? Right now we don't have a "turn kits on and off" but maybe we should introduce that setting alongside the in person kits setting.

ui-participant/src/hub/kit/KitsPage.tsx
```diff
@@ -0,0 +1,155 @@
...
+    <div className="pb-4">
+      Sample collection kits are a valuable part of the study process and can help provide researchers with
+      important insights. Below you will find the status of all kits that have been provided to you.
+    </div>
+    { isInPersonKitEnabled &&
+      <>
+        <h3>Provide a sample in-person</h3>
+        <div className="pb-2">
+          This study is currently offering the option to complete a sample collection kit in-person.
+          Click below for more information about this process.
```
connorlbark: Is the `Click below` supposed to direct them to click `Complete a kit in-person`? I was a bit confused because I thought I should be looking for a link to a page with documentation. Maybe `Click below if you are attending an in-person event with <study name>` ? (or something like that)

connorlbark (approved):
Very nice! This is cleaning up to be an awesome new addition to the participant experience

ui-core/src/participant/ParticipantNavbar.tsx
devonbush: concur -- this could confuse HeartHive folks if they see a "no sample kits" message, even though they have done a kit just not through Juniper

devonbush (changes requested):
Looks great, just need to hide this from HeartHive :)

ui-core/src/participant/ParticipantNavbar.tsx
MatthewBemis: That's a great callout- I just updated the dropdown to only display the "Sample Kits" button if at least one kit type has been configured for the given study environment

ui-participant/src/hub/kit/KitsPage.tsx
MatthewBemis: That was the intention, but I think I'll get rid of the entire "Click below for more information about this process.", it seems redundant

ui-participant/src/hub/HubRouter.tsx
```diff
@@ -34,6 +35,8 @@ export default function HubRouter() {
       <Route path="profile/:ppUserId?"
         element={<ParticipantProfile/>}/>
       <Route path="kits"
+        element={<KitsPage/>}/>
+      <Route path="kits/in_person"
```
devonbush: FWIW, it's generally best practice to use dashes rather than underscores in URLs.  They're easier to see in links, and apparently search engines like them better

ui-core/src/participant/ParticipantNavbar.tsx
```diff
@@ -398,6 +405,13 @@ export const AccountOptionsDropdown = (
               {i18n('manageProfiles')}
             </button>
           </NavLink>}
+          { studyEnv && studyEnv.kitTypes.length > 0 &&
```
devonbush, line 408: nice

devonbush (approved):
one small tweak, then this is good to go.  Love the UI -- very useful!

PR #1117, opened 2024-09-30
[JN-1382] [member's choice] Reference survey answers from other surveys

connorlbark:

If you use a variable of format `{<survey stable id>.<question stable id>}` in a survey, then the answer for that question will be sent to the frontend which will create it as a variable. Only referenced questions will be loaded.

- populate demo
- Fill out required surveys and, in the basics, put your address as massachusetts.
- Go to `Massachusetts survey` and make sure you see your name render properly in the first question.
- run `select * from survey` and check that the `referenced_questions` columns looks reasonable


core/src/main/java/bio/terra/pearl/core/dao/survey/AnswerDao.java
```diff
@@ -76,4 +76,20 @@ public Answer findForEnrolleeByQuestion(UUID enrolleeID, String surveyStableId,
                         .orElse(null)
         );
     }
+
+    public List<Answer> findAllForEnrolleeByQuestion(UUID enrolleeID, String surveyStableId, List<String> questionStableId) {
+        return jdbi.withHandle(handle ->
+                handle.createQuery("""
+                                   SELECT DISTINCT ON (question_stable_id) * FROM answer
```
devonbush: can you add a comment reminding the reader how distinct on filters duplicates?  IIRC it grabs the "first", which is why you have it ordered by last_updated_at to get the most recent?

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
```diff
@@ -131,6 +131,26 @@ public static SurveyQuestionDefinition unmarshalSurveyQuestion(
         return definition;
     }
 
+    private static Pattern surveyQuestionRegexPattern = Pattern.compile("\\{\\s*\\w+\\.[\\w.]+\\s*}");
```
devonbush: needs comment

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
```diff
@@ -131,6 +131,26 @@ public static SurveyQuestionDefinition unmarshalSurveyQuestion(
...
+            String cleaned = result.group().substring(1, result.group().length() - 1).trim();
+
+            int periodIndex = cleaned.indexOf('.');
+
+            String surveyStableId = cleaned.substring(0, periodIndex);
+            String questionStableId = cleaned.substring(periodIndex + 1);
+
+            return Map.entry(surveyStableId, questionStableId);
+        }).filter(
+                entry -> !List.of("profile", "proxyProfile").contains(entry.getKey())
```
devonbush: pull this list out into a named constant -- and make sure there's a comment around line 315 of surveyUtil.js noting that any extraVariables that are objects need to be added here so they are not attempted to be fetched

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
```diff
@@ -131,6 +131,26 @@ public static SurveyQuestionDefinition unmarshalSurveyQuestion(
...
+    private static Pattern surveyQuestionRegexPattern = Pattern.compile("\\{\\s*\\w+\\.[\\w.]+\\s*}");
+
+    public static Map<String, List<String>> findReferencedSurveyQuestions(Survey survey) {
+        return surveyQuestionRegexPattern.matcher(survey.getContent()).results().map(result -> {
+            String cleaned = result.group().substring(1, result.group().length() - 1).trim();
+
+            int periodIndex = cleaned.indexOf('.');
+
+            String surveyStableId = cleaned.substring(0, periodIndex);
+            String questionStableId = cleaned.substring(periodIndex + 1);
```
devonbush: let's use `split(".")` to make it clearer what's going on and to avoid off-by-ones

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
```diff
@@ -131,6 +131,26 @@ public static SurveyQuestionDefinition unmarshalSurveyQuestion(
...
+
+            return Map.entry(surveyStableId, questionStableId);
+        }).filter(
+                entry -> !List.of("profile", "proxyProfile").contains(entry.getKey())
+        ).reduce(new HashMap<>(), (acc, entry) -> {
```
devonbush: please add comment describing the desired map this reduce should make

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
```diff
@@ -131,6 +131,26 @@ public static SurveyQuestionDefinition unmarshalSurveyQuestion(
         return definition;
     }
 
+    private static Pattern surveyQuestionRegexPattern = Pattern.compile("\\{\\s*\\w+\\.[\\w.]+\\s*}");
+
+    public static Map<String, List<String>> findReferencedSurveyQuestions(Survey survey) {
+        return surveyQuestionRegexPattern.matcher(survey.getContent()).results().map(result -> {
```
devonbush: long chains are fun, but I think this method would be more readable if you broke out some intermediate variables for the `map` and `filter`.  Just seeing the variable names helps understand the intention

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -108,10 +108,23 @@ public SurveyWithResponse findWithActiveResponse(UUID studyEnvId, UUID portalId,
                 .stream().findFirst().orElseThrow(() -> new NotFoundException("no active survey found"));
         configSurvey.setSurvey(form);
         return new SurveyWithResponse(
-                configSurvey, lastResponse
+                configSurvey, lastResponse, getRelevantAnswers(enrollee, form)
         );
     }
 
+    private List<Answer> getRelevantAnswers(Enrollee enrollee, Survey survey) {
```
devonbush:
```suggestion
    private List<Answer> getReferencedAnswers(Enrollee enrollee, Survey survey) {
```

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
```diff
@@ -131,6 +131,26 @@ public static SurveyQuestionDefinition unmarshalSurveyQuestion(
...
+            String surveyStableId = cleaned.substring(0, periodIndex);
+            String questionStableId = cleaned.substring(periodIndex + 1);
+
+            return Map.entry(surveyStableId, questionStableId);
+        }).filter(
+                entry -> !List.of("profile", "proxyProfile").contains(entry.getKey())
+        ).reduce(new HashMap<>(), (acc, entry) -> {
+            acc.computeIfAbsent(entry.getKey(), k -> new ArrayList<>()).add(entry.getValue());
+            return acc;
+        }, (a, b) -> a);
```
devonbush: this might be a complex and expensive enough operation that it's worth computing it when the survey is created and storing it in the DB.  That could help with debugging and making this more performant

devonbush (changes requested):
This is a really nice way of accomplishing this -- good balance of simplicity and performance.  I would like to consider putting this in the DB, since parsing the whole survey server-side every time we need to display it seems expensive, and it's nice to be able to debug the referenced answers at creation time, rather than have to wait for a response

core/src/main/java/bio/terra/pearl/core/dao/survey/AnswerDao.java
connorlbark: ~Yep, that's exactly it! I'll leave a comment explaining that.~
This code ended up being deleted

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
connorlbark: It's definitely better, but I'm not sure it's much cleaner (the issue being that, e.g. `survey1.matrixQuestion.specificQuestion` is possible, so we have to account for multiple periods in question names)

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyService.java
```diff
@@ -94,6 +102,27 @@ public Survey create(Survey survey) {
...
+                if (questionDef.isPresent()) {
+                    ReferencedQuestion referencedQuestion = new ReferencedQuestion();
+                    referencedQuestion.setSurveyId(survey.getId());
+
+                    referencedQuestion.setReferencedSurveyStableId(potentialSurveyStableId);
+                    referencedQuestion.setReferencedQuestionStableId(potentialQuestionId);
+
+                    referencedQuestionDao.create(referencedQuestion);
+                }
+            });
```
connorlbark: I'm debating if this check is worth it.

On the one hand, it's nice to be able to confirm that we're not loading junk data every time someone opens a survey - for example, if someone uses a matrix question `{myMatrixQuestion.myAnswer}`, we don't want a referenced answer to be created for that. Without this check, it would be created, with the survey stable id of `myMatrixQuestion`.

On the other hand, this introduces the possibility of a race condition on populate (i haven't had any issues yet, though) and general confusion. It could be that we just need to make good UX around this so that if you reference a question it's clear in the UI if the reference was picked up by the system or not.

core/src/main/resources/db/changelog/changesets/2024_09_30_referenced_questions.yaml
```diff
@@ -0,0 +1,20 @@
+databaseChangeLog:
+  - changeSet:
+      id: "referenced_questions"
```
devonbush, line 3: since this is a computed cache-y value that's only used for internal purposes, I'd lean toward having it be a column on Survey directly.  I'd recommend using a `text[]` column (Jdbi will handle autoconverting it into a `List<String>`, the only difficulty is that you have to add the column with sql since `text[]` isn't recognized by liquibase `addColumn` yaml) and just storing a list of `surveyStableId.questionStableId` strings.  That avoids the race conditions in populate and makes the fetching less complicated

devonbush (changes requested):
great changes, this is getting close!  and definitely worth getting over the finish line

core/src/main/resources/db/changelog/changesets/2024_09_30_referenced_questions.yaml
connorlbark: I think the race condition still exists, as long as we verify on survey creation that the other questions exist. But I didn't know about `text[]`! Great idea.

devonbush: I would be ok skipping the "other question exists" validation.  It might be nice to let survey creators create surveys in either order

connorlbark: Fair enough - this will put a bit more work into the parsing side, since we already have quite a few collisions with existing questions (e.g., we use `hd_hd_addressValidation.addressValidationResult` which gets picked up as a question reference)

devonbush (approved):
nice!

MatthewBemis (approved):
very cool!

PR #1121, opened 2024-10-02
[JN-1396] templatizing k8s manifests with helm

connorlbark:

Originally, I wanted to use Kustomize, but helm's Go templating makes our lives so much easier. Initial attempts to do our dns/cert stuff with Kustomize were... ugly. I really hope we can get away with pretty must just using helm as a basic template/deploy system. 

The nice thing is that helm has good integration into pretty much any CICD system. It seems possible to just use cloud build to trigger building and deploying this helm chart on merge, which would be pretty much the exact same as our current setup with argo. Though imo this is pretty low priority since in the meantime we can just do deploys manually. Perhaps the subject of the next member's choice 😄 

In the future, any additional changes will be as simple as authentication to `kubectl` and running:
```
helm upgrade -f environments/<dev/prod>.yml juniper .
```

if not installed
```
brew install helm
```

```
cd terraform/gcp/k8s
helm template . -f environments/dev.yaml > test.yml
```
review content of `test.yml` to make sure everything looks good / templating is effective

test https://juniper-cmi.dev / https://sandbox.demo.juniper-cmi.dev - everything on k8s was destroyed and redeployed using helm


terraform/gcp/k8s/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: juniper
+description: Juniper Helm chart
+type: application
+# this is the chart version, increment with each change
+version: 0.0.1
```
MatthewBemis, line 6: is this something we have to remember to update manually?

is there a way we can incorporate this into the templating process so it gets auto-incremented every time we make changes?

terraform/gcp/k8s/values.yaml
```diff
@@ -0,0 +1,18 @@
+adminUrl: ""
```
MatthewBemis, line 1: for my own understanding, since all of the values here are blank- what's the purpose of this file?

MatthewBemis (approved):
looks good, I just have a couple of questions

terraform/gcp/k8s/Chart.yaml
connorlbark: Oops, I meant to delete that comment. I've been able to redeploy without changing it, so I honestly think we can just not touch it. Since we're not deploying this helm chart remotely anywhere / giving it to any customers, I don't think the `version` here has any salient meaning.

terraform/gcp/k8s/values.yaml
connorlbark: It is the "base" values file, and your environments overlay onto this. It does two important things. First, it informs helm the structure of our `.Values` so it can give us more intelligent errors. Second, and more importantly, is that it enables autocomplete in the IDE, ahaha. 

It won't without this file because helm doesn't know where we keep our environment-specific variable files. If you think about it, helm is really meant for making charts that you send out to the public so that your application is reproducible. In that case, having an 'example, default' values file is really important because you wouldn't typically ship it with an actual complete values file - that's for the end customer to fill out. For us, it isn't so useful, but it's the way helm works, so we have to make it.

MatthewBemis: Gotcha, that makes sense. Thank you for the detailed explanation!

devonbush (approved):

PR #1122, opened 2024-10-02
[JN-1397] fix atcp key ordering problem

connorlbark:

final quick fix before kiara does a deep dive into everything that should be changed.

in a couple spots, A-T surveys translated things weirdly because there were keys that were conflicting depending on the order of execution (e.g. `diagnosis_location` and `diagnosis_location_country`). this makes sure it tries the longest key first 

- n/a, if you want, go check out the contacting physician A-T portal in demo. it's the most up to date, the countries should render as expected. (before it was like "Center for A-T, Center For A-T_country", whereas now it's "Center For A-T, United States", for example)


devonbush (approved):
thanks for the comment

MatthewBemis (approved):

PR #1123, opened 2024-10-02
[JN-1353] fixing pop errors

devonbush:

We were getting occasional role population errors in tests.  I suspect the cause is that this startup script was running any time the application was starting, but because it triggers on applicationReady, it might not be within the transactional wrapper of the test.  At least, that's the theory.  So we're disabling this method for unit tests.  

Ordinarily, detecting whether you're inside a test and behaving differently is a bad idea, but this seems like a good case for a limited exception to that rule

1.  see if unit tests don't sporadically fail over the next 2 weeks 😄

### DIFF
--- a/core/src/main/resources/db/changelog/db.changelog-master.yaml
+++ b/core/src/main/resources/db/changelog/db.changelog-master.yaml
@@ -213,9 +213,6 @@ databaseChangeLog:
       file: changesets/2024_02_23_rename_site_image.yaml
       relativeToChangelogFile: true
   - include:
-      file: changesets/2024_02_09_permission_populate.yaml
-      relativeToChangelogFile: true
-  - include:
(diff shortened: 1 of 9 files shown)


connorlbark (approved):
🙏

core/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -212,9 +212,6 @@ databaseChangeLog:
   - include:
       file: changesets/2024_02_23_rename_site_image.yaml
       relativeToChangelogFile: true
-  - include:
-      file: changesets/2024_02_09_permission_populate.yaml
-      relativeToChangelogFile: true
```
MatthewBemis, line 217: this was a great catch!

MatthewBemis (approved):
🤞

PR #1128, opened 2024-10-04
[JN-1402] update surveyjs

connorlbark:

There's a list of breaking changes [here](https://surveyjs.io/stay-updated/breaking-changes). Most of these shouldn't be a big deal to us (i think)

I believe I've improved a lot of the e2e tests. We relied very very heavily on complex xpath and css locators which surveyjs has no incentive to keep compatible. Playwright also heavily discourages using xpath/css wherever possible. I refactored what broke to make it more `getByText`-y rather than css-y. But I didn't touch what I didn't have to to make it work - I think there's still a lot we could do, if we want, to make these less brittle.

- try going through demo in dev and locally side by side and make sure nothing seems off
- try editing surveys in the admin editor and make sure previews/etc still work fine


ui-core/src/surveyjs/surveyJsStyle.scss
```diff
@@ -25,7 +24,6 @@
 }
 
 .sd-html p:only-child {
-  padding-top: 2em;
```
connorlbark, line 28: We no longer need to add this padding - surveyjs html blocks seem to have been updated to include this padding by default

e2e-tests/src/tests/studies/ourhealth/new-registration-age18.test.ts
```diff
@@ -66,18 +69,18 @@ test.describe('Home page', () => {
...
+      await expect(progress).toHaveText('Page 2 of 3')
       await consent.agree()
       await consent.clickByRole('button', 'Next')
 
 
       progress = await consent.progress()
-      expect(progress).toStrictEqual('Page 3 of 3')
-      await consent.fillIn(data.QLabel.FullName, 'Tony Junior Stark', page.locator('body')) // TODO randomize
+      await expect(progress).toHaveText('Page 3 of 3')
+      await page.locator('body').getByLabel('Full legal name *').fill('Tony Junior Stark')
```
connorlbark, line 63: This could have been accomplished with a `new Question` -> `question.fill` but imo I think we should refactor all of our tests to just do something like this. I think the nested interfaces we have made me spend at least an hour longer than I needed to to refactor these... but if we end up having a lot of e2e tests maybe the DRY-ness of the interface approach is better.

devonbush (approved):
approving, but we might want to consider waiting to merge this until after the Oct. 15 release so we get a full week of it

MatthewBemis (approved):

PR #1130, opened 2024-10-06
[JN-1403] basic merge service

devonbush:

No API or UX yet, but wanted to push this up to get thoughts on the approach.  The basic idea is to have one service that returns a "plan" for the merge, and then another service that executes the plan.  This is similar to how Environment publishing currently works. 

My main worry about this merge service is:
1. the amount of possible merge capabilities is vast and very complicated
2. It'll be too easy for an engineer to add a new data model to enrollees and then forget to update the merge service to operate on the new models. 

My long-term strategy for #1 is to keep the merge logic relatively basic, and invest more in admin capabilities to edit participant data  like deleting/assigning/completing tasks manually.  

For #2  Dennis and I had a conversation about some possible "EnrolleeProcessable" interfaces that could be put on services to make sure they implement merge & withdraw capability.  I spent a few hours trying to work something like that out, but it just didn't work -- there was too much difference and interconnectedness across the various services as to how they interface with withdraw/merge operations. So any suggestions are welcome.

As a drive-by, this pulls the test 'bundle' classes into their own files, which shortened a lot of test code.

1. look at the ParticipantMergeServiceTests, confirm you can follow what's happening


core/src/main/java/bio/terra/pearl/core/service/participant/merge/ParticipantUserMerge.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/** a record of an enrollee merge operation */
+@Getter
+@Setter
+@Builder
+public class ParticipantUserMerge {
+    private MergeAction<ParticipantUser, ?> users;
+    private MergeAction<PortalParticipantUser, ?> ppUsers;
```
connorlbark, line 19: Should this be a list?

core/src/main/java/bio/terra/pearl/core/service/participant/merge/MergePair.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public PairType getPairType() {
+        if (source == null && target != null) {
+            return PairType.TARGET_ONLY;
+        } else if (source != null && target == null) {
+            return PairType.SOURCE_ONLY;
+        } else {
+            return PairType.BOTH;
+        }
+    }
```
connorlbark, line 47: IMO I find having a `pairType` enum less readable than just doing simple null checks - it makes me think there's something more complex going on under the hood, and, at least for me, my eyes glaze over when I see the `Action` enum and `PairType` enum together in the same function (though I do appreciate that it's shorter)

core/src/main/java/bio/terra/pearl/core/service/participant/merge/MergeAction.java
```diff
@@ -0,0 +1,51 @@
...
+    public T getTarget() {
+        return pair.getTarget();
+    }
+
+    public enum Action {
+        MOVE_SOURCE,
+        NO_ACTION,
+        MERGE,
+        DELETE_SOURCE,
+        MOVE_SOURCE_DELETE_TARGET
```
connorlbark, line 49: It took me a while to fully grok the approach here. I would find it to be super helpful to have a basic `README.md` explaining the overall merge approach and also some comments here to explain the purpose of each `Action` (e.g., MOVE_SOURCE_DELETE_TARGET is for when target has no data but source does, etc). I think there's a lot of intentional scoping here to make it not too complex, but it's not really clear what that scoping is, so if I had to touch this code I think I would have to ask a lot of questions without just a bit more documentation

core/src/main/java/bio/terra/pearl/core/service/participant/merge/ParticipantMergeService.java
```diff
@@ -0,0 +1,199 @@
...
+        for (MergeAction<ParticipantTask,?> action : mergePlan.getTasks()) {
+            if (action.getAction().equals(MergeAction.Action.DELETE_SOURCE)) {
+                deleteTask(action.getSource(), auditInfo);
+            } else if (action.getAction().equals(MergeAction.Action.MOVE_SOURCE) ||
+                    action.getAction().equals(MergeAction.Action.MERGE) ||
+                    action.getAction().equals(MergeAction.Action.MOVE_SOURCE_DELETE_TARGET)) {
+                moveTask(action.getSource().getId(), target, targetPpUser, auditInfo);
+                if (action.getAction().equals(MergeAction.Action.MOVE_SOURCE_DELETE_TARGET)) {
+                    deleteTask(action.getTarget(), auditInfo);
+                }
```
connorlbark, line 101: Do we need to also delete the target on merge? Otherwise we'll end up with two tasks right?

core/src/main/java/bio/terra/pearl/core/service/participant/merge/ParticipantMergeService.java
```diff
@@ -0,0 +1,199 @@
...
+            deleteMergedEnrollee(mergeAction.getSource(), auditInfo);
+        }
+    }
+
+    protected void deleteMergedEnrollee(Enrollee enrollee, DataAuditInfo auditInfo) {
+        withdrawnEnrolleeService.withdrawEnrollee(enrollee, auditInfo);
+    }
+
+    /** merges data from one enrollee into another.  Does not delete the source enrollee */
+    protected void mergeEnrolleeData(Enrollee source, Enrollee target, EnrolleeMerge mergePlan, PortalParticipantUser targetPpUser, DataAuditInfo auditInfo) {
```
connorlbark, line 91: where do survey responses get merged together? e.g., if an enrollee has some responses in source and other responses in target? or is that not something we're worried about here? I'm kind of confused because I see things being moved but I can't find where things are actually merged 2 objects to one; what's the difference between `move` and `merge` in this context?

connorlbark (commented):
some initial clarifying questions/comments

core/src/main/java/bio/terra/pearl/core/service/participant/merge/ParticipantUserMerge.java
devonbush: for now, no.  The idea is that the merge is only taking place in a single portal (we don't want to give admins the ability to impact things outside their portal.  I could see eventually this becoming a list for a superuser super-merge, but that seems out of scope for now

core/src/main/java/bio/terra/pearl/core/service/participant/merge/MergePair.java
devonbush: Agree with the sentiment--I initially was coding it without this enum.  But then I proceeded to make so many copy-paste and logic typos getting `!= null` and `== null` switched that I put in the enum in.  And the rate of errors declined dramatically.

core/src/main/java/bio/terra/pearl/core/service/participant/merge/ParticipantMergeService.java
devonbush: Eventually, yes.  But since we're not merging the SurveyResponses (yet), we have to keep both tasks.  I'll add a comment to clarify

core/src/main/java/bio/terra/pearl/core/service/participant/merge/ParticipantMergeService.java
devonbush: Right now, we're not merging responses -- so the merge service treats MERGE the same as MOVE_SOURCE.  I see that's confused you in multiple places -- I'll see if I can come up with a clearer structure.

MatthewBemis (approved):
100% agree with the approach of keeping the merge functionality simple and building out admin edit functionality- that seems to line up with the functionality that we want to provide staff anyway and solves some amount of the maintenance burden that comes with having this type of merge functionality

I see that `applyMerge` takes in a `DataAuditInfo` object- is the eventual intent to add an additional audit record with some details of the merge plan?

devonbush:
yeah, we'll store that a merge happened.  And also we need a dataAuditInfo for a lot of the underlying service calls

connorlbark (approved):

devonbush:
Haven't forgotten about the comments here, just merging to avoid conflicts with tests in the user login PR.  Will handle comments in the later PR adding the API/UX

PR #1131, opened 2024-10-07
[JN-1404] fix proxy profile being populated even when not proxy environment

connorlbark:

`{isGovernedUser}` is derived from `proxyProfile` existing, and we were always populating it. The other approach here would be to keep `proxyProfile` always populated and either derive `isGovernedUser` from if `profile.id != proxyProfile.id` or there's another `isGovernedUser` prop on the paged survey view.

As an aside, I think it would be worth doing a code cleanup to have a `SurveyJsVariables` type with all of these things in it, and maybe even a `generateSurveyJsVariables(e: Enrollee, ...)` method. If you guys like that idea, I can do that here.

- extract cmi portal from dev
- upload cmi portal
- create non-proxy adult user
- ensure that consent shows correct language ("I" not "my child")

### DIFF
--- a/ui-participant/src/hub/survey/SurveyView.tsx
+++ b/ui-participant/src/hub/survey/SurveyView.tsx
@@ -40,7 +40,7 @@ ReactQuestionFactory.Instance.registerQuestion('addressvalidation', props => {
 /** handles loading the survey form and responses from the server */
 function SurveyView({ showHeaders = true }: { showHeaders?: boolean }) {
   const { portal, portalEnv } = usePortalEnv()
-  const { enrollees } = useActiveUser()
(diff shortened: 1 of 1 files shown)


devonbush (approved):
feels like this could use an automated test

MatthewBemis (approved):

connorlbark:
> feels like this could use an automated test

Trying this out, it feels more difficult than might be worth it - to test the regression that caused this issue, you have to test the highest level of survey component, which has no current tests and requires mocking way more than normal. I'm gonna add a ticket to the backlog to clean this up with a `SurveyJSVariable` type and generate surveyjsvariable function - that way we can test it separately from the rest of the survey things.

PR #1132, opened 2024-10-07
[JN-1375] Fix enrollee lookup + prevent COLLECTED_BY_STAFF status from being reset

MatthewBemis:

Previously, enrollee lookups took in portal/study/env as parameters, but never checked against them to make sure that the enrollee actually existed in the given study environment (or portal, for that matter...). Now we'll throw an error if the enrollee doesn't exist in the specified study env.

Also fixes an issue where kit requests with status COLLECTED_BY_STAFF would be set back to SENT status when we fetch latest statuses from DSM. COLLECTED_BY_STAFF is a status that's unique to Juniper, so we have to have a little bit of special logic.

Go to the kit scanning tool (live environment) https://localhost:3000/demo/studies/heartdemo/env/live/kits/scan
Try loading an enrollee shortcode that exists in sandbox but isn't in live (i.e. HDSALK)
Confirm you get an error
Try switching to sandbox and loading HDSALK
Confirm it loads

For kit status fix:

Ensure OurHealth sandbox is using live dsm and NOT mock DSM
Assign a new kit to OHSALK and then collect it
Go to the kit status page: https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/requested/collected
Click "Refresh"
Confirm that the in person kit remains in "COLLECTED_BY_STAFF" status

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeExtService.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeExtService.java
@@ -7,31 +7,37 @@ import bio.terra.pearl.core.model.audit.DataAuditInfo;
 import bio.terra.pearl.core.model.audit.ParticipantDataChange;
 import bio.terra.pearl.core.model.participant.Enrollee;
 import bio.terra.pearl.core.model.participant.WithdrawnEnrollee;
+import bio.terra.pearl.core.model.study.StudyEnvironment;
 import bio.terra.pearl.core.service.exception.NotFoundException;
 import bio.terra.pearl.core.service.participant.EnrolleeService;
 import bio.terra.pearl.core.service.participant.WithdrawnEnrolleeService;
+import bio.terra.pearl.core.service.study.StudyEnvironmentService;
 import bio.terra.pearl.core.service.workflow.ParticipantDataChangeService;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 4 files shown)


api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeExtService.java
```diff
@@ -43,7 +49,12 @@ public List<Enrollee> findForKitManagement(
     return enrolleeService.findForKitManagement(studyShortcode, environmentName);
   }
 
-  public Enrollee findWithAdminLoad(AdminUser operator, String enrolleeShortcodeOrId) {
+  public Enrollee findWithAdminLoad(
```
devonbush, line 52: Let's convert this to use @EnforcePortalStudyEnvPermission(perm= "participant_data_view").  Or, if you're ambitious, create a new @EnforceStudyEnvEnrolleePermission annotation that also handles checking the enrollee is in the study environment

devonbush (approved):
Good catch--that's a hole that's existed for a long time

api-admin/src/main/java/bio/terra/pearl/api/admin/service/enrollee/EnrolleeExtService.java
MatthewBemis: if it's okay with you, I'm going to defer this for a later PR. I'd rather tackle this entire service all at once and implement the `EnforceStudyEnvEnrolleePermission` annotation that you proposed- and I'd like to keep this PR very small because I want to guarantee that it gets released tomorrow prior to the live-testing of in-person kits later this week

connorlbark (approved):

PR #1133, opened 2024-10-07
[JN-1331] Vanta API changes, error handling, and slack chattiness

zyme:

A few things going on here for JN-1331:
1. Vanta changed their update API in breaking fashion.  Fixing this required a few new bean-y classes to properly model the inputs and outputs.
2. The breaking change wasn't immediately clear from the logs.  It turns out that we had a bug that resulted in our update calls hanging forever because our error handling wasn't quite right.  It's been fixed now, so future breaking API changes should explode loudly and be visible from the logs.  I tested this by mangling some calls to vanta and verifying that we logged the 400 instead of hanging in perpetuity.
3. The slack messages that this script generates are only interesting when stuff changes.  Having 2x daily messages that basically say "everything is fine" was a distraction, so the code now only sends a message to slack if something has changed.
4. A few log changes along the way.

These changes have been deployed for a week or two.  The logs show it's doing the right thing, and the state of vanta also shows that we're keeping up with personnel changes.


compliance/src/main/java/bio/terra/pearl/compliance/SyncVantaUsers.java
```diff
@@ -177,24 +186,18 @@ public void run(String... args) throws Exception {
...
             int delayInSeconds;
             if (!header.isEmpty()) {
                 delayInSeconds = Integer.parseInt(header.get(0));
             } else {
-                delayInSeconds = 70;
+                delayInSeconds = 61;
             }
             log.info("Vanta rate limit exceeded; waiting for {}s", delayInSeconds);
-            try {
-                Thread.sleep(Duration.ofSeconds(delayInSeconds));
```
zyme, line 191: Instead of sleeping here, we use the more idiomatic `Retry.fixedDelay`.

compliance/src/main/java/bio/terra/pearl/compliance/SyncVantaUsers.java
```diff
@@ -208,25 +211,17 @@ private <T extends VantaObject> Collection<T> collectAllData(WebClient wc, Param
...
+            VantaResultsResponse<T> response = (wc.get().uri(uriBuilder -> {
                         if (cursorCount.get() > 0) {
                             uriBuilder.queryParam("pageCursor", results.get().getPageInfo().getEndCursor());
                         }
                         uriBuilder.queryParam("pageSize", 100);
                         return uriBuilder.build();
-                    }).retrieve())
+                    }).retrieve()).onStatus(HttpStatus.TOO_MANY_REQUESTS::equals, get429StatusHander())
                     .bodyToMono(type)
-                    .retryWhen(Retry.withThrowable(throwableFlux -> throwableFlux.filter(RateLimitException.class::isInstance).map(t ->
```
zyme, line 228: Recently vanta made a breaking change to one of their APIs, which exposed a bug in our code that led to the job hanging because it couldn't respond to a 400.  This bug is now fixed.

compliance/src/main/java/bio/terra/pearl/compliance/SyncVantaUsers.java
```diff
@@ -208,25 +211,17 @@ private <T extends VantaObject> Collection<T> collectAllData(WebClient wc, Param
         AtomicInteger cursorCount = new AtomicInteger(0);
 
         while (cursorCount.get() == 0 || results.get().getPageInfo().isHasNextPage()) {
-            log.debug("Getting cursor number {}", cursorCount);
+            log.info("Getting {} cursor number {}", type.getType().getTypeName(), cursorCount);
```
zyme, line 215: More often than not, it's helpful to see the progress through the cursors in the logs.  So this will be `info`.

compliance/src/main/java/bio/terra/pearl/compliance/model/SingleUpdate.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+@Getter
+@ToString
+public class SingleUpdate {
+
+    private String id;
+
+    private boolean inScope;
```
zyme, line 12: One of the changes vanta made to the API was to allow you to set `inScope` for each `id` in the list.  Prior to this change, all resources in the payload had to have the same scope.  Although our code currently splits the list of resources into `inScope=true` and `inScope=false` lists, I decided to leave the splitting code alone since it's working just fine.

devonbush (approved):
nice

MatthewBemis (approved):

PR #1134, opened 2024-10-08
[JN-1299] refine at translation script

connorlbark:

I was overzealous to think that proxy would be able to happen here - AT actually requires >1 governed user per proxy so my initial plan wouldn't work. 

This PR cleans up survey / profile data to the point that, from my initial testing, it seems good, just needs some much deeper QA than I have done. AFAIK the only major thing left is proxy account creation.

If you want to test, then you can:

Download data & atcp portal:
https://drive.google.com/file/d/1b7x-YzmRVwH3yO6GFkZoMjk-tvH07Dcy/view?usp=drive_link, https://drive.google.com/file/d/1aIsCiWLuxdOUU6fnpM5LM2Vc4X8OLppO/view?usp=drive_link

- Import portal
- Extract data zip file
- cd into the data zip file contents
- translate:
```bash
python3 ~/path/to/juniper/pepper-import/scripts/translate.py -DDj juniper_data_dict.xlsx -DDd pepper_data_dict.xlsx -I pepper_data.tsv -O juniper_data.csv -T translation_overrides.csv --limit 50
```
- delete admin notification in atcp portal
- import the `juniper_data.csv` file created by the translation script
- ensure that there's no FAILED participants
- click into the top couple participants and ensure there are no parse errors or anything
- verify profiles are properly filled out
- run `select * from answer where survey_stable_id = 'PREQUAL';` and verify pre-enroll answers are saved
   - NOTE: right now pre-enrollment responses in the UI are being displayed from the json blob even though refactors have made it to where pre-enrollment questions are actually saved as full surveyresponse objects. since this import infra only saves them as surveyresponses and doesn't fill in the json blob, these questions don't show up in the UI. IMO I think the UI should be refactored to use the surveyresponse object since (eventually) we'll want pre-enrolls to be full surveys anyway.


pepper-import/scripts/translate.py
```diff
@@ -508,6 +505,10 @@ def apply_translations(data: list[dict[str, Any]], translations: list[Translatio
     for row in data:
         new_row = {}
         for translation in translations:
+            # certain modules (e.g. kit requests, families, relations) can be repeated, so we need
+            # to see how many repeats there are and apply the translation to each one
+            if is_in_repeatable_juniper_module(translation.juniper_question_definition.stable_id):
+                apply_repeatable_translation(row, new_row, translation)
```
connorlbark, line 511: A-T doesn't technically use this, but I implemented it thinking they had kits.  future studies will need it anyway, so 🤷

devonbush (approved):
Looks good, please remove the import PR stuff that got (I assume) inadvertently included though

connorlbark:
> Looks good, please remove the import PR stuff that got (I assume) inadvertently included though

Import stuff is intentional - it's for importing pre-enroll information, which is critical for A-T because it includes all of their profile information. Also, I believe import is broken until this gets added because of the fact that the pre-enroll module will be included by default but the code to support it doesn't exist.

MatthewBemis (approved):

PR #1136, opened 2024-10-09
[JN-1331] Add support for Duo and Google Workspace

zyme:

We recently enabled Duo and Google Workspace integration in Vanta, so this PR adds the ability to de-scope newly added users identified by the integrations.  Along the way, I improved error handling for Vanta's `Resource of type resource not found with 'id' identifier matching 'xyz'`, which pops up for new integrations for as-of-yet unknown reasons (vanta support ticket is in the works).

Cloud function has been deployed for a few days and seems to be behaving itself.


compliance/src/main/java/bio/terra/pearl/compliance/SyncVantaUsers.java
```diff
@@ -380,6 +388,7 @@ private void setInScope(String accessTokeen, Collection<VantaObject> vantaObject
         try {
             String updateResult = getWebClientForIntegration(accessTokeen, integrationId, resourceKind)
                     .patch().bodyValue(updateMetadata).retrieve().onStatus(HttpStatus.TOO_MANY_REQUESTS::equals, get429StatusHander())
+                    .onStatus(HttpStatus.UNPROCESSABLE_ENTITY::equals, res -> res.bodyToMono(String.class).map(VantaUpdateException::new))
```
zyme, line 391: `422` comes back from Vanta when it thinks a resource id isn't found (even though the resource id is one that we got back from a previous `get`) so that we can log more information like `Resource of type resource not found with 'id' identifier matching '[id]'`.

devonbush (approved):

MatthewBemis (approved):

PR #1137, opened 2024-10-09
[JN-1419]  tracking login on PortalParticipantUser

devonbush:

As prep for TRCC, we can no longer assume a participantUser only has one PortalParticipantUser.  so we need to track logins at the PortalParticipantUser level.  this adds that ability, and then updates the search engine to use PortalParticipantUser.lastLogin as the login time.

Along the way, this does some refactoring of EnrolleeSearchExpressionDao to make it easier to add future things.

1. redeploy apiAdminApp
2. repopulate demo
3. confirm jsalk still shows as having logged in 5 hours ago  at https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants


core/src/main/java/bio/terra/pearl/core/dao/search/EnrolleeSearchExpressionDao.java
```diff
@@ -26,20 +26,34 @@
...
+    /** list of mappers for the various modules that can be included in a search result */
+    protected static final List<SearchModuleMapper<? extends BaseEntity>> moduleMappers = List.of(
+            new SearchModuleMapper<>("enrollee", Enrollee.class, EnrolleeSearchExpressionResult::setEnrollee),
+            new SearchModuleMapper<>("profile", Profile.class, EnrolleeSearchExpressionResult::setProfile),
+            new SearchModuleMapper<>("portalUser", PortalParticipantUser.class, EnrolleeSearchExpressionResult::setPortalParticipantUser),
+            new SearchModuleMapper<>("participant_user", ParticipantUser.class, EnrolleeSearchExpressionResult::setParticipantUser),
+            new SearchModuleMapper<>("mailing_address", MailingAddress.class, EnrolleeSearchExpressionResult::setMailingAddress),
+            new SearchModuleMapper<>("latest_kit", KitRequest.class, EnrolleeSearchExpressionResult::setLatestKit),
+            new SearchModuleCollectionMapper<>("answer", Answer.class, (result, answer) -> result.getAnswers().add(answer)),
+            new SearchModuleCollectionMapper<>("task", ParticipantTask.class, (result, task) -> result.getTasks().add(task)));
```
connorlbark, line 54: love this!!!

core/src/main/java/bio/terra/pearl/core/dao/search/EnrolleeSearchExpressionDao.java
```diff
@@ -65,11 +79,16 @@ public List<EnrolleeSearchExpressionResult> executeSearch(EnrolleeSearchQueryBui
...
+            StopWatch stopWatch = new StopWatch();
+            stopWatch.start();
+            var result = query
                     .registerRowMapper(Family.class, BeanMapper.of(Family.class, "family"))
                     .registerRowMapper(EnrolleeSearchExpressionResult.class, new EnrolleeSearchResultMapper())
                     .reduceRows(new EnrolleeSearchResultReducer())
                     .toList();
+            stopWatch.stop();
+            log.info("Search took {} ms", stopWatch.getTotalTimeMillis());
+            return result;
```
connorlbark: Is this for testing or are we logging to keep track over time?

core/src/main/java/bio/terra/pearl/core/service/search/terms/PortalUserTerm.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public SearchValueTypeDefinition type() {
+        return FIELDS.get(field);
+    }
+
+    public static final Map<String, SearchValueTypeDefinition> FIELDS = Map.ofEntries(
+            Map.entry("createdAt", SearchValueTypeDefinition.builder().type(DATE).build()),
+            Map.entry("lastLogin", SearchValueTypeDefinition.builder().type(DATE).build()));
+
```
connorlbark: These need to also be added to the `EnrolleeSearchService.getExpressionSearchFacetsForStudyEnv` method so that the query builder knows about it. Very open to ideas for refactoring here, I even forgot to add the FamilyTerm here, it's so easy to forget!

connorlbark (approved):

core/src/main/java/bio/terra/pearl/core/dao/search/EnrolleeSearchExpressionDao.java
devonbush: ah, good catch, yeah, I wanted to make sure I wasn't unduly impacting performance.  But I can take it out

core/src/main/java/bio/terra/pearl/core/service/search/terms/PortalUserTerm.java
devonbush: Good point ,although we don't support searching on this just yet.  (Doing so would require some tweaks to the eventPublishing to make sure PPUsers get attached).

core/src/main/java/bio/terra/pearl/core/dao/search/EnrolleeSearchExpressionDao.java
devonbush: thanks!  I spent a little bit of time seeing if I could avoid having these listed here -- if the query parser returned to this service a list of terms, then we'd both eliminate this list and reduce some unneccesary result set processing

core/src/main/java/bio/terra/pearl/core/service/search/terms/PortalUserTerm.java
connorlbark: Oh, I just noticed that that method wasn't implemented. I'm not sure what you mean that we need to change event publishing. Actually, as-is right now, search would work in most places that just use raw SQL instead of the extract search. But I'm curious - why not use use `portalParticipantUserDao` like we do in the `LatestKitTerm`/other terms to grab it? 

Also, thinking on this more, how are we choosing a specific `portalParticipantUser` in the complex (and non-existent right now) case there could be multiple? It doesn't seem like we have enough information since all that's passed down is the enrollee profileId / studyEnvId.

Imo this could be a good incentive to refactor searching to be more portal-aware, we've been discussing making search work across studies in a portal. Right now it's very strictly within a studyenv, which complicates something like this (and I don't think it's possible to "correctly" implement this, unless I'm missing something). I think if we restricted search by using a `portalEnvId` instead of a `studyEnvId` and passed that value down to the terms that needed it, that would be the ideal way to implement this. Then, if you wanted enrollees of a specific study, you could have a new term, either `study.shortcode` `enrollee.studyShortcode`

core/src/main/resources/db/changelog/changesets/2024_10_08_portal_last_login.yaml
```diff
@@ -0,0 +1,11 @@
+databaseChangeLog:  # This script is DEPRECATED -- it has been replaced by RolePopulator.  This has been removed from the changelog
```
MatthewBemis, line 1:
```suggestion
databaseChangeLog: 
```

leftover copypaste

MatthewBemis (approved):
👍

PR #1142, opened 2024-10-15
[JN-1421] import proxy accounts

connorlbark:

General approach:
- New property `proxy.username` which is both exported and imported
- When importing, before doing any imports, group into `AccountImport`s which include primary enrollee and proxy enrollees. This grouping occurs by matching `proxy.username` and `account.username`
- If there's a proxy and no enrollee is found with `account.username` the same as the proxy's `proxy.username`, then a non-subject proxy account is created for proxy relationships
- If an account is found where `account.username` = `proxy.username`, then the governed user is linked to this account instead of creating a new non-subject one.

Things of note
- This approach CANNOT handle the situation where there are >1 proxies for a governed user. If this is a big deal, I can figure out how to refactor things, but I think the simplicity of this approach justifies this limitation, especially if we build out admin tools for managing accounts. I'm assuming this is a pretty rare scenario, if it ever even happens.
- Proxy accounts are not effectively merged with any existing participants in juniper. This is because the existing infrastructure relies on checking uniqueness with `account.username`. This works fine for normal enrollees, but since proxy usernames are randomly generated, they will never match with anything in the system. I think this is OK since we're fleshing out merge functionality.

- observe tests
- go to demo data export and observe `proxy.username`
- try importing:
[test_import_2.csv](https://github.com/user-attachments/files/17382664/test_import_2.csv)
[test_import.csv](https://github.com/user-attachments/files/17382665/test_import.csv)
- in both cases, 3 accounts should be created. the first should create a proxy that is a subject. the second should create a non-subject enrollee.


core/src/main/java/bio/terra/pearl/core/service/export/formatters/item/PropertyItemFormatter.java
```diff
@@ -55,6 +55,8 @@ public  Object getRawExportValue(T bean) {
         Object value = null;
         try {
             value = PropertyUtils.getNestedProperty(bean, propertyName);
+        } catch (NullPointerException | NestedNullException e) {
+            // this is expected if the property is null
```
connorlbark, line 59: drive by fix for the "family.shortcode" null error that keeps popping up in logs (I really wish getNestedProperty would just return null if it didn't exist instead of throwing...)

core/src/main/java/bio/terra/pearl/core/dao/participant/ParticipantUserDao.java
```diff
@@ -26,6 +26,17 @@ public Optional<ParticipantUser> findOneByShortcode(String shortcode) {
       return findByProperty("shortcode", shortcode);
    }
 
+   public Optional<ParticipantUser> findByEnrolleeId(UUID enrolleeId) {
```
devonbush, line 34: I'm surprised we've never had this method before!  thanks for adding

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
```diff
@@ -167,6 +149,65 @@ public Import importEnrollees(String portalShortcode, String studyShortcode, Stu
...
+        return accountData;
+    }
+
+    AccountImportData findExistingAccount(List<AccountImportData> accountData, String email) {
+        for (AccountImportData account : accountData) {
+            if (account.getEmail().equals(email)) {
+                return account;
+            }
+        }
+        return null;
```
devonbush:
```suggestion
       return accountData.stream().filter(account -> account.getEmail().equals(email)).findFirst().orElse(null);
```

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
```diff
@@ -167,6 +149,65 @@ public Import importEnrollees(String portalShortcode, String studyShortcode, Stu
...
+        List<AccountImportData> accountData = new ArrayList<>();
+
+        for (Map<String, String> enrolleeMap : enrolleeMaps) {
+            String email = enrolleeMap.get("account.username");
+            String proxyEmail = enrolleeMap.get("proxy.username");
+            if (email == null || email.isBlank()) {
+                email = proxyEmail;
+            }
+
+            AccountImportData existingAccount = findExistingAccount(accountData, email);
```
devonbush: let's rename this to `findAccount` -- `findExisting` made me think it was going to search the DB for already-imported participants.  Or better yet just inline it

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
```diff
@@ -167,6 +149,65 @@ public Import importEnrollees(String portalShortcode, String studyShortcode, Stu
...
+            } else {
+                if (existingAccount == null) {
+                    AccountImportData newAccount = new AccountImportData();
+                    newAccount.setEmail(email);
+                    newAccount.getProxyData().add(enrolleeMap);
+                    accountData.add(newAccount);
+                } else {
+                    existingAccount.getProxyData().add(enrolleeMap);
+                }
+            }
```
devonbush: Shorter and IMHO easier-to-read
```suggestion
            if (existingAccount == null) {
                AccountImportData newAccount = new AccountImportData();
                newAccount.setEmail(email);
                if (StringUtils.isBlank(proxyEmail) {
                  newAccount.setEnrolleeData(enrolleeMap);
                }  else {
                  newAccount.getProxyData().add(enrolleeMap);
                }
                accountData.add(newAccount);        
            } else {
                if (StringUtils.isBlank(proxyEmail) {
                    existingAccount.setEnrolleeData(enrolleeMap);
                } else {
                    existingAccount.getProxyData().add(enrolleeMap);
                }
            }
```

devonbush (approved):
looks good, some codesmithing suggestions

MatthewBemis (approved):

PR #1143, opened 2024-10-15
[JN-1436] adding field includes to data export

devonbush:

Adds the ability to specify a list of fields to limit the columns in the export to just those fields.

image

along the way, this 

-  refactored some of the Formatter class structure to make so the filtering logic could be somewhat centralized.  
-  converted the export APIs from GET to POST -- having them as GETs was a nicer representation of what they were, but as the params got more complex and longer, a POST felt more appropriate.
-  Updated BaseJdbDao to automatically recognize List<String> as a type that can be persisted.  Because this uses a different field-instrospection library, the order in which the fields are returned is slightly different.  This required updating some tests which had hardcoded sql equality checks--I confirmed the new and old sql was functionally identical, just with different field ordering. 
- Implemented a "save" method for ExportIntegrations

1. redeploy ApiAdminApp
4. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/export/dataBrowser
5. click Download, and then under advanced options, specify one or more field names (the field names are the same as the column headers in the export, e.g. "enrollee.shortcode"
6. confirm the download contains only those columns


wluken (approved):
Concept and design approved

MatthewBemis (approved):
nice, worked great

PR #1144, opened 2024-10-15
[JN-1379] make search value null resilient

connorlbark:

Improves search value comparison to be more null resilient, and adds some tests for the class.

- I am not even sure exactly what caused the null value - I don't think there's a good way to test this. Just refer to the unit tests.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/search/terms/SearchValue.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/search/terms/SearchValue.java
@@ -101,6 +101,13 @@ public class SearchValue {
             return right.equals(this);
         }
 
+        Object leftValue = this.getValue();
+        Object rightValue = right.getValue();
+
+        if (leftValue == null || rightValue == null) {
+            return leftValue == rightValue;
+        }
+
         if (this.searchValueType != right.searchValueType && !this.searchValueType.equals(SearchValueType.ARRAY)) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


core/src/main/java/bio/terra/pearl/core/service/search/terms/SearchValue.java
```diff
@@ -83,38 +83,57 @@ public static SearchValue ofNestedProperty(Object object, String field, SearchVa
...
         if (!this.searchValueType.equals(SearchValueType.ARRAY) && right.searchValueType.equals(SearchValueType.ARRAY)) {
             // flip the comparison if the right side is an array so that we hit the array comparison logic
             return right.equals(this);
         }
 
+        Object leftValue = this.getValue();
+        Object rightValue = right.getValue();
+
+        if (leftValue == null || rightValue == null) {
+            return leftValue == rightValue;
```
devonbush: `return leftValue == null && rightValue == null` has the same effect and might be easier to read

core/src/main/java/bio/terra/pearl/core/service/search/terms/SearchValue.java
```diff
@@ -83,38 +83,57 @@ public static SearchValue ofNestedProperty(Object object, String field, SearchVa
...
+            case INSTANT -> this.instantValue;
+            case DATE -> this.dateValue;
+            case BOOLEAN -> this.booleanValue;
+            case ARRAY -> this.arrayValue;
+            default -> null;
+        };
+    }
+
     public boolean equals(SearchValue right) {
         if (!this.searchValueType.equals(SearchValueType.ARRAY) && right.searchValueType.equals(SearchValueType.ARRAY)) {
```
devonbush: This wasn't from this PR but this strikes me as a bit too clever and makes the below code seem incomplete.  I'd recommend two helper functions "isUseArrayCompare()" and "arrayEquals"

devonbush (commented):
thanks for tackling this and adding the tests!

devonbush (approved):

MatthewBemis (approved):

PR #1146, opened 2024-10-16
[JN-1437] Adding import source parameter

devonbush:

The end goal is to not send task reminders to enollees who haven't yet logged in -- we want them to use the invitation email pathway.  Along the way, it made sense to just add a "source" field for enrollees.  

This does not yet surface the source anywhere in the UX, other than the export. 

Testing the reminder logic probably isn't worth the time (it's just an if statement), but confirming the source gets set correctly is
1.  redeploy apiAdminApp
2. repopulate demo
3.  go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/export/dataBrowser
4. download the results
5. confirm everyone except HDINVI has a source of 'PARTICIPANT'.  HDINVI has a source of IMPORT

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/model/participant/EnrolleeSourceType.java
+++ b/core/src/main/java/bio/terra/pearl/core/model/participant/EnrolleeSourceType.java
@@ -1,0 +1,6 @@
+package bio.terra.pearl.core.model.participant;
+
+public enum EnrolleeSourceType {
+    IMPORT, // the enrollee data was imported from an external source
+    PARTICIPANT // default, they came in through standard signup flow
+}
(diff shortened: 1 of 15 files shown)


core/src/main/java/bio/terra/pearl/core/model/participant/EnrolleeSourceType.java
```diff
@@ -0,0 +1,6 @@
+package bio.terra.pearl.core.model.participant;
+
+public enum EnrolleeSourceType {
+    IMPORT, // the enrollee data was imported from an external source
+    PARTICIPANT // default, they came in through standard signup flow
```
MatthewBemis: what do you think about calling this `ENROLLMENT` instead of `PARTICIPANT`? It seems like we want this to tell us which action resulted in the creation of the enrollee, rather than who created the enrollee, does that sound that right?

MatthewBemis (approved):
Looks great, just one naming suggestion

core/src/main/java/bio/terra/pearl/core/model/participant/EnrolleeSourceType.java
devonbush: Good thought -- just talked with Whitney, who suggested `PORTAL_SITE` to reflect they came through the UI

PR #1148, opened 2024-10-17
[JN-1425] merge modal

devonbush:

Upgrades the accounts list to have tabs for possible dupes and merging.
image

image

image

1. redeploy apiAdminApp
2. repopulate demo
3. go to `https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants/accounts/all`
4. confirm you see the user list
5. clikc the 'possible duplicates' tab
6. confirm you see a couple of users who have the same first&last name
7. click 'Preview Merge'
8. review the merge plan, click the 'switch' icon to make consented@test.com be the email to keep
9. click "execute merge"
10. on the user list, confirm consentedOtherEmail no longer appears. and that HDCOND appears in the 'withdrawn' list


ui-admin/src/study/participants/merge/ParticipantMergeView.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+const mergeArrow = <FontAwesomeIcon icon={faArrowRight} className="px-2"/>
+
+
+const TASK_STATUS_ICONS: Record<ParticipantTaskStatus, React.ReactNode> = {
+  'COMPLETE': <FontAwesomeIcon icon={faCheck}/>,
+  'IN_PROGRESS': <FontAwesomeIcon icon={faBatteryHalf}/>,
+  'NEW': <FontAwesomeIcon icon={faMinus}/>,
+  'REJECTED': <FontAwesomeIcon icon={faTimes}/>,
+  'VIEWED': <FontAwesomeIcon icon={faMinus}/>
```
connorlbark: IMO I think we might be relying too much on icons here. I think we should either replace it with just the status text - `new` `in progress` etc or use the same icons that we use in the enrollee view page (`EnrolleeView -> statusDisplayMap`) and write the task status out in the tooltip.

ui-admin/src/study/participants/merge/ParticipantMergeView.tsx
```diff
@@ -0,0 +1,232 @@
...
+function renderResultSummary(taskMerge: MergeAction<ParticipantTask, object>) {
+  if (taskMerge.pair.source && taskMerge.pair.target) {
+    if (taskMerge.action === 'MOVE_SOURCE') {
+      return <span>keep both</span>
+    } else if (taskMerge.action === 'MERGE') {
+      return <span>merge</span>
+    } else if (taskMerge.action === 'MOVE_SOURCE_DELETE_TARGET') {
+      return <span>keep duplicate, delete original</span>
+    } else if (taskMerge.action === 'DELETE_SOURCE') {
+      return <span>keep original, delete duplicate</span>
```
connorlbark:
```suggestion
      return <span>keep duplicate</span>
    } else if (taskMerge.action === 'DELETE_SOURCE') {
      return <span>keep original</span>
```
nit: I think i prefer just using `keep` language, frees up a lot of whitespace and I find it easier to scan

ui-admin/src/study/participants/participantList/PortalUserList.tsx
```diff
@@ -110,21 +117,52 @@ export default function PortalUserList({ studyEnvContext }:
...
+      </div>
+    },
+    {
+      ...useParticipantDupeTab({ users, studyEnvContext, onUpdate: reload }),
+      path: 'dupes'
+    },
+    {
+      name: 'Account merging', path: 'merge', component: <ParticipantMergeView studyEnvContext={studyEnvContext}
+        onUpdate={reload}/>
+    }
```
connorlbark, line 139: I think it would be a really helpful if the textfields in this tab persisted as you switched tabs. I kept trying to copy from the main tab into this tab, then go back and find the other account, and copy it into the other field. But it would clear each time, so I had to copy the emails in a note doc to be able to keep track of them to put them in.

api-admin/src/main/resources/api/openapi.yml
```diff
@@ -538,6 +538,40 @@ paths:
...
+      operationId: plan
+      parameters:
+        - *portalShortcodeParam
+        - *envNameParam
+      requestBody:
+        required: true
+        content: *jsonContent
+      responses:
+        '200':
+          description: list of participant users and list of enrollees
```
connorlbark: nit: copy paste incorrect description? I think same for other endpoint.

connorlbark (approved):
nice! worked well in my testing, but I have some UIUX ideas/concerns. no blocking issues, though!

ui-admin/src/study/participants/merge/ParticipantMergeView.tsx
devonbush: agree it's wordy, but I like emphasizing that something is getting removed.  Changed to "keep duplicate only"

ui-admin/src/study/participants/participantList/PortalUserList.tsx
devonbush: bah, it doesn't look like there's an easy way to do this witht he current routing setup.  I'll keep this in mind if we revisit this page

core/src/main/resources/db/changelog/changesets/2024_10_17_withdrawal_reason.yaml
```diff
@@ -0,0 +1,19 @@
...
+        - addColumn:
+            tableName: withdrawn_enrollee
+            columns:
+              - column: { name: reason, type: text }
+        - addColumn:
+            tableName: withdrawn_enrollee
+            columns:
+              - column: { name: note, type: text }
+        - sql:
+            sql: update withdrawn_enrollee set reason = 'TESTING'; # as of now, no enrollees have requested to be withdrawn
```
MatthewBemis: I just want to double check this- I thought a number of OurHealth enrollees have requested to be withdrawn? Am I mixing that up with something else?

MatthewBemis (approved):
Looks great- just one question about previous OH withdrawals

core/src/main/resources/db/changelog/changesets/2024_10_17_withdrawal_reason.yaml
devonbush: ah, you're right.  I hadn't realized.  I'll reverse this and update everyone to be a legit withdrawal, and then manually update the testing ones

PR #1152, opened 2024-10-18
[JN-632] adding StudyEnvAwareDao interface

devonbush:

I think that adding interfaces like this StudyEnvAttached will be a step in helping us build out better infrastructure for staying on top of all the "forgot to support all the populate/publish/import/extract/delete/withdraw operations".  In the short term, they'll also just help clean out some boilerplate methods so that it's easier to see the really interesting methods in DAOs. 

I haven't updated all the DAOs that could be StudyEnvAttached yet, wanted to run this by you first to get any thoughts.  If you all approve, I'll update the rest before merging

 I had to do a bit of cleverness to make StudyEnvAttachedDao an interface, rather than an abstract class.  but I think that that's worth it to avoid try to cram everything into an inheritance of e.g. MutableStudyEnvAttachDao...

1.  no functional changes, just confirm tests still pass.
2. read a couple of the DAOs, confirm you like the way the code looks


connorlbark (approved):
Super cool!! I really like this.

MatthewBemis (approved):

PR #1154, opened 2024-10-18
[JN-1358] HOC for refreshing state on environment switch

MatthewBemis:

Adds a helper higher order component that wraps components that need to have their state reset when switching environments. This simplifies the usages of these components because we no longer need to remember to manually add a `key` prop or reset the state in some other way, which should reduce state-related bugs.

I applied this to LoadedSiteSettings, which was the most recent example of needing to do this.

Go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/settings
Edit some values
Confirm that switching between the environments clears any unsaved changes and displays the correct values for the selected env

### DIFF
--- a/ui-admin/src/util/withStateResetOnEnvChange.test.tsx
+++ b/ui-admin/src/util/withStateResetOnEnvChange.test.tsx
@@ -1,24 +1,29 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { render, screen } from '@testing-library/react'
 import { withStateResetOnEnvChange } from './withStateResetOnEnvChange'
 import { mockStudyEnvContext } from 'test-utils/mocking-utils'
+import { userEvent } from '@testing-library/user-event'
 
-const MockComponent = ({ text }: { text: string }) => {
-  return <div data-testid="mock-component">{text}</div>
+const MockComponent = () => {
+  const [input, setInput] = useState<string>('')
+  return <div data-testid="mock-component">
+    <input data-testid="mock-input" value={input} onChange={e => setInput(e.target.value)} />
+  </div>
 }
 
 const WrappedComponent = withStateResetOnEnvChange(MockComponent)
 
-test('resets component state when studyEnvContext changes', () => {
+test('resets component state when studyEnvContext changes', async () => {
   const { rerender } = render(
-    <WrappedComponent studyEnvContext={{ ...mockStudyEnvContext(), currentEnvPath: 'mystudy/sandbox' }} text="sandbox"/>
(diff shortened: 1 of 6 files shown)


ui-admin/src/util/withStateResetOnEnvChange.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { ComponentType, useEffect, useState } from 'react'
+import { StudyEnvContextT } from 'study/StudyEnvironmentRouter'
+
+export const withStateResetOnEnvChange = <T, >(WrappedComponent: ComponentType<T>) => {
+  const WithStateResetOnEnvChange = (props: T & { studyEnvContext: StudyEnvContextT }) => {
+    const { studyEnvContext } = props
+    const [key, setKey] = useState(studyEnvContext.currentEnvPath)
+
+    useEffect(() => {
```
devonbush: I don't understand why we need this hook -- if studyEnvContext.currentEnvPath is what we're listening for for changes, why can't we use that as the `key` directly?

ui-admin/src/util/withStateResetOnEnvChange.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+const WrappedComponent = withStateResetOnEnvChange(MockComponent)
+
+test('resets component state when studyEnvContext changes', () => {
+  const { rerender } = render(
+    <WrappedComponent studyEnvContext={{ ...mockStudyEnvContext(), currentEnvPath: 'mystudy/sandbox' }} text="sandbox"/>
+  )
+
+  expect(screen.getByTestId('mock-component')).toHaveTextContent('sandbox')
+
+  rerender(
```
devonbush, line 24: this test isn't testing the wrapper -- since the 'text' argument changes, the component will rerender anyway.  I think what you'll want to do is have the mock component have some state (e.g. a textbox) and then you userevent.type some characters to update the mock component state.  Then check that changing the studyenvironment resets the component state

devonbush (commented):
Thanks for adding this--I agree this will be better than having to remember to add `key` everywhere -- the components can own their own refreshability.   Just needs some tweaks to make sure we're testing the intended behavior

ui-admin/src/util/withStateResetOnEnvChange.tsx
MatthewBemis: ah you're totally right- this was some leftover from an earlier iteration. just cleaned this up

connorlbark (approved):

devonbush (approved):

PR #1159, opened 2024-10-23
[JN-1444] Improve GCP HTTPS security

connorlbark:

Adds HTTP->HTTPS redirect, HSTS headers, CAA records, and forces modern TLS/SSL protocols. This might mean some very old browsers (~2013-14) won't be compatible with our website, but old TLS protocols are insecure.

As a bonus, I also added cloud armor policies. They basically evaluate premade security checks before they pass the ingress. You can read about them [here](https://cloud.google.com/armor/docs/waf-rules) and compare with the ones we're using. Our rules are just a copy paste from what Pepper does.

- go to http://juniper-cmi.dev / http://juniperdemostudy.dev and ensure they both redirect to https
- run ssl test on juniper-cmi.dev (i've been using https://www.ssllabs.com/ssltest/) and ensure it gives a good score
- test cloud armor by going to https://juniper-cmi.dev/?test=Gifts%27+OR+1=1--%22 and observe 502 bad gateway
    - we will need to be careful - I suspect we might be able to trigger these rules in certain parts of our application, esp. search expressions


MatthewBemis (approved):

terraform/gcp/network.tf
```diff
@@ -68,3 +68,50 @@ module "cloud-nat" {
   name                               = "juniper-nat-config"
   source_subnetwork_ip_ranges_to_nat = "ALL_SUBNETWORKS_ALL_IP_RANGES"
 }
+
```
devonbush, line 71: can you add a comment here explaining what's going on or a link to docs?  the priority numbers are, um, interesting :)

devonbush (approved):
could use some comments/pointers for the arcane stuff

terraform/gcp/network.tf
connorlbark: alaric never got back to me, so I'll make this a TODO in the future

PR #1162, opened 2024-10-25
[JN-1455] refine trigger editor

connorlbark:

Requested review for quick thoughts and opinions since I finished up with this quickly.

My goal was to make the trigger editing page much more comprehensible. My thought was that there's really two separate things a trigger does - define when the thing happens, and define what the thing does. I hope this UI makes it really clear what's going on when you look at it.

image

Screenshot 2024-10-25 at 2 46 18 PM
Screenshot 2024-10-25 at 2 46 26 PM
Screenshot 2024-10-25 at 2 46 52 PM
image

I haven't actually deleted the old code yet, but it isn't being used anywhere.

Test for regressions - there's all kinds of tiny things I could have forgotten.


ui-admin/src/util/formatQueryBuilderAsSearchExp.tsx
```diff
@@ -50,10 +51,18 @@ export const createEnrolleeSearchExpressionRuleProcessor = (facets: ExpressionSe
...
+            } else {
+              processedValue = parseNumber(value, { parseNumbers: true })
+            }
           } else if (typeDefinition.type === 'BOOLEAN') {
-            processedValue = isString(value) ? value.toLowerCase() === 'true' : false
+            if (isBoolean(value)) {
+              processedValue = value
+            } else {
+              processedValue = isString(value) ? value.toLowerCase() === 'true' : false
+            }
```
connorlbark, line 65: ran into some issues with the `{profile.doNotEmailSolicit}` search expression we use in demo for testing conditional rules on triggers

core/src/main/java/bio/terra/pearl/core/service/search/terms/ProfileTerm.java
```diff
@@ -102,6 +101,8 @@ public SearchValueTypeDefinition type() {
             Map.entry("phoneNumber", SearchValueTypeDefinition.builder().type(STRING).build()),
             Map.entry("birthDate", SearchValueTypeDefinition.builder().type(DATE).build()),
             Map.entry("sexAtBirth", SearchValueTypeDefinition.builder().type(STRING).build()),
+            Map.entry("doNotEmail", SearchValueTypeDefinition.builder().type(BOOLEAN).build()),
+            Map.entry("doNotEmailSolicit", SearchValueTypeDefinition.builder().type(BOOLEAN).build()),
```
connorlbark, line 105: adds `doNotEmailSolicit` as an official field - it technically worked fine because the field did exist in the database, but now the UI won't be broken for it.

devonbush (commented):
love it!

devonbush (approved):
Nice, needs a frontend test or two, and also please delete the old code

ui-admin/src/components/InfoCard.tsx
```diff
@@ -47,7 +47,7 @@ import { isEmpty } from 'lodash'
  * ```
  */
 export function InfoCard({ children }: { children: React.ReactNode }) {
-  return <div className="card w-75 border shadow-sm mb-3">
+  return <div className="card w-100 border shadow-sm mb-3">
```
MatthewBemis, line 50: nice- in the portal dashboard redesign PR i hedged here and kept 75 as the default, allowing overrides. but i think you're right to just make it 100

MatthewBemis (approved):
very nice!

PR #1164, opened 2024-10-25
[JN-1452] simple publishing ux

devonbush:

Now that Aarushi and Christine have publishing access, I want to simplify the UX

image

this reflects that the vast majority of publishing is sandbox->irb->live.  You could still publish live -> irb by reversiing diff on the diff view page, but in 1+ years of Juniper, we've only done that once.

1. go to https://localhost:3000/demo/studies/heartdemo/publishing
2. confirm publishing flows still work


ui-admin/src/portal/publish/PortalEnvDiffView.tsx
```diff
@@ -124,6 +109,8 @@ export default function PortalEnvDiffView(
     { userHasPermission(user, portal.id, 'publish') &&
       <span>Select changes to publish</span>
     }
+    <Link to={studyDiffPath(portal.shortcode, studyShortcode!, destEnvName, sourceEnvName)}
+      className="ms-3">reverse diff</Link>
```
connorlbark, line 113: I wonder if we should even advertise doing this, just because it's getting into real "power-user" stuff. I could see this being here for superusers when things blow up and they ask for help from us. But I'm not sure, I don't really know when you'd need to use this anyway.

ui-admin/src/portal/publish/PortalPublishingView.tsx
```diff
@@ -0,0 +1,97 @@
...
+      <p>
+        Juniper manages three environments for each study: sandbox, irb, and live.
+        The sandbox environment is for testing, the irb environment is for IRB review,
+        and the live environment is for real participant use.
+      </p>
+      <p>
+       Once you have made changes to your study in the sandbox, you can publish those changes to the irb environment.
+       If these changes require IRB review, you can then continue to make other changes in the sandbox while
+       you wait for approval.  Once the changes are approved (or if they do not require IRB review), you can publish
+       from the IRB environment to the live environment.
```
connorlbark:
```suggestion
       Once you have made changes to your study in the sandbox environment, you can publish those changes to the irb environment.
       If these changes require IRB review, you can then continue to make other changes in the sandbox environment while
       you wait for approval.  Once the changes are approved (or if they do not require IRB review), you can publish
       from the IRB environment to the live environment.
```

connorlbark (approved):
Nice! Much easier to understand where to click

ui-admin/src/portal/publish/PortalEnvDiffView.tsx
devonbush: yeah, and I wouldn't expect them to necessarily know what "reverse diff" means.  I'll leave it for now, and we can see what Rachel and friends think

MatthewBemis (approved):
slick!

PR #1165, opened 2024-10-28
[JN-1456] make it easier to add new search terms

connorlbark:

I'm not 100% sure I prefer this way of doing this, it adds a lot of boilerplate. But it should be easier to add a new search expression. There are only 3 things you need to add: term, term parser, and the module mapper in the dao. Any of these missing will be very obvious, so, even if there's still a bunch of places to add things, there shouldn't be mysteriously missing things.


devonbush (approved):
Cool, agree that it adds lines, but I think it's more easily understandable, and keeps terms more isolated

MatthewBemis (approved):
nice!

PR #1166, opened 2024-10-28
[JN-1461] reference other study survey answers in search expressions

connorlbark:

Thought process: We keep search expressions study specific but certain terms can optionally be study-switchable. For example, `{answer.my_survey.diagnosis = 'something'} and {answer["my_other_study"].my_other_survey.diagnosis} = 'something else'`. This means "look for any enrollee with the same profile id as the enrollee in this study and enrolled in "my_other_study" that has a response to this question". This keeps the search expressions (imo) pretty intuitive. Currently, answer is the only term that's study-switchable, but other ones could be, too. They just need to add `parse(studyname, variable)` to their search term parser.

The UI is totally unaware of this new syntax, so you have to use the advanced editor to use it. In the future, we can probably figure out a way to get our facets to also look at other studies in the same portal environment.

- observe tests


core/src/main/java/bio/terra/pearl/core/dao/survey/AnswerDao.java
```diff
@@ -79,4 +79,24 @@ public Optional<Answer> findForEnrolleeByQuestion(UUID enrolleeID, String survey
                         .findFirst()
         );
     }
+
+    public Optional<Answer> findByProfileIdByStudyAndQuestion(UUID profileId, String studyName, String surveyStableId, String questionStableId) {
+        return jdbi.withHandle(handle ->
+                handle.createQuery("select a.* from " + tableName + " a " +
```
devonbush: please use multiline """ string and .formatted

core/src/main/java/bio/terra/pearl/core/dao/survey/AnswerDao.java
```diff
@@ -79,4 +79,24 @@ public Optional<Answer> findForEnrolleeByQuestion(UUID enrolleeID, String survey
                         .findFirst()
         );
     }
+
+    public Optional<Answer> findByProfileIdByStudyAndQuestion(UUID profileId, String studyName, String surveyStableId, String questionStableId) {
```
devonbush: this is a tough one to name...  but the two 'bys' threw me off

```suggestion
    public Optional<Answer> findByProfileIdStudyAndQuestion(UUID profileId, String studyName, String surveyStableId, String questionStableId) {
```

core/src/main/java/bio/terra/pearl/core/service/search/terms/SearchTermParser.java
```diff
@@ -16,14 +16,26 @@ public T parseVariable(String variable) {
 
         List<String> splitModelName = List.of(variableNoBraces.split("\\.", 2));
 
-        if (!splitModelName.getFirst().equals(this.getTermName())) {
+        String modelName = splitModelName.get(0);
+        String studyName = null;
+
+        if (modelName.contains("[")) {
```
devonbush, line 22: this feels like parser logic that belongs more in the antlr parser.  Is it possible to have a VARIABLE_WITH_STUDY term?  or otherwise update our parsing so that the syntax for how studies are specified is handled there?

devonbush (approved):
looks good, some codesmithing suggestions.

core/src/main/java/bio/terra/pearl/core/service/search/terms/SearchTermParser.java
connorlbark: I totally agree and I have ideas on how to do this but my antlr skills are not nearly refined enough to accomplish them. I tried doing
```antlr
expr:  term OPERATOR term | expr AND expr | expr OR expr | NOT expr | INCLUDE PAR_OPEN term PAR_CLOSE | PAR_OPEN expr PAR_CLOSE;
term: value | function | variable;
value: NUMBER | STRING  | BOOLEAN | NULL;
function: FUNCTION_NAME PAR_OPEN term (',' term)* PAR_CLOSE;
variable: VAR_OPEN VAR_MODEL (VAR_STUDY)? (VAR_ARGUMENTS)? VAR_CLOSE;

// Lexer rules
NUMBER: [0-9]+ ('.' [0-9]+)?;
STRING: '\'' (~[\\'\r\n])* '\'';
VAR_STUDY: '["'[a-zA-Z0-9_]+'"]';
VAR_MODEL: [a-zA-Z0-9_]+;
VAR_ARGUMENTS: '.'([a-zA-Z0-9_]|'.')+;
VAR_SEPARATOR: '.';
VAR_OPEN: '{';
VAR_CLOSE: '}';
```
But there's some weird execution order issues where it's trying to read functions as variables even though there isn't a {} surrounding them. I can't figure out why... I might punt this off as a future improvement.

MatthewBemis (approved):

PR #1167, opened 2024-10-28
[JN-1458] Toggle superuser view [member's choice]

MatthewBemis:

We have historically used a pattern where we would hide functionality from non-superusers. Recently we've been better, but there are still a bunch of things that are hidden or behave differently for superusers. This adds a very quick and not-at-all robust way of previewing the site as a non-superuser when you're actively logged in as a superuser.

Again, this is not meant to provide any sort of robust functionality- previewing the site as other roles could quickly get confusing (and hopefully isn't even needed in the future). To avoid adding a bunch of additional debug-related state to the UserProvider, in order to reset it, you'll just need to refresh the page.

Until we make our UX consistent in how we display superuser only functionality, I think this is good for sanity checking what our users actually see

Log in as a super user
Go to https://localhost:3000/ourhealth
Click the account dropdown
Uncheck the box to disable superuser view
Confirm that superuser only functionality is hidden
Click through the app and note other things are now invisible


connorlbark (approved):
Sweet! I really like how easy this is to use and I didn't notice any weirdness in my testing.

ui-admin/src/user/UserProvider.tsx
```diff
@@ -9,15 +9,17 @@ export type UserContextT = {
...
 
 /** current user object context */
 export const UserContext = React.createContext<UserContextT>({
   user: null,
   loginUser: () => { throw new Error('context not yet initialized') },
   loginUserUnauthed: () => { throw new Error('context not yet initialized') },
-  logoutUser: () =>  { throw new Error('context not yet initialized') }
+  logoutUser: () =>  { throw new Error('context not yet initialized') },
+  toggleSuperuserOverride: () => { throw new Error('context not yet initialized') }
 })
```
devonbush: we have a util for this (that should also have been used above)
```suggestion
  logoutUser: emptyContextAlertFunction,
  toggleSuperuserOverride: emptyContextAlertFunction
})
```

devonbush (approved):
Cool, agree this is a great first step.  Eventually we extend this out so that superusers can adjust their own roles in the UX

PR #1170, opened 2024-10-29
[JN-1139] Portal dashboard redesign

MatthewBemis:

Screenshot 2024-11-06 at 2 12 09 PM

Click into a portal, any portal (try a multi-study like HH to make it moderately more interesting)
Play around with stuff and confirm it behaves as you'd expect
Note any UX oddities


ui-admin/src/portal/dashboard/widgets/WebsiteWidget.tsx
```diff
@@ -0,0 +1,94 @@
...
+          </div>
+        </div>
+      </InfoCardHeader>
+      <InfoCardBody>
+        <div className="d-flex">
+          <div style={{ minHeight: '200px' }}>
+            { livePortalUrl ?
+              <img className="border rounded-3"
+                style={{ maxWidth: '100%', height: 'auto' }}
+                src={`https://api.urlbox.io/v1/A8hJem0DKIAE8mDm/png?url=${livePortalUrl}`}
```
MatthewBemis: my trial expired, so the screenshots don't work anymore =/ but it's not expensive ($50-100/month)

there might be cheaper alternatives, i can put a placeholder button here for now. i don't think this should block merging

connorlbark: FWIW I could get 90% of the way there (with a bit of weirdness) with just CSS scaling properties on a div around the website preview component. I also wonder if we used something like https://github.com/bubkoo/html-to-image we could get it to work.

ui-admin/src/portal/dashboard/widgets/StudyWidget.tsx
```diff
@@ -0,0 +1,114 @@
...
+          </Link>
+          {study.shortcode === primaryStudy &&
+            <span className="badge bg-secondary ms-2">Primary Study</span>}
+        </div>
+        <div className="dropdown">
+          <Button
+            data-bs-toggle='dropdown'
+            className="dropdown-toggle border m-1"
+            type="button"
+            id={`editStudyMenu-${study.shortcode}`}
```
devonbush: why does this button need an id?

ui-admin/src/portal/dashboard/PortalDashboard.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { StudyWidget } from './widgets/StudyWidget'
+import { WebsiteWidget } from './widgets/WebsiteWidget'
+import { TutorialsWidget } from './widgets/TutorialsWidget'
+import { MailingListWidget } from './widgets/MailingListWidget'
+import { StudyTeamWidget } from './widgets/StudyTeamWidget'
+import { StudyTrendsWidget } from './widgets/StudyTrendsWidget'
+import classNames from 'classnames'
+
+export default function PortalDashboard({ portal }: {portal: Portal}) {
```
devonbush, line 15: I love the modular design of this - let's wrap each widget in an ErrorBoundary to add extra stability

devonbush (changes requested):
great architecture -- love the look.  Only blocker is to wrap the widgets in error boundaries so that the user isn't blocked if some random widget fails.  Also, this could use at least one front-end test :)

ui-admin/src/portal/dashboard/widgets/StudyWidget.tsx
MatthewBemis: hmmm, im not sure, this part was copied over from the previous study list component. doesn't seem to be necessary. i'll remove it

ui-admin/src/portal/dashboard/PortalDashboard.tsx
MatthewBemis: good call!

ui-admin/src/navbar/StudySidebar.tsx
```diff
@@ -103,10 +96,6 @@ export const StudySidebar = ({ study, portalList, portalShortcode }:
         }
       </ul>}/>
       <CollapsableMenu header={'Design & Build'} content={<ul className="list-unstyled">
-        <li className="mb-2">
-          <NavLink to={studyEnvSiteContentPath(portalShortcode, study.shortcode, 'sandbox')}
-            className={sidebarNavLinkClasses} style={navStyleFunc}>Website</NavLink>
-        </li>
```
connorlbark, line 109: I feel like Website should stay on the sidebar but be there regardless of the study being selected or not. I feel like I think of every important study/portal building page in Juniper being accessible from the sidebar so my brain feels weird about this page not being there. In the future, we could maybe make it clear which parts of the sidebar are study-specific and which are portal-specific. Maybe we could add titles to the sidebar to separate into sections:

```
*ourhealth Portal*
 > Website

*OurHealth Study*
> Forms & Surveys
....

**Comment by MatthewBemis** [comment]
i forgot the tests! will add shortly

File: ui-admin/src/portal/dashboard/widgets/WebsiteWidget.tsx [hunk 5]
```diff
@@ -0,0 +1,102 @@
...
+        <DropdownButton
+          onClick={() => navigate(siteContentPath(portal.shortcode, 'sandbox'))}
+          icon={faPencil}
+          label="Edit website"
+          description="Design your portal website"
+        />
+        <div className="dropdown-divider my-1"></div>
+        {/* rendering the existing randing modal within the context of the portal dashboard
+            isn't particularly clean, from a user experience perspective
+            for now, we'll direct users to the sandbox site editor to edit branding */}
```
connorlbark: I found this confusing - I vote on just not having this as a button. Or maybe the above one should be `Edit website & branding`

connorlbark (approved):

ui-admin/src/navbar/StudySidebar.tsx
MatthewBemis: Totally agree with you, might be too premature to remove Website until the actual navigation and flow between portals<->studies is fully baked.

I'll add Website back into the original sidebar slot for the time being

devonbush (approved):

PR #1172, opened 2024-10-30
[JN-1387] Shortcode profanity filter

MatthewBemis:

Adds a basic profanity filter, so we don't generate shortcodes with offensive words in them.

There wasn't a good pre-existing Java library for this, so unfortunately this pulls in a shortened list of words that are banned by Google: https://github.com/coffee-and-fun/google-profanity-words

Not a great way to test this without modifying code or looking at the unit tests, but you can regression test to make sure that we're still generating shortcodes as expected (enroll a user, etc)


devonbush (approved):
nice, appreciate the thorough tests.  Agree that it's a bummer we have to have the wordlist in our repo, but I don't think it's worth the infrastructure to e.g. load the list from a google bucket

core/src/test/java/bio/terra/pearl/core/service/participant/ShortcodeServiceTest.java
```diff
@@ -36,4 +37,50 @@ void testFailsToGenerateShortcode() {
...
+                                ShortcodeService.SHORTCODE_LENGTH,
+                                ShortcodeService.SHORTCODE_ALLOWED_CHARS))
+                .thenReturn("BANNED")
+                .thenReturn("ALLOWED");
+        ShortcodeService shortcodeService = Mockito.spy(new ShortcodeService(randomUtilService));
+
+        String shortcode = shortcodeService.generateShortcode(null, (s) -> Optional.empty());
+        assertNotNull(shortcode);
+        assertEquals("ALLOWED", shortcode);
+    }
```
connorlbark, line 85: 🤯  great test

connorlbark (approved):

PR #1176, opened 2024-10-31
[JN-1428] make enrollee population more realistic

connorlbark:

Makes completed enrollees select random choices, if present.

Generated completed enrollee, check responses.

### DIFF
--- a/populate/src/main/java/bio/terra/pearl/populate/service/EnrolleePopulator.java
+++ b/populate/src/main/java/bio/terra/pearl/populate/service/EnrolleePopulator.java
@@ -54,6 +54,7 @@ import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.RandomUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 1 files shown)


connorlbark:
sonar failure is for using pseudorandom generator, which is totally OK here

MatthewBemis (approved):

populate/src/main/java/bio/terra/pearl/populate/service/EnrolleePopulator.java
```diff
@@ -613,6 +609,40 @@ public void autoCompleteSurvey(ParticipantTask task, Enrollee enrollee, Particip
...
 
+    private AnswerPopDto createAnswerFromQuestionDef(SurveyQuestionDefinition questionDef) {
+        AnswerPopDto answer = AnswerPopDto.builder()
+                .questionStableId(questionDef.getQuestionStableId())
+                .build();
+
+        if (!StringUtils.isBlank(questionDef.getChoices())) {
+            answer.setStringValue(parseAndSelectRandomChoice(questionDef.getChoices(), "blah"));
+        } else {
+            answer.setStringValue("blah");
```
devonbush: might as well make the text input random too. 
```suggestion
            answer.setStringValue(RandomStringUtils.randomAlphabetic(RandomUtils.nextInt(5,30));
```

devonbush (approved):
nice!  One future thing that would be cool is if we had a way to make the choices/text random but deterministic based on the enrollee shortcode and question stableId.  That way we could, e.g.  specify in jsalk.json a survey response with `"generateAutoResponses": true`, and then not have to specify out a bunch of values that we don't care what they are, but we do want them to be consistent

PR #1184, opened 2024-10-31
[JN-1410] file upload

connorlbark:

It turns out putting binary into postgres is kind of annoying. Since I thought it would be useful to stub out the file uplaods for local development anyway, I figured it would be helpful to to take cues from the address validation client provider structure for the file upload backend structure. So, however we want to accomplish file uploads, we can cleanly swap the code out.

This is totally agnostic to how we will actually assign the file upload task. That's been the subject of much debate. Plan is to come up with some ideas and then present them Thursday at demo.

(includes survey type additions that devon approved from other branch)

- observe tests


core/src/main/java/bio/terra/pearl/core/model/fileupload/ParticipantFileUpload.java
```diff
@@ -0,0 +1,27 @@
...
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class ParticipantFileUpload extends BaseEntity {
+    private String fileName;
+    private String fileType; // e.g. pdf, docx, etc.
+
+    private UUID uploadedFileId; // files are uploaded as UUIDs
+
+    private UUID creatingPortalParticipantUserId; // the user who uploaded the file; could be proxy
```
devonbush: I think these should be attached to enrollee and studyEnvironment (perhaps instead of portalParticipantUserId, but at least in addition to).   Participants upload files after consenting to a particular study, so we have to track which consent the files were uploaded in accordance with.  We'll eventually want to give enrollees the ability to easily share/propagate their files across multiple studies, but IIRC the default should always be study-scoped, and in the "file hub" we'll need to clearly indicate which study(ies) the files belong to

core/src/main/java/bio/terra/pearl/core/service/fileupload/FileStorageConfig.java
```diff
@@ -0,0 +1,19 @@
+package bio.terra.pearl.core.service.fileupload;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.core.env.Environment;
+import org.springframework.stereotype.Component;
+
+@Getter
+@Setter
+@Component
```
devonbush: I think @Configuration would be a more descriptive annotation here

core/src/main/java/bio/terra/pearl/core/service/fileupload/ParticipantFileUploadService.java
```diff
@@ -0,0 +1,28 @@
...
+    private final FileStorageBackend fileStorageBackend;
+
+    public ParticipantFileUploadService(ParticipantFileUploadDao dao, FileStorageBackendProvider fileStorageBackendProvider) {
+        super(dao);
+
+        this.fileStorageBackend = fileStorageBackendProvider.get();
+    }
+
+    public ParticipantFileUpload uploadFileAndCreate(ParticipantFileUpload participantFileUpload, InputStream file) {
+        participantFileUpload.setUploadedFileId(fileStorageBackend.uploadFile(file));
```
devonbush: let's not hide the most significant operation inside a `set` call
```suggestion
        UUID fileId = fileStorageBackend.uploadFile(file)
        participantFileUpload.setUploadedFileId(fileId);
```

api-admin/src/main/resources/application.yml
```diff
@@ -45,7 +45,9 @@ env:
     token: ${MIXPANEL_TOKEN:}
   airtable:
     authToken: ${AIRTABLE_AUTH_TOKEN:foobar}
-
+  fileUpload:
+    backend: LocalFileStorageBackend
+    localFileStoragePath: ${LOCAL_FILE_STORAGE_PATH:/tmp/juniper}
```
devonbush:
```suggestion
    localFileStoragePath: ${LOCAL_FILE_STORAGE_PATH:/tmp/juniper/fileupload}
```

api-participant/src/main/resources/application.yml
```diff
@@ -33,7 +33,9 @@ env:
   mixpanel:
     enabled: ${MIXPANEL_ENABLED:false}
     token: ${MIXPANEL_TOKEN:}
-
+  fileUpload:
+    backend: LocalFileStorageBackend
+    localFileStoragePath: ${LOCAL_FILE_STORAGE_PATH:/tmp/juniper}
```
devonbush:
```suggestion
    localFileStoragePath: ${LOCAL_FILE_STORAGE_PATH:/tmp/juniper/fileupload}
```

core/src/main/java/bio/terra/pearl/core/model/fileupload/ParticipantFileUpload.java
```diff
@@ -0,0 +1,27 @@
...
+
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class ParticipantFileUpload extends BaseEntity {
+    private String fileName;
+    private String fileType; // e.g. pdf, docx, etc.
+
+    private UUID uploadedFileId; // files are uploaded as UUIDs
```
devonbush:
```suggestion
    private UUID externalFileId; // files are uploaded as UUIDs
```

devonbush: `uploaded` is ambiguous here, so I'd go with `external` or `backend`.  I could also see `externalFileName` since it's just a name in the external system

core/src/main/java/bio/terra/pearl/core/model/fileupload/ParticipantFileUpload.java
```diff
@@ -0,0 +1,27 @@
+package bio.terra.pearl.core.model.fileupload;
+
+import bio.terra.pearl.core.model.BaseEntity;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.experimental.SuperBuilder;
+
+import java.util.UUID;
+
```
devonbush:
```suggestion
/**
  a file stored externally that relates to a participant.  May be uploaded by either participant or staff 
  */
```

core/src/main/java/bio/terra/pearl/core/model/fileupload/ParticipantFileUpload.java
```diff
@@ -0,0 +1,27 @@
...
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class ParticipantFileUpload extends BaseEntity {
```
devonbush: I'd just as soon call this "participantFile". "ParticipantFileUpload" sounds more like it is a record fo the upload process, not a pointer to the file.  And then we can use this to track files that might not have come in via direct upload
```suggestion
public class ParticipantFile extends BaseEntity {
```

devonbush (commented):
Good stuff.  100% agree with using file storage instead of the database for the local option.  The thing that I think needs the most thought is how uploaded files are associated with studies, either instead of or in addition to being associated with a portal.

connorlbark:
> The thing that I think needs the most thought is how uploaded files are associated with studies, either instead of or in addition to being associated with a portal.

Something I had considering doing was keeping files at a portal level but a participant can link them to studies with a `EnrolleeParticipantFile` joiner table. We could also just copy the file over when we consent for it in another study, with the caveat that if any notes were changed or added to the file, then they wouldn't also propagate to the copy.

devonbush (approved):

MatthewBemis (approved):

PR #1185, opened 2024-11-01
[JN-1478] generic task assignment

connorlbark:

Trying out making task assignment generic. This adds a new `TaskConfig` concept. Anything that implements `TaskConfig` can then use the `TaskConfigDispatcher` which will handle all of the fancy task assignment features we want. You then extend it with your own task dispatcher (e.g. SurveyTaskDispatcher) and listen on all the events you care about, with the bulk of the logic just going through TaskConfigDispatcher.

- test no regressions in survey task assignment & publishing functionality.  existing tests were pretty thorough in this regard.


core/src/main/java/bio/terra/pearl/core/model/survey/StudyEnvironmentSurvey.java
```diff
@@ -33,4 +35,77 @@ public Versioned versionedEntity() {
...
+        if (survey == null) {
+            return null;
+        }
+        return survey.getVersion();
+    }
+
+    @Override
+    public RecurrenceType getRecurrenceType() {
+        if (survey == null) {
+            return null;
```
connorlbark, line 59: I'm not super happy with this - I think it's awkward to have all the null checks. But, at the same time, I think most tasks wouldn't be portal attached like surveys are, so I think it makes most sense to treat everything as a flat TaskConfig with surveys being the odd one out.

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
```diff
@@ -0,0 +1,347 @@
...
+                taskConfig.getStudyEnvironmentId(),
+                taskConfig.getStableId(),
+                Duration.of(taskConfig.getRecurrenceIntervalDays(), ChronoUnit.DAYS));
+        assign(enrollees, taskConfig, false, new ResponsibleEntity(DataAuditInfo.systemProcessName(getClass(), "assignRecurringSurvey")));
+    }
+
+    /**
+     * will assign a delayed task to enrollees who have never taken it, but are due to take it now
+     */
+    public void assignDelayedSurvey(T taskConfig) {
```
devonbush:
```suggestion
    public void assignDelayed(T taskConfig) {
```

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
```diff
@@ -0,0 +1,347 @@
...
+            return;
+        }
+        updateTasksFromEnrolleeEvent(enrolleeEvent);
+    }
+
+    protected abstract List<T> findTaskConfigsByStudyEnvironment(UUID studyEnvId);
+
+    protected abstract T findTaskConfigByStableId(UUID studyEnvironmentId, String stableId, Integer version);
+
+    protected abstract void copyTaskData(ParticipantTask newTask, ParticipantTask oldTask, T taskDispatchConfig);
```
devonbush: I would make this a non-abstract method, and implement it with a BeanUtils.copyProperties excluding id, createdAt, lastUpdatedAt, etc....  (or explicitly include the 'data' fields).  I think it's ok for this class to know about the schema of ParticipantTask

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    /**
+     * survey responses can update what surveys a person is eligible for -- recompute as needed
+     */
+    public void handleSurveyEvent(EnrolleeSurveyEvent enrolleeEvent) {
+        /** for now, only recompute on updates involving a completed survey.  This will
+         * avoid assigning surveys based on an answer that was quickly changed, since we don't
+         * yet have functions for unassigning surveys */
+        if (!enrolleeEvent.getSurveyResponse().isComplete()) {
```
devonbush: this seems like logic that we want at the dispatcher subclass level -- I'd move it there, and then delete this method

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
```diff
@@ -0,0 +1,347 @@
...
+     */
+    public void handleNewTaskConfigEvent(E newEvent) {
+        T newTaskConfig = findTaskConfigByStableId(newEvent.getStudyEnvironmentId(), newEvent.getStableId(), newEvent.getVersion());
+        handleNewTaskConfig(newTaskConfig);
+    }
+
+    /**
+     * create the survey tasks for an enrollee's initial creation
+     */
+    public void handleNewEnrolleeEvent(EnrolleeCreationEvent enrolleeEvent) {
```
devonbush: I think we can delete this method

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
```diff
@@ -0,0 +1,347 @@
...
+        if (taskOpt.isPresent()) {
+            ParticipantTask task = taskOpt.get();
+            log.info("Task creation: enrollee {}  -- task {}, target {}", event.getEnrollee().getShortcode(),
+                    task.getTaskType(), task.getTargetStableId());
+            task = participantTaskService.create(task, auditInfo);
+            event.getEnrollee().getParticipantTasks().add(task);
+        }
+    }
+
+    public void handleNewTaskConfig(T newTaskConfig) {
```
devonbush:
```suggestion
    public void updateTasksFromNewTaskConfig(T newTaskConfig) {
```

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
devonbush: In general, I'd support a convention where subclasses of this class are the only ones with `handle...` methods, since they're doing the listening.

core/src/main/java/bio/terra/pearl/core/service/publishing/Publishable.java
```diff
@@ -13,7 +13,7 @@
 public interface Publishable {
     public static final List<String> CONFIG_IGNORE_PROPS = List.of("id", "createdAt", "lastUpdatedAt", "class",
             "studyEnvironmentId", "portalEnvironmentId", "emailTemplateId", "emailTemplate",
-            "consentFormId", "consentForm", "surveyId", "survey", "versionedEntity", "trigger");
+            "consentFormId", "consentForm", "surveyId", "survey", "versionedEntity", "trigger", "version");
```
devonbush: I think we need version?

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigCreatedEvent.java
```diff
@@ -0,0 +1,9 @@
+package bio.terra.pearl.core.service.workflow;
+
+import java.util.UUID;
+
+public interface TaskConfigCreatedEvent {
```
devonbush: I don't think this should have stableId/version on it -- that won't be applicable, to e.g. kit assignment configs

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
```diff
@@ -0,0 +1,347 @@
...
+        this.enrolleeSearchExpressionParser = enrolleeSearchExpressionParser;
+        this.enrolleeContextService = enrolleeContextService;
+        this.portalParticipantUserService = portalParticipantUserService;
+    }
+
+    /**
+     * handle new task config creation/publishing
+     */
+    public void handleNewTaskConfigEvent(E newEvent) {
+        T newTaskConfig = findTaskConfigByStableId(newEvent.getStudyEnvironmentId(), newEvent.getStableId(), newEvent.getVersion());
```
devonbush: let's delete this method.  the only logic it has is finding the config, but that requires that configs always have stableIds and versions, which probably won't be the case (e.g. for kits)

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
```diff
@@ -0,0 +1,347 @@
...
+            assign(assignDto, newTaskConfig.getStudyEnvironmentId(),
+                    new ResponsibleEntity(DataAuditInfo.systemProcessName(getClass(), "handleNewAssignableTask.assignToExistingEnrollees")));
+
+        }
+    }
+
+    /** builds any tasks that the enrollee is eligible for that are not duplicates
+     *  Does not add them to the event or persist them.
+     *  */
+    public Optional<ParticipantTask> buildTaskIfApplicable(Enrollee enrollee,
```
devonbush:
```suggestion
    protected Optional<ParticipantTask> buildTaskIfApplicable(Enrollee enrollee,
```

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
```diff
@@ -0,0 +1,347 @@
...
+                        existingTask.getTargetStableId().equals(task.getTargetStableId()) &&
+                                !isRecurrenceWindowOpen(taskDispatchConfig, existingTask))
+                .toList().isEmpty();
+    }
+
+    /**
+     * whether or not sufficient time has passed since a previous instance of a task being assigned to assign
+     * a new one
+     */
+    public boolean isRecurrenceWindowOpen(T taskDispatchConfig, ParticipantTask pastTask) {
```
devonbush:
```suggestion
    protected boolean isRecurrenceWindowOpen(T taskDispatchConfig, ParticipantTask pastTask) {
```

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
```diff
@@ -0,0 +1,347 @@
...
+            return enrolleeService.findAll(assignDto.enrolleeIds());
+        }
+    }
+
+    /**
+     * To avoid accidentally assigning the same tasks or outreach activity to a participant multiple times,
+     * confirm that if the stableId matches an existing task, the existing task must be complete and the
+     * configured survey must allow recurrence.
+     */
+    public boolean isDuplicateTask(T taskDispatchConfig, ParticipantTask task,
```
devonbush:
```suggestion
    protected boolean isDuplicateTask(T taskDispatchConfig, ParticipantTask task,
```

devonbush (changes requested):
Nice!  this is definitely a step in the right direction.  Major thoughts are:

1.  I think the attempt to specify fields in TaskConfigCreatedEvent and extract the "getTaskConfigByStableId" isn't worth the extra complexity right now.  I'd make TaskConfigCreatedEvent a marker interface, and leave fetching the configs up to the subclass.  If you want, you could maybe have a `protected abstract findConfigsForEvent(UUID studynEnvironmentId, E event)`
2. I'd strip out the `handle` methods from TaskConfigDispatcher.  And then rearrange the remaining methods so that the methods expected to be invoked by subclasses (assignScheduledTasks, assignRecurring, assignDelayed, assign) are at the top and everything else is protected

core/src/main/java/bio/terra/pearl/core/service/publishing/Publishable.java
connorlbark: leftover from a bug i ran into

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigCreatedEvent.java
connorlbark: I would actually argue that maybe it should. For example, for surveys, we have `Survey` which specifies what the survey asks, the `Task`, and the enrollee's `SurveyRepsonse`. But with kits, we only have analogues for the `Task` and `SurveyResponse`. I would argue that perhaps there should be a `KitRequestConfig` to match `Survey`.  It probably wouldn't be worth the refactor, but I could see `KitRequest` being renamed to `Kit` and `KitRequest` is the specification for how `Kit`s are assigned. That new table is where the configuration - what kind of kit (blood etc.) & who to assign it to - would live.

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
```diff
@@ -0,0 +1,338 @@
...
+    public void assignScheduledTasks() {
+        List<StudyEnvironment> studyEnvironments = studyEnvironmentService.findAll();
+        for (StudyEnvironment studyEnv : studyEnvironments) {
+            assignScheduledTasks(studyEnv);
+        }
+    }
+
+    public List<ParticipantTask> assign(ParticipantTaskAssignDto assignDto,
+                                        UUID studyEnvironmentId,
+                                        ResponsibleEntity operator) {
```
connorlbark: Something we could do is also add a `boolean match(AssignDto)` function and in our `ParticipantTaskExtService` iterate through the `List<TaskConfigDispatcher>` provided by spring. That way assigndtos can automatically "just work" with new tasks. That could be also how we do manual task assignment for document requests.

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
devonbush: don't forget this

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
```diff
@@ -0,0 +1,338 @@
...
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
+import java.util.*;
+
+/**
+ * Handles dispatching tasks to enrollees based on task configurations.
+ */
+@Service
+@Slf4j
+public abstract class TaskConfigDispatcher<T extends TaskConfig> {
```
devonbush:
```suggestion
public abstract class TaskDispatcher<T extends TaskConfig> {
```

devonbush: otherwise it looks like it's dispatching configs

devonbush (approved):
nice, one code cleanup you missed, and then this is good to go

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskConfigDispatcher.java
connorlbark: Oops, sorry, I forgot to leave a comment. This is actually being used in a few `SurveyTaskDispatcher` tests, so I thought it would be easier to just keep it public rather than delete the tests.

MatthewBemis (approved):

PR #1188, opened 2024-11-05
[JN-1407] doc request task

connorlbark:


devonbush (commented):
Seeing this written out -- I'm having too many flashbacks to ConsentForms.    Let's talk -- seeing this code I'm leaning very strongly that DocumentRequest should just be a type of survey

devonbush:
oops, misclicked, didn't mean to close :)

PR #1191, opened 2024-11-07
[JN-1407] document request survey type

connorlbark:

Idea: Document requests are a new kind of survey. We attach files to document requests like we attach answers. This slightly muddies the intuitive "in progress" logic, as it could be that there is an answer OR there is a participant file. Otherwise, I think it neatly fits into survey logic.

What's left is that we need to create an API for uploading files. Then, the frontend for a document request will interact directly with the file upload API and then attach those files to the survey response. When the survey response is sent to the backend, the new code from this PR will attach it to the survey response.

Potentially, a far in the future world could have these files being referenced in answers. For example, there's a multi-upload file, and you can match a particular upload to a particular clinician, or something. But that's definitely an advanced need that won't be common. But it would be supported with this approach.

- n/a, verify unit tests make sense
- follow up admin UX pr will add admin view of this + synthetic data & populate capability

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
@@ -4,6 +4,7 @@ import bio.terra.pearl.core.dao.survey.SurveyResponseDao;
 import bio.terra.pearl.core.model.audit.DataAuditInfo;
 import bio.terra.pearl.core.model.audit.ParticipantDataChange;
 import bio.terra.pearl.core.model.audit.ResponsibleEntity;
+import bio.terra.pearl.core.model.fileupload.ParticipantFile;
 import bio.terra.pearl.core.model.participant.Enrollee;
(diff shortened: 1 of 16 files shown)


core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -335,6 +345,34 @@ private Answer createAnswer(Answer answer, SurveyResponse response,
         return answerService.create(answer);
     }
 
+    @Transactional
+    protected List<ParticipantFile> createOrUpdateParticipantFiles(SurveyResponse response, List<ParticipantFile> participantFiles, ResponsibleEntity operator) {
```
devonbush, line 349: This method surprised me in that it didn't handle persistence/deletion of the underlying files.  So how about renaming to "updateParticipantFileLinks" and adding a comment to clarify that it doesn't manage persistence for the files themselves

devonbush (approved):
looks good, one codesmith suggestion

PR #1195, opened 2024-11-07
[JN-1446] recurring survey participant

devonbush:

Adds participant capability for recurring surveys.  As discussed in demo, the iconography isn't perfect.  I've switched from chevron to vertical ellipsis to avoid visual confusion with other checkmarks.  I tried moving it to the left, but Connor and I agreed it made the most sense in its current position.  I also tried moving it to the right of the "Completed" text, but then it either made the right margin look uneven, or unaligned the text
 
image

on expansion, dates are shown
image

1. redeploy apiParticipant and apiAdminApp 
2. repopulate demo
3. enable development mode signin for participant flow
4. go to https://sandbox.demo.localhost:3001/
5. log in as `recurrenceDue@test.com`
6. confirm the screen appears as above, and that you can take the updated survey instance


core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -97,13 +97,6 @@ public SurveyWithResponse findWithActiveResponse(UUID studyEnvId, UUID portalId,
             lastResponse = dao.findOneWithAnswers(task.getSurveyResponseId()).orElse(null);
         }
```
devonbush, line 112: this was legacy code from before we had the task system we do now

ui-participant/src/hub/TaskLink.tsx
```diff
@@ -38,49 +43,67 @@ const statusDisplayMap: Record<ParticipantTaskStatus, StatusDisplayInfo> = {
...
+            </Link>
+            : i18n(`${task.targetStableId}:${task.targetAssignedVersion}`, { defaultValue: task.targetName })}
+          { (task.status === 'NEW' && !!history?.length) && <sup className="ms-1">{i18n('taskNew')}</sup>}
+          {(isExpanded || nested) && <span className="text-muted fst-italic ms-2">
+            ({instantToDateString(task.createdAt)})
+          </span> }
+        </div>
+        { history && history.length > 0 && <button onClick={() => setIsExpanded(!isExpanded)}
+          className="btn btn-link py-0" title={i18n('taskHistory')}>
+          <FontAwesomeIcon icon={faEllipsisV}/>
```
connorlbark: Oooo idea. What if it's a `faPlus`? I'm not sold on the ellipse because it strongly implies a context menu to me. But a plus symbol makes it feel like it would be an expansion and it wouldn't conflict with the checkmark. Then, when expanded, it becomes an `faMinus`.

For an alternate placement idea, I also think I could see it to the left of the task status icon.

image

Idk, just some ideas!

devonbush: hmm... I kind of like the plus.  I worry that having it to the left makes it look like that's what we want people to click on -- it's the first thing people will scan

MatthewBemis: Carets might work better while still fulfilling the spirit of the chevron.

`<FontAwesomeIcon icon={isExpanded ? faCaretUp : faCaretDown}/>`

screencapture-sandbox-demo-localhost-3001-hub-2024-11-13-13_49_52

devonbush: yeah, that's the right stuff -- indicates a dropdown but doesn't get confused with a checkmark like a chevron

MatthewBemis (approved):

api-admin/src/main/java/bio/terra/pearl/api/admin/service/forms/ScheduledSurveyAssignmentService.java
```diff
@@ -0,0 +1,30 @@
...
+      initialDelay = 5 * 1000) // wait an hour between executions, start after 5 seconds
+  @SchedulerLock(
+      name = "ScheduledSurveyAssignmentService.assignScheduledSurveys",
+      lockAtMostFor = "500s",
+      lockAtLeastFor = "10s")
+  public void assignScheduledSurveys() {
+    log.info("Scheduled survey processing beginning");
+    surveyTaskDispatcher.assignScheduledTasks();
+    log.info("Scheduled survey processing complete");
+  }
```
connorlbark: This is already being accomplished in `ScheduledAssignmentService` - we don't need this class.

connorlbark (approved):
Pending deletion of ScheduledSurveyAssignmentService, 👍

PR #1196, opened 2024-11-08
[JN-1471] adding scheduled export service

devonbush:

Adds a scheduled service to run all active exports at 1:00AM.  this also groups all the scheduled services together in a single package.  IMHO this makes it easier to get a sense of what runs when, and we really don't want the scheduled services to have business logic in them anywhere.

1.  edit the cron in the ScheduledExportService to a time coming up
2. redeploy ApiAdminApp
3. confirm the job runs at the time you specified


connorlbark (approved):

MatthewBemis (approved):
grouping the scheduled services together in one place is a great idea, i like it 👍

PR #1205, opened 2024-11-13
[JN-1231] extract only active content

connorlbark:

While there are more versioned objects, this _only_ deals with site content and surveys. They're really the only things that are making the extracts huge.

- populate hearthive
- for extra completeness, make some changes to surveys and to site content to create versions in different environments
- extract with `only active content` checked & repopulate
- observe that the downloaded zip file was small & that there is no history in the repopoulated portal

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/internal/PopulateController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/internal/PopulateController.java
@@ -179,11 +179,13 @@ public class PopulateController implements PopulateApi {
   }
 
   @Override
-  public ResponseEntity<Resource> extractPortal(String portalShortcode) {
+  public ResponseEntity<Resource> extractPortal(
+      String portalShortcode, Boolean extractPublishedVersionsOnly) {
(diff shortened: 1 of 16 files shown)


devonbush (changes requested):
Functionality looks good--just needs some renaming.  Elsewhere in the codebase we use "active" to delineate surveys/configs that are currently being used in an environment.  "Published" is used to mean something that made it out of sandbox.  So I think the concept you're going for is "extract only active".    Also, see if you can  leverage `findAllByStudyEnvIdWithSurvey` from StudyEnvrionmentSurveyService rather than having to write new DAO methods.

connorlbark:
> Also, see if you can leverage `findAllByStudyEnvIdWithSurvey` from StudyEnvrionmentSurveyService rather than having to write new DAO methods.

I'm not quite sure how to use this method since survey extraction happens at the portal level, unless we loaded all study environments for a portal, ran `findAllByStudyEnvIdWithSurvey` on each, and then got all the unique results. But I assume that's not what you mean.

devonbush (approved):

MatthewBemis (approved):

PR #1207, opened 2024-11-13
[JN-1475] Low-hanging fruit memory/query improvements for enrollee export

MatthewBemis:

Takes care of some low-hanging fruit in enrollee export by batch loading several modules. Obviously this drastically reduces the number of queries, but the more important bit is that the memory footprint for `loadEnrolleeExportData` was reduced by about 30-40% in OurHealth (with ~1000 enrollees, very little data, but the same survey configurations as production, so the same number of columns). The results were consistently repeatable, here's one before/after screenshot:

Before:
Screenshot 2024-11-13 at 12 53 19 PM

After:
Screenshot 2024-11-13 at 12 54 20 PM

Obviously from the screenshots, there are other areas we can improve on. Notably:

* `CSVPrinter.printRecord` is very memory intensive- requiring about half a gig in this test (to write a measley 5.4MB TSV...). I didn't see an immediate obvious way to reduce this so I left it alone for the time being.

Confirm data export still works as expected

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
@@ -4,16 +4,16 @@ import bio.terra.pearl.core.dao.search.EnrolleeSearchExpressionDao;
 import bio.terra.pearl.core.dao.survey.AnswerDao;
 import bio.terra.pearl.core.dao.survey.SurveyQuestionDefinitionDao;
 import bio.terra.pearl.core.model.export.ExportOptions;
-import bio.terra.pearl.core.model.participant.Enrollee;
-import bio.terra.pearl.core.model.participant.EnrolleeRelation;
-import bio.terra.pearl.core.model.participant.ParticipantUser;
-import bio.terra.pearl.core.model.participant.RelationshipType;
+import bio.terra.pearl.core.model.kit.KitRequest;
+import bio.terra.pearl.core.model.participant.*;
 import bio.terra.pearl.core.model.search.EnrolleeSearchExpressionResult;
 import bio.terra.pearl.core.model.study.Study;
 import bio.terra.pearl.core.model.study.StudyEnvironment;
 import bio.terra.pearl.core.model.study.StudyEnvironmentConfig;
 import bio.terra.pearl.core.model.survey.*;
+import bio.terra.pearl.core.model.workflow.ParticipantTask;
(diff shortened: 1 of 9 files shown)


core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -222,32 +214,47 @@ protected List<SurveyFormatter> generateSurveyModules(ExportOptions exportOption
...
+    public List<EnrolleeExportData> loadEnrolleeExportData(UUID studyEnvironmentId, ExportOptionsWithExpression exportOptions) {
+        Study study = studyService.findByStudyEnvironmentId(studyEnvironmentId).orElseThrow();
+        List<Enrollee> enrollees = loadEnrollees(studyEnvironmentId, exportOptions.getFilterExpression(), exportOptions.getRowLimit());
+
+        List<UUID> enrolleeIds = enrollees.stream().map(Enrollee::getId).toList();
+        List<UUID> profileIds = enrollees.stream().map(Enrollee::getProfileId).toList();
+        List<UUID> participantUserIds = enrollees.stream().map(Enrollee::getParticipantUserId).toList();
+
+        // batch load the following modules to reduce the number of queries and reduce the memory footprint of data exports.
+        // eventually, the in-clauses of these sql queries will be too large, and we'll need to batch load these in smaller chunks
```
connorlbark, line 226: just to make sure I understand, when you say batch load, is the idea to refactor our export like so?
- create CSV/TSV object early/at beginning
- load, e.g. 100 enrollees, add to CSV
- load next 100 enrollees, add to CSV, etc. until we're finished?

connorlbark (approved):
works great!

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
MatthewBemis: the key idea would be to just limit the in-clauses to some number of enrollee ids (we can probably get away with an in-clause containing a few thousand ids)

how we decide to assemble it all is still TBD. once we get to the point of having to worry about in-clause size limits, we probably want to start thinking about streaming the tsv back to the user as we assemble it (and that might look a lot like what you described)

devonbush:
wow, very unintuitive that batching would improve both number of queries and memory usage.  I would have thought that this would have increased performance but at the cost of memory usage.

devonbush (approved):

PR #1210, opened 2024-11-14
[no ticket yet] make gcp artifact upload more secure

connorlbark:

Some random slack thread pointed out that you can actually, technically, use these SA keys to change our artifacts from any branch unless you specify a ref. So, this ref check makes our artifact registry more secure because some random person can't come in and open a PR and then trigger uploading some malicious images. With this, it would fail - image uploads must come from the development branch.

- n/a, we'll have to see if it works when I merge or if it fails to upload!


connorlbark:
closed because we don't need this added security measure; only broad members can make PRs

PR #1211, opened 2024-11-14
[JN-1449] Soft maintenance mode

MatthewBemis:

Included in this PR:

* Adds a new maintenance mode wrapper component that checks if the application (admin and ptp) is maintenance mode, and displays a configurable message if so. Allows for a superuser to bypass it.
* Adds a new "System Settings" concept for superusers. Currently supports configuring maintenance mode-related settings and disabling scheduled jobs (see admin UX screenshot below)
* Adds a new annotation that we can use on scheduled tasks to check if the task should run. This is a global setting for now, but this paves the way for us to eventually get granular and specify which tasks we'd like to disable

I didn't want to hammer the backend with requests, so the participant and admin apps are configured to poll for system setting changes every 5 minutes. If a user is actively using the application when we enable maintenance mode, it could take up to 5 minutes for them to see the message unless they fully refresh the app in the meantime.

Admin UX:
screencapture-localhost-3000-system-settings-2024-11-18-15_06_10

User UX (admin and participant see the same UX):
screencapture-localhost-3000-maintenance-2024-11-14-16_40_34

* Go to the new system settings superuser page https://localhost:3000/system/settings
* Enable maintenance mode and add a message and a bypass phrase
* Go to a participant site and confirm you see the message: https://sandbox.demo.localhost:3001/
* Refresh the admin app and confirm you see the message
* Confirm the bypass phrase works
* Play around with disabling scheduled jobs. You should see messages like `Scheduled tasks have been disabled. Skipping task: sendReminderEmails`. You don't need to restart the app to see these, although that can speed up the process. You just need to wait for the next scheduled job to attempt to run
* Also confirm the 5 minute system setting polling works: make sure maintenance mode is **off*, open a participant site, enable maintenance mode, and then wait up to 5 minutes for the participant site to automatically display the banner


ui-admin/src/App.tsx
```diff
@@ -50,42 +52,49 @@ mixpanel.register({ application: 'ADMIN_UI' })
...
-                    <Route path='redirect-from-oauth' element={<RedirectFromOAuth/>}/>
-                  </Routes>
-                </BrowserRouter>
-              </div>
-            </UserProvider>
-          </AuthProvider>
-        }
-      </ConfigConsumer>
-    </ConfigProvider>
+    <ApiProvider api={previewApi('', '')}>
```
MatthewBemis, line 55: we may want to get slightly more sophisticated here and periodically poll the maintenance mode endpoint in case users are actively using the system when we enter maintenance mode. as it stands right now, only a hard-refresh would reload the latest settings

ui-core/src/maintenance/MaintenanceMode.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { MaintenanceModeSettings } from 'src/types/maintenance'
+
+export function MaintenanceMode({ children }: { children: React.ReactNode }) {
+  const [settings, setSettings] = useState<MaintenanceModeSettings>()
+  const [bypassMaintenanceMode, setBypassMaintenanceMode] = useState(false)
+  const [password, setPassword] = useState('')
+  const Api = useApiContext()
+
+  useEffect(() => {
+    const loadSettings = async () => {
```
MatthewBemis, line 15: we may want to get slightly more sophisticated here and periodically poll the maintenance mode endpoint in case users are actively using the system when we enter maintenance mode. as it stands right now, only a hard-refresh would reload the latest settings

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/PublicApiController.java
```diff
@@ -55,6 +56,22 @@ public ResponseEntity<Object> getConfig() {
     return ResponseEntity.ok(config);
   }
 
+  @Override
+  public ResponseEntity<Object> getMaintenanceModeSettings() {
+    Map<String, Object> config = new HashMap<>();
```
MatthewBemis, line 61: todo

api-admin/src/main/java/bio/terra/pearl/api/admin/service/system/CheckDisableScheduledTaskAspect.java
```diff
@@ -0,0 +1,35 @@
...
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.reflect.MethodSignature;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Aspect
+@Component
+public class CheckDisableScheduledTaskAspect {
```
connorlbark, line 14: cool!!!!

api-admin/src/main/java/bio/terra/pearl/api/admin/service/system/SystemSettingsExtService.java
```diff
@@ -0,0 +1,27 @@
...
+
+  public SystemSettingsExtService(SystemSettingsService systemSettingsService) {
+    this.systemSettingsService = systemSettingsService;
+  }
+
+  @SuperuserOnly
+  public SystemSettings updateSystemSettings(
+      OperatorAuthContext authContext, SystemSettings updatedSystemSettings) {
+    SystemSettings currentSystemSettings = systemSettingsService.getSystemSettings();
+    BeanUtils.copyProperties(updatedSystemSettings, currentSystemSettings);
```
connorlbark: should we exclude `id` `lastupdatedat` etc. ?

core/src/main/java/bio/terra/pearl/core/service/system/SystemSettingsService.java
```diff
@@ -0,0 +1,49 @@
...
+                    .maintenanceModeBypassPhrase(null)
+                    .maintenanceModeEnabled(false)
+                    .disableScheduledJobs(false)
+                    .maintenanceModeMessage(null).build();
+        }
+
+        return settings.getFirst();
+    }
+
+    public SystemSettings updateMaintenanceModeSettings(SystemSettings systemSettings) {
```
connorlbark: should this be `updateSystemSettings`?

ui-admin/src/system/SystemSettingsEditor.tsx
```diff
@@ -0,0 +1,98 @@
...
+    doApiLoad(async () => {
+      await Api.updateSystemSettings(settings)
+      Store.addNotification(successNotification('System settings updated'))
+    }, { setIsLoading: setIsSaving })
+  }
+
+  return <RequireUserPermission superuser>
+    <div className={'mx-3'}><form onSubmit={e => {
+      e.preventDefault()
+    }}>
```
connorlbark: curious: why does it need the `<form>`?

ui-core/src/maintenance/MaintenanceMode.tsx
```diff
@@ -0,0 +1,72 @@
...
+                onChange={e => {
+                  setPassword(e.target.value)
+                }
+                }/>
+              <button
+                className={'btn btn-primary mx-2 border border-1'}
+                onClick={() => {
+                  if (password === settings?.maintenanceModeBypassPhrase) {
+                    setBypassMaintenanceMode(true)
+                  }
```
connorlbark, line 66: Almost certain it's fine, but is there any concern with the fact that anyone would be able to check the network tab to get the password and bypass the wall?

connorlbark (approved):

ui-core/src/maintenance/MaintenanceMode.tsx
MatthewBemis: good question. i agree that it's fine, mostly because this is just a deterrent. if someone goes out of their way to bypass this, they shouldn't be surprised if some of their data gets lost as a result (i.e. we do a smoke test post GCP-cutover and realize we need to re-migrate the database)

we could do some naive password checking on the backend if we wanted to tighten it up a little bit, but i don't think that's necessary for our needs right now

ui-admin/src/system/SystemSettingsEditor.tsx
MatthewBemis: Doesn't need to be, this was leftover. Just removed it

devonbush (changes requested):
Overall, like the approach, and the code is solid.  I don't like that we've added yet another serial round trip that delays the participant UI from loading.  Can we load system settings along with the stuff retrieved by the ConfigProvider?

api-admin/src/main/java/bio/terra/pearl/api/admin/service/system/CheckDisableScheduledTaskAspect.java
devonbush: This is super cool -- not at all an annotation I would have thought to write, but I really like it

devonbush (approved):

PR #1217, opened 2024-11-15
[JN-1479] annotating more services

devonbush:

Adds a new EnforcePortalEnvPermission, and updates the dashboard, note, portal, and siteContent services

1. confirm participant notes still work
2. confirm sitecontent editor still works
3. confirm portal settings still work


MatthewBemis (approved):

connorlbark (changes requested):
I logged in locally as heartdemostaff and could not edit site content - the auth context seems to be returning a null portalEnv

Also, there's an issue with the admin sidebar on the homepage when not a superuser. I think I know why (unrelated to this PR). I'll put up a fix.

connorlbark (approved):
I wasn't able to change portal settings as heartdemostaff - but I think that's intentional, so 👍

PR #1225, opened 2024-11-19
[JN-1439] Fix Dependabot for api and ui modules, add terraform module

MatthewBemis:

Fixes dependabot checks for our various frontend and backend modules
Also adds checks for the terraform folder

There isn't a good way to validate this pre-merge. Look over the config and make sure it looks good

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,39 @@ updates:
     package-ecosystem: "github-actions"
     commit-message:
       prefix: "[JN-1457]"
+  - schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


devonbush (changes requested):
Awesome, thanks for fixing this.  I don't think we want to include pepper-import in our checks though.  That's only run offline, and brings in a ton of pepper dependencies which we have no power to update

connorlbark (approved):
We also need to look at `terraform/gcp/k8s` for the helm chart

devonbush (approved):

PR #1230, opened 2024-11-21
[JN-1506] Fix admin sidebar for non superusers

connorlbark:

Before, going to the homepage of juniper (where you select a portal) would be broken for non-superusers. This is because react will freak out if you `return` early (there can't be any hooks that are conditional - since we had a useEffect between the returns, react didn't like it).

- log in as heartdemostaff
- click "Home" & select portal


MatthewBemis (approved):
Thank you for fixing this!

devonbush (approved):
good catch!

PR #1266, opened 2024-11-21
[JN-1494] gcp deployments (+ docs)

connorlbark:

Also adds the ability to generate documentation with `terraform-docs` and general comments on each file explaining what the file does. I also created a "Production Release" doc: https://broadworkbench.atlassian.net/wiki/spaces/PEARL/pages/3519414286/Production+Deployment+GCP

- n/a, if you wanna see the triggers, go to cloud console, search for triggers, and look at `deploy-juniper`


terraform/gcp/cloud_build_deploy.tf
```diff
@@ -0,0 +1,114 @@
...
+  service_account = "projects/${var.project}/serviceAccounts/${google_service_account.juniper_cloudbuild_service_account.email}"
+
+  substitutions = {
+    _ENV = var.environment
+    _NAMESPACE = var.k8s_namespace
+  }
+
+  approval_config {
+    approval_required = var.environment == "prod" ? true : false
+  }
```
connorlbark, line 117: What I wanted to do was create a manual trigger that could be applied to any tag that matches our semvar tags. However, if you create a manual trigger, it cannot accept regex for the tag name, it can only be a static tag. So I settled for the in-between solution of creating an automatic trigger that requires approval. So, every time we merge a PR & create a new tag, it's gonna create a new job but not actually execute it. We can either accept the job we want to run or manually create a new job on the tag we want.

devonbush (approved):
thanks for creating the new wiki page

MatthewBemis (approved):

PR #1268, opened 2024-11-22
[JN-1401 Optionally hide navbar on pages

MatthewBemis:

https://github.com/user-attachments/assets/ea59070e-979a-4287-889e-8c1561a4bade

Go to the site editor
Create a new page and check the Hide Navbar option
Add the link to your site navbar so you can navigate to the new page (or just navigate to the url directly)
Confirm you only see the logo and the login button (and the language dropdown- if you're in Heart Demo)


ui-admin/src/portal/siteContent/AddPageModal.tsx
```diff
@@ -82,20 +85,37 @@ const AddPageModal = ({ portalEnv, portalShortcode, insertNewPage, onDismiss }:
...
+
+        <div className="d-flex mt-2">
+          <Checkbox
+            label={'Hide Navbar'}
+            checked={page.hideNavbar}
+            onChange={checked => {
+              setPage({ ...page, hideNavbar: checked })
             }}/>
+          <InfoPopup title="Page Path" content={
+            `If checked, navbar content will be hidden on this page and only the logo,
```
MatthewBemis: kinda wondering now...should we also hide footer navigation too?

devonbush: great question -- I suspect no, since the footer has legally required stuff.  and mainly we just want to avoid them hitting "register"

core/src/main/resources/db/changelog/changesets/2024_11_22_hide_navbar.yaml
```diff
@@ -0,0 +1,9 @@
+databaseChangeLog:
+  - changeSet:
+      id: "hide_navbar"
+      author: mbemis
+      changes:
+        - addColumn:
+            tableName: html_page
+            columns:
+              - column: { name: hide_navbar, type: boolean, defaultValueBoolean: false }
```
devonbush: let's make this column non-nullable--that's best practice for columns that are mapped to `boolean` rather than `Boolean`

devonbush (approved):
looks good, one schema nit

connorlbark (approved):

PR #1279, opened 2024-11-26
[JN-1513] fix markdown issues

connorlbark:

AT aggressively uses markdown. Dunno if it broke before or after surveyjs upgrade but our markdown was wonky and only sometimes got applied.

image

- add markdown to different quesitons and make sure it always applies
- make sure existing surveys look good / the same


MatthewBemis (approved):

devonbush (approved):
makes sense, but we'll want to regression test this a LOT.

PR #1284, opened 2024-11-27
[JN-1479] annotate remaining ext services

connorlbark:

- log in as `heartdemostaff@gmail.com` and try out:
    - add/remove emails to mailing list
    - upload/update site media 
    - look at metrics
- log in as superuser and try out:
     - create & delete a study
     - read internal configs

### DIFF
--- a/api-admin/src/test/java/bio/terra/pearl/api/admin/AuthAnnotationSpec.java
+++ b/api-admin/src/test/java/bio/terra/pearl/api/admin/AuthAnnotationSpec.java
@@ -56,4 +56,8 @@ public record AuthAnnotationSpec(
       List<Class<? extends Annotation>> otherAnnotations) {
     return new AuthAnnotationSpec(null, null, otherAnnotations);
   }
+
+  public static AuthAnnotationSpec noAnnotations() {
+    return new AuthAnnotationSpec(null, null, List.of());
+  }
 }
--- a/api-admin/src/test/java/bio/terra/pearl/api/admin/AuthAnnotationSpec.java
+++ b/api-admin/src/test/java/bio/terra/pearl/api/admin/AuthAnnotationSpec.java
@@ -57,7 +57,7 @@ public record AuthAnnotationSpec(
     return new AuthAnnotationSpec(null, null, otherAnnotations);
   }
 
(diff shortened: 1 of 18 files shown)


api-admin/src/test/java/bio/terra/pearl/api/admin/AuthTestUtils.java
```diff
@@ -22,6 +23,7 @@ public static void assertAllMethodsAnnotated(
             .filter(
                 method ->
                     Modifier.isPublic(method.getModifiers())
+                        && !Modifier.isStatic(method.getModifiers())
```
connorlbark: This skips over all of the static methods when running `assertAllMethodsAnnotated` - not sure if this is considered "correct" but one of the ext classes had a static method that was just a simple string modifier. I assumed we would not count static methods since they would never have access to the DB.

devonbush: nice, can you put this description in a comment in the code?

api-admin/src/test/java/bio/terra/pearl/api/admin/AuthAnnotationSpec.java
```diff
@@ -56,4 +56,8 @@ public static AuthAnnotationSpec withOtherAnnotations(
       List<Class<? extends Annotation>> otherAnnotations) {
     return new AuthAnnotationSpec(null, null, otherAnnotations);
   }
+
```
devonbush, line 59: This works, but I'd consider adding an `@Public` or `@Unrestricted` annotation.  That way it's clearer when reading the service that the lack of permissions is intentional.  (The annotations won't actually do anything yet, but might be useful later for logging purposes)

devonbush (approved):
Nice, a couple optional thoughts

MatthewBemis (approved):

api-admin/src/test/java/bio/terra/pearl/api/admin/AuthTestUtils.java
connorlbark: I ended up making this `@Public` annotated - it seems like a special case and I think it's probably better to be explicit

PR #1285, opened 2024-11-27
[JN-1515] Remove legacy survey designer

MatthewBemis:

I think at this point, we're pretty committed to the v2 designer. The legacy designer is a _ton_ of dormant code for us to be responsible for, so I think we should just remove it.

Confirm split designer still works


connorlbark (approved):

devonbush (approved):
wooooo!   there needs to be a special approval badge for changes that get rid of 1000+ lines of code

PR #1287, opened 2024-11-27
[JN-565] fix trigger editing for inactive triggers

connorlbark:

Makes you unable to change inactive triggers. 

image

- Edit a trigger
- Hit the back button
- Observe that it shows you a helpful message and disables saving

### DIFF
--- a/ui-admin/src/study/notifications/TriggerList.tsx
+++ b/ui-admin/src/study/notifications/TriggerList.tsx
@@ -104,7 +104,7 @@ export default function TriggerList({ studyEnvContext, portalContext }:
           </li> }
         </ul>
       </div> }
-      <div className="flex-grow-1 bg-white p-3">
+      <div className="flex-grow-1 bg-white px-3">
         <Routes>
           <Route path="triggers/:triggerId"
             element={<TriggerDesigner studyEnvContext={studyEnvContext}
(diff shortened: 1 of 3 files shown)


devonbush (changes requested):
If what we're trying to prevent is editing inactive triggers, let's make that more explicit.  Add a check in the `replace` and `create` methods in this service for if the received object is inactive, and throw an IllegalArgumentException.   And maybe an affordance in the UX for, if a trigger is inactive,  show an alert banner "this trigger is not active"

Otherwise, this will cause unexpected errors when looking at a participant's notification history and clikcing on the trigger.  If the trigger is inactive, we still want to be able to see the config that caused the email

ui-admin/src/study/notifications/TriggerList.tsx
```diff
@@ -104,7 +104,7 @@ export default function TriggerList({ studyEnvContext, portalContext }:
           </li> }
         </ul>
       </div> }
-      <div className="flex-grow-1 bg-white p-3">
+      <div className="flex-grow-1 bg-white px-3">
```
connorlbark, line 107: drive by alignment fix

devonbush (approved):
nice! thanks for the revisions

MatthewBemis (approved):

PR #1301, opened 2024-12-03
[JN-1507] Server unavailable error page

MatthewBemis:

Admin and participant UX is the same, with the omission of the error message in the participant UI:
screencapture-sandbox-demo-localhost-3001-not-found-2024-12-04-10_35_48

Since the error requires the config provider to set the error, I found the quickest way to induce that was to just change the useEffect in ConfigProvider.tsx to:

```
  useEffect(() => {
    Api.getConfig().then(fetchedConfig => {
      setConfig(fetchedConfig)
      setIsLoaded(true)
    }).then(() => {
      setError(`Unable to load config: Syntax Error: Unexpected token '<', "<!DOCTYPE"... is not valid JSON`)
    })
  }, [])
```

Do that in both the admin or participant apps, and confirm you see the above screenshotted error page.

### DIFF
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/i18n/I18nController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/i18n/I18nController.java
@@ -6,8 +6,6 @@ import bio.terra.pearl.core.service.exception.NotFoundException;
 import bio.terra.pearl.core.service.i18n.LanguageTextService;
 import bio.terra.pearl.core.service.portal.PortalService;
 import java.util.HashMap;
-import java.util.Objects;
-import java.util.Optional;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
 
--- a/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/i18n/I18nController.java
+++ b/api-admin/src/main/java/bio/terra/pearl/api/admin/controller/i18n/I18nController.java
@@ -2,6 +2,7 @@ package bio.terra.pearl.api.admin.controller.i18n;
 
 import bio.terra.pearl.api.admin.api.I18nApi;
 import bio.terra.pearl.core.model.portal.Portal;
+import bio.terra.pearl.core.service.exception.NotFoundException;
 import bio.terra.pearl.core.service.i18n.LanguageTextService;
 import bio.terra.pearl.core.service.portal.PortalService;
 import java.util.HashMap;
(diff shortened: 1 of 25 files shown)


ui-admin/src/HomePage.tsx
```diff
@@ -57,7 +57,7 @@ function HomePage() {
       <div className="d-flex justify-content-center mt-3">
         <div className="alert alert-warning" role="alert">
             You do not have access to any portals or studies. If this is an error, please
-            contact <a href="mailto:support@juniper.terra.bio">support@juniper.terra.bio</a>
+            contact <a href={`mailto:${SUPPORT_EMAIL_ADDRESS}`}>{SUPPORT_EMAIL_ADDRESS}</a>
```
MatthewBemis, line 60: we've talked about doing this, but since our support address will probably change soon once we migrate away from the terra.bio domain, now seemed like a good time to quickly do it

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/i18n/I18nController.java
```diff
@@ -21,17 +22,18 @@ public I18nController(LanguageTextService languageTextService, PortalService por
...
+    Optional<Portal> portal = portalService.findOneByShortcode(portalShortcode);
+
+    // default to English if no language is provided
+    String lang = Objects.requireNonNullElse(language, "en");
 
     HashMap<String, String> languageTexts;
-    if (language != null) {
-      languageTexts = languageTextService.getLanguageTextMapForLanguage(portal.getId(), language);
+    // if the portal is not found or not specified, at least return the global/system language texts
+    // it's possible to need language texts without a loaded portal context in the admin tool
```
MatthewBemis: we need to load language texts a little bit earlier now, so we don't necessarily have a portal context yet. hence this slight refactor

ui-admin/src/HomePage.tsx
MatthewBemis: whoops, the whole reason this came up was because i meant to put the support email address in the Application unavailable page, which i forgot to do 🤦 .

will update that...

edit: done.

ui-admin/src/providers/ConfigProvider.tsx
```diff
@@ -42,7 +43,7 @@ export default function ConfigProvider({ children }: { children: React.ReactNode
     })
   }, [])
   if (error) {
-    return <p>{error}</p>
+    return <ServiceUnavailable error={error}/>
```
devonbush, line 46: I think this might be a case where we don't want the admin and participant UIs to be the same.   (and then this also saves having to wrap the admin UI in i18nProvider).

devonbush (changes requested):
I think the "server unavailable" message should not rely on fetching translations from the DB--that fetch will likely fail if the server is unavailable.  I'm fine with the unavailable message being English-only.  It's mostly developers who will see this since if the server is down in production, the frontend won't load at all.  And this will save a lot of i18n provider refactoring

connorlbark (approved):

devonbush (approved):
Nice!  thanks for the revisions--I like how much simpler this got.

PR #1302, opened 2024-12-03
[JN-1502] Fix infinite spinner when clicking Register from pre-enroll

MatthewBemis:

This is the quicker fix, and I think it's also probably the right one for the moment. If a user is already in the middle of a pre-enroll and they click on Register again, this prevents them from losing their current progress in the form or registration flow.

The alternative is to muck with the StudyEnrollRouter quite a bit, which I didn't have much luck with. And it's a bit hairy anyway so this drastically reduces the surface area/risk. We may want to refactor that eventually, though.

Go to https://sandbox.demo.localhost:3001/studies/heartdemo/join/preEnroll
Click Register
Confirm you don't see a loading spinner or lose registration progress


devonbush (approved):
nice fix.  Can you add a code comment explaining what's happening?

connorlbark (approved):

PR #1305, opened 2024-12-05
[JN-1523] preserving study sidebar during portal navigation

devonbush:

Updates the portal dashboard to auto-select a study so that we can always have the study sidebar available.  Also changes some sidebar links so the study context is preserved while, e.g., navigating to edit the portal

1. go to https://localhost:3000/
2. select a portal
3. observe the full study sidebar now appears
4. click on Design -> Website
5. observe that the study context is preserved
6. go back to the portal dashboard
7. click on "Manage" links in the widgets.  Confirm study context is preserved


MatthewBemis (approved):
this pr is such a relief 🎉

ui-admin/src/navbar/StudySidebar.tsx
```diff
@@ -120,6 +124,12 @@ export const StudySidebar = ({ study, portalList, portalShortcode }:
             className={sidebarNavLinkClasses} style={navStyleFunc}>Site Settings</NavLink>
         </li>
       </ul>}/>
+      <CollapsableMenu header={'Manage'} content={<ul className="list-unstyled">
```
connorlbark, line 127: `Manage` sounds generic to me but I can't think of anything I'm super in love with... Maybe the section could be called `Study Staff` and the page is `Accounts`?

(As an aside, right now `Team members` has a different casing than the rest of the sidebar)

connorlbark (approved):
Nice!

ui-admin/src/navbar/StudySidebar.tsx
MatthewBemis: fwiw, this section used to be called `Administration` before i removed it in the redesign

also good catch with the differing casing- looks like Mailing list differs too

PR #1306, opened 2024-12-06
[JN-1528] cloud build alerts

connorlbark:

Has everything needed for alerts on start and success of deploys, but only creates failed deploy alerts. Because the messages are not customizable by default 😢 they all look the same and the channel quickly becomes super noisy. There's some options for making messages more customized.... but for now this just helps us catch issues. Only applied in dev, will apply in prod after merge.

- n/a


terraform/gcp/.terraform.lock.hcl
```diff
@@ -2,39 +2,40 @@
 # Manual edits may be lost in future updates.
 
 provider "registry.terraform.io/hashicorp/google" {
-  version = "6.4.0"
```
connorlbark, line 5: upgraded our google terraform for new features

terraform/gcp/juniper_alerts.tf
```diff
@@ -14,10 +14,6 @@ EOT
   }
 }
 
-# NOTE: ANYTIME YOU EDIT THIS, MAKE SURE YOU DISABLE
-#       "notify on incident closure" MANUALLY VIA THE UI
-#        There is no way to disable this via terraform
-#        and it adds noise to the channel.
```
connorlbark, line 20: I think this is solved with the terraform upgrade!

MatthewBemis (approved):

devonbush (approved):

PR #1323, opened 2024-12-10
[JN-1533] Consistent env var access

MatthewBemis:

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/logging/MixpanelService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/logging/MixpanelService.java
@@ -69,13 +69,14 @@ public class MixpanelService {
     }
 
     protected JSONObject buildEvent(JSONObject event) {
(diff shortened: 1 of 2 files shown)


connorlbark (approved):

core/src/main/java/bio/terra/pearl/core/service/logging/MixpanelService.java
```diff
@@ -69,13 +69,13 @@ public void logEvent(String data) {
     }
 
     protected JSONObject buildEvent(JSONObject event) {
-        MessageBuilder messageBuilder = new MessageBuilder(env.getProperty("mixpanel.token"));
+        MessageBuilder messageBuilder = new MessageBuilder(env.getProperty("env.mixpanel.token"));
```
devonbush: while you're doing this, might as well pull it into a constant since we have to use it twice

devonbush (approved):

PR #1331, opened 2024-12-11
[JN-1525] adds subject to participant list

devonbush:

Adds subject to the participant list 
image

This is hidden by default since it will only be useful for A-T and other studies with proxies

1. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants
2. confirm 'subject' is not initially shown 
3. Add it from the column select, confirm it appears

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,9 @@
   },
   "scripts": {
     "test": "npm -w ui-core test && npm -w ui-admin test && npm -w ui-participant test",
-    "lint": "npm -w ui-core run build && eslint --max-warnings=0 e2e-tests/src ui-admin/src ui-core/src ui-participant/src"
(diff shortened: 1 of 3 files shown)


connorlbark (commented):
Hmm... So right now the backend doesn't load proxies at all for the participant list. By default it only loads subjects. I would find this UI confusing because this will always display no proxies unless someone also goes in and edits their advanced search filters as well as this column filter.

Perhaps we should remove that automatically added search filter and instead do the filtering in the frontend. That way this column filter will work as expected.

Alternatively, we could have this column filter control whether or not that filter gets added to the search expression sent to the backend.

devonbush:
agree with Connor's thought.  We can go ahead and merge this PR to add the column, and then follow-up with a later PR to change the default filtering

connorlbark (approved):

MatthewBemis (approved):

PR #1332, opened 2024-12-12
[JN-1536] adding typescript live-checking

devonbush:

This adds back real-time (ish) typescript checking to the build.  
image

I've gone through and corrected some of the errors (notably, we had an entire duplicated section of api functions), but there are still a lot more to go.  I think we should go ahead and merge this, and live with the temporary inconvenience of having to dismiss the window while we clean up the rest, rather than wait for one massive PR to resolve it all

1. run `npm install`
2. restart ui-admin
3. confirm a window pops up showing you the boatload of TS errors we've accumulated since moving to Vite 😛


connorlbark (approved):
I noticed that it only seems to show errors as they load into the browser. For example, when I started the admin UI without any backend running, it displayed no errors. When I started the API without anything populated, it showed no errors. Only when I populated demo did it show any errors. That makes me worry there might be weird nested/lazy loaded bugs that won't show up. Not a huge deal because we'll find them eventually, but I definitely could see this hiding bugs from us in areas of the codebase we don't often go to locally. 

I wonder if we could supplement this, after we fix all of the errors, with a github action that statically checks if there are any typescript errors left in the codebase.

MatthewBemis (approved):
nice!

also, sorry for the merge conflicts from removing the legacy designer :(

PR #1333, opened 2024-12-12
[JN-1529] adding trcc config placeholders

devonbush:

so that https://sandbox.trccproject.juniper-cmi.org/ will be accessible.  I'm also opening a ticket so that if b2cconfig isn't found, the application doesn't wholesale crash


devonbush:
dupe

PR #1364, opened 2025-01-06
[JN-1416] Add language texts for 10 new languages

MatthewBemis:

Per LanguageScientific, this adds 10 new languages and updates our Spanish language texts.

Restart AdminApiApp
Confirm language texts repopulate without error
`select count(*) from language_text;` should yield about 1800 rows


populate/src/main/resources/seed/i18n/es/languageTexts.json
```diff
@@ -1,44 +1,43 @@
 [
-  {"keyName": "navbarDashboard", "text": "Tablero", "language": "es"},
-  {"keyName": "navbarChangePassword", "text": "Cambiar Contraseña", "language": "es"},
-  {"keyName": "navbarLogout", "text": "Cerrar Sesión", "language": "es"},
+  {"keyName": "navbarDashboard", "text": "Panel de control", "language": "es"},
```
MatthewBemis, line 2: This file in particular is interesting to look at because it basically shows us Google Translate versus a professional translation service

connorlbark (approved):
nice!

maybe this is totally redundant, but would it be worth making a test that makes sure there are at least some language texts populated for all of these languages? just to make sure something doesn't blow up and we lose a language.

devonbush:
Yeah, that sounds like a reasonable test.  I was surprised that we're now escaping `/` in translation files as ` \/ `   I think React handles that escaping for us, but I guess it can't hurt

devonbush (approved):

PR #1366, opened 2025-01-06
[JN-1551] checkboxes  with multiple 'other' questions

connorlbark:

By default, surveyjs only allows 1 other question per checkbox group. This PR adds the ability to make checkboxes that have multiple other questions attached, with different titles for each.

I'm not in love with this approach - it adds a new "magic" way to add questions to surveys that aren't really fully surveyjs questions but have their own stable id. But aside from some hacky things like creating separate questions elsewhere and somehow rendering them inside of another question's checkboxes, I felt like this was the most elegant solution with the caveat of some weirdness if you rely on surveyjs's question list.

image

image

- Create a participant, complete basics with an address in Massachussetts
- in mass survey, interact with the "Schools" question
- try multiple languages, verify that it loads the languages properly
- Fill out the details
- Right now, the answers for these questions don't show up on the enrollee's page, as they aren't really their own surveyjs question, so the page doesn't know about them. I think it'll be a slightly more than trivial refactor, so I'll do a cleanup PR after this to fix that if we are on board with this approach.
- Look at data export preview, observe that the questions show up there

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
@@ -59,20 +59,19 @@ public class SurveyParseUtils {
         List<JsonNode> subQuestions = new ArrayList<>();
 
         if (parent.get("type").asText().equals("paneldynamic") && parent.has("templateElements")) {
-            subQuestions = getPanelDynamicSubQuestions(parent);
+            // get subquestions and add parent stable id to each
+            subQuestions = getPanelDynamicSubQuestions(parent)
+                    .stream()
+                    .map(q -> (JsonNode) q.deepCopy())
+                    .map(q -> {
+                        ((ObjectNode) q).put("parent", parent.get("name").asText());
+                        return q;
+                    })
+                    .toList();
         }
 
-        // keep track of the parent stableid
-        subQuestions = subQuestions
-                .stream()
-                .map(q -> (JsonNode) q.deepCopy())
-                .map(q -> {
-                    ((ObjectNode) q).put("parent", parent.get("name").asText());
-                    return q;
-                })
-                .toList();
-
         if (parent.get("type").asText().equals("checkbox") && parent.has("choices")) {
+            // not a derived value, so doesn't need parent
             subQuestions = getCheckboxOtherSubquestions(parent);
         }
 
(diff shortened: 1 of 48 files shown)


scripts/generate_release_notes.sh
```diff
@@ -4,10 +4,10 @@
 git fetch --tags
 
 # Check what versions demo and production are currently running
-demoResponse=$(curl -s 'https://admin-d2p.ddp-dev.envs.broadinstitute.org/version')
+demoResponse=$(curl -s 'https://juniper-cmi.dev/version')
 demoGitTag=$(echo $demoResponse | jq -r '.gitTag')
 
-prodResponse=$(curl -s 'https://juniper.terra.bio/version')
```
connorlbark, line 10: drive by fix

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
```diff
@@ -134,6 +162,14 @@ public static SurveyQuestionDefinition unmarshalSurveyQuestion(
...
     }
 
+    private static String nodeToString(JsonNode node) {
+        if (node.getNodeType().equals(JsonNodeType.STRING)) {
+            return node.asText();
+        } else {
+            return node.toString();
+        }
+    }
+
```
connorlbark, line 171: another drive by fix - most of our survey question definitions were actually blank because if you pass a json object through the normal `asText`  it will just be "" since it isn't 'text'. This was affecting all questions that were translated - now, the name will be, e.g. `{ "en": "Title", "es": ... `

populate/src/main/resources/seed/portals/demo/studies/heartdemo/surveys/massachusettsSurvey.json
```diff
@@ -43,6 +43,71 @@
                 "value": "unsure"
               }
             ]
+          },
+          {
+            "name": "schools",
+            "type": "checkbox",
+            "title": "What schools did you attend?",
+            "renderAs": "checkbox-multiple-other",
```
connorlbark, line 51: Note: checkboxes are unchanged unless you include this

devonbush (approved):
This looks about as clean as it could be -- especially on the javascript side

MatthewBemis (approved):

PR #1369, opened 2025-01-07
[JN-1409] Document library [participant UX]

MatthewBemis:

screencapture-sandbox-demo-localhost-3001-hub-documents-2025-01-10-12_23_38

screencapture-sandbox-demo-localhost-3001-hub-documents-2025-01-10-12_24_00

Restart participant API
Repopulate demo
Log in as jsalk@test.com
Go to the new "Documents" page in the account dropdown
Observe 3 files of differing types
Confirm you can download them


populate/src/main/resources/seed/i18n/de/languageTexts.json
```diff
@@ -136,5 +136,18 @@
   {"keyName": "kitsInPersonCollectedThankYou", "text": "Vielen Dank für Ihre Teilnahme.", "language": "de"},
   {"keyName": "kitTypeSaliva", "text": "Speichelkit", "language": "de"},
   {"keyName": "kitTypeStool", "text": "Stuhlkit", "language": "de"},
-  {"keyName": "kitTypeBlood", "text": "Blutkit", "language": "de"}
+  {"keyName": "kitTypeBlood", "text": "Blutkit", "language": "de"},
```
MatthewBemis, line 139: member's choice idea: write a script where you pass in a json blob of new english language texts and it calls google/microsoft translation APIs and returns translated values for all of our supported languages

connorlbark: Should we add a comment to section this off as "machine translated" so it's clear next time we need to do any new translations which ones need to be verified?

populate/src/main/resources/seed/portals/demo/studies/heartdemo/enrollees/jsalk.json
```diff
@@ -122,7 +132,7 @@
       "surveyVersion": 1,
       "complete": true,
       "answerPopDtos": [
-        {"questionStableId": "medical_record_document_request", "format": "FILE_NAME", "stringValue":  "my_file.txt"}
```
connorlbark, line 125: I found it confusing that the UI let me delete this file, which then deleted this answer, without telling me it would. IMO it might be best to just prevent deleting if it's referenced by any answers, because if you delete it, then you might have a completed document request task with no document attached to it. It would probably be better to force them to go back to the document request, change the file, then allow them to delete it.

ui-participant/src/hub/documents/DocumentLibrary.tsx
```diff
@@ -0,0 +1,179 @@
...
+                  participantFile={participantFile}
+                  enrollee={enrollee}/>
+              </td>
+            </tr>
+          ))}
+        </tbody>
+      </table> }
+      {participantFiles.length === 0 &&
+        <div className="text-muted fst-italic my-3">{i18n('documentsListNone')}</div>
+      }
```
connorlbark, line 104: Do we also need to be able to upload things here? E.g. if study staff ask them to upload a file over the phone or something?

connorlbark (approved):
Looks really nice!

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/file/ParticipantFileController.java
```diff
@@ -0,0 +1,97 @@
...
+    this.request = request;
+    this.objectMapper = objectMapper;
+    this.requestUtilService = requestUtilService;
+  }
+
+  @Override
+  public ResponseEntity<Resource> download(
+      String portalShortcode,
+      String envName,
+      String studyShortcode,
```
connorlbark, line 42: Oh also, I think we should do a check to verify that the participant is in this study env

ui-participant/src/hub/documents/DocumentLibrary.tsx
MatthewBemis: I was thinking for now we'd keep upload limited to document request forms. Is that reasonable?

populate/src/main/resources/seed/i18n/de/languageTexts.json
MatthewBemis: nice idea- i've added `"machineTranslated": true` as a property on these for now

populate/src/main/resources/seed/portals/demo/studies/heartdemo/enrollees/jsalk.json
MatthewBemis: that's a good point. based on this and feedback during demo, i've removed the ability to delete for now. from discussion it seems like it's a case that we should support but perhaps only through the admin flow

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/file/ParticipantFileController.java
```diff
@@ -0,0 +1,110 @@
...
+      String enrolleeShortcode,
+      String fileName) {
+    ParticipantUser participantUser = requestUtilService.requireUser(request);
+    authUtilService.authParticipantToPortal(
```
devonbush: This check feels like it should be in the ExtService, rather than the controller

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/file/ParticipantFileController.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public ResponseEntity<Object> upload(
+      String portalShortcode,
+      String envName,
+      String studyShortcode,
+      String enrolleeShortcode,
+      MultipartFile participantFile) {
+    ParticipantUser participantUser = requestUtilService.requireUser(request);
+    authUtilService.authParticipantToPortal(
```
devonbush: same as above, this auth feels like an extService concern

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/file/ParticipantFileController.java
```diff
@@ -0,0 +1,110 @@
...
+        participantFileExtService.uploadFile(participantUser, enrolleeShortcode, participantFile);
+
+    return ResponseEntity.ok(created);
+  }
+
+  @Override
+  public ResponseEntity<Object> list(
+      String portalShortcode, String envName, String studyShortcode, String enrolleeShortcode) {
+    ParticipantUser participantUser = requestUtilService.requireUser(request);
+    authUtilService.authParticipantToPortal(
```
devonbush: same

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/file/ParticipantFileController.java
```diff
@@ -0,0 +1,110 @@
...
+        participantUser.getId(), portalShortcode, EnvironmentName.valueOf(envName));
+
+    return ResponseEntity.ok(participantFileExtService.list(participantUser, enrolleeShortcode));
```
devonbush: I like having the ext service call on it's own line since that's doing most of the "work"
```suggestion
    List<ParticipantFile> files = participantFileExtService.list(participantUser, enrolleeShortcode)
    return ResponseEntity.ok(files);
```

api-participant/src/main/java/bio/terra/pearl/api/participant/service/file/ParticipantFileExtService.java
```diff
@@ -0,0 +1,100 @@
...
+      FileStorageBackendProvider fileStorageBackendProvider) {
+    this.participantFileService = participantFileService;
+    this.authUtilService = authUtilService;
+    this.fileStorageBackend = fileStorageBackendProvider.get();
+  }
+
+  public ParticipantFile get(
+      ParticipantUser participantUser, String enrolleeShortcode, String fileName) {
+    Enrollee enrollee =
+        authUtilService.authParticipantUserToEnrollee(participantUser.getId(), enrolleeShortcode);
```
devonbush, line 45: We don't have permissions on the PArticipantApi, but we probably should eventually move to annotation-based auth like the admin tool so logic like this gets more foolproof and easier to test.  I'll make a ticket

api-participant/src/main/java/bio/terra/pearl/api/participant/service/file/ParticipantFileExtService.java
```diff
@@ -0,0 +1,100 @@
...
+      throw new RuntimeException("Error uploading file");
+    }
+  }
+
+  public List<ParticipantFile> list(ParticipantUser participantUser, String enrolleeShortcode) {
+    Enrollee enrollee =
+        authUtilService.authParticipantUserToEnrollee(participantUser.getId(), enrolleeShortcode);
+    return participantFileService.findByEnrolleeId(enrollee.getId());
+  }
+
```
devonbush, line 101: this method could use a comment and possibly a rename, since I at first thought it was doing something similar to the cleanFileName in SiteContentImage.

ui-participant/src/hub/documents/DocumentLibrary.tsx
```diff
@@ -0,0 +1,121 @@
...
+    </h1>
+    <div className="pb-4">
+      {i18n('documentsPageMessage')}
+    </div>
+    <h3>{i18n('documentsPageUploadedDocumentsTitle')} ({participantFiles.length})</h3>
+    <div className="d-flex flex-column">
+      { participantFiles.length > 0 && <table className="table">
+        <thead>
+          <tr>
+            <th>{i18n('documentsListFileName')}</th>
```
devonbush: UX nit -- I don't think we need the "File Name" header--it's obvious

devonbush (approved):
Looks great.  just some minor non-blocking UX and codesmithing

PR #1370, opened 2025-01-08
[JN-1552] import survey response times/status

connorlbark:

Allows importing response time and completion time.

1. Download synthetic AT data
[juniper_data_new.csv](https://github.com/user-attachments/files/18361603/juniper_data_new.csv)
2. Repopulate AT from dev configs
3. Import participants
4. Observe medical history, etc. all have dates in the past


core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
```diff
@@ -547,6 +548,18 @@ protected SurveyResponse importSurveyResponse(UUID portalId, SurveyFormatter for
...
+        // make sure the task reflects created and completion status
+        // so that recurrences are properly scheduled
+        String completedAtKey = formatter.getModuleName() + ExportFormatUtils.COLUMN_NAME_DELIMITER + "completedAt";
+        String createdAtKey = formatter.getModuleName() + ExportFormatUtils.COLUMN_NAME_DELIMITER + "createdAt";
+        if (enrolleeMap.containsKey(completedAtKey)) {
+            relatedTask.setCompletedAt(ExportFormatUtils.importInstant(enrolleeMap.get(completedAtKey)));
+        }
+        if (enrolleeMap.containsKey(createdAtKey)) {
+            relatedTask.setCreatedAt(ExportFormatUtils.importInstant(enrolleeMap.get(createdAtKey)));
+        }
```
connorlbark: I needed to modify the created/found task because tasks are not exported/imported - which makes sense but I did find mildly surprising. Especially with the complexity of recurring tasks, would it make sense to start exporting tasks separately from exporting survey responses?

This would also make this PR a bit simpler, because I had to add `completedAt` as a field on survey response because it isn't exported/imported by default. If tasks were a separate concept, then that extra linkage wouldn't be needed.

devonbush: I'd rather not have tasks as a separate export concept--I think that will make the export file harder to use for users.  I'd lean towards the solution that you have here, where the "surveyResponse" export columns are a combination of the surveyResponse and participantTask fields.  What this probably entails is creating a ResponseWithTask dto object, and then having the import/export operate on that object, rather than operating on just the surveyResponse object.  (this will also avoid having to make the change you did to SurveyResponse, which as you point out is confusing because it's unclear where/how that gets set)

devonbush: To be clear this means you'll update SurveyFormatter to be extends ModuleFormatter<SurveyResponseWithTask, ItemFormatter<SurveyResponseWithTask>> ...

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -296,5 +299,40 @@ protected BaseExporter getExporter(ExportFileFormat fileFormat, List<ModuleForma
...
+            }
+
+            for (SurveyResponse surveyResponse : surveyResponses) {
+                SurveyResponseWithTaskDto surveyResponseWithTaskDto = new SurveyResponseWithTaskDto();
+
+                try {
+                    BeanUtils.copyProperties(surveyResponseWithTaskDto, surveyResponse);
+                } catch (Exception e) {
+                    throw new IllegalStateException("Error copying properties from surveyResponse to surveyResponseWithTaskDto", e);
+                }
```
devonbush: How about making a SurveyResponseWithTaskDto constructor that takes a SurveyResponse and a ParticipantTask as arguments?  That will make the split between task-finding and dto assembly clearer

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -296,5 +299,40 @@ protected BaseExporter getExporter(ExportFileFormat fileFormat, List<ModuleForma
         return new TsvExporter(moduleFormatters, enrolleeMaps, fileFormat, columnSorting);
     }
 
+    private Map<UUID, List<SurveyResponseWithTaskDto>> attachTasksToSurveyResponses(Map<UUID, List<ParticipantTask>> taskMap, Map<UUID, List<SurveyResponse>> surveyResponseMap) {
+        Map<UUID, List<SurveyResponseWithTaskDto>> surveyResponseWithTaskMap = new HashMap<>();
+
+        for (UUID enrolleeId : surveyResponseMap.keySet()) {
+            List<ParticipantTask> tasks = taskMap.get(enrolleeId);
```
devonbush:
```suggestion
            List<ParticipantTask> tasks = taskMap.getOrDefault(enrolleeId, new ArrayList());
```

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -296,5 +299,40 @@ protected BaseExporter getExporter(ExportFileFormat fileFormat, List<ModuleForma
...
+            List<SurveyResponse> surveyResponses = surveyResponseMap.get(enrolleeId);
+            List<SurveyResponseWithTaskDto> surveyResponseWithTaskDtos = new ArrayList<>();
+
+            if (tasks == null) {
```
devonbush: this check can be removed with the above change

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
```diff
@@ -547,9 +548,42 @@ protected SurveyResponse importSurveyResponse(UUID portalId, SurveyFormatter for
...
                 "Imported", ppUser, enrollee, relatedTask.getId(), portalId).getResponse();
+
+
+        shiftTime(surveyResponse, relatedTask, formatter, enrolleeMap);
+
+        return surveyResponse;
+    }
+
+    // preserving response creation and task creation/completion
+    // is important for recurring surveys
```
devonbush, line 565: I like that you pulled this into a separate method

core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/SurveyFormatter.java
```diff
@@ -53,11 +51,15 @@ public SurveyFormatter(ExportOptions exportOptions,
...
-    protected List<ItemFormatter<SurveyResponse>> generateItemFormatters(ExportOptions options) {
+    protected List<ItemFormatter<SurveyResponseWithTaskDto>> generateItemFormatters(ExportOptions options) {
         // Note that we generate and add answer formatters to this list later in the constructor
-        List<ItemFormatter<SurveyResponse>> formatters = new ArrayList<>();
-        formatters.add(new PropertyItemFormatter<>("lastUpdatedAt", SurveyResponse.class));
-        formatters.add(new PropertyItemFormatter<>("complete", SurveyResponse.class));
+        List<ItemFormatter<SurveyResponseWithTaskDto>> formatters = new ArrayList<>();
+        formatters.add(new PropertyItemFormatter<>("lastUpdatedAt", SurveyResponseWithTaskDto.class));
+        formatters.add(new PropertyItemFormatter<>("createdAt", SurveyResponseWithTaskDto.class));
+        PropertyItemFormatter<SurveyResponseWithTaskDto> formatter = new PropertyItemFormatter<>("task.completedAt", SurveyResponseWithTaskDto.class, "completedAt");
```
devonbush:
```suggestion
        PropertyItemFormatter<SurveyResponseWithTaskDto> taskCompletedFormatter = new PropertyItemFormatter<>("task.completedAt", SurveyResponseWithTaskDto.class, "completedAt");
```

core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/SurveyFormatter.java
```diff
@@ -467,6 +469,10 @@ public SurveyResponse fromStringMap(UUID studyEnvironmentId, Map<String, String>
                 specifiedComplete = true;
             }
 
+            if (!itemFormatter.isImportable()) {
```
devonbush, line 474: this is a nice handling of this

devonbush (approved):
Nice -- a few code-smithing suggestions, but nothing blocking.

MatthewBemis (approved):

PR #1373, opened 2025-01-09
[JN-1551] cleanup issues with multiple 'other' on checkboxes

connorlbark:

Since the other questions weren't "real" questions they weren't showing up in the enrollee overview page for surveys. Since we relied on full surveyjs questions, I refactored it to use a more condensed type of only the things we need to know - this has the advantage of a lot less ugly casting and typescript workarounds across the page.

Also, variables (profile.givenName, etc.) weren't being rendered

image

- Fill out mass survey, filling out some other questions for schools
- Observe that they show up on the enrollee overview page (with history)


devonbush (approved):
nice

MatthewBemis (approved):

PR #1375, opened 2025-01-09
[JN-1565] backend for custom language texts

connorlbark:

Adds the backend needed to support generic custom translations for any system texts. There's no frontend for this yet - I considered doing the minimum I need to support A-T's needs, but I figure this will open up a ton of doors (most notably, future studies can add support for new languages in juniper without us paying for it / making changes to global text that other studies rely on). So, I figure it's worth it to spend a day making a generic interface for system translation overrides.

- repopulate demo
- ensure that it says `Join Us` in sandbox and irb environments but still says `Register` in the live environment
- publish site content irb->live
- make sure it now says `Join Us` on the live environment as well
- General smoke test of i18n because a lot got refactored - e.g. survey titles still translated, etc.


devonbush (changes requested):
This is good, but I think we want a more general solution.  What I think we want is for LocalSiteContent to have a list of LanguageText "translationOverrides" that handle changing any of the incidental text.  And then we would just change the i18n helper function to look for those translation overrides. That will save us having to model every bit of incidental text in schema, and plug into our existing LanguageText system.

That said, if you think that will take too long to implement, I'd be fine keeping this just to get A-T online

devonbush:
Related, we originally put portalId on LanguageText with the idea of supporting this.  I'm realizing now that that won't work given our desire to support publishing flows.  So I would lean towards removing the portal_id column from languageText, and replacing it with a local_site_content_id column

connorlbark:
> This is good, but I think we want a more general solution. What I think we want is for LocalSiteContent to have a list of LanguageText "translationOverrides"

Ooo I really like this solution. I'll put it into draft and do this.

MatthewBemis (approved):
nice, this will be incredibly useful! I didn't notice any issues with i18n when testing it out 👍

core/src/main/java/bio/terra/pearl/core/dao/i18n/LanguageTextDao.java
```diff
@@ -29,14 +29,26 @@ public void deleteByKeyNameAndPortal(String keyName, UUID portalId) {
...
                 handle
                     .createQuery(
-                        "SELECT * FROM language_text WHERE (portal_id = :portalId OR portal_id IS NULL) AND language = :language")
-                    .bind("portalId", portalId)
+                            // distinct on key_name and order by localized_site_content_id is used to ensure that
+                            // if there are conflicting keys, the one with localized_site_content_id will be used
+                            """
+                                    SELECT DISTINCT ON(lt.key_name) lt.* FROM language_text lt
+                                    LEFT JOIN localized_site_content lsc ON lsc.id = lt.localized_site_content_id
+                                    LEFT JOIN site_content sc ON sc.id = lsc.site_content_id
```
devonbush: I don't think you need/want either of these joins.  I think you made the correct decision to use subqueries instead

core/src/main/java/bio/terra/pearl/core/dao/i18n/LanguageTextDao.java
```diff
@@ -29,14 +29,26 @@ public void deleteByKeyNameAndPortal(String keyName, UUID portalId) {
...
+                                    WHERE (
+                                        sc.id = (select site_content_id from portal_environment where id = :portalEnvId)
+                                        OR lt.portal_id = (select portal_id from portal_environment where id = :portalEnvId)
+                                        OR (lt.localized_site_content_id IS NULL and lt.portal_id IS NULL)
+                                    ) AND lt.language = :language
+                                    ORDER BY lt.key_name, lt.localized_site_content_id ASC
```
devonbush: I think you might also want to add `lt.portal_id` to this sort just to make sure those override defaults (although for now they're just survey names, so they won't)

core/src/main/java/bio/terra/pearl/core/service/site/LocalizedSiteContentService.java
```diff
@@ -60,6 +64,16 @@ public LocalizedSiteContent create(LocalizedSiteContent localSite) {
             NavbarItem savedItem = navbarItemService.create(navItem);
             savedSite.getNavbarItems().add(savedItem);
         }
+
+        for (int i = 0; i < localSite.getLanguageTextOverrides().size(); i++) {
```
devonbush: I don't think you need the index, so you can just use the regualr iteration
```suggestion
        for (LanguageText langText : localSite.getLanguageTextOverrides()) {
```

devonbush: It would be nice if these LanguageTexts also had the portalId set on them.  Not sure what's the best way to do that given that the LocalizedSiteContent doesn't have a portal Id to reference.

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyParseUtils.java
```diff
@@ -1,6 +1,7 @@
 package bio.terra.pearl.core.service.survey;
 
 import bio.terra.pearl.core.model.i18n.LanguageText;
+import bio.terra.pearl.core.model.site.LocalizedSiteContent;
```
devonbush: remove?

populate/src/test/java/bio/terra/pearl/populate/PopulateDemoTest.java
```diff
@@ -271,7 +271,7 @@ private void checkSurveyPopulation(UUID studyEnvironmentId) {
         List<StudyEnvironmentSurvey> socialHealthSurveys = studyEnvironmentSurveys.stream().filter((ses) -> ses.getSurvey().getStableId().equals("hd_hd_socialHealth")).toList();
         // check that the past versions of the social health survey are included
         assertThat(socialHealthSurveys, hasSize(3));
-        assertThat(socialHealthSurveys.stream().map(ses -> ses.getSurvey().getVersion()).toList(), contains(1, 2, 3));
+        assertThat(socialHealthSurveys.stream().map(ses -> ses.getSurvey().getVersion()).collect(Collectors.toSet()), contains(1, 2, 3));
```
devonbush, line 274: why was this necessary?  if the findSurveys is returning duplicates, we might have deeper problems

devonbush (commented):
looks good, a bunch of nice-to-have codesmithing changes

devonbush (approved):

populate/src/test/java/bio/terra/pearl/populate/PopulateDemoTest.java
connorlbark: I ran into a random test failure where it returned 3, 2, 1. It happened just once, though, 🤷

core/src/main/java/bio/terra/pearl/core/service/site/LocalizedSiteContentService.java
connorlbark: I'm not sure the cleanest way to do this and I also worry it muddies up the model because right now, if portal is specified, then that language text is always global to the portal, whereas if we specified both then the SQL would have to do a bunch of null/not null checks. Maybe that's the correct thing to do, but i do like the simplicity of them being "attached" to the portal in two different ways - globally or locally.

PR #1377, opened 2025-01-09
[JN-1588] Defer to DSM as source of truth for kit mf barcodes

MatthewBemis:

Resets the kit label field to match whatever DSM returns during kit status refresh.

Needed for OurHealth- they had 4 kits collected with "n/a" kit labels. Sampath set these to the correct values in DSM so GP can scan them is, and this adds logic to automatically pick those changes up.

It's unclear to me what workflow they're using and how they got into this state. Would be curious to follow up with them.

Enable live DSM for OurHealth sandbox
Assign a new in-person kit and collect it (collecting it is the step that creates the kit request in DSM)
Update the kit label to be `n/a` i.e. `update kit_request set kit_label='n/a' where id='...';`
Refresh kit statuses for OurHealth sandbox
Confirm kit label is set back to what it originally was


connorlbark (approved):
Looking forward to direct GP integration! 😄

devonbush (approved):

PR #1383, opened 2025-01-13
[JN-1568] import survey response history

connorlbark:

After this, I'll fix up translate.py to copy over response history from pepper.

- I don't have any example imports, that will come with the followup translate.py fixups, but for now check out the unit tests and make sure no regressions have occurred in the import process


core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
```diff
@@ -524,14 +525,26 @@ protected SurveyResponse importPreEnrollResponse(Survey preEnroll,
...
+            SurveyResponseWithTaskDto response = responses.get(i);
+            imported.add(importSurveyResponse(ppUser, enrollee, studyEnv, formatter, response, portalId, auditInfo, enrolleeMap, i + 1));
+        }
+
+        return imported;
+    }
+
+    private SurveyResponse importSurveyResponse(PortalParticipantUser ppUser, Enrollee enrollee, StudyEnvironment studyEnv, SurveyFormatter formatter, SurveyResponseWithTaskDto response, UUID portalId, DataAuditInfo auditInfo, Map<String, String> enrolleeMap, Integer repeatNum) {
+        // if its the first repeat, attempt to find existing task to update. other repeats are previous responses and need their own task
+        ParticipantTask relatedTask = repeatNum == 1 ? participantTaskService.findTaskForActivity(ppUser.getId(), studyEnv.getId(), formatter.getModuleName())
```
connorlbark: I am not sure if this is the correct, desired logic.

Currently, survey response import can also function to update existing responses, i.e. it is idempotent. However, if we import response history, this gets tricky - there would be multiple tasks in the database with no clear way to get the one you want to update.

So, this ternary right here makes survey response no longer idempotent - if you import the same participant multiple times, they will get an increasingly large number of new surveyresponse/tasks.

I think this is OK, though.

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
```diff
@@ -524,14 +525,26 @@ protected SurveyResponse importPreEnrollResponse(Survey preEnroll,
         return created;
     }
 
-    protected SurveyResponse importSurveyResponse(UUID portalId, SurveyFormatter formatter, Map<String, String> enrolleeMap, ExportOptions exportOptions,
+    protected List<SurveyResponse> importSurveyResponses(UUID portalId, SurveyFormatter formatter, Map<String, String> enrolleeMap, ExportOptions exportOptions,
                                                   StudyEnvironment studyEnv, PortalParticipantUser ppUser, Enrollee enrollee, DataAuditInfo auditInfo) {
-        SurveyResponse response = formatter.fromStringMap(studyEnv.getId(), enrolleeMap);
-        if (response == null) {
+        List<SurveyResponseWithTaskDto> responses = formatter.listFromStringMap(studyEnv.getId(), enrolleeMap);
+        if (responses == null || responses.isEmpty()) {
```
devonbush, line 531: I'd return an empty list here

core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/BeanModuleFormatter.java
```diff
@@ -40,7 +39,7 @@ public Map<String, String> toStringMap(EnrolleeExportData enrolleeExportData) {
     }
 
     @Override
-    public T fromStringMap(UUID studyEnvironmentId, Map<String, String> enrolleeMap) {
+    public T fromStringMap(UUID studyEnvironmentId, Map<String, String> enrolleeMap, int moduleRepeatNum) {
         T bean = newBean();
         for (PropertyItemFormatter<T> itemInfo : getItemFormatters()) {
             String columnName = getColumnKey(itemInfo, false, null, 1);
```
devonbush:
```suggestion
            String columnName = getColumnKey(itemInfo, false, null, moduleRepeatNum);
```

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
devonbush: Hmm... making this non-idempotent means fixing up an existing import will be a lot harder.  Can the logic be something like:
```
if (repeatNum == 1 && the response doesn't have an explicit completed at date ) {
 auto-match with findTaskForActivity
} else if (response has an explicit completedAt date) {
 match task based on stableId and completedAt
} else {
  find all the tasks for that stableId sorted by date, and take the repeatNumTh task.  If there isn't one, create a new one.
}

**Review by devonbush — CHANGES_REQUESTED** [review]
Nice, I think you're right that the task-matching logic needs to be upgraded.  General approach looks great, thanks for coding this up

File: core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java (hunk 1 above)
Comment (connorlbark) [reply core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java]: I ended up settling for something a bit simpler:


`if repeatNum=1 { get latest existing task }`
- I can't match according to the completion date because it was really hard to dedupe the situation where on creation a participant gets auto assigned a task. When we import responses the auto created task will have no completion date,  so if we match only according to explicit created at, there will _always_ be an extra uncompleted task leftover on import of auto assigned surveys.

`else {require a completion date; find task with given completion date}`
- date matching is fuzzy within 24 hours

`if task is still null, then create new task`

File: core/src/main/java/bio/terra/pearl/core/dao/workflow/ParticipantTaskDao.java [hunk 4]
```diff
@@ -71,6 +72,45 @@ public Optional<ParticipantTask> findTaskForActivity(UUID ppUserId, UUID studyEn
...
+
+    public Optional<ParticipantTask> findTaskForActivityWithCompletionTime(UUID ppUserId, UUID studyEnvironmentId, String activityStableId, Instant completedAt) {
+        return jdbi.withHandle(handle ->
+                // Note: 43200 seconds is 12 hours. Any task that was completed within 12 hours of the given time is considered a match
+                handle.createQuery("""
+                                select * from %s
+                                where portal_participant_user_id = :ppUserId
+                                and target_stable_id = :activityStableId
+                                and study_environment_id = :studyEnvironmentId
+                                and ABS(EXTRACT(EPOCH FROM completed_at) - EXTRACT(EPOCH FROM :completedAt::timestamp)) < 43200
```
devonbush: can we pull this bit of query string into a helper function?  I suspect we'll want "queryStringForApproximateDateMatch()" in some other places too

devonbush (approved):
nice, appreciate the attention to detail in the logic.  One codesmithing suggestion

MatthewBemis (approved):

PR #1396, opened 2025-01-15
[JN-1527] adding timezone to study/export

devonbush:

As usual, anything involving time takes about 100 more lines of code than expected.  This adds a "time zone" to study environments (defaulting to US/New York) that gets used by default when exporting.  

image

For now, I haven't added the option to configure this specifically in the export UX, because we want to encourage consistent timezone usage, and so it should be at the study level.  

1. redeploy apiAdminApp
2. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants
3. add a new synthetic participant of "consented" type
4. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/export/dataBrowser?showPreview=true
5. confirm the new enrollee shows up with a createdAt time that matches your local time (assuming you are in EDT)

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
@@ -10,6 +10,7 @@ import bio.terra.pearl.core.model.kit.KitRequest;
 import bio.terra.pearl.core.model.kit.KitType;
 import bio.terra.pearl.core.model.participant.*;
 import bio.terra.pearl.core.model.study.StudyEnvironment;
+import bio.terra.pearl.core.model.study.StudyEnvironmentConfig;
 import bio.terra.pearl.core.model.survey.Survey;
 import bio.terra.pearl.core.model.survey.SurveyResponse;
 import bio.terra.pearl.core.model.workflow.HubResponse;
@@ -24,6 +25,7 @@ import bio.terra.pearl.core.service.kit.KitRequestDto;
 import bio.terra.pearl.core.service.kit.KitRequestService;
 import bio.terra.pearl.core.service.participant.*;
 import bio.terra.pearl.core.service.portal.PortalService;
+import bio.terra.pearl.core.service.study.StudyEnvironmentConfigService;
 import bio.terra.pearl.core.service.survey.AnswerProcessingService;
 import bio.terra.pearl.core.service.survey.SurveyResponseService;
 import bio.terra.pearl.core.service.survey.SurveyService;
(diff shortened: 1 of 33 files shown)


MatthewBemis (approved):
nice solution!

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
```diff
@@ -146,6 +151,8 @@ public Import importEnrollees(String portalShortcode, String studyShortcode, Stu
         if (ImportFileFormat.CSV.equals(fileFormat)) {
             exportOptions = IMPORT_OPTIONS_CSV;
         }
+        StudyEnvironmentConfig studyEnvConfig = studyEnvironmentConfigService.findByStudyEnvironmentId(studyEnv.getId());
+        exportOptions.setTimeZone(studyEnvConfig.getTimeZone());
```
connorlbark, line 155: I'm not entirely sure if I feel this is correct. I like that you can change timezone for export, but it would be cool if import prompted you to specify what timezone incoming dates are e.g. with A-T using UTC, and perhaps only take the default if timeZone isn't specified.
```suggestion
        if (exportOptions.getTimeZone() == null) {
                StudyEnvironmentConfig studyEnvConfig = studyEnvironmentConfigService.findByStudyEnvironmentId(studyEnv.getId());
                exportOptions.setTimeZone(studyEnvConfig.getTimeZone());
        }
```

connorlbark (approved):
Nice! This is awesome and I love that you can globally configure it.

It also feels like this is something we should export, but given that every row would have the same value, it would be kind of weird to add a `studyConfig.timeZone` attribute. That being said, I think it would be cool to be able, on import, to specify how to interpret dates - if we had just a `studyConfig.timeZone` it would make it trivial to then import A-T timezones properly (which are UTC). And I could see us running into situations where we don't know what timezone we exported with, so having that extra column to verify could be nice. Idk, just a thought!

core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeImportService.java
devonbush: Yup that idea definitely makes sense, it would be nice to be able to say "this import is in a specific timezone".  Was just beyond the scope of this feature.  

What we'll want to do for A-T is set their study timezone to UTC before we import anything, and then everything will work, and their exports will be consistent with their past exports.

devonbush:
For documenting the timezone -- that's a great point -- I'll make a ticket to update the data dictionary (I think you're right that it doesn't belong in every row of the export data.  That's metadata that would be useful there.

PR #1401, opened 2025-01-17
[JN-1560] fix missing trcc export fields

connorlbark:

We weren't actually getting the latest survey question definition, and earlier versions of these questions were `html` so it was skipping them on export.

- import what trcc is on prod
- go to sandbox data export
- observe that researchConsentForm.consentSignatureDate and researchAssentForm.assentSignatureDate show up

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/SurveyFormatter.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/SurveyFormatter.java
@@ -15,7 +15,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.BeanUtils;
 
-import java.time.ZoneId;
 import java.util.*;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


core/src/main/java/bio/terra/pearl/core/service/export/formatters/module/SurveyFormatter.java
```diff
@@ -77,7 +76,13 @@ private void generateAnswerItemFormatters(
         )).values().stream().sorted(Comparator.comparingInt(a -> a.get(0).getExportOrder())).toList();
 
         for (List<SurveyQuestionDefinition> questionVersions : questionDefsByStableId) {
-            SurveyQuestionDefinition mostRecent = questionVersions.get(0);
+
+            Optional<SurveyQuestionDefinition> mostRecentOpt = getLatest(questionVersions);
```
devonbush: Instead of doing this, I think it's cleaner to update the code above to sort the lists as they're made.  

I think you can pass a second argument to groupingBy of a collector that sorts the sublists. See https://stackoverflow.com/questions/35872236/sorting-lists-after-groupingby

devonbush (approved):
nice fix, one small suggestion

MatthewBemis (approved):

PR #1417, opened 2025-01-24
[JN-1581] Prevent duplicate kit labels during in-person kit assignment

MatthewBemis:

Prevents duplicate kit labels during in-person kit assignment

Re-start admin API
Go to https://localhost:3000/ourhealth/studies/ourheart/env/sandbox/kits/scan
Using manual kit override, assign a kit to OHSALK with label `foo`
Repeat that flow, and confirm you receive an error "Kit with label foo already exists"
Now confirm you can still collect the `foo` kit


connorlbark (approved):

devonbush (changes requested):
I like the explicit check.  But if our application depends on labels being unique, we should add that as a DB constraint too

MatthewBemis:
@devonbush I'll add the db constraint- we'll need to be mindful during the next release and make sure that no duplicate labels were created in the meantime, otherwise the Liquibase will fail. Small risk, but wanted to note it

devonbush (approved):

PR #1419, opened 2025-01-24
[JN-1571] adding custom columns to searches

devonbush:

Allows adding additional columns to both the participant search and kit request pages.
image

image

This also consolidates the search infrastructure between the kits and participant search pages.  It does this by adding a new "includes" property to SearchOptions.  This is handled separately from `include({someField})` logic in search expressions.  This seemed appropriate given that we're including entire collections, not just single fields.  Also it seemed healthier to split that collection retrieval processing away from the search/filter logic.  Adding all that collection retrieval to the searches would result in a ton of duplicate rows processed within the search query.

1.  redeploy apiAdminApp
2. repopulate demo
3. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/kits/eligible
4. Click "columns"
5. in the dropdown, click "answer.adminKits.kitNotes"
6. select ok
7. confirm the kitNotes column appears

8. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants
9. in "search filters" add a custom filter for answer.adminKits.kitNotes = 'yes'
10. confirm the filter works and the column appears in the search
image


core/src/main/java/bio/terra/pearl/core/model/search/EnrolleeSearchExpressionResult.java
```diff
@@ -24,7 +26,10 @@ public class EnrolleeSearchExpressionResult {
     private PortalParticipantUser portalParticipantUser;
     private MailingAddress mailingAddress;
     private final List<Answer> answers = new ArrayList<>();
+    @Builder.Default
     private final List<ParticipantTask> tasks = new ArrayList<>();
     private final List<Family> families = new ArrayList<>();
```
connorlbark: should answers and families also have a Builder.Default?

ui-admin/src/util/table/columnUtils.tsx
```diff
@@ -0,0 +1,181 @@
...
+import { checkboxColumnCell } from './tableUtils'
+import { instantToDefaultString } from '@juniper/ui-core'
+import _startCase from 'lodash/startCase'
+
+
+/**
+ * adapted from https://tanstack.com/table/v8/docs/examples/react/column-visibility
+ * For now, this control assumes that all the headers are simple strings.
+ * */
+export function ColumnVisibilityControl<T>({ table, dynamicColOpts }: {table: Table<T>,
```
MatthewBemis, line 20: nice, thanks for breaking this out into its own util file

ui-admin/src/study/participants/enrolleeView/ParticipantNotesView.tsx
```diff
@@ -9,7 +9,7 @@ import {
   ParticipantTask
 } from '@juniper/ui-core'
 import { ParticipantNoteModal } from './ParticipantNoteModal'
-import { renderEmptyMessage } from 'util/tableUtils'
+import { renderEmptyMessage } from '../../../util/table/tableUtils'
```
MatthewBemis: looks like a bunch of these imports got rewritten as part of the refactor.

definitely not a blocker- i'll leave it up to you if you want to clean them up

core/src/main/java/bio/terra/pearl/core/dao/search/EnrolleeSearchExpressionDao.java
```diff
@@ -85,6 +92,31 @@ public List<EnrolleeSearchExpressionResult> executeSearch(EnrolleeSearchQueryBui
...
+                    attachTasks(searchResult);
+                }
+            }
+        }
+
+        return searchResult;
+    }
+
+    private void attachKitRequests(List<EnrolleeSearchExpressionResult> result) {
+        Map<UUID, List<KitRequestDto>> kitsByEnrolleeId = kitRequestService.findByEnrollees(result.stream().map(EnrolleeSearchExpressionResult::getEnrollee).toList());
```
connorlbark, line 109: should this use the `kitRequestDao` since we're in a dao?

core/src/main/java/bio/terra/pearl/core/service/search/expressions/IncludeExpression.java
```diff
@@ -37,8 +37,9 @@ public EnrolleeSearchQueryBuilder generateQueryBuilder(UUID studyEnvId) {
         EnrolleeSearchQueryBuilder enrolleeSearchQueryBuilder = new EnrolleeSearchQueryBuilder(enrolleeDao, profileDao, studyEnvId);
 
         // Add the join, select, and condition clauses from the inner expression
-        // to the query builder.
-        inner.requiredJoinClauses().forEach(enrolleeSearchQueryBuilder::addJoinClause);
+        // to the query builder, but add the joins as left joins.
+        inner.requiredJoinClauses().forEach(join -> enrolleeSearchQueryBuilder.addJoinClause(
+                new EnrolleeSearchQueryBuilder.JoinClause(join, "left")));
```
connorlbark, line 42: So this is overriding all join clauses to always have left joins. I think this is OK but we should remove `joinType` from `joinClause` in this case since it's always being overridden to `left` anyway

core/src/main/java/bio/terra/pearl/core/service/search/terms/AnswerTerm.java
```diff
@@ -69,16 +69,25 @@ public List<EnrolleeSearchQueryBuilder.JoinClause> requiredJoinClauses() {
                     .joinClausesForStudy(studyName);
 
             joinClauses.add(
-                    new EnrolleeSearchQueryBuilder.JoinClause("answer", alias(), "%s.id = %s.enrollee_id".formatted(
+                    new EnrolleeSearchQueryBuilder.JoinClause("answer", alias(), """
+                            %s.id = %s.enrollee_id \
+                            and %s.survey_stable_id = '%s' \
+                            and %s.question_stable_id = '%s'\
```
connorlbark, line 76: should we leave a comment warning that it's ONLY ok that we did this because of the fact that we've sanitized survey_stable_id and question_stable_id to being alphanumeric? This would otherwise introduce possibility of SQL injection.

MatthewBemis (approved):
nice, worked great. i really like that the facets are persisted in the url params too, it's a great UX

ui-admin/src/util/participantSearchUtils.tsx
```diff
@@ -39,42 +51,65 @@ export const ParticipantSearchStateLabels: { [key in keyof ParticipantSearchStat
...
       ...searchState,
       [field]: value
     })
   }
+  return { searchState, searchExpression, updateSearchState, setSearchState }
+}
 
+/** handles loading and caching of search facets.  Returns a parsed version of the searchState identifying
+ * any included facets */
+export const useParticipantSearchFacets = (searchState: ParticipantSearchState, studyEnvParams: StudyEnvParams) => {
```
connorlbark, line 89: I found this hard to follow until I did some digging, both because I think I always find the term `facet` confusing and moreso that I struggled to follow the fact that there are now _two_ copies of participant search state, regular search state and faceted search state. I didn't find it immediately obvious what the difference was between the two types. IMO I think it would be clearer if we did:

```
const customQueryTypeDefs = findTypeDefsFromExpression(typeDefs: KeyedSearchValueTypeDef[], searchExp: string)
const includeTypeDefs = findTypeDefsForIncludeKeys(searchFacets: KeyedSearchValueTypeDef[], includeKeys: string[])
```

then maybe it was like

```
const {
    customQueryTypeDefs,
    includeTypeDefs
} = useParseCustomSearchStateTypeDefs(searchState)
```

That way it's clear that it's giving us type definitions based on the custom state and it's separate from the source of truth of the search state. I think this would clean up a lot of the typing, too.

connorlbark (approved):
Works great! Approving because I trust your choices and it works great as-is, but I do find the frontend typing/data flow a bit confusing

core/src/main/java/bio/terra/pearl/core/dao/search/EnrolleeSearchExpressionDao.java
devonbush: this confused me too at first.  But we need KitRequestService because we need kitRequestDtos, not KitRequests.  And converting to the DTO is a service-level concern

core/src/main/java/bio/terra/pearl/core/service/search/expressions/IncludeExpression.java
devonbush: is it?  there are a few other places (e.g. EnrolleeTermComparisonFacet) that don't

ui-admin/src/util/participantSearchUtils.tsx
devonbush: hmmm... I agree that it's tricky.  The hard part is we need a nice way to handle facets so the page isn't blocked on loading all the facet definitions, and also a unified state object so that it's clear they depend on each other.  And I thought it was useful to keep them separate since it made isolated testing easier.  But I'll have another go at consolidating them into a single hook

connorlbark: I see what you mean. I think, upon reflection, I don't mind this solution, but I think `Faceted` threw me off. Perhaps `SearchStateWithTypeDefs` would help clarify what the difference is - and that the added "type defs" are not a _part_ of the search state, just added metadata.

core/src/main/java/bio/terra/pearl/core/service/search/expressions/IncludeExpression.java
connorlbark: Oops yeah, in my head those were also left joins like this, but I realize that they can't always be left joins. Nevermind

ui-admin/src/util/participantSearchUtils.tsx
devonbush: ok, just pushed it up with a consolidated hook.  still not perfect, but better than before.

PR #1421, opened 2025-01-27
[JN-1547] allow ternary in email templates

connorlbark:

Allow ternary e.g. `{isSubject ? "your" : "your child's"}` in email templates. Not super robust (doesn't allow escaping quotes, doesn't allow nesting) but should work for 99.9% of use cases.

- Sign up a non-proxy user; observe you get an email that says "your signed consent" when you consent
- Sign up a proxy user; observe you get an email that says "your child's signed consent" when you consent


core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
```diff
@@ -45,6 +45,12 @@ protected EnrolleeEmailSubstitutor(EnrolleeContext ruleData, NotificationContext
         valueMap.put("siteMediaBaseUrl", getImageBaseUrl(contextInfo.portalEnv(), contextInfo.portalEnvConfig(), contextInfo.portal().getShortcode()));
         valueMap.put("siteImageBaseUrl", getImageBaseUrl(contextInfo.portalEnv(), contextInfo.portalEnvConfig(), contextInfo.portal().getShortcode()));
         valueMap.put("profile", enrolleeContext.getProfile());
+        if (enrolleeContext.getParticipantUser().getUsername().contains("-prox-")) {
```
connorlbark: I couldn't find another way to glean this information from the enrollee context. I think this is probably OK but obviously not foolproof, so let me know if there's a better way to do this.

As an aside, seeing this (+ Matt's participant document PR where he does some profileId checks) makes me think we should have some kind of `EnrolleeAccountType` enum where the values are something like `PROXY_ONLY` `GOVERNED_USER` `SELF_ENROLL` and `SELF_AND_PROXY`

core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
```diff
@@ -45,6 +46,13 @@ protected EnrolleeEmailSubstitutor(EnrolleeContext ruleData, NotificationContext
         valueMap.put("siteMediaBaseUrl", getImageBaseUrl(contextInfo.portalEnv(), contextInfo.portalEnvConfig(), contextInfo.portal().getShortcode()));
         valueMap.put("siteImageBaseUrl", getImageBaseUrl(contextInfo.portalEnv(), contextInfo.portalEnvConfig(), contextInfo.portal().getShortcode()));
         valueMap.put("profile", enrolleeContext.getProfile());
+        if (Objects.nonNull(enrolleeContext.getParticipantUser())
+                && StringUtils.contains(enrolleeContext.getParticipantUser().getUsername(), "-prox-")) {
```
devonbush: we'll want to make this more robust -- this might not work in the context of age-up protocols, there may be a period of time when they have their real email address as their account name but are still proxied. 

I'd add an "isGoverned" field to EnrolleeContext, and a query in EnrolleeContextService to populate it

core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
```diff
@@ -45,6 +46,13 @@ protected EnrolleeEmailSubstitutor(EnrolleeContext ruleData, NotificationContext
...
+            valueMap.put("isProxy", "true");
+        } else {
+            valueMap.put("isSubject", "true");
```
devonbush: let's go ahead and just put the whole `enrolleee` on the valueMap -- that way we have `subject` and `consented` available

core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
```diff
@@ -80,6 +90,37 @@ public String lookup(String key) {
...
+    private record Ternary(String condition, String left, String right) {
+        public static Ternary fromString(String key) {
+            String[] parts = key.split("\\?");
+            String condition = parts[0];
+            String[] values = parts[1].split(":");
+            return new Ternary(condition.trim(), values[0].trim(), values[1].trim());
+        }
+
+        public static boolean isTernary(String key) {
+            return key.contains("?") && key.contains(":") && key.indexOf('?') < key.indexOf(':');
```
devonbush: Might be just my personal preference, but I lean towards (commented) regexes for stuff like this -- something like`/[^:]+\?.+:`

core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
```diff
@@ -80,6 +90,37 @@ public String lookup(String key) {
         return "";
     }
 
+    private record Ternary(String condition, String left, String right) {
```
devonbush, line 105: Nice, I like how you made this as a record -- good encapsulation of logic

devonbush (changes requested):
Love the implementation of the ternaries -- just needs some tweaks to handling the isProxy/Governed logic

core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
connorlbark: I found it simpler to just fetch all the enrollee relations and populate them - perhaps a heavyweight solution, though

connorlbark:
test failure is a fluke and sonarcloud is flagging the regex which shouldn't be a security problem

core/src/main/java/bio/terra/pearl/core/dao/participant/EnrolleeRelationDao.java
```diff
@@ -113,4 +113,15 @@ public List<EnrolleeRelation> findAllFamilyRelationshipsByEitherEnrollee(UUID en
                         .list()
         );
     }
+
+    public List<EnrolleeRelation> findAllByEnrolleeOrTargetIds(List<UUID> enrolleeIds) {
+        return jdbi.withHandle(handle ->
+                handle.createQuery(
+                                "SELECT * FROM enrollee_relation " +
```
devonbush, line 120: I just realized that we don't have indexes on those columns.  Our enrollee_relation table probably won't have over 1000 entries anytime soon, but if we're querying it every time we're evaluating rules (which is often) we should add an index those columns

devonbush (approved):
looks good, and yeah, I could see the decision to just fetch all the relations going either way.  Since we mostly have small N for relations, I like your choice to keep it simple and keep the proxy logic in java rather than in a query.

That said, please add an upgrade script to put indexes on the enrollee_id and target_enrollee_id columns -- that will help with a bunch of family API queries too

MatthewBemis (approved):
Looks great! I went into Sonar and marked the regex security hotspot as safe so that check should be green now ✅

PR #1429, opened 2025-01-27
[JN-1587] make instants more robust from search exp UI

connorlbark:

Search expression UI would create instants that wouldn't work with Instant.parse but would work with SQL, so sometimes they would work and sometimes they wouldn't. They also wouldn't provide any timezone. Now, they will provide ISO compliant dates with UTC timezone.

- play around with search exp UI for dates
- e.g. add a search exp to a survey using enrollee.createdAt with the basic UI (e.g. {createdAt} > yesterday) and make sure it works when you enroll a new user

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/search/terms/SearchTermParser.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/search/terms/SearchTermParser.java
@@ -45,7 +45,6 @@ public abstract class SearchTermParser<T extends SearchTerm> {
     protected abstract T parse(String arguments);
 
     /**
-     <<<<<<< HEAD
      * Parse the term string into a search term for the given study environment.
      */
     protected T parse(String study, String arguments) {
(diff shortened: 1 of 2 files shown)


devonbush (approved):

core/src/main/java/bio/terra/pearl/core/service/search/terms/SearchTermParser.java
```diff
@@ -45,26 +45,20 @@ public T parseVariable(String variable) {
...
-     <<<<<<< HEAD
      * Parse the term string into a search term for the given study environment.
      */
     protected T parse(String study, String arguments) {
         throw new IllegalArgumentException("The " + getTermName() + " term does not support searching across study environments.");
     }
 
     /**
-     =======
-     >>>>>>> origin/development
```
MatthewBemis, line 57: 😬

MatthewBemis (approved):

PR #1433, opened 2025-01-28
[JN-TBD] Refactoring pre-enroll surveys

devonbush:

Updates pre-enrollment surveys to be a survey type, rather than have a dedicated data model attribute.  this refactor will need to be done in two stages in order to avoid downtime. Stage 1 (this PR) migrates existing pre-enroll surveys and updates the frontend to read the migrated versions, but does not remove the data fields.  Stage 2 would remove the data field

The big advantage of this is an overall simplification of our data model, and corresponding improvements in the UX.  For example, this gives us the ability to publish pre-enroll surveys individually, which we did not have before.  the downside is that this no longer enforces that a study env can only have a single pre-enrollment survey, so we'll have to think about what that means.

Note that this temporarily punts the issue of being able to create multiple pre-enroll surveys in the admin tool.  We'll clean up the admin tool when we also remove all the old preEnrollSurveyId questions

1. redeploy apiParticipantApp
2. go to https://sandbox.demo.localhost:3001 
3. join, confirm pre-enroll processes correctly.
4. redeploy apiAdminApp
5. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms
6. confirm you can view/edit/publish the pre-enroll survey

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/study/StudyDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/study/StudyDao.java
@@ -84,9 +84,9 @@ public class StudyDao  extends BaseMutableJdbiDao<Study> {
      * So, for example, if a portal has two studies, this might return the 'sandbox' environment for
      * both studies
      */
-    public List<Study> findWithPreregContent(String portalShortcode, EnvironmentName envName) {
+    public List<Study> findWithPreEnrollContent(String portalShortcode, EnvironmentName envName) {
         List<Study> studies = findByPortal(portalShortcode);
-        List<StudyEnvironment> studyEnvs =  studyEnvironmentDao.findWithPreregContent(portalShortcode, envName);
+        List<StudyEnvironment> studyEnvs =  studyEnvironmentDao.findWithPreEnrollContent(portalShortcode, envName);
         for (Study study : studies) {
             Optional<StudyEnvironment> studyEnvOpt = studyEnvs.stream()
                     .filter(studyEnv -> studyEnv.getStudyId().equals(study.getId())).findFirst();
--- a/core/src/main/java/bio/terra/pearl/core/model/survey/SurveyType.java
+++ b/core/src/main/java/bio/terra/pearl/core/model/survey/SurveyType.java
@@ -1,6 +1,7 @@
 package bio.terra.pearl.core.model.survey;
 
 public enum SurveyType {
+    PRE_ENROLL, // for prenrollment surveys
     CONSENT, // for consent forms
     RESEARCH, // for surveys intended to be included in the research dataset of a study
     OUTREACH, // for surveys intended for outreach purposes, e.g. to collect marketing/feedback information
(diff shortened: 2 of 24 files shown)


core/src/main/java/bio/terra/pearl/core/model/survey/SurveyType.java
```diff
@@ -1,6 +1,7 @@
 package bio.terra.pearl.core.model.survey;
 
 public enum SurveyType {
+    PRE_ENROLL, // for prenrollment surveys
```
MatthewBemis, line 4: we'll need to add this as an allowed `SurveyType` in `forms.ts` too

core/src/main/java/bio/terra/pearl/core/dao/study/StudyDao.java
```diff
@@ -84,9 +84,9 @@ public Optional<Study> findOneFullLoad(UUID id) {
      * So, for example, if a portal has two studies, this might return the 'sandbox' environment for
      * both studies
      */
-    public List<Study> findWithPreregContent(String portalShortcode, EnvironmentName envName) {
+    public List<Study> findWithPreEnrollContent(String portalShortcode, EnvironmentName envName) {
```
MatthewBemis, line 87: Pre-reg is portal-scoped, pre-enroll is study scoped, right? So this was just renaming it to be correct?

(Just making sure I've got the terminology straight in my head- I mix these up a bunch)

ui-admin/src/study/StudyContent.tsx
```diff
@@ -85,45 +79,35 @@ function StudyContent({ studyEnvContext }: {studyEnvContext: StudyEnvContextT})
...
-                      </ul>
-                    </div>
-                  </div>}
-                </li>
-              </ul>}
-              {(!preEnrollSurvey && !isReadOnlyEnv) && <Button variant="secondary"
-                data-testid={'addPreEnroll'}
-                onClick={() => {
-                  setShowCreatePreEnrollModal(!showCreatePreEnrollSurveyModal)
+              <SurveyEnvironmentTable
```
MatthewBemis, line 92: For pre-enrolls, should we hide certain task-related functionality like "see participant assignment", "assign to {studyEnv} participants", and other settings like recurrence interval?

MatthewBemis (commented):
looking good! just have a few comments/questions to start

core/src/main/java/bio/terra/pearl/core/dao/study/StudyDao.java
devonbush: yes, this is renaming to be correct, not changing behavior

ui-admin/src/study/StudyContent.tsx
devonbush: yeah, right on.  We can tackle that in the subsequent schema and admin cleanup PR

MatthewBemis (approved):

connorlbark (commented):
works great! I love how little code it took to do this.

I did notice, though, it broke one aspect of proxy functionality. If you create a proxy user, click on the participant dropdown,t hen click `Add New Participant` it doesn't seem to laod the pre-enroll. I suspect it thinks that the pre-enroll was already completed for this user, so it dumps them back at the dashboard.

devonbush:
Thanks for flagging that proxy issue, connor.  will investigate

devonbush:
@connorlbark proxy add participant works for me locally.  I wonder if it's an issue with how the migration scripts/redeploys/UX reloads get sequenced in development environment.  Since we don't have any currently running proxy studies, I propose we merge this, and then I'll confirm the migrations worked on demo.

connorlbark (approved):

PR #1441, opened 2025-01-30
[JN-1598] Disable HeartDemo export integrations by default

MatthewBemis:

This is the source of a bunch of logspam in #juniper-dev-notifications

Repopulate heartdemo
Confirm export integrations are disabled

### DIFF
--- a/populate/src/main/resources/seed/portals/demo/studies/heartdemo/study.json
+++ b/populate/src/main/resources/seed/portals/demo/studies/heartdemo/study.json
@@ -176,7 +176,7 @@
         {
           "name": "Airtable participant table",
           "destinationType": "AIRTABLE",
-          "enabled": true,
+          "enabled": false,
           "exportOptions": {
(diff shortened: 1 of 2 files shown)


devonbush (approved):
thanks!

connorlbark (approved):

populate/src/main/resources/seed/portals/demo/studies/heartdemo/study.json
```diff
@@ -176,7 +176,7 @@
         {
           "name": "Airtable participant table",
           "destinationType": "AIRTABLE",
-          "enabled": true,
```
connorlbark, line 179: Ooo if we could also have the hostname changed to `juniperdemostudy.dev` by default that would be great as well, right now it has to be manually added every time.

MatthewBemis: good thinking! i'll add that

PR #1443, opened 2025-01-31
[JN-1600] customer-specific mixpanel export

devonbush:

This adds a mixpanel token to portal environment configs.  If set, and a custom domain is present, events in that domain will also be logged to the given token.

This also adds spring caching so that we don't have to load and assemble the config map every time an event is called.  this is probably overkill for our purposes, but I thought it was a good excuse to dip our toes into using spring caching.  Down the road, I could imagine us implementing a simple postgres-backed Spring cache manager so that we wouldn't have to do crude cache evictions based on timeouts.

This is tricky to test in dev...
1.  redeploy admin app, redeploy participantApiApp with MIXPANEL_ENABLED=true
3. go to https://localhost:3000/demo/studies/heartdemo/env/live/settings/website
4. set the participant hostname to "demo.localhost"
5. set the mixpanel token value to anything
6. save the configs
7. put a breakpoint in line 84 of MixpanelService
8. go to https://demo.localhost:3001/
9. after a few seconds, the breakpoint should be triggered, and you should see an event being created with your custom token

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/logging/LoggingConfigCache.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/logging/LoggingConfigCache.java
@@ -1,0 +1,40 @@
+package bio.terra.pearl.core.service.logging;
+
+import bio.terra.pearl.core.model.portal.PortalEnvironmentConfig;
+import bio.terra.pearl.core.service.portal.PortalEnvironmentConfigService;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.cache.annotation.CacheEvict;
+import org.springframework.cache.annotation.Cacheable;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Service;
+
+import java.util.Map;
+
+/**
+ * we don't want to have to go to the database every time an event is logged in our system to get the configs,
+ * and they change very rarely. So we cache them here
+ */
+@Service
+@Slf4j
+public class LoggingConfigCache {
+    private final PortalEnvironmentConfigService portalEnvironmentConfigService;
+
+    public static final String CONFIGS_WITH_DOMAIN_CACHE_KEY = "portalEnvironmentConfigsWithDomain";
+
+    public LoggingConfigCache(PortalEnvironmentConfigService portalEnvironmentConfigService) {
+        this.portalEnvironmentConfigService = portalEnvironmentConfigService;
+    }
+
+    @Cacheable(value = CONFIGS_WITH_DOMAIN_CACHE_KEY)
+    public Map<String, PortalEnvironmentConfig> getConfigsWithDomain() {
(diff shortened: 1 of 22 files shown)


core/src/main/java/bio/terra/pearl/core/service/logging/MixpanelService.java
```diff
@@ -47,59 +53,100 @@ public String filterEventData(String data) {
...
 
         ClientDelivery delivery = new ClientDelivery();
+        ClientDelivery customDelivery = null;
 
         for (int i = 0; i < events.length(); i++) {
-            JSONObject mixpanelEvent = buildEvent(events.getJSONObject(i));
+            JSONObject event = events.getJSONObject(i);
+            JSONObject mixpanelEvent = buildEvent(event, mixpanelConfig.token);
             delivery.addMessage(mixpanelEvent);
+            String eventDomain = getEventCurrentDomain(event);
```
connorlbark, line 74: Why not send the portal shortcode when tracking to avoid domain matching?

ui-admin/src/study/settings/portal/WebsiteSettings.tsx
```diff
@@ -13,30 +14,39 @@ export const WebsiteSettings = (
...
+      <label className="form-label mt-2">
         participant hostname
         <input type="text" className="form-control" value={config.participantHostname ?? ''}
           onChange={e => updateConfig('participantHostname', e.target.value)}/>
       </label>
     </div>
+    <div>
+      <label className="form-label mt-2">
+        Mixpanel api token <InfoPopup content={<span>
+        If provided, events for participant interactions with the website will
```
connorlbark, line 45: should we specify that the events contain no identifiable information?

connorlbark (approved):
Nice! I like that we still have our own mixpanel but our customers don't have to rely on us for insights.

core/src/main/java/bio/terra/pearl/core/service/logging/MixpanelService.java
devonbush: I suppose we could send portal shortcode and environment (we'd need the environment to not log sandbox events to their mixpanel)

core/src/main/java/bio/terra/pearl/core/service/logging/LoggingConfigCache.java
```diff
@@ -0,0 +1,40 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.cache.annotation.CacheEvict;
+import org.springframework.cache.annotation.Cacheable;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Service;
+
+import java.util.Map;
+
+/**
+ * we don't want to have to go to the database every time an event is logged in our system to get the configs,
```
MatthewBemis, line 14: great decision to cache this

MatthewBemis (approved):
nice! 🎉

PR #1452, opened 2025-02-03
[JN-1608][member's choice] editable sidebars

connorlbark:

Adds ability to customize what you see on your sidebar. Could be cool to also be able to drag n drop reorder but that's definitely a lot harder (but possible!)

Right now, this is just stored in local storage. However, I could imagine a world where this is sent to the backend as part of the user's settings just as a json blob that gets refreshed upon logging in.

Screenshot 2025-02-03 at 2 36 01 PM

Screenshot 2025-02-03 at 2 36 04 PM

- populate at least 2 studies
- edit sidebar (i.e. hide some stuff) in one study
- go to another study, observe stuff isn't hidden then
- hide some other stuff in the other study
- switch between them and observe that state is study-specific
- refresh, exit chrome, etc., ensure it remembers state

### DIFF
--- a/ui-admin/src/navbar/SidebarSection.tsx
+++ b/ui-admin/src/navbar/SidebarSection.tsx
@@ -1,0 +1,71 @@
+import { NavLink } from 'react-router-dom'
+import React from 'react'
+import { sidebarNavLinkClasses } from 'navbar/AdminSidebar'
+import CollapsableMenu from 'navbar/CollapsableMenu'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import {
+  faEye,
+  faEyeSlash
+} from '@fortawesome/free-solid-svg-icons'
+
+export type SidebarItemT = {
+  key: string
+  label: string
+  link: string
+}
+
+export type SidebarSectionT = {
+  key: string
+  label: string
+  items: SidebarItemT[]
+}
+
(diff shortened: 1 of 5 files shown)


devonbush (commented):
This is super cool.  And I think we might be able to use this functionality it in other portions of our app too.  But given that Juniper is going to be on a bit of a feature freeze soon, I'm leaning against merging this since it could lead to some user confusion where their screen doesn't match the documentation, or Whitney is confused why things look different in one study versus another (especially considering that menu items may also be hidden for permissions reasons)

  I'll be curious to get Whitney and Parker's take -- if they say they want it, I'm on board.  Otherwise maybe we we icebox this PR until the next time we add a new item to the sidebar?

ui-admin/src/navbar/SidebarSection.tsx
```diff
@@ -0,0 +1,72 @@
...
+  key: string
+  label: string
+  items: SidebarItemT[]
+}
+
+const navStyleFunc = ({ isActive }: { isActive: boolean }) => {
+  return isActive ? { background: 'rgba(255, 255, 255, 0.3)' } : {}
+}
+
+export const SidebarSection = ({
```
MatthewBemis, line 27: really nice refactor

MatthewBemis (approved):
Sorry the delay- I could have sworn I submitted the review for this a few days ago

devonbush (approved):

PR #1454, opened 2025-02-04
[JN-1543] Send emails to enrollees with tasks

connorlbark:

(or, how I learned to stop worrying about automation and learned to love ad-hoc notifications 😄 )

This is probably the simplest way to accomplish launch emails. Shoutout to Matt for coming up with the idea .

I could have done it without _any_ backend changes if I instead used the search results in the frontend to get the list of shortcodes instead of sending the search expression to the backend. But I figured the "send email by search expression" functionality was generally useful and easy to implement, so I went with that approach.

I don't know if this is really the right place to put the functionality, let me know if you think there's a better place.

Screenshot 2025-02-04 at 12 40 01 PM

Screenshot 2025-02-04 at 12 40 09 PM

- restart admin api
- go to flow page
- click dropdown next to lost interest
- send email to lost interest folks
- if local, check email to make sure you got email for expected 2 enrollees. in dev, check the history for the email you used to make sure it was only sent to 2 people.


ui-admin/src/study/notifications/SendTaskEmailModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+        triggerId: selectedConfig.id
+      })
+      Store.addNotification(successNotification('email processed'))
+    } catch (e) {
+      Store.addNotification(failureNotification('email processing failed'))
+    }
+    onClose()
+  }
+
+  const taskStatusOptions: { value: ParticipantTaskStatus, label: string}[] = [
```
devonbush: Let's consolidate this somewhere with the options in TaskStatusFacet

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/notifications/TriggerController.java
```diff
@@ -105,13 +106,26 @@ public ResponseEntity<Object> adHoc(
...
     AdHocNotification adHoc = objectMapper.convertValue(body, AdHocNotification.class);
     EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
-    Trigger configUsed =
-        notificationExtService.sendAdHoc(
-            PortalStudyEnvAuthContext.of(
-                adminUser, portalShortcode, studyShortcode, environmentName),
-            adHoc.enrolleeShortcodes,
-            adHoc.customMessages,
-            adHoc.triggerId);
+    Trigger configUsed;
```
devonbush, line 109: nice, I really like how you didn't create a new endpoint for this

ui-admin/src/study/notifications/SendTaskEmailModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+                .map(status =>
+                  taskStatusOptions
+                    .find(opt => opt.value === status))}
+              onChange={newVals => changeTaskStatus(
+                newVals
+                  ?.filter(newVal => !isNil(newVal))
+                  .map(newVal => newVal!.value))}/>
+          </label>
+        </div>
+
```
devonbush: this is a big batch of copy from AdHocEmailModal.   I'd just as soon centralize this into AdHocEmailModal by having it take an "assignedTaskStableId" option.  
That way it's clearer in code where all ad hoc emails come from, and later we could also build a task selector into that modal so you could also accomplish this directly from a general "send email to people" modal.

devonbush (changes requested):
LOVE this functionality.  I think you did an amazing job at planning this out and not unduly sticking to the first implementation either one of us thought of.  Just a couple of refactoring cleanups are needed

I think this dialog should also be linked from the Forms tables.  Should be very easy to add the link to the menu in SurveyTableEnvColumn in SurveyEnvironmentTable.tsx

One of the long term design goals for Juniper is thinking in terms of processing cohorts.  Both "I have a group of people, now I want to do stuff with them"  and "I'm doing something, now let me pick the group of people this applies to".  I like that this continues to advance our functionality in that regard.  As we continue to develop features like this, please keep your eyes peeled for opportunities to generalize and centralize this cohort handling

ui-admin/src/study/notifications/SendTaskEmailModal.tsx
connorlbark: Good idea - this reduced overall code. The modal is now a little bit on the complex side but the work to make it more modular seemed to be harder than would be worth it - if there end up being more recipient types, though, a refactor might be called for.

devonbush (approved):

MatthewBemis (approved):
nice!!

PR #1455, opened 2025-02-04
[JN-1591] Show responses associated with participant documents

MatthewBemis:
#### DESCRIPTION (include screenshots, and mobile screenshots for participant UX)\

screencapture-sandbox-demo-localhost-3001-hub-documents-2025-02-04-13_54_04

screencapture-localhost-3000-demo-studies-heartdemo-env-sandbox-participants-HDSALK-documents-2025-02-04-13_54_23

* Restart admin and participant API apps
* Log into sandbox demo as jsalk@test.com and go to the document hub: https://sandbox.demo.localhost:3001/hub/documents
* Note that medical_notes.txt is associated with the Medical Record Request task
* As an admin, go to "View all documents" for HDSALK and confirm that you can see the same associated task information: https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants/HDSALK

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/model/file/ParticipantFile.java
+++ b/core/src/main/java/bio/terra/pearl/core/model/file/ParticipantFile.java
@@ -1,11 +1,14 @@
 package bio.terra.pearl.core.model.file;
 
 import bio.terra.pearl.core.model.BaseEntity;
+import lombok.Builder;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
 import lombok.experimental.SuperBuilder;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.UUID;
 
 /**
@@ -28,4 +31,7 @@ public class ParticipantFile extends BaseEntity {
     private UUID enrolleeId;
 
     private String notes;
+
+    @Builder.Default
+    private List<UUID> surveyResponseIds = new ArrayList<>(); //list of survey responses that are associated with this file
 }
--- a/core/src/main/java/bio/terra/pearl/core/model/file/ParticipantFile.java
+++ b/core/src/main/java/bio/terra/pearl/core/model/file/ParticipantFile.java
@@ -1,6 +1,7 @@
 package bio.terra.pearl.core.model.file;
 
 import bio.terra.pearl.core.model.BaseEntity;
+import bio.terra.pearl.core.model.survey.Answer;
 import lombok.Builder;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
@@ -33,5 +34,5 @@ public class ParticipantFile extends BaseEntity {
     private String notes;
 
     @Builder.Default
-    private List<UUID> surveyResponseIds = new ArrayList<>(); //list of survey responses that are associated with this file
+    private List<Answer> associatedAnswers = new ArrayList<>(); //list of answers that are associated with this file
 }
--- a/ui-participant/src/hub/documents/DocumentLibrary.tsx
+++ b/ui-participant/src/hub/documents/DocumentLibrary.tsx
@@ -133,12 +133,12 @@ const surveyResponseIdsToTaskNames = (
   }).filter(task => task !== undefined)
 
   if (associatedTasks.length === 0) {
-    return null
+    return <div className={'mt-2 fst-italic'}>This document is not associated with any tasks</div>
   }
 
   return (
-    <div className={'mt-2'}>
-      <span className={'fw-medium'}>Survey Responses</span>
+    <div className={'mt-2 text-muted'}>
+      <span>Document shared in response to:</span>
       <ul>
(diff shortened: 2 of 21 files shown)


core/src/main/java/bio/terra/pearl/core/model/file/ParticipantFile.java
```diff
@@ -28,4 +31,7 @@ public class ParticipantFile extends BaseEntity {
     private UUID enrolleeId;
 
     private String notes;
+
```
devonbush, line 35: I understand the need to send this association to the frontend, but I worry this obfuscates the underlying data model of response -> answer -> file too much. 
How about we make this a `List<Answer>` instead?  then those answers will still have the SurveyResponse id you need, and you'll also get other juicy tidbits too and a reminder aof how it's actually attached

devonbush: Also, while you're at it, please add a comment on the `List<ParticipantFile> `on SurveyResponse talking about how that relation is actually stored on the answers

devonbush: I think that will make the fetch logic simpler too -- instead of a nested for match and a new dao method in SurveyResponse, you'll just need to create a `findWithAnswers` method in ParticipantFileDao

ui-participant/src/hub/documents/DocumentLibrary.tsx
```diff
@@ -126,6 +124,34 @@ const DocumentsList = ({ studyName, studyEnvParams, enrollee }: {
...
+    return enrollee.participantTasks.find(task => task.surveyResponseId === surveyResponseId)
+  }).filter((task): task is ParticipantTask => task !== undefined)
+
+  if (associatedTasks.length === 0) {
+    return <div className={'mt-2 fst-italic'}>This document is not associated with any tasks</div>
+  }
+
+  return (
+    <div className={'mt-2 text-muted'}>
+      <span>Document shared in response to:</span>
```
devonbush:
```suggestion
      <span>shared in response to:</span>
```

ui-participant/src/hub/documents/DocumentLibrary.tsx
```diff
@@ -126,6 +124,34 @@ const DocumentsList = ({ studyName, studyEnvParams, enrollee }: {
...
+const surveyResponseIdsToTaskNames = (
+  i18n: (key: string, options?: I18nOptions) => string, studyEnvParams: StudyEnvParams,
+  enrollee: Enrollee, surveyResponseIds: string[]
+) => {
+  const associatedTasks = surveyResponseIds.map(surveyResponseId => {
+    return enrollee.participantTasks.find(task => task.surveyResponseId === surveyResponseId)
+  }).filter((task): task is ParticipantTask => task !== undefined)
+
+  if (associatedTasks.length === 0) {
+    return <div className={'mt-2 fst-italic'}>This document is not associated with any tasks</div>
```
devonbush:
```suggestion
    return <div className={'mt-2 fst-italic'}>not associated with any tasks</div>
```

devonbush (changes requested):
Like the UX a lot.  Just would like some tweaks so that I don't have to go down a schema rabbit hole reminding myself "wait, how are files and responses tied together again?" next time :)

ui-participant/src/hub/documents/DocumentLibrary.tsx
```diff
@@ -126,6 +129,34 @@ const DocumentsList = ({ studyName, studyEnvParams, enrollee }: {
...
+    return enrollee.participantTasks.find(task => task.surveyResponseId === surveyResponseId)
+  }).filter((task): task is ParticipantTask => task !== undefined)
+
+  if (associatedTasks.length === 0) {
+    return <div className={'mt-2 fst-italic text-muted'}>not associated with any tasks</div>
+  }
+
+  return (
+    <div className={'mt-2 text-muted'}>
+      <span>shared in response to:</span>
```
connorlbark, line 146: Do we use `task` in other places participant facing? It feels like admin language, but I could be wrong. IMO I would just return `<></>` and specify in the description up above that study staff can see these files even if they aren't associated with a survey.

Also, this needs to be i18n'd.

connorlbark (approved):
Looks great!

ui-participant/src/hub/documents/DocumentLibrary.tsx
MatthewBemis: Good point re: the language, I'll make that more participant friendly.

Regarding i18n- I'm planning to do the i18n at the very end after the UX for everything related to file upload is 100% solid

MatthewBemis: Updated the UX to remove the task-related language in https://github.com/broadinstitute/juniper/pull/1478

devonbush (approved):
nice!

PR #1457, opened 2025-02-04
[JN-1581] Speculative fix for DSM connection errors

MatthewBemis:

This is a speculative fixe based on discussion in https://github.com/reactor/reactor-netty/issues/1774

Unless there are objections to the actual code changes, I'd like to merge this ASAP because it's not possible to test if this actually fixes the problem locally, so we need to try it out on demo (and even then, it's difficult to predictably reproduce- but we can at least have some degree of confidence). This also gives us more than enough time to test before pushing to prod during next week's release.

Configure local OurHealth to use live DSM
Try sending a kit
Confirm it succeeds and there aren't any regressions requesting kits from DSM
Check DSM to be doubly-sure: https://dsm-dev.datadonationplatform.org/juniper-ourheart/shippingUploaded


devonbush (approved):
love it, thanks for digging in on this

connorlbark (approved):

PR #1461, opened 2025-02-07
[JN-1615] remove only include most recent

connorlbark:

Removes the button to only include most recent in export (which does not function and is non-trivial to fix). Export currently does the behavior that would be expected by A-T - which is to include all responses.


devonbush (approved):
nice, I like the idea of keeping it one the backend in case we want to support it later

MatthewBemis (approved):

PR #1476, opened 2025-02-11
[JN-1526] Add option to prepopulate longitudinal survey responses with previous answers

MatthewBemis:

Surfaces the prepopulate option on surveys if the recurrence type is longitudinal.

Screenshot 2025-02-11 at 12 17 58 PM

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcher.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcher.java
@@ -139,29 +139,33 @@ public class SurveyTaskDispatcher extends TaskDispatcher<SurveyTaskConfigDto> {
     public void copyTaskData(ParticipantTask newTask, ParticipantTask oldTask, SurveyTaskConfigDto taskDispatchConfig) {
         super.copyTaskData(newTask, oldTask, taskDispatchConfig);
 
-        //todo check if task config is set to prepopulate
-        Optional<SurveyResponse> priorResponse = surveyResponseService.findOneWithAnswers(oldTask.getSurveyResponseId());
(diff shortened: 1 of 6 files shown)


core/src/main/java/bio/terra/pearl/core/service/survey/SurveyTaskDispatcher.java
```diff
@@ -131,4 +135,43 @@ protected Optional<SurveyTaskConfigDto> findTaskConfigByStableId(UUID studyEnvir
...
+        if(taskDispatchConfig.getSurvey().isPrepopulate()) {
+            Optional<SurveyResponse> priorResponse = surveyResponseService.findOneWithAnswers(oldTask.getSurveyResponseId());
+
+            if (priorResponse.isPresent() && taskDispatchConfig.getSurvey().isPrepopulate()) {
+                SurveyResponse createdResponse = createPrepopulatedSurveyResponse(priorResponse.get());
+                newTask.setStatus(TaskStatus.NEW);
+                newTask.setSurveyResponseId(createdResponse.getId());
+            }
+        }
+    }
```
connorlbark, line 157: I feel like this would be useful to have a test in place, I could see this code getting moved/forgotten

connorlbark (changes requested):
I couldn't get this to work - I deleted jsalk's most recent lifestyle response and made the lifestyle survey prepopulate in the files and it didn't bring any answers forward. 

As an aside, I think it would be super useful to have a synthetic enrollee we can use to test this in demo.

devonbush (approved):
+1 to connor's point about adding configs/synthetic participant for testing.  Approving the general code approach -- looks clean

connorlbark (approved):
Good news! I'm working on the "assign new instance of task" functionality which will actually hit this code path, so we will be able to test that it works.

PR #1478, opened 2025-02-12
[JN-1411] Allow participants to delete unused documents

MatthewBemis:

screencapture-sandbox-demo-localhost-3001-hub-documents-2025-02-12-13_59_37

Screenshot 2025-02-12 at 1 59 45 PM
Screenshot 2025-02-12 at 2 03 21 PM
Screenshot 2025-02-12 at 1 59 49 PM

Log into heart demo as jsalk@test.com
Go to Document library
Try to delete a document with no associated responses, confirm it works
Try to delete a document with an associated response, confirm you see a warning modal

### DIFF
--- a/ui-participant/src/hub/documents/DocumentLibrary.tsx
+++ b/ui-participant/src/hub/documents/DocumentLibrary.tsx
@@ -209,7 +209,11 @@ const FileOptionsDropdown = ({ studyEnvParams, participantFile, enrollee, loadDo
       </Modal.Header>
       <Modal.Body>
         {participantFile.associatedAnswers.length === 0 ?
-          <p className="m-0">Are you sure you want to delete this document? This cannot be undone.</p> :
+          <p className="m-0">
+            Are you sure you want to delete this document? This cannot be undone. Please note that if a member
+            of the study staff has already downloaded this document, it will still be accessible to them. Please
+            contact a member of the study staff if you would like to have the document fully removed from all records.
+          </p> :
           <p className="m-0">This document is currently shared in response to at least one survey. Please remove it
                   from the survey response(s) before deleting it.</p>
         }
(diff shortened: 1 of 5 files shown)


devonbush (approved):

ui-participant/src/hub/documents/DocumentLibrary.tsx
```diff
@@ -170,3 +166,74 @@ const fileTypeToIcon = (fileType: string) => {
...
+      <Modal.Header>
+        <Modal.Title>
+          <h2 className="fw-bold pb-0 mb-0">
+            {participantFile.associatedAnswers.length === 0 ? 'Are you sure?' : 'This document is in use'}
+          </h2>
+        </Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        {participantFile.associatedAnswers.length === 0 ?
+          <p className="m-0">Are you sure you want to delete this document? This cannot be undone.</p> :
```
connorlbark: IIRC I think one of the things that was mentioned during demo was that the language here needs to be clear that this doesn't guarantee the files are actually gone - study staff might have saved them elsewhere, etc. Maybe some language here that, if you want to fully scrub from the system, they need to email staff, or something.

connorlbark (approved):

ui-participant/src/hub/documents/DocumentLibrary.tsx
```diff
@@ -170,3 +166,74 @@ const fileTypeToIcon = (fileType: string) => {
...
+          </a>
+        </li>
+      </ul>
+    </li>
+    {showConfirmDelete && <ThemedModal show={true}
+      onHide={() => setShowConfirmDelete(false)} size={'lg'} animation={true}>
+      <Modal.Header>
+        <Modal.Title>
+          <h2 className="fw-bold pb-0 mb-0">
+            {participantFile.associatedAnswers.length === 0 ? 'Are you sure?' : 'This document is in use'}
```
connorlbark, line 206: should we also check this on the backend? it looks like it naively will just delete it

MatthewBemis: you're right, that would be a good idea. i'll add that check to the backend

PR #1481, opened 2025-02-14
[JN-1607] add filter to study staff emails

connorlbark:

Adds the ability to filter which emails to send to.

I couldn't figure out how to slice it to easily mute during enrollee import. But this way, if you remove all the emails, it just won't send. So we can temporarily "mute" it by removing all the emails while importing. Then add the emails afterwards.

The other thing mentioned on the ticket was (I believe) to link to the enrollee on sent notifications, which it already does.

image

- Sign up a user in demo and say "no" to receiving emails
- Go to sent notifications for the admin notif trigger
- Notice it only sent an email to heartdemostaff and not power user
- Click on the enrollee to go to your new enrollee's page

### DIFF
--- a/core/src/main/resources/db/changelog/changesets/2025_02_14_trigger_admin_email_filter.yaml
+++ b/core/src/main/resources/db/changelog/changesets/2025_02_14_trigger_admin_email_filter.yaml
@@ -1,11 +1,11 @@
 databaseChangeLog:
   - changeSet:
-      id: "trigger_admin_email_filter"
+      id: "trigger_target_emails"
       author: connorlbark
       changes:
         - addColumn:
             tableName: trigger
             columns:
               - column:
-                  name: admin_email_filter
+                  name: target_emails
                   type: text
--- a/core/src/main/resources/db/changelog/changesets/2025_02_14_trigger_admin_email_filter.yaml
+++ b/core/src/main/resources/db/changelog/changesets/2025_02_14_trigger_admin_email_filter.yaml
@@ -1,0 +1,11 @@
+databaseChangeLog:
+  - changeSet:
+      id: "trigger_admin_email_filter"
+      author: connorlbark
+      changes:
+        - addColumn:
+            tableName: trigger
+            columns:
+              - column:
+                  name: admin_email_filter
+                  type: text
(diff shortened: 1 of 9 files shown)


core/src/main/resources/db/changelog/changesets/2025_02_14_trigger_admin_email_filter.yaml
```diff
@@ -0,0 +1,11 @@
+databaseChangeLog:
+  - changeSet:
+      id: "trigger_admin_email_filter"
+      author: connorlbark
+      changes:
+        - addColumn:
+            tableName: trigger
+            columns:
+              - column:
+                  name: admin_email_filter
```
devonbush: I could maybe see a more general column name like "target_emails"

ui-admin/src/study/notifications/TriggerDesignerEditor.tsx
```diff
@@ -424,6 +440,46 @@ const NotificationEditor = (
...
+) => {
+  const [adminUsers, setAdminUsers] = useState<string[]>([])
+
+  const { isLoading } = useLoadingEffect(async () => {
+    const adminUsers = await Api.fetchAdminUsersByPortal(studyEnvContext.portal.shortcode)
+    setAdminUsers(adminUsers.map(user => user.username))
+  }, [])
+
+
+  if (isLoading) {
```
devonbush, line 463: any reason we're not using the LoadingSpinner component?

devonbush (approved):
nice, a couple nits.  Also, please do a quick check in prod to make sure we don't have any existing admin triggers that we need to update after this PR

ui-admin/src/study/notifications/TriggerDesignerEditor.tsx
connorlbark: good catch, it's just that copilot decided not to 😆

connorlbark:
> Also, please do a quick check in prod to make sure we don't have any existing admin triggers that we need to update after this PR

Only study staff email is in trcc, so no worries there I think.

MatthewBemis (approved):

PR #1483, opened 2025-02-14
[JN-1619] Create new response and task for user if response was considered stale

MatthewBemis:

Adds the option for study staff to configure a longitudinal survey to create a new response and task if a response is edited after X days

Admin UX:
Screenshot 2025-02-14 at 2 16 35 PM

Restart admin and participant API apps
Repopulate demo
Log in as jsalk@test.com
Edit the response to the lifestyle survey
Since the survey is configured to create a new response if the last edit was >1 day ago, you should see a new task and response created with the new changes
Continue editing and confirm that the new edits are saved to the just-now-created task/response

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
@@ -141,13 +141,13 @@ public class SurveyResponseService extends CrudService<SurveyResponse, SurveyRes
     }
 
     //if the cutoff time has passed, create a new task and response
-    private boolean shouldCreateNewLongtitudinalTaskAndResponse(Integer createNewResponseAfterDays, ParticipantTask task) {
+    private boolean shouldCreateNewLongitudinalTaskAndResponse(Integer createNewResponseAfterDays, Instant surveyResponseLastUpdatedAt) {
         if(createNewResponseAfterDays == null) {
             return false;
         }
         Instant cutoffTime = ZonedDateTime.now(ZoneOffset.UTC)
-                .minusSeconds(createNewResponseAfterDays).toInstant(); //todo change back to minusDays
-        return task.getLastUpdatedAt().isBefore(cutoffTime);
+                .minusDays(createNewResponseAfterDays).toInstant();
+        return surveyResponseLastUpdatedAt.isBefore(cutoffTime);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -131,6 +135,16 @@ private List<Answer> getReferencedAnswers(Enrollee enrollee, Survey survey) {
...
+    //if the cutoff time has passed, create a new task and response
+    private boolean shouldCreateNewLongitudinalTaskAndResponse(Integer createNewResponseAfterDays, Instant surveyResponseLastUpdatedAt) {
+        if(createNewResponseAfterDays == null) {
+            return false;
+        }
+        Instant cutoffTime = ZonedDateTime.now(ZoneOffset.UTC)
+                .minusDays(createNewResponseAfterDays).toInstant();
+        return surveyResponseLastUpdatedAt.isBefore(cutoffTime);
+    }
+
```
connorlbark, line 152: I was thinking about this ticket and another point of trickiness that I think we need to account for is - what happens if you want to edit a previous response?

Like, let's say you have 5 responses, and you realize the second one is incorrect - if you edit that second one, I'm assuming it should edit in place. But if you attempt to edit the _latest_ one, it should make a new one.

I dunno if that's the right approach, but it feels right to me.

connorlbark (commented):
I'm having trouble getting this to work.
See: 
https://github.com/user-attachments/assets/ce1fa052-cb64-4cb2-9429-2bc957346463
It looks like for me the task id isn't updating, so it keeps creating new responses from the old task. Then, when I go back to the dashboard, it breaks the dashboard til I reload.

connorlbark (approved):
nice! works great! I'll add some tests for this alongside tests for the read-only history PR i'm starting on

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -152,8 +166,23 @@ public HubResponse<SurveyResponse> updateResponse(SurveyResponse responseDto, Re
                 task.getTargetAssignedVersion(), portalId).get();
         validateResponse(survey, task, responseDto.getAnswers());
 
-        // find or create the SurveyResponse object to attach the snapshot
-        SurveyResponse response = findOrCreateResponse(task, enrollee, enrollee.getParticipantUserId(), responseDto, portalId, operator);
+
+        SurveyResponse priorResponse = dao.findOneWithAnswers(task.getSurveyResponseId()).orElse(null);
+        SurveyResponse response;
+        //if the survey is longitudinal and we're past the cutoff point for updating an existing response, we need to create a new response and task
+        if (survey.getRecurrenceType() == RecurrenceType.LONGITUDINAL && priorResponse != null && shouldCreateNewLongitudinalTaskAndResponse(survey.getCreateNewResponseAfterDays(), priorResponse.getLastUpdatedAt())) {
```
devonbush, line 173: It may be cleaner if you put more of this logic in the helper method, and also use a helper method for the task/response logic
```
if (shouldCreateNewTaskAndResponse(...)) {
    taskAndResponse = createNewTaskAndResponse(...) 
   response = taskAndResponse.response
  task = taskAndResponse.task
} else {
   response = findOrCreateResponse(task, enrollee, enrollee.getParticipantUserId(), responseDto, portalId, operator);
}
```

ui-core/src/components/forms/PagedSurveyView.tsx
```diff
@@ -102,6 +105,7 @@ export function PagedSurveyView({
         participantTasks: response.tasks,
         profile: response.profile
       }
```
devonbush, line 107:
```suggestion
      }
      // update the taskId in case this is an update to a longitudinal survey which will create a new task & response
```

devonbush (approved):
looks good, I like how this didn't require a lot of new code to accomplish.  Some non-blocking code style nits

PR #1507, opened 2025-02-24
[JN-1630] Only dispatch survey events if survey is complete

MatthewBemis:

Survey completion notifications would be sent on every autosave, not just completions. Now this only sends an email if the survey has been completed.

Repopulate demo
Log in as jsalk@test.com
Go to the Other Medical History survey
Fill out some answers - make sure an autosave is triggered
Confirm you don't get an email
Autofill and submit the rest of the survey
Confirm you _do_ get an email

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/workflow/TriggerActionService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/workflow/TriggerActionService.java
@@ -72,7 +72,7 @@ public class TriggerActionService {
                 //if event is EnrolleeSurveyEvent, we need to check if the event was a survey
                 //completion or not to determine if we should send the notification
                 if (event instanceof EnrolleeSurveyEvent surveyEvent) {
-                    if (surveyEvent.isComplete()) {
+                    if (surveyEvent.getSurveyResponse().isComplete()) {
                         notificationDispatcher.dispatchNotificationAsync(trigger, event.getEnrolleeContext(),
                                 event.getPortalParticipantUser().getPortalEnvironmentId());
                     }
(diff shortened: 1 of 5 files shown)


core/src/main/java/bio/terra/pearl/core/service/workflow/TriggerActionService.java
```diff
@@ -68,8 +69,17 @@ public void handleEvent(EnrolleeEvent event) {
 
         for (Trigger trigger: applicableTriggers) {
             if (TriggerActionType.NOTIFICATION.equals(trigger.getActionType())) {
-                notificationDispatcher.dispatchNotificationAsync(trigger, event.getEnrolleeContext(),
-                        event.getPortalParticipantUser().getPortalEnvironmentId());
+                //if event is EnrolleeSurveyEvent, we need to check if the event was a survey
```
MatthewBemis: this _could_ have been a new event type like `SURVEY_RESPONSE_UPDATED`, but given that autosave triggers a survey response update every few seconds, i really don't think there's a good way for a study admin to capitalize on an event type like that

devonbush: This is kind of duplicating the logic in lines 79-81 of SurveyTaskDispatcher.  I wonder if we should just add logic around line 178 so that we only publish EnrolleeSurveyEvents when a survey is complete, which I think fits into what you're saying of this event is now really SURVEY_COMPLETE (or an update to an already-completed survey).  We might also consider an event "COMPLETE_FOR_FIRST_TIME" or somesuch.

devonbush (approved):
good fix.  Some thoughts about a more general/communicative solution, but might not be worth the refactor right now

connorlbark (approved):

PR #1496, opened 2025-02-20
[JN-1625] Document upload bucket terraform

MatthewBemis:
#### DESCRIPTION (include screenshots, and mobile screenshots for participant UX)

Adds new terraform resources for the participant document upload buckets.

tf plan for dev:

```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create
  ~ update in-place

Terraform will perform the following actions:

  # google_compute_subnetwork.juniper_subnetwork will be updated in-place
  ~ resource "google_compute_subnetwork" "juniper_subnetwork" {
        id                         = "projects/broad-juniper-dev/regions/us-central1/subnetworks/juniper-cluster-subnetwork"
        name                       = "juniper-cluster-subnetwork"
      ~ private_ipv6_google_access = "DISABLE_GOOGLE_ACCESS" -> "ENABLE_GOOGLE_ACCESS"
        # (18 unchanged attributes hidden)

        # (2 unchanged blocks hidden)
    }

  # google_storage_bucket.participant_documents will be created
  + resource "google_storage_bucket" "participant_documents" {
      + effective_labels            = {
          + "goog-terraform-provisioned" = "true"
        }
      + force_destroy               = false
      + id                          = (known after apply)
      + location                    = "US-CENTRAL1"
      + name                        = "participant-documents-dev"
      + project                     = (known after apply)
      + project_number              = (known after apply)
      + public_access_prevention    = (known after apply)
      + rpo                         = (known after apply)
      + self_link                   = (known after apply)
      + storage_class               = "STANDARD"
      + terraform_labels            = {
          + "goog-terraform-provisioned" = "true"
        }
      + uniform_bucket_level_access = (known after apply)
      + url                         = (known after apply)

      + soft_delete_policy (known after apply)

      + versioning {
          + enabled = true
        }

      + website (known after apply)
    }

Plan: 1 to add, 1 to change, 0 to destroy.
```

PR #1509, opened 2025-02-24
[JN-1632] GCS backend for document upload

MatthewBemis:

Adds a GCS file backend for document uploads.

In general, local development should still use the tmp file storage, but I've included instructions for how to test this locally.

1) Update application.yml for participant and admin APIs to use `GCSFileStorageBackend`
2) Add two new environment variables to admin and participant run configurations:

```
GCS_FILE_STORAGE_BUCKET_NAME=mb-test-juniper-upload-bucket;GCS_FILE_STORAGE_PROJECT_ID=broad-juniper-dev
```
(the above uses a test bucket I made- not the real bucket for demo)

3) Repopulate demo
4) Confirm you can see the 3 new objects created and uploaded as part of the populate process: https://console.cloud.google.com/storage/browser/mb-test-juniper-upload-bucket;tab=permissions?forceOnBucketsSortingFiltering=true&project=broad-juniper-dev&prefix=&forceOnObjectsSortingFiltering=false&inv=1&invt=Abqnsg
5) Log in to Heart Demo as jsalk, go to the Documents page and confirm download and delete still work
6) Confirm you can upload a new document and the file can be found in the bucket

Also note testing note's in Connor's PR for virus scanning: https://github.com/broadinstitute/juniper/pull/1548

### DIFF
--- a/api-admin/src/test/resources/application.yml
+++ b/api-admin/src/test/resources/application.yml
@@ -18,6 +18,15 @@ env:
   email:
     sendgridApiKey: ${SENDGRID_API_KEY:}
 
+otel:
+  sdk:
+    disabled: true
+  instrumentation:
+    spring-webflux:
+      enabled: false
+    spring-web:
+      enabled: false
+
 # Below here is non-deployment-specific
 
 # When the target is 'local' the write-config.sh script will generate this properties file. It
--- a/terraform/gcp/k8s/environments/prod.yaml
+++ b/terraform/gcp/k8s/environments/prod.yaml
@@ -5,6 +5,7 @@ deploymentZone: prod
 replicas: 3
 dsmUrl: https://dsm.datadonationplatform.org/dsm
 dsmIssuer: juniper.terra.bio
+gcsFileStorageBucketName: juniper-participant-documents-prod
 # "portals" adds certificates for each portal - both for the admin subdomains and the custom domain
 portals:
   - name: demo
--- a/api-admin/build.gradle
+++ b/api-admin/build.gradle
@@ -38,6 +38,7 @@ dependencies {
     implementation 'com.auth0:java-jwt:4.5.0'
     implementation 'org.liquibase:liquibase-core:4.31.0'
     implementation 'io.jsonwebtoken:jjwt-api:0.12.6'
+    implementation 'com.google.guava:guava:33.4.0-jre'
 
     implementation 'net.javacrumbs.shedlock:shedlock-spring:6.2.0'
     implementation 'net.javacrumbs.shedlock:shedlock-provider-jdbc-template:6.2.0'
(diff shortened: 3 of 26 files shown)


api-admin/src/test/resources/application.yml
```diff
@@ -18,6 +18,15 @@ env:
   email:
     sendgridApiKey: ${SENDGRID_API_KEY:}
 
+otel:
```
MatthewBemis, line 21: for some reason, this is needed now otherwise tests start to instantiate opentelemetry from terra-common-lib

terraform/gcp/k8s/environments/prod.yaml
```diff
@@ -5,6 +5,7 @@ deploymentZone: prod
 replicas: 3
 dsmUrl: https://dsm.datadonationplatform.org/dsm
 dsmIssuer: juniper.terra.bio
+gcsFileStorageBucketName: juniper-participant-documents-prod
```
devonbush, line 8: how does auth work?  is the application automatically allowed to write in the bucket because it's running in the same project?  If so, I'd add some comments specifying that

devonbush (approved):
Nice, one question about auth that I just want to confirm

terraform/gcp/k8s/environments/prod.yaml
connorlbark: I'm not entirely sure how it works locally, but I can speak to how (I believe) it would work on the GKE side. Our pod is authenticated to a service account (`juniper-app-gsa`) which is how it can read secrets and stuff. I don't think it has enough perms, though, becuase I think it needs `roles/storage.objectUser` (see [docs](https://cloud.google.com/storage/docs/access-control/iam-roles), change needed in juniper_service_account.tf). Here you can see them testing out authetnication from a [pod](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#verify) with the same kind of authentication we use - it looks like the auth is injected into the pod itself, so (I think) it should automatically be picked up.

connorlbark (approved):
Works locally as expected - confirmed it _does_ use ADC from local gcloud to authenticate.

However, I think it needs a few things before it will work in dev/prod:
1. application-gcp.yml needs to be updated for both admin and participant to specify the GCS backend for fileupload
2. In juniper_service_account.tf, we need to add:
```terraform
resource "google_project_iam_binding" "access-fileupload-bucket" {
  project = var.project
  role    = "roles/storage.objectUser"
  members = [
    "serviceAccount:${google_service_account.junper-app-gsa.email}"
  ]
}
```

See [bucket roles](https://cloud.google.com/storage/docs/access-control/iam-roles)



Finally, after this is merged, I think we'll also wanna look into virus scanning on buckets. There's some [docs](https://cloud.google.com/architecture/automate-malware-scanning-for-documents-uploaded-to-cloud-storage) that exist from GCP that could be interesting.

MatthewBemis:
@connorlbark thanks for calling that out! I'll add the terraform change to https://github.com/broadinstitute/juniper/pull/1496

PR #1513, opened 2025-02-26
[JN-1636] Don't sync deactivated kits

MatthewBemis:

Once a kit is deactivated in Juniper, we don't want to keep updating the statuses for those kits.

Motivation: gVasc has a bunch of kits that they accidentally sent out. Participants returned completed kits back to GP, but gVasc wants to exclude these kits from sequencing (presumably to save $). Things have already been handled on the GP-side, and we set these kits to Deactivated in Juniper, but because we sync statuses with DSM nightly, the statuses keep getting reset to "Received". This is confusing for gVasc and has generated a decent amount of support chatter.

One downside of this is that we could have statuses that differ between DSM and Juniper, but I think that's okay if it means we don't have to have someone go into both Juniper _and_ DSM to deactivate kits.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
@@ -407,7 +407,14 @@ public class KitRequestService extends CrudService<KitRequest, KitRequestDao> {
             // if the kit has been deactivated in Juniper, we no longer need to update the status
             // based on what DSM thinks. this allows us to independently deactivate kits in Juniper
             // without having to worry about DSM status updates.
-            log.info("Skipping status update for deactivated kit request %s".formatted(kitRequest.getId()));
+            if(!PepperKitStatus.mapToKitRequestStatus(pepperKit.getCurrentStatus()).equals(priorStatus)) {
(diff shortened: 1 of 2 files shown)


connorlbark (approved):

core/src/main/java/bio/terra/pearl/core/service/kit/KitRequestService.java
```diff
@@ -402,6 +402,15 @@ protected String stringifyPepperAddress(PepperKitAddress kitAddress) {
      */
     private void saveKitStatus(KitRequest kitRequest, PepperKit pepperKit, Instant pepperStatusFetchedAt) {
         KitRequestStatus priorStatus = kitRequest.getStatus();
+
+        if(priorStatus.equals(KitRequestStatus.DEACTIVATED)) {
+            // if the kit has been deactivated in Juniper, we no longer need to update the status
+            // based on what DSM thinks. this allows us to independently deactivate kits in Juniper
+            // without having to worry about DSM status updates.
+            log.info("Skipping status update for deactivated kit request %s".formatted(kitRequest.getId()));
```
devonbush: can we change this to not log anything (or maybe debug) if the DSM status matches, but to log a warning if the statuses do not match?

devonbush (approved):

PR #1526, opened 2025-03-07
[JN-1637] allow overriding translations for study & portal names

connorlbark:

Adds the ability to override translations for study.name and portal.name. Does not add any new fields or anything to SQL. Instead, uses language text overrides to accomplish this in a simple/publishable way.

The frontend looks to see if any i18n override keys with the format `study:<portalshortcode>.<studyshortcode>` exist and uses that key instead of `study.name`, if it exists. Otherwise just take whatever's in `study.name` as the default. Same for portal, with `portal:<portalshortcode>` replacing `portal.name`.

Also uses the translation for all usages of  `{study.name}` in enrollee email substitutor. The portal name used in the `sender` field will also appropriately reflect the translation override.

- Sign up a user in spanish
- Observe that the name of the study is in spanish on your dashboard
- Send the ad-hoc invitation email to this enrollee
- Observe the email they receive also uses the spanish name of the study & the sender is in spanish

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
@@ -141,7 +141,7 @@ public class EnrolleeEmailService implements NotificationSender {
                     .findBySiteContentLanguageAndKey(
                             contextInfo.portalEnv().getSiteContentId(),
                             preferredLanguage,
-                            "study:" + contextInfo.portal().getShortcode() + "." + contextInfo.study().getShortcode())
+                            LanguageTextService.formatStudyNameTranslationKey(contextInfo.portal().getShortcode(), contextInfo.study().getShortcode()))
                     .ifPresent(studyNameText -> contextInfo.study().setName(studyNameText.getText()));
         }
 
@@ -151,7 +151,7 @@ public class EnrolleeEmailService implements NotificationSender {
                     .findBySiteContentLanguageAndKey(
                             contextInfo.portalEnv().getSiteContentId(),
                             preferredLanguage,
-                            "portal:" + contextInfo.portal().getShortcode())
+                            LanguageTextService.formatPortalNameTranslationKey(contextInfo.portal().getShortcode()))
                     .ifPresent(portalNameText -> contextInfo.portal().setName(portalNameText.getText()));
         }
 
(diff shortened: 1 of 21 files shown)


core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
```diff
@@ -140,6 +143,17 @@ protected Mail buildEmail(NotificationContextInfo contextInfo, EnrolleeContext r
...
 
+
+        if (contextInfo.study() != null) {
+            // makes sure study.name returns the name of the study in the preferred language
+            languageTextService
+                    .findBySiteContentLanguageAndKey(
+                            contextInfo.portalEnv().getSiteContentId(),
+                            preferredLanguage,
+                            contextInfo.portal().getShortcode() + "." + contextInfo.study().getShortcode())
+                    .ifPresent(studyNameText -> contextInfo.study().setName(studyNameText.getText()));
```
connorlbark, line 145: I don't know if mutating Study is the optimal way to go. This has the advantage of "feeling correct" for the end user, but it feels weird from a code perspective.

core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
```diff
@@ -131,6 +134,28 @@ protected Mail buildEmail(NotificationContextInfo contextInfo, EnrolleeContext r
...
             fromAddress = routingPaths.getSupportEmailAddress();
         }
+
+        if (contextInfo.study() != null) {
+            // makes sure study.name returns the name of the study in the preferred language
+            languageTextService
+                    .findBySiteContentLanguageAndKey(
+                            contextInfo.portalEnv().getSiteContentId(),
+                            preferredLanguage,
+                            "study:" + contextInfo.portal().getShortcode() + "." + contextInfo.study().getShortcode())
```
devonbush: let's make this a method in LanguageTextService - `findTranslatedStudyName(portalShortcode, studyShortcode)` so the syntax is in a more findable place

core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
```diff
@@ -131,6 +134,28 @@ protected Mail buildEmail(NotificationContextInfo contextInfo, EnrolleeContext r
...
+                    .ifPresent(studyNameText -> contextInfo.study().setName(studyNameText.getText()));
+        }
+
+        if (contextInfo.portal() != null) {
+            // makes sure portal.name returns the name of the portal in the preferred language
+            languageTextService
+                    .findBySiteContentLanguageAndKey(
```
devonbush, line 151: same, make a method in LanguageTextService to do this lookup

core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
devonbush: yeah, I think this is the right idea.  To make it feel better from a code perspective, you could rename the `contextInfo` variable to `translatedContextInfo`?

devonbush (approved):
nice! some non-blocking codesmithing suggestions

MatthewBemis (approved):

PR #1547, opened 2025-03-17
[JN-1644] virus scanning infrastructure

connorlbark:

Originally, I thought it would be best to run this code in Eng infra. After some reflection (and it proving to be more difficult than I thought), I realized it might be better to keep prod/dev virus scanning separate.

```bash
# file has to have some content otherwise it won't get scanned
echo "test" > <unique filename>.txt
gsutil cp <unique filename>.txt gs://juniper-participant-documents-dev-unscanned/<unique filename>.txt
# should move pretty much instantly
gsutil ls gs://juniper-participant-documents-dev-unscanned
gsutil ls gs://juniper-participant-documents-dev-clean
```


terraform/gcp/envs/dev.tfvars
```diff
@@ -9,6 +9,12 @@ environment = "dev"
 portals = ["demo", "atcp", "ourhealth", "hearthive", "rgp", "cmi"]
 k8s_namespace = "juniper-dev"
 
+artifact_registry = "juniper"
+artifact_registry_project = "broad-juniper-eng-infra"
+artifact_registry_location = "us-central1"
```
MatthewBemis: do these need to be explicitly set here since they're the defaults in variables.tf?

MatthewBemis (approved):
Nice. Just left one quick question.

I'll close out [this PR](https://github.com/broadinstitute/juniper/pull/1496) too since it no longer brings anything to the table.

terraform/gcp/envs/dev.tfvars
connorlbark: good point, they don't. I'll remove them when I apply to prod.

devonbush (approved):

PR #1550, opened 2025-03-20
[no ticket yet] add translation script

connorlbark:

I got annoyed and wagered it would be not that hard to write this script. Luckily I was right!

- test it out:
`./scripts/auto_translate.sh "test" "Hello world!"`
- observe changes; first use will have a ton of newlines/changes as jq formats json a bit differently (I tried to fix this, but jq isn't really that smart at pretty-printing.)


MatthewBemis (approved):
NICE 🥇

PR #1551, opened 2025-03-20
[JN-1650] redirect user back to pre-enroll if enrollment fails

connorlbark:

Sends user back to pre-enroll if something goes wrong.

Tons of changes because of reformatting on language text keys; see [this pr](https://github.com/broadinstitute/juniper/pull/1550).

 I don't think there's an easy way to test this in dev. For local testing:
- Run participant ui WITH b2c
- Comment out line 115 of StudyEnrollRouter.tsx:
```
  function updatePreEnrollResponseId(preEnrollId: string | null) {
    // setPreEnrollResponseId(preEnrollId)
    setPreEnrollSatisfied(!!preEnrollId)
  }
```
- Submit pre-enroll; observe that no errors happen. This is because pre-enroll is not required.
- Set pre-enroll to required
- Enroll again, with it commented out still
- During b2c process, uncomment the line (so that second pre-enroll submission will work)
- After creating a password, observe that you get an alert
- Fill out pre-enroll again
- Observe you get taken to the consent

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
@@ -148,6 +148,10 @@ public class EnrollmentService {
         if (!studyEnvConfig.isAcceptingEnrollment()) {
             throw new IllegalArgumentException("study %s is not accepting enrollment".formatted(studyShortcode));
         }
+        if (preEnrollResponseId == null && hasRequiredPreEnroll(studyEnv.getId())) {
+            throw new IllegalArgumentException("user did not complete required pre-enrollment survey");
+        }
+
         PreEnrollmentResponse preEnrollResponse = validatePreEnrollResponse(operator, studyEnv, preEnrollResponseId, user.getId(), isSubject);
 
         // if the user is signed up, but not a subject, we can just return the existing enrollee,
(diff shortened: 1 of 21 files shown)


core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
```diff
@@ -148,6 +148,10 @@ public HubResponse<Enrollee> enroll(PortalParticipantUser operator,
         if (!studyEnvConfig.isAcceptingEnrollment()) {
             throw new IllegalArgumentException("study %s is not accepting enrollment".formatted(studyShortcode));
         }
+        if (preEnrollResponseId == null && hasRequiredPreEnroll(studyEnv.getId())) {
+            throw new IllegalArgumentException("user did not complete required pre-enrollment survey");
```
devonbush, line 152: FYI: this will be a change for OurHealth.  We may want to notify them or mark their preEnroll as not required

devonbush (approved):

core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
connorlbark: Their preEnroll is already not required; this is true of all existing studies AFAIK, so I _think_ this should operate the same for them

MatthewBemis (approved):

PR #1552, opened 2025-03-21
[JN-1651] previous responses read only

connorlbark:

Cleans up admin experience as well.

- Log in as jsalk
- Check out previous lifestyle surveys
- Go to latest
- Update it; ensure it saves as new row
- Go to admin tool
- Update jsalk's previous responses
- Ensure that it doesn't trigger the same mechanisms for admins (updates in-place always, can update previous responses)


devonbush (changes requested):
Looks good, but I think we need to still let admins edit prior survey responses.

connorlbark:
> Looks good, but I think we need to still let admins edit prior survey responses.

great catch

devonbush (approved):
nice

MatthewBemis (approved):

PR #1588, opened 2025-04-11
[JN-1663] save new responses as in-progress

connorlbark:

Cleans up longitudinal surveys to make a bit more sense
- Editing an old survey saves an In Progress response 
    - From a-t's perspective (which makes sense to me), you can't guarantee that some random edit is, from the enrollee's perspective, a complete response, and could lead to misleading data if they decide to leave for a while then come back and end up making a whole new response. So we should force them to hit complete. If they don't, they will get survey reminders and/or study staff can update the status.
- Doesn't create new longitudinal response if editing in-progress survey, only creates new response for (old) complete surveys. 
    - That way if a response is lingering for a while and an enrollee goes back to edit it, it doesn't start adding unexpected columns to their data until the enrollee confirms the survey is completed.

- log in as jsalk
- edit lifestyle 
- observe it becomes in progress
- observe read-only survey logic still works as expected


devonbush (approved):
Thanks for working this logic out with A-T!

MatthewBemis (approved):

PR #1597, opened 2025-04-18
[JN-1665 & JN-1668] fix invitation process for proxies and non-english enrollees

connorlbark:

Now, invitations will prefill properly for proxies. Also, `preferredLanguage` will propagate to the proxy, which will now also be maintained across the import/invitation process.

- import this spanish enrollee with proxy:
[2025-04-18-enrollees.tsv.zip](https://github.com/user-attachments/files/19814909/2025-04-18-enrollees.tsv.zip)
(if on demo server search and replace `testspanishproxy@test.com` with your email)
- send invitation email to imported governed user
- observe that it prefills with the correct email (no `-prox-ASDF`) and in the correct language
- observe upon landing that it shows the correct language

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
@@ -48,7 +48,8 @@ public class EnrolleeEmailSubstitutor implements StringLookup {
         valueMap.put("participantSupportEmailLink", getParticipantSupportEmailLink(contextInfo.portalEnv(), contextInfo.portalEnvConfig()));
         valueMap.put("siteMediaBaseUrl", getImageBaseUrl(contextInfo.portalEnv(), contextInfo.portalEnvConfig(), contextInfo.portal().getShortcode()));
         valueMap.put("siteImageBaseUrl", getImageBaseUrl(contextInfo.portalEnv(), contextInfo.portalEnvConfig(), contextInfo.portal().getShortcode()));
-        if (isProxy(ruleData)) {
+        boolean isProxy = isProxy(ruleData);
+        if (isProxy) {
             valueMap.put("isProxy", "true");
         } else {
             valueMap.put("isProxy", "false");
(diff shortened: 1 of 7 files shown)


ui-participant/src/Navbar.tsx
```diff
@@ -31,6 +31,13 @@ export default function Navbar(props: NavbarProps) {
     localContent
   } = usePortalEnv()
 
+  useEffect(() => {
+    if (profile && profile.preferredLanguage && selectedLanguage !== profile.preferredLanguage) {
+      changeLanguage(profile.preferredLanguage)
+      reloadPortal()
+    }
+  }, [profile?.preferredLanguage, selectedLanguage, changeLanguage])
+
```
connorlbark, line 40: So, if you logged into a user who has never been to Juniper before, they won't have the right preferred language in session storage. It takes a full refresh for it to populate - for some reason. Everything related to managing language state seems to be a can of worms, so this was my perhaps hamfisted solution to make sure it is synced on initial load.

core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
```diff
@@ -173,15 +175,39 @@ public String getParticipantSupportEmailLink(PortalEnvironment portalEnvironment
...
+            }
+            return url;
         } catch (UnsupportedEncodingException e) {
             throw new IOInternalException("unable to encode username");
         }
     }
+
+    private String removeProxySuffix(String email) {
+        // if the email ends with -prox-XXXX, remove it
+        Pattern proxySuffix = Pattern.compile("-prox-[A-Z]{4}$");
```
connorlbark: I thought about getting the email from the main account... but it made `EnrolleeContext` more complicated and it's already gotten really big.

MatthewBemis (approved):

devonbush (approved):

core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
devonbush: Let's put this method in RegistrationService so that the correlation between it and `getGovernedUsername` is more overt.  (or consider moving both to a `ProxyUsernameUtils` or somesuch)

PR #1615, opened 2025-05-09
[JN-1679] fix not equals sql behavior

connorlbark:

Search expressions didn't work exactly as you expected when executing SQL because they used `!=` instead of `IS DISTINCT FROM`.... so e.g. `'test' != null` -> Unknown -> false.

This was causing an annoyance for A-T who need to filter by `deceased != true` but many people had nothing for deceased.

- test making != on questions that are blank for many enrollees.


devonbush (approved):
nice!

MatthewBemis (approved):

PR #1622, opened 2025-05-16
[JN-1686] handle removed surveys in longitudinal surveys

connorlbark:

This was a headache; across the app we don't really intelligently have "REMOVED"/"REJECTED" checks. We ran into some issues importing some participants, and ended up removing some instances of surveys that were superfluous. This caused a variety of issues that essentially just completely locked a user up once they had a removed survey.

We also were doing some weird broken logic (from the perspective of longitudinal surveys) to determine if a task was accessible.

- export/import latest A-T configs
- sign up a new enrollee
- before filling out medical history, re-assign it to get two medical histories... then remove the medical history you just assigned
- make sure when you log in you only see one medical history and it's editable
- complete it
- add another removed survey - make sure it doesn't show up in dashboard and doesn't "lock" the following surveys (which would happen before)
- make sure previous longitudinal responses are still read-only.


core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -149,16 +149,40 @@ private boolean shouldCreateNewLongitudinalTaskAndResponse(Integer createNewResp
         return surveyResponseLastUpdatedAt.isBefore(cutoffTime);
     }
 
-    private boolean isMostRecentResponse(SurveyResponse surveyResponse) {
-        List<SurveyResponse> surveyResponses = dao.findAllByEnrolleeAndSurveyId(
-                surveyResponse.getEnrolleeId(), surveyResponse.getSurveyId());
+    /**
+     * Longitudinal tasks are editable if:
```
devonbush:
```suggestion
     * Longitudinal tasks are editable by the participant if:
```

ui-core/src/util/taskUtils.ts
```diff
@@ -43,24 +43,57 @@ export function isTaskAccessible(task: ParticipantTask, enrollee: Enrollee) {
...
+    }
+
+    if (openRequiredTasks[i].targetStableId === lastTaskStableId) {
+      if (!lastTaskIsComplete) {
+        lastTaskIsComplete = openRequiredTasks[i].status === 'COMPLETE'
+      }
+    } else {
+      lastTaskStableId = openRequiredTasks[i].targetStableId || ''
+      lastTaskIsComplete = openRequiredTasks[i].status === 'COMPLETE'
+    }
```
devonbush: I think this logic is really hard to follow.  Try something like below-- the intermediate variable names will help explain what's happening
```
const openRequiredStableIds = _uniq(openRequiredTasks.map(task => task.stableId))
const neverCompletedStableIds = openRequiredStableIds.filter(stableId => !openRequiredTasks.some(task => task.targetStableId === stableId && task.status === 'COMPLETE'))
return neverCompletedStableIds.length === 0 ||  task.targetStableId === neverCompletedStableIds[0]
```

devonbush (approved):
looks good, but please clean up the logic so it's easier to edit in the future

ui-core/src/util/taskUtils.ts
connorlbark: ahh, that's much nicer!

MatthewBemis (approved):

PR #1623, opened 2025-05-16
[JN-1687] recur on completion date

connorlbark:

From A-T's perspective they only want a new recurrence to come out if the task is completed, and it's one year after the completion date. Since hearthive wants it the other way, (recur every year no matter if completed or not based on creation date), it's configurable with a new option:

image

Also makes it more robust to removed tasks and makes testing more complete.

- I did some testing locally, but I think it will be easiest to make sure the unit tests make sense and then on merge I'll add a bunch of test participants to demo to make sure it works as expected before we release it to production.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/workflow/ParticipantTaskDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/workflow/ParticipantTaskDao.java
@@ -50,18 +50,24 @@ public class ParticipantTaskDao extends BaseMutableJdbiDao<ParticipantTask> impl
                 .stream().collect(Collectors.groupingBy(ParticipantTask::getEnrolleeId, Collectors.toList()));
     }
 
-    public Map<UUID, ParticipantTask> findLatestNotRemovedByEnrolleeIds(Collection<UUID> enrolleeIds) {
+    public Map<UUID, ParticipantTask> findLatestNotRemovedByEnrolleeIds(Collection<UUID> enrolleeIds, String stableId) {
+        if (enrolleeIds.isEmpty()) {
+            return Collections.emptyMap();
+        }
+
         return jdbi.withHandle(handle ->
                 handle.createQuery("""
                                 select * from %s
                                 where enrollee_id in (<enrolleeIds>)
                                 and status not in ('REMOVED', 'REJECTED')
+                                and target_stable_id = :stableId
                                 order by created_at desc
                                 """.formatted(tableName))
                         .bindList("enrolleeIds", enrolleeIds)
+                        .bind("stableId", stableId)
                         .mapTo(clazz)
                         .stream()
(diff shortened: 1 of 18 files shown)


core/src/main/java/bio/terra/pearl/core/service/workflow/TaskDispatcher.java
```diff
@@ -193,11 +192,21 @@ private void assignScheduledTasks(StudyEnvironment studyEnv) {
      * will assign a recurring task to enrollees who have already taken it at least once, but are due to take it again
      */
     private void assignRecurring(T taskConfig) {
-        List<Enrollee> enrollees = enrolleeService.findWithTaskInPast(
-                taskConfig.getStudyEnvironmentId(),
-                taskConfig.getStableId(),
```
connorlbark, line 198: I tried updating `findWithTaskInPast` but encoding all of the new logic we need into a single SQL query proved to be frustrating, so I opted for less efficient but most correct solution of loading all enrollees and then just calling `isRecurrenceWindowOpen`

MatthewBemis (approved):

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskDispatcher.java
devonbush: this makes me nervous for the 2000+ OurHealth enrollees

devonbush: Can we keep this the way it was, but just pass a RecurOn enum to findWithTaskInPast, and then just swap created_at/completed_at in the query based on the enum?

core/src/main/java/bio/terra/pearl/core/dao/workflow/ParticipantTaskDao.java
```diff
@@ -50,6 +50,27 @@ public Map<UUID, List<ParticipantTask>> findByEnrolleeIds(Collection<UUID> enrol
...
+        }
+
+        return jdbi.withHandle(handle ->
+                handle.createQuery("""
+                                select * from %s
+                                where enrollee_id in (<enrolleeIds>)
+                                and status not in ('REMOVED', 'REJECTED')
+                                and target_stable_id = :stableId
+                                order by created_at desc
+                                """.formatted(tableName))
```
devonbush, line 65:
```suggestion
                                select distinct on (enrollee_id, target_stable_id) * from %s
                                where enrollee_id in (<enrolleeIds>)
                                and status not in ('REMOVED', 'REJECTED')
                                and target_stable_id = :stableId
                                order by enrollee_id, target_stable_id, created_at desc
                                """.formatted(tableName))
```
will only return the rows you want -- there might be a lot of tasks for studies with a lot of recurrence like A-T

devonbush (changes requested):
would like to think about the SQL issues

devonbush (approved):

PR #1628, opened 2025-05-20
[JN-1690] fix multiple reminders not sending

connorlbark:

Adds more tests (mostly just for my own sanity), including one just to make sure longitudinal reminders work as i expect. 

Also, fixes a case where A-T has multiple unrelated reminders with filters; only one of the reminders would send because the reminders would be filtered out if ANY notification has been sent recently. This PR makes reminder debouncing trigger-specific in the case that we have a filter, so that multiple reminders can send out at once in these cases.

- Create multiple zero-hour reminders with filters on the stable ids
- Enroll
- Make sure you get all of them

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/workflow/ParticipantTaskDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/workflow/ParticipantTaskDao.java
@@ -213,7 +213,7 @@ public class ParticipantTaskDao extends BaseMutableJdbiDao<ParticipantTask> impl
                                     and created_at > :maxTimeSinceCreationInstant 
                                     and status in (<statuses>)
                                     %s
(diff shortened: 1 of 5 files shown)


MatthewBemis (approved):

core/src/main/java/bio/terra/pearl/core/dao/workflow/ParticipantTaskDao.java
```diff
@@ -190,11 +195,12 @@ public List<EnrolleeWithTasks> findByStatusAndTime(UUID studyEnvironmentId,
             return new ArrayList<>();
         }
         String stableIdString = targetStableIds == null ? "" : "and target_stable_id IN (<targetStableIds>)";
+        String triggerScopeString = triggerScopeId == null ? "" : "and trigger_id = :triggerScopeId";
```
devonbush, line 198: nice

devonbush (approved):
nice!

PR #1636, opened 2025-05-28
[JN-1697] fix proxies not being able to edit longitudinal imported surveys

connorlbark:

This one was tough; imported surveys would have the ppUserId of the _governed user_. Since this logic relied on finding the activities for the active pp user (which is incorrect, but is easy to not notice because in most proxy environments, the proxy pp user does everything). Since the proxy pp user doesn't have any tasks, it breaks, even though the enrollee had tasks.

So, this fixes the logic to do the correct thing - grab all tasks for the enrollee, not the active pp user.

- import a proxy user into a study with a longitudinal survey
- ensure you can edit it

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/workflow/TaskDispatcher.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/workflow/TaskDispatcher.java
@@ -364,7 +364,7 @@ public abstract class TaskDispatcher<T extends TaskConfig> {
                 ? latestTask.getCompletedAt()
                 : latestTask.getCreatedAt();
 
-        if (date == null) {
+        if (date == null || (taskDispatchConfig.getRecurOn() == RecurOn.COMPLETION && latestTask.getStatus() != TaskStatus.COMPLETE)) {
             return false; // never recur on incomplete tasks
         }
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/bio/terra/pearl/core/service/workflow/TaskDispatcher.java
```diff
@@ -364,7 +364,7 @@ private boolean isRecurrenceWindowOpen(T taskDispatchConfig, ParticipantTask lat
                 ? latestTask.getCompletedAt()
                 : latestTask.getCreatedAt();
 
-        if (date == null) {
+        if (date == null || (taskDispatchConfig.getRecurOn() == RecurOn.COMPLETION && latestTask.getStatus() != TaskStatus.COMPLETE)) {
```
devonbush: is this still needed given your fixup from yestderay?

devonbush (approved):
one non-blocking question because it still seems safe

MatthewBemis (approved):

core/src/main/java/bio/terra/pearl/core/service/workflow/TaskDispatcher.java
connorlbark: It's not - but I think it is better to double check this assumption regardless. It's a real sneaky bug when it happens.

PR #1642, opened 2025-06-02
[JN-1701] make search params globally shared via context

connorlbark:

I finally figured out how to reproduce a bug that had been causing issues for A-T.

If someone updates an old longitudinal survey, a new task ID is returned, which updates the URL via useSearchParams. The issue is that if you then _go to the next page_ then you run into issues. When you go to the next page, there's another hook that updates the URL via a different instance of useSearchParams.

Since they didn't have the same underlying searchParams instance, they would overwrite each other and the old task id would return, and then would cause any new updates to fail since that old task is now read only.

I think this is the cleanest way to solve this problem; the other way would be to drill down a useSearchParams across a few children, which I think would be pretty ugly.

- populate demo
- make basics longitudinal and make it save completed as new repsonse after 1 day
- update all in sandbox to this new version of the basics
- log in as jsalk
- update basics
- observe task id change
- go to next page
- continue making changes; does not cause issues.

### DIFF
--- a/ui-admin/src/App.tsx
+++ b/ui-admin/src/App.tsx
@@ -5,7 +5,13 @@ import 'survey-core/defaultV2.min.css'
 import './App.css'
 import './print.css'
 
-import { BrowserRouter, Outlet, Route, Routes, useLocation } from 'react-router-dom'
+import {
+  BrowserRouter,
+  Outlet,
+  Route,
+  Routes,
+  useLocation
+} from 'react-router-dom'
 import { ReactNotifications } from 'react-notifications-component'
 
 import { Config } from 'api/api'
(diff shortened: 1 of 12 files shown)


devonbush (changes requested):
This feels like a classic React bug of a component making an update with stale state.  Search params are already 'global state' in that they are tied to the URL.  
Instead of what you have here, I would try finding all the instances of `setSearchParams(params)`  and replacing them with 
```
setSearchParams(oldParams => {
 return {...oldParams, ...params}
}
```
so that the set operation is always using the latest state.  I believe that's the correct React pattern

devonbush (approved):

PR #1643, opened 2025-06-02
[JN-1700] make sure new longitudinal responses stay in-progress

connorlbark:

before, it would keep the "complete" of the old response, so on next save it would go from in-progress to complete.

- do the same testing as https://github.com/broadinstitute/juniper/pull/1642
- make sure after filling oiut basics completely, it still says "in progress" on dashboard
- hit complete
- make sure it is now complete
- edit the completed response
- ensure no errors, stays complete


MatthewBemis (approved):

devonbush (changes requested):
I'm not sure I agree on this one (although I could be convinced otherwise).  It seems like the issue is this component gets passed a response, makes an update to that response, and then needs to make sure that updated is reflected in its own display.  

You've fixed that here by storing response.complete in a separate place. But I think the correct way to fix this is for this component to communicate to its parent that the passed-in object has changed (which I thought was what `updateResponseMap(form.stableId, response.response)` was supposed to do...

devonbush (approved):

PR #1644, opened 2025-06-02
[JN-1703] import & export proxy profile

connorlbark:

Adds proxy profile to import and export systems.

Exports 1-many proxy profiles, but import is limited to only one proxy, so import only imports 1 proxy profile.

import this file into demo portal
[2025-04-18-enrollees 2.tsv.zip](https://github.com/user-attachments/files/20592938/2025-04-18-enrollees.2.tsv.zip)
observe they have a name & contact email

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
@@ -151,7 +151,8 @@ public class EnrolleeExportService {
                 new KitRequestFormatter(exportOptions),
                 new EnrolleeRelationFormatter(exportOptions),
                 new FamilyFormatter(exportOptions),
-                new ProxyFormatter(exportOptions));
+                new ProxyFormatter(exportOptions),
+                new ProxyProfileFormatter(exportOptions));
 
         List<ModuleFormatter> moduleFormatters = allSimpleFormatters.stream().filter(
                 moduleFormatter -> !exportOptions.getExcludeModules().contains(moduleFormatter.getModuleName())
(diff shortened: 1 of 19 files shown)


core/src/main/java/bio/terra/pearl/core/service/export/EnrolleeExportService.java
```diff
@@ -231,12 +232,16 @@ protected EnrolleeExportData loadEnrolleeData(Study study, StudyEnvironmentConfi
 
         List<EnrolleeRelation> enrolleeRelations = loadRelations(config, enrollee);
         List<ParticipantUser> proxies = loadProxyUsers(config, enrolleeRelations, enrollee);
+        // technically, there could be more than one proxy, but in practice this does not happen.
+        // if it does, there's no way to tell which profile belongs to which proxy.
+        List<Profile> proxyProfiles = loadProxyProfiles(config, enrolleeRelations, enrollee);
```
connorlbark, line 237: this absolutely could have performance implications for an A-T export - it's still fast now, but we will see with O(1000) enrollees how it goes.

devonbush: I'm not worried -- it's just another DB query -- we make lots of those :)

devonbush (approved):
Nice!

core/src/main/java/bio/terra/pearl/core/service/participant/ProfileService.java
```diff
@@ -45,6 +44,15 @@ public Map<UUID, Profile> loadAllWithMailingAddress(List<UUID> profileIds) {
                 .collect(Collectors.toMap(Profile::getId, Function.identity()));
     }
 
+    public List<Profile> loadAllByEnrolleeIdsWithMailingList(List<UUID> enrolleeIds) {
```
MatthewBemis: should this be called `loadAllByEnrolleeIdsWithMailingAddress`?

MatthewBemis (approved):
just had one naming question

core/src/main/java/bio/terra/pearl/core/service/participant/ProfileService.java
connorlbark: 🤦  nice catch

PR #1645, opened 2025-06-04
[JN-1704] remove old in-progress surveys during import 

connorlbark:

If we don't remove them, it causes weird edge cases here and there, I've found. Since it's kind of garbage data, anyway, I think it makes sense to set status to REMOVED - this indicates it's valid data actual put in by a participant, it can be brought back if needed, but it's cruft that may not be actually meaningful. in most cases, these in-progress responses are followed up almost immediately by a completed full response.

I debated if this is "overkill" - I _could_ put this logic in the transplate.py script. But I feel like this both makes it easier to test, and prevents a "gotcha" scenario that almost certainly will happen if/when future DSM studies migrate.

- import this file into demo
[2025-06-04-enrollees.tsv.zip](https://github.com/user-attachments/files/20594047/2025-06-04-enrollees.tsv.zip)
- observe that there are 3 responses to the basics: the latest is in-progress, the middle is removed, and the oldest is completed.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/workflow/ParticipantTaskDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/workflow/ParticipantTaskDao.java
@@ -293,6 +293,21 @@ public class ParticipantTaskDao extends BaseMutableJdbiDao<ParticipantTask> impl
         return findByProperty("kit_request_id", kitRequestId);
     }
 
+    public List<ParticipantTask> findTasksByEnrolleeAndSurveyResponse(UUID enrolleeId, List<UUID> surveyResponseId) {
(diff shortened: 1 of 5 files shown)


devonbush (approved):
this seems like a nice way of handling this so the data will still appear in export (right?) but not in the main application.  Although is that what you want?

connorlbark:
> this seems like a nice way of handling this so the data will still appear in export (right?) but not in the main application. Although is that what you want?

The followup here is that I think A-T will definitely want the ability to only export `COMPLETE` surveys. Then I think it will work exactly: they can export everything if they want it, they can bring back anything if they want it, but if they want a "clean" export, they can get that, too.

MatthewBemis (approved):

PR #1647, opened 2025-06-07
[JN-1698]: allow editing of survey blurbs

devonbush:

Enables editing of the blurb for Outreach surveys.  Also does some drive-by upgrades of font awesome dependencies to fix typescript warnings

Also contains a fix for JN-1706

1. go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms/surveys/depressionOutreach
2. confirm you can see and edit the blurb under "configuration"
3. confirm the blurb is not visible/editable for non-outreach surveys

### DIFF
--- a/ui-admin/src/util/downloadUtils.ts
+++ b/ui-admin/src/util/downloadUtils.ts
@@ -1,6 +1,6 @@
 /** escapes double quotes with an extra ", and adds double quotes around any values that contain commas or newlines */
 export const escapeCsvValue = (value?: string) => {
-  if (!value) {
+  if (!value || !value.replaceAll) {
     return ''
   }
   value = value.replaceAll('"', '""')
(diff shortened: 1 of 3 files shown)


MatthewBemis (approved):

ui-admin/src/util/downloadUtils.ts
```diff
@@ -1,6 +1,6 @@
 /** escapes double quotes with an extra ", and adds double quotes around any values that contain commas or newlines */
 export const escapeCsvValue = (value?: string) => {
-  if (!value) {
+  if (!value || !value.replaceAll) {
```
connorlbark, line 3: Why might `!value.replaceAll` be true, but `!value` is false? Just curious

connorlbark (approved):

ui-admin/src/util/downloadUtils.ts
devonbush: it's for catching weird edges where value isn't a string

PR #1670, opened 2025-07-10
[JN-1716] failed login event

connorlbark:

Adds endpoint to track failed logins, plus an enrollee event to match.

I have a theoretical change for B2c that would make this work, but I can't really test it until this is on dev.

locally / before b2c:
- create trigger to send participant email on failed login event
- execute:
```bash
curl --header "Content-Type: application/json" \
  --request POST \
  --data '{"username": "jsalk@test.com"}' \
  https://localhost:3001/api/public/portals/v1/demo/env/sandbox/externalEvent/failedLogin
```
- go to jsalk timeline: https://localhost:3000/demo/studies/heartdemo/env/sandbox/participants/HDSALK/timeline
- observe event, email, and see email in your inbox
- execute curl again, observe email again
- go to trigger, make "minimum minutes since last notification" 5
- trigger again multiple times; observe it skips when expected

on demo / after b2c setup:
- sign up participant
- log out
- create trigger with event on failed login to send email to participant & min minutes since last notification of 2
- try to log in as participant with wrong password multiple times, wait a couple minutes, fail to log in a few times
- observe expected number of emails sent / skipped

### DIFF
--- a/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/ExternalEventController.java
+++ b/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/ExternalEventController.java
@@ -7,6 +7,8 @@ import bio.terra.pearl.core.model.EnvironmentName;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.CrossOrigin;
+import org.springframework.web.bind.annotation.RequestMethod;
 
 @Controller
 @Slf4j
@@ -19,6 +21,24 @@ public class ExternalEventController implements ExternalEventApi {
   }
 
   @Override
+  @CrossOrigin(
+      origins = {
+        "https://juniperdemodev.b2clogin.com", // Heart Demo (demo)
+        "https://juniperdemoprod.b2clogin.com", // Demo Portal (prod)
+        "https://junipercmidemo.b2clogin.com", // CMI (demo only)
+        "https://juniperrgpdemo.b2clogin.com", // RGP (demo only)
+        "https://ourhealthdev.b2clogin.com", // OurHealth (demo)
+        "https://ourhealthstudy.b2clogin.com", // OurHealth (prod)
+        "https://hearthivedev.b2clogin.com", // HeartHive (demo)
+        "https://hearthive.b2clogin.com", // HeartHive (prod)
+        "https://gvascdev.b2clogin.com", // gVASC (demo)
+        "https://gvascprod.b2clogin.com", // gVASC (prod)
+        "https://juniperatcpdev.b2clogin.com", // ATCP (demo)
+        "https://juniperatcp.b2clogin.com", // ATCP (prod)
+        "https://trccproject.b2clogin.com" // tRCC (prod)
+      },
+      maxAge = 3600,
+      methods = {RequestMethod.GET, RequestMethod.OPTIONS})
   public ResponseEntity<Void> trackFailedLogin(
       String portalShortcode, String envName, ExternalEventFailedLoginBody body) {
     try {
@@ -26,8 +46,6 @@ public class ExternalEventController implements ExternalEventApi {
       externalEventExtService.trackFailedLoginEventAsync(
           portalShortcode, EnvironmentName.valueOfCaseInsensitive(envName), body.getUsername());
     } catch (Exception e) {
-      // Log the exception if tracking fails
-      // Note: Logging is not shown here, but should be implemented in a real application
       log.error("Failed to call failed login async method: {}", e.getMessage());
     }
 
--- a/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/ExternalEventController.java
+++ b/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/ExternalEventController.java
@@ -1,0 +1,36 @@
+package bio.terra.pearl.api.participant.controller;
+
+import bio.terra.pearl.api.participant.api.ExternalEventApi;
+import bio.terra.pearl.api.participant.model.ExternalEventFailedLoginBody;
+import bio.terra.pearl.api.participant.service.ExternalEventExtService;
+import bio.terra.pearl.core.model.EnvironmentName;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+
+@Controller
+@Slf4j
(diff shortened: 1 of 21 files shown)


api-participant/src/main/java/bio/terra/pearl/api/participant/service/ExternalEventExtService.java
```diff
@@ -0,0 +1,94 @@
...
+   * tracks failed logins for registered users.
+   */
+  @Async
+  public void trackFailedLoginEventAsync(
+      String portalShortcode, EnvironmentName envName, String username) {
+    try {
+      trackFailedLogin(portalShortcode, envName, username);
+    } catch (Exception e) {
+      log.error(
+          "Error tracking failed login for user {} in portal {} on environment {}: {}",
```
connorlbark: is it bad to log username, considering it's PII/PHI?

api-participant/src/main/java/bio/terra/pearl/api/participant/service/ExternalEventExtService.java
```diff
@@ -0,0 +1,94 @@
...
+
+    List<Enrollee> enrollees = enrolleeService.findByPortalParticipantUser(ppUser);
+
+    if (enrollees.isEmpty()) {
+      return; // no enrollee to track
+    }
+
+    for (Enrollee enrollee : enrollees) {
+      eventService.publishEnrolleeFailedLoginEvent(enrollee, ppUser);
+    }
```
connorlbark: should work just fine for proxies even though we aren't grabbing governed users - the failed logins will just be attached to the proxy enrollee instead of governed. which, to me, seems correct.

core/src/main/java/bio/terra/pearl/core/model/notification/Trigger.java
```diff
@@ -53,6 +53,7 @@ public class Trigger extends BaseEntity implements VersionedEntityConfig, StudyE
     private UUID emailTemplateId;
     private EmailTemplate emailTemplate;
     private String rule;
+    private Integer minMinutesSinceLastNotification;
```
connorlbark, line 57: new optional trigger field to avoid spamming failed login emails too often.

api-participant/src/main/java/bio/terra/pearl/api/participant/service/ExternalEventExtService.java
MatthewBemis: Yeah, I think the general guideline we've been following is to avoid logging this info since it's PII like you said. I think we can get this information from the audit logs in the b2c tenant anyway, in case we want to still log a generic error here and grab the info from the audit logs as needed.

MatthewBemis (approved):

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/ExternalEventController.java
```diff
@@ -0,0 +1,54 @@
...
+import bio.terra.pearl.api.participant.model.ExternalEventFailedLoginBody;
+import bio.terra.pearl.api.participant.service.ExternalEventExtService;
+import bio.terra.pearl.core.model.EnvironmentName;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.CrossOrigin;
+import org.springframework.web.bind.annotation.RequestMethod;
+
+@Controller
```
devonbush: this really needs a comment for what "ExternalEvents" are.  I might scope this more narrowly -- "B2CEventsController"?

core/src/main/java/bio/terra/pearl/core/dao/notification/NotificationDao.java
```diff
@@ -61,4 +61,17 @@ public void deleteByEnrolleeId(UUID enrolleeId) {
     public List<Notification> findAllByConfigId(UUID configId) {
         return findAllByProperty("trigger_id", configId);
     }
+
+    public Optional<Notification> findMostRecentSentByEnrolleeAndTriggerId(UUID enrolleeId, UUID triggerId) {
+        return jdbi.withHandle(handle ->
+                handle.createQuery("SELECT * FROM notification " +
```
devonbush: let's use multiline string templates instead of String addition

core/src/main/java/bio/terra/pearl/core/model/notification/Trigger.java
devonbush: good comment, please add it to the code :)

core/src/main/java/bio/terra/pearl/core/service/notification/email/EnrolleeEmailService.java
```diff
@@ -196,6 +199,24 @@ public boolean shouldSendEmail(Trigger config,
...
+                if (timeSinceLastSent.toMinutes() < config.getMinMinutesSinceLastNotification()) {
+                    log.info("skipping email, last email sent at {}, which is less at than {} minutes ago. enrollee {}, triggerId: {}, portalEnv: {}",
+                            lastSentTime,
+                            config.getMinMinutesSinceLastNotification(),
+                            enrolleeContext.getEnrollee().getShortcode(),
+                            config.getId(), config.getPortalEnvironmentId());
+                    return false;  // too soon to send another email
+                }
+            }
+        }
```
devonbush: this needs to get refactored into a helper method:
"hasNotificationWithinTimeWindow" or somesuch

ui-admin/src/study/notifications/TriggerDesignerEditor.tsx
```diff
@@ -412,6 +412,25 @@ const NotificationEditor = (
       <Select options={deliveryTypeOptions} isDisabled={true}
         value={deliveryTypeOptions.find(opt => opt.value === trigger.deliveryType)}/>
     </label>
+    {trigger.actionType === 'NOTIFICATION' && <div className='w-50 mb-2 d-flex flex-row'>
+      <label className="form-label">
+            Minimum minutes since last notification
```
devonbush: this could use an info popup explaining in more detail that this is about debouncing

devonbush (approved):
A few codesmithing suggestions, nothing that should block you getting this up on dev

api-participant/src/main/java/bio/terra/pearl/api/participant/controller/ExternalEventController.java
connorlbark: Funnily enough, that was almost exactly my original name for it... I waffled ("what if we one day stop using b2c?") but in retrospect it's much more clear this way.

PR #1675, opened 2025-07-17
[JN-1716] reset password page + link in email substitutor

connorlbark:

review in parallel with: https://github.com/broadinstitute/terraform-ap-deployments/pull/1913

- sign up user
- set up reset password email and use ${resetPasswordLink}
- log out
- open network tab
- attempt to sign in with random email
- see network tab sends failedLog
- attempt to sign in with correct email but use wrong password
- see network tab sends failedLogin 
- check that you received an email
- click on the reset password link in the email
- follow the reset password flow
- log out
- log in using new password

### DIFF
--- a/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/B2cEventController.java
+++ b/api-participant/src/main/java/bio/terra/pearl/api/participant/controller/B2cEventController.java
@@ -1,6 +1,6 @@
 package bio.terra.pearl.api.participant.controller;
 
-import bio.terra.pearl.api.participant.api.ExternalEventApi;
+import bio.terra.pearl.api.participant.api.B2cEventApi;
 import bio.terra.pearl.api.participant.model.ExternalEventFailedLoginBody;
 import bio.terra.pearl.api.participant.service.B2cEventExtService;
 import bio.terra.pearl.core.model.EnvironmentName;
@@ -12,7 +12,7 @@ import org.springframework.web.bind.annotation.RequestMethod;
 
 @Controller
 @Slf4j
-public class B2cEventController implements ExternalEventApi {
+public class B2cEventController implements B2cEventApi {
 
   private final B2cEventExtService b2cEventExtService;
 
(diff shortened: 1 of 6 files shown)


api-participant/src/main/java/bio/terra/pearl/api/participant/controller/B2cEventController.java
```diff
@@ -37,14 +35,14 @@ public B2cEventController(B2cEventExtService b2cEventExtService) {
...
+      allowedHeaders = "*",
+      maxAge = 3600)
   public ResponseEntity<Void> trackFailedLogin(
-      String portalShortcode, String envName, ExternalEventFailedLoginBody body) {
+      String portalShortcode, String envName, String email) {
     try {
       // track asynchronously; caller doesn't care if it succeeds or fails
       b2cEventExtService.trackFailedLoginEventAsync(
-          portalShortcode, EnvironmentName.valueOfCaseInsensitive(envName), body.getUsername());
+          portalShortcode, EnvironmentName.valueOfCaseInsensitive(envName), email);
```
connorlbark, line 45: separately: there was a CORS issue with the failed login tracking endpoint.

I _think_ it has something to do with the fact that the preflight response was not set up properly. Preflight requests are only sent when the request is considered "complex" by the browser. GET requests are never complex (unless they have unknown headers) and POST requests are not complex IF they use certain content-types. Using application/json is considered complex, but, text/plain isn't.

once I changed the endpoint to just take the email directly instead of using JSON, the cors issue magically disappeared.

ui-participant/src/landing/registration/ResetPasswordPage.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const accountName = searchParams.get('accountName') || ''
+  const preferredLanguage = searchParams.get('preferredLanguage') || ''
+  const envSpec = getEnvSpec()
+
+  const createAccount = async () => {
+    auth.signinRedirect({
+      redirectMethod: 'replace',
+
+      extraQueryParams: {
+        option: 'resetpassword',
```
connorlbark, line 22: this `option` only works because of the corresponding terraform-ap-deployments PR; we already had `option: signup`, so it's set up similarly.

core/src/main/java/bio/terra/pearl/core/service/notification/substitutors/EnrolleeEmailSubstitutor.java
```diff
@@ -177,20 +196,23 @@ public String getParticipantSupportEmailLink(PortalEnvironment portalEnvironment
     }
 
     /**
-     * gets a link the participant can use to create their b2c account, given that they already exist in Juniper
+     * gets a b2c redirect link; e.g., invitation or reset password link. Path must be from
```
devonbush: I'm not sure I'm following what's happening with this method--can you refine the comment?

devonbush (approved):
looks good, just would like some clarification on one of the methods

MatthewBemis (approved):

PR #1682, opened 2025-07-22
[JN-1712] send sex at birth to dsm

connorlbark:

Not sure this is the cleanest way to go about it.

I figured making no changes to existing study ops is the way to go - e.g., for trcc they determine sex at birth through other processes and I could see us including it for them causing confusion. Thus, it's a study env config option.

as mentioned in the code comments, I think it would make sense for this config option to be configurable per-kit-type. considering how much work this would be (right now, the study env kit type get/update methods are hardcoded to just a string) and how rare it would be, i went with the simpler option.

- honestly... not easy to test becuase DSM does not send us this value back. I created a few kits locally in the dev ourheart and Kiara was able to check in the DSM database that the values were serialized properly. otherwise, I think we just need to make sure nothing broke.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/model/study/StudyEnvironmentConfig.java
+++ b/core/src/main/java/bio/terra/pearl/core/model/study/StudyEnvironmentConfig.java
@@ -52,6 +52,14 @@ public class StudyEnvironmentConfig extends BaseEntity {
     private boolean enableInPersonKits = false;
 
     /**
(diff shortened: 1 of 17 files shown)


devonbush (approved):
Nice, agree that per-kit-type isn't needed.  It feels weird that this is hardcoded as a boolean, rather than a list of strings "additional properties to include".  But that's a refactoring we can do when/if we add any other additional properties

MatthewBemis (approved):

PR #1685, opened 2025-07-28
[JN-1729] refresh proxy profile as well as governed profile on survey submit

connorlbark:

trcc ran into this because a survey would submit but because the hub response did not include the proxy profile subsequent surveys that utilized `proxyProfile` would be missing the value.

- add question & answer mapping for proxyProfile.givenName in the basics survey
- enroll proxy user & complete basics. WITHOUT refreshing, click on the user dropdown and make sure it shows the proxy name you provided


MatthewBemis (approved):
thanks for tracking this down!

devonbush (approved):

PR #1699, opened 2025-08-12
[JN-1731] Adding sleep between DSM requests

devonbush:

This adds a small sleep between DSM requests.  @MatthewBemis  I looked for any "connection reset by peer" logs, and didn't see any, so I don't *think* this is the same error you previously fixed, but curious if anything about this rings a bell or if you have better ideas for the fix.

1.  select >1 participants for kit requests
2. send the requests
3. confirm they send


connorlbark (approved):

MatthewBemis (approved):
Seems reasonable to me. This exact error mode doesn't ring a bell 🤔. 

Related thought: did the DSM machine recently get downsized or something? I know Broad as a whole has been trying to right-size machines for cost savings, I wonder if the DSM machine got reduced and now it doesn't have enough memory or something to handle these spikes?

PR #1701, opened 2025-08-14
[JN-1734] - pre-enroll branches on signed-in status

devonbush:

Adds the user (ParticipantUser) to the survey variables that can be branched on.   Also updates the removeCompleteButton logic to handle if the expression was true at the beginning.  
image

In order to test this (and also to support testing of other multi-study stuff), I've added a second "sleep" study to the demo.  I think this also correctly reflects that multi-study is now Juniper's main use case.

1.  repopulate demo
2.  go to https://sandbox.demo.localhost:3001/sleep -- make sure you are *not* signed in
3. click "Join sleep study"
4. Observe that you can't join, and the complete button does not appear
5. sign in as jsalk@test.com
6. click on the "Join Sleep study" again
7. confirm you see a one-question survey now
8. answer "yes"
9. hit submit
10. confirm you have enrolled in the sleep study

### DIFF
--- a/ui-core/src/surveyUtils.tsx
+++ b/ui-core/src/surveyUtils.tsx
@@ -440,6 +440,14 @@ export function useSurveyJSModel(
         newSurveyModel.showCompleteButton = !valAsBool
       }
     })
+    newSurveyModel.onAfterRenderSurvey.add((_model, event) => {
+      event.survey.calculatedValues.forEach(val => {
+        const valAsBool = isString(val.value) ? val.value === 'true' : !!val.value
+        if (val.name.toLowerCase() === 'removecompletebutton' && valAsBool) {
+          newSurveyModel.showCompleteButton = false
+        }
+      })
+    })
 
     if (readonly) {
       newSurveyModel.mode = 'display'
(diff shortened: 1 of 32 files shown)


MatthewBemis (approved):

ui-core/src/surveyUtils.tsx
```diff
@@ -95,6 +95,7 @@ export const surveyJSModelFromFormContent = (formContent: FormContent): SurveyMo
 }
 
 export type SurveyJsVariableContext = {
+  user?: ParticipantUser | null,
```
connorlbark, line 98: am i missing something? I don't see this being used right now
i'd assuming we need a:

```
  surveyModel.setVariable('loggedIn', !isNil(user))
```

ui-core/src/surveyUtils.tsx
```diff
@@ -95,6 +95,7 @@ export const surveyJSModelFromFormContent = (formContent: FormContent): SurveyMo
 }
 
 export type SurveyJsVariableContext = {
```
connorlbark, line 97: I was surprised changing this type has fewer side effects; I thought it was being used in more places (but could be wrong)

connorlbark (approved):
quick review with some questions but otherwise LG and will test more deeply later today on dev

ui-core/src/surveyUtils.tsx
devonbush: the way I'm using it is a pre-enroll survey with a check for `{user.username} empty`

connorlbark: I see - but where does that get added to the actual variable context? I don't see it being used in  `applySurveyJsVariables` and since it's not in `extraVariables` it wouldn't get automatically added - but if it's working maybe I'm missing something

PR #1708, opened 2025-08-25
[JN-1740] Adding study eligibility rule

devonbush:

Adds the ability to configure an eligibility rule for a study.  
image

A new endpoint has been added to check eligibility.  The pre-enrollment survey calls this endpoint if a rule exists for the study, and loads the answer into the available survey variables

NOTE: This refactors the rule logic to use study shortcode, rather than name (otherwise the parsing errors if the study name has spaces, and 'Heart Hive' has spaces).  I did a check in prod for any cross-study rules and didn't find any

1.  go to https://sandbox.demo.localhost:3001/sleep   make sure you are not logged in 
2.  click "join sleep study"
3. confirm you see a message saying you must log in
4. log in as jsalk@test.com
5.  go back to https://sandbox.demo.localhost:3001/sleep and cliick join
6. confirm you get a different message saying enrollment is limited to participants with dpression symptoms
7. log out, log in as lostInterest@test.com
8. try to join the sleep study again
9. confirm the pre-enrollment survye appears, and you can join

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
@@ -439,7 +439,7 @@ public class EnrollmentService {
                         .build());
         EnrolleeSearchContext context = new EnrolleeSearchContext(
                 primaryStudyEnrollee,
-                profileService.find(ppUser.getProfileId()).orElse(new Profile())
+                profileService.loadWithMailingAddress(ppUser.getProfileId()).orElse(new Profile())
         );
         return enrolleeSearchExpressionParser
                 .parseRule(envConfig.getStudyEligibilityRule())
--- a/core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
@@ -12,10 +12,11 @@ import bio.terra.pearl.core.model.study.StudyEnvironmentConfig;
 import bio.terra.pearl.core.model.survey.*;
 import bio.terra.pearl.core.model.workflow.HubResponse;
 import bio.terra.pearl.core.service.exception.NotFoundException;
-import bio.terra.pearl.core.service.participant.EnrolleeRelationService;
-import bio.terra.pearl.core.service.participant.EnrolleeService;
-import bio.terra.pearl.core.service.participant.ParticipantUserService;
-import bio.terra.pearl.core.service.participant.PortalParticipantUserService;
+import bio.terra.pearl.core.service.participant.*;
+import bio.terra.pearl.core.service.rule.EnrolleeContext;
+import bio.terra.pearl.core.service.rule.EnrolleeRuleEvaluator;
+import bio.terra.pearl.core.service.search.EnrolleeSearchContext;
+import bio.terra.pearl.core.service.search.EnrolleeSearchExpressionParser;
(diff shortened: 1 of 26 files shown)


core/src/main/java/bio/terra/pearl/core/service/workflow/EnrollmentService.java
```diff
@@ -421,4 +426,20 @@ public HubResponse enroll(EnvironmentName environmentName, String studyShortcode
...
+                Enrollee.builder()
+                        .participantUserId(ppUser.getParticipantUserId())
+                        .profileId(ppUser.getProfileId())
+                        .build(),
+                profileService.find(ppUser.getProfileId()).orElse(new Profile())
+        );
+        return enrolleeSearchExpressionParser
+                .parseRule(envConfig.getStudyEligibilityRule())
+                .evaluate(context);
+    }
```
connorlbark: I think this will only work in a limited number of contexts;  cross-study answers happen to work because:
```
                this.studyName == null
                        ? answerDao.findForEnrolleeByQuestion(context.getEnrollee().getId(), surveyStableId, questionStableId)
                        : answerDao.findByProfileIdStudyAndQuestion(context.getEnrollee().getProfileId(), studyName, surveyStableId, questionStableId);
```
when you specify a study, it uses profile id instead of enrollee id.

But I think this would be a confusing gotcha for people who wouldn't use it just like it's being used here. I think it would be a lot more clear if you had to also specify which study the search expression is being executed from, and we load the full enrollee context from it. Or, default to the primary study and document that on the settings page.

ui-admin/src/study/settings/study/StudyEnrollmentSettings.tsx
```diff
@@ -75,6 +75,12 @@ export const StudyEnrollmentSettings = (
           onChange={e => updateConfig('acceptingProxyEnrollment', e.target.checked)}/>
       </label>
     </div>
+    <div>
+      <label className="form-label">
+        eligibility rule <input type="text" className="form-control" size={80} value={config.studyEligibilityRule ?? ''}
+          onChange={e => updateConfig('studyEligibilityRule', e.target.value)}/>
+      </label>
+    </div>
```
connorlbark, line 83: should we use `LazySearchQueryBuilder`?

ui-participant/src/studies/enroll/PreEnroll.tsx
```diff
@@ -61,7 +105,10 @@ export default function PreEnrollView({ enrollContext, survey }:
       studyEnvParams: { envName: studyEnv.environmentName, studyShortcode, portalShortcode },
       enrolleeShortcode: enrollee?.shortcode || '',
       referencedAnswers: [],
-      extraVariables: { isProxyEnrollment, isSubjectEnrollment, user }
+      extraVariables: {
+        isProxyEnrollment, isSubjectEnrollment, user,
+        eligibilityResult
```
connorlbark, line 110: nit: should we do `eligible: eligibilityResult.eligible` so it's a bit shorter/easier to remember?

connorlbark (commented):
couple quick thoughts

ui-participant/src/studies/enroll/PreEnroll.tsx
devonbush: I wanted this as an object because I expect it will get complicated with stuff like "eligibleForProxy"

MatthewBemis (approved):

connorlbark (approved):

PR #1720, opened 2025-09-05
[JN-1747] cleanup post cert release issues -> must merge before next release

connorlbark:

last release had 2 issues:
1. I accidentally put the portal name for pedihcc as `pedihcc` instead of `pedihccproject`. This means that `pedihccproject.juniper-cmi.org` does not have a certificate. It also meant that I had to add `pedicc.juniper-cmi.org` to our DNS
2. I accidentally left the ingress referencing `admin-cert` instead of `{{ .Values.adminCertName }}` which, for prod, is now `admin-cert-v2`, so I had to manually update the ingress for it to sync.

This PR will:
1. Update ingress's admin cert name reference to point to `{{ .Values.adminCertName }}`
2. Remove pedihcc.juniper-cmi.org from admin cert
3. Remove _old_ trccproject cert (will be OK cause there's a new CMI cert w/ both trcc and pedicc URLs)
4. Remove pedihcc.juniper-cmi.org from DNS
5. Add pedihccproject.juniper-cmi.org TO the admin cert

I also think we don't need to do the `admin-cert-v2` renaming on each regen... I think the issue last time was unrelated to it needing to be a new cert entirely... So we could probably remove that if we wanted.

Steps:
1. Merge this to dev
2. Deploy to prod
3. Wait for _new_ admin cert to regen with correct pedihcc.juniper-cmi.org domain
4. Re-apply terraform on development for prod to delete pedicc.juniper-cmi.org as subdomain (cannot do before deploying because it might break admin cert)

Will have admin downtime again while cert regens. We can avoid this by just keeping the vestigial pedihcc.juniper-cmi.org DNS records and accepting that pedihccproject.juniper-cmi.org will not have valid cert. Which I don't think is a big deal cause we have the public URL. No strong feelings.

### DIFF
--- a/terraform/gcp/envs/prod.tfvars
+++ b/terraform/gcp/envs/prod.tfvars
@@ -8,7 +8,8 @@ admin_url = "juniper-cmi.org"
 environment = "prod"
 # note: automatically creates DNS records for these portals under the admin domain
 
-portals = ["demo", "atcp", "ourhealth", "hearthive", "rgp", "cmi", "trccproject", "gvasc", "pedihccproject"]
+# TODO: remove pedihcc next release
+portals = ["demo", "atcp", "ourhealth", "hearthive", "rgp", "cmi", "trccproject", "gvasc", "pedihccproject", "pedihcc"]
 
 admin_dnssec = "on"
 k8s_namespace = "juniper-prod"
--- a/terraform/gcp/k8s/templates/ingress.yml
+++ b/terraform/gcp/k8s/templates/ingress.yml
@@ -13,7 +13,7 @@ metadata:
   name: juniper-ingress
   annotations:
     kubernetes.io/ingress.global-static-ip-name: admin-ip
-    networking.gke.io/managed-certificates: "admin-cert,{{ template "customer-certs" . }}"
+    networking.gke.io/managed-certificates: "{{ .Values.adminCertName }},{{ template "customer-certs" . }}"
     networking.gke.io/v1beta1.FrontendConfig: "juniper-frontend-config"
     # If the class annotation is not specified it defaults to "gce".
     kubernetes.io/ingress.class: "gce"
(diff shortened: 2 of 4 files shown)


terraform/gcp/envs/prod.tfvars
```diff
@@ -8,7 +8,8 @@ admin_url = "juniper-cmi.org"
 environment = "prod"
 # note: automatically creates DNS records for these portals under the admin domain
 
-portals = ["demo", "atcp", "ourhealth", "hearthive", "rgp", "cmi", "trccproject", "gvasc", "pedihccproject"]
+# TODO: remove pedihcc next release
+portals = ["demo", "atcp", "ourhealth", "hearthive", "rgp", "cmi", "trccproject", "gvasc", "pedihccproject", "pedihcc"]
```
connorlbark, line 12: I used the wrong portal name `pedihcc` on the admin cert so it failed to generate because the DNS would not resolve

could have re-releaed to fix but... Meh. just added the DNS record so it would generate.

terraform/gcp/k8s/templates/ingress.yml
```diff
@@ -13,7 +13,7 @@ metadata:
   name: juniper-ingress
   annotations:
     kubernetes.io/ingress.global-static-ip-name: admin-ip
-    networking.gke.io/managed-certificates: "admin-cert,{{ template "customer-certs" . }}"
+    networking.gke.io/managed-certificates: "{{ .Values.adminCertName }},{{ template "customer-certs" . }}"
```
connorlbark, line 16: I forgot to update the reference to the cert name, so I had to update it manually

In retrospect: I don't think this was actually necessary. I think it _can_ regen in place, _but_ I had to manually clean up a bunch of partially deleted certs that were lingering on the ingress (but did not exist) - possible caused by race condition from deleting a bunch all at once?

I think if I had fixed that issue first, this admin-renaming thing wouldn't have been necessary.

devonbush (approved):

MatthewBemis (approved):

PR #1730, opened 2025-09-16
[JN-1750] handle empty surveys elegantly on import

connorlbark:

Would love to have someone else double check my thought process.

Changes:
1. Instead of importing until `answers.size() == 0` (current behavior), import all survey response rows that have any values (even if it's just createdAt, etc.)
2. If we see there's an empty survey response that precedes a COMPLETE or IN_PROGRESS response for the task, remove it

These is _necessary_ because:
1. DSM export included random empty NEW survey responses (often as the LATEST row)
3. It's a big problem when we receive an empty response as the latest row because:
    1. The `answers.size() == 0` check would prevent later rows from being added to the participant's data
    2. Even _if_ we had the full response history, on longitudinal recurrence, the answers are copied from the latest response.... which would be empty for quite a few participants

This was harder than I first thought, for a few reasons:

1. Our entire import system is centered around `SurveyResponse`s
2. However, typically, in juniper, a survey response is not created until the participant edits the survey for the first time (i.e., a NEW survey has no survey response yet)
3. Thus, it should never be the case that a Survey Response with zero answers should exist (AFAIK)
4. The original import script would never process survey responses with no answers, so it preserved this behavior
5. Now that we're processing empty surveys, the import system will explicity _create_ empty survey responses
6. We _could_ have refactored the import logic to be task-centric instead of surveyresponse centric- but this felt like a bigger change than was needed to get the job done

Thus, imported participants now will ALWAYS have survey responses for all imported tasks, even if the response is empty. All of these empty survey responses should end up `REMOVED` in the end, _except_ in the case that it's the only task the participant has for the survey / the survey is not longitudinal. I think it's not a huge deal, but maybe it will cause some strange edge cases.

An alternate path would be to IGNORE any survey responses that are empty. I feel like it's more accurate to import them but hide them, but I could be convinced here.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/dao/dataimport/TimeShiftDao.java
+++ b/core/src/main/java/bio/terra/pearl/core/dao/dataimport/TimeShiftDao.java
@@ -79,6 +79,16 @@ public class TimeShiftDao {
         );
     }
 
+    public void changeTaskLastUpdatedTime(UUID taskId, Instant lastUpdatedTime) {
+        jdbi.withHandle(handle ->
+                handle.createUpdate("update participant_task set last_updated_at = :lastUpdatedTime where id = :taskId;")
+                        .bind("taskId", taskId)
+                        .bind("lastUpdatedTime", lastUpdatedTime)
+                        .execute()
+        );
+    }
+
+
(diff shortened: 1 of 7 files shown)


devonbush (approved):
looks good, appreciate the comments

MatthewBemis (approved):

PR #1745, opened 2025-10-22
[JN-1758] add hubspot records for trcc

connorlbark:

### DIFF
--- a/terraform/gcp/envs/prod.tfvars
+++ b/terraform/gcp/envs/prod.tfvars
@@ -124,7 +124,37 @@ customer_urls = {
         domain_prefix = ""
         type = "TXT"
         ttl = 3600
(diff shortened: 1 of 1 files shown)


terraform/gcp/envs/prod.tfvars
```diff
@@ -176,7 +206,7 @@ customer_urls = {
         record_values=["j.domainkey.u33588015.wl016.sendgrid.net."]
       },
       {
-        name="sendgrid_domainkey1"
+        name="sendgrid_domainkey2"
```
connorlbark, line 209: I don't know how this worked? renaming it this way doesn't seem to force recreation of the record, so 🤷

devonbush (approved):

MatthewBemis (approved):

PR #1754, opened 2025-11-25
[JN-1763] research ids

connorlbark:

Adds new participant identifier, "Research ID". This identifier is hidden from all participant views, which eliminates the risk of re-identification. It also adds the option to send this in place of the shortcode for kit requests, which is important if you use the kitting identifiers in published research (as A-T does). Also, you can override research IDs with any value you like.

There will be a follow up PR to this, after it's been released, which will remove the GENERATE_RESEARCH_IDS command as well as make the column required. Also, will import legacy DDP IDs for A-T so that the research IDs match the identifiers they use in their existing published research.

- populate demo on develop
- switch to this branch
- observe nobody has research ids
- populate ourhealth
- observe newly created enrollees have research ids
- use the GENERATE_RESEARCH_IDS populate command to generate all research ids for people in demo
- log in as participant
- observe that no network calls include research ids
- test enabling/disabling sending kits with research ids
- enable research ids as kit identifier and disable mocking, send kit in ourhealth realm. go to DSM and observe that ourhealth kit was created with research identifier instead of shortcode.


core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
```diff
@@ -217,6 +222,9 @@ public Enrollee create(Enrollee enrollee) {
         if (enrollee.getShortcode() == null) {
             enrollee.setShortcode(shortcodeService.generateShortcode(null, dao::findOneByShortcode));
         }
+        if (enrollee.getResearchId() == null) {
```
devonbush: might want to use StringUtils.isEmpty here

core/src/main/java/bio/terra/pearl/core/service/participant/EnrolleeService.java
```diff
@@ -229,6 +237,19 @@ public void updateConsented(UUID enrolleeId, boolean consented) {
         logger.info("Updated enrollee consent status: enrollee: {}, consented {}", enrolleeId, consented);
     }
 
+    @Transactional
+    public void updateResearchId(UUID enrolleeId, String researchId) {
+        Optional<Enrollee> enrolleeOpt = dao.find(enrolleeId);
+        if (enrolleeOpt.isPresent()) {
+            // make sure that new research ID is unique, but given that legacy IDs might follow similar formats
+            // across studies (e.g., simple counters), only check uniqueness within the study environment.
```
devonbush, line 246: great comment, thanks!

api-participant/src/test/java/bio/terra/pearl/api/participant/config/MappingConfigurationTest.java
```diff
@@ -0,0 +1,31 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class MappingConfigurationTest extends BaseSpringBootTest {
+  @Autowired ObjectMapper objectMapper;
+
+  @Test
+  public void testObjectMapperDoesNotRenderResearchId() throws JsonProcessingException {
```
devonbush, line 17: At first I thought we might also need a test on the actual API endpoint, but I think what you've done here of testing the wired mapped is quite nice

devonbush (approved):
Great work, worked well locally for me

api-participant/src/test/java/bio/terra/pearl/api/participant/config/MappingConfigurationTest.java
connorlbark: I first tried to make a test on the endpoint, but surprisingly, there's no built-in way to say `get me what rendered json you'd give to the user`, just `give me the object that you would render, and traverse it in a json-like way` - but it never actually uses the object mapper, so any mixins are irrelevant! quite strange.

populate/src/main/java/bio/terra/pearl/populate/service/EnrolleePopulator.java
```diff
@@ -554,6 +554,37 @@ public void createTimeShiftedTasks(Enrollee enrollee, PortalParticipantUser ppUs
...
+    }
+
+
+    public int generateResearchIdsInStudyEnv(StudyEnvironment studyEnv) {
+        List<Enrollee> enrollees = enrolleeService.findByStudyEnvironment(studyEnv.getId());
+
+        int numGenerated = 0;
+
+        for (Enrollee enrollee : enrollees) {
+            if (enrollee.getResearchId() != null) {
```
ashin-broad: Do we need to check for empty here as well?

ui-admin/src/util/zendeskUtils.tsx
```diff
@@ -22,7 +23,8 @@ const ZENDESK_PAGES: Record<DocsKey, string> = {
   PROXY_ENROLLMENT: 'https://broad-juniper.zendesk.com/hc/en-us/articles/26099797590555-Proxy-Enrollment',
   WITHDRAWAL: 'https://broad-juniper.zendesk.com/hc/en-us/articles/20172265139995-How-to-withdraw-an-enrollee',
   DERIVED_VALUES: 'https://broad-juniper.zendesk.com/hc/en-us/articles/29123290381723-Derived-Values',
-  EXPORT_FORMATS: 'https://broad-juniper.zendesk.com/hc/en-us/articles/18259824756123'
+  EXPORT_FORMATS: 'https://broad-juniper.zendesk.com/hc/en-us/articles/18259824756123',
+  RESEARCH_IDS: '' // TODO
```
ashin-broad: Was this intentional?

ashin-broad (approved):

PR #1758, opened 2026-01-06
[JN-1764] persist lang query param in url when changed

connorlbark:

when someone changes their language, add it to the url (e.g., `?lang=es`) to make it easier to share URLs in other languages

### DIFF
--- a/ui-core/src/participant/ParticipantNavbar.tsx
+++ b/ui-core/src/participant/ParticipantNavbar.tsx
@@ -334,13 +334,11 @@ export function LanguageDropdown({ languageOptions, selectedLanguage, changeLang
             return (
               <button key={index} className="dropdown-item" aria-label={lang.languageName}
                 onClick={() => {
-                  //If the user selects a language from the dropdown, remove the lang query param from the URL
-                  if (langQueryParam) {
-                    searchParams.delete('lang')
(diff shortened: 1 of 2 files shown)


ui-core/src/participant/ParticipantNavbar.tsx
```diff
@@ -334,13 +334,13 @@ export function LanguageDropdown({ languageOptions, selectedLanguage, changeLang
             return (
               <button key={index} className="dropdown-item" aria-label={lang.languageName}
                 onClick={() => {
-                  //If the user selects a language from the dropdown, remove the lang query param from the URL
-                  if (langQueryParam) {
-                    searchParams.delete('lang')
-                    navigate({ search: searchParams.toString() })
-                  }
                   changeLanguage(lang.languageCode)
-                  reloadPortal()
```
connorlbark, line 343: this call previously was necessary because simply changing the language here would not always propagate to trigger a full reload of portal content (e.g., surveys, etc.).

however, the portal content provider will always refresh everything if it detects that the `lang` query param differs from the current loaded language, so the call became redundant once we force `navigate`d.

devonbush (approved):

ashin-broad (approved):

PR #1762, opened 2026-02-10
[JN-1769] removing survey links for files uploaded on admin forms

devonbush:

We don't want to show links to participants to admin forms -- they get "hub path not found" messages

image

0. repopulate demo
1. from the admin tool, upload a file for jsalk for the "kit history" admin form
2. log in to participant view as jsalk, confirm the file appears, but not linked to the survey

### DIFF
--- a/ui-participant/src/hub/documents/DocumentLibrary.tsx
+++ b/ui-participant/src/hub/documents/DocumentLibrary.tsx
@@ -168,13 +168,19 @@ const surveyResponseIdsToTaskNames = (
     <div className={'mt-2 text-muted'}>
       <span>{i18n('documentSharedInResponseTo')}:</span>
       <ul>
-        {associatedTasks.map(task =>
-          <li key={task.id}>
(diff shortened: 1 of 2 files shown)


connorlbark (approved):
looks good as is, but do we even want to show study staff forms? I would argue study staff forms should never be visible, even in name only, but I could see it either way.

ui-participant/src/hub/documents/DocumentLibrary.tsx
connorlbark, line 161:
```suggestion
  }).filter((task): task is ParticipantTask => task !== undefined && task.taskType !== 'ADMIN-FORM')
```
(if we wanted to not display admin forms)

devonbush: yeah, I could see this either way -- for OurHealth it works since the study staff form is called PRS upload, so they liked the text,  just wanted the link removed

ashin-broad (approved):

PR #1765, opened 2026-03-11
[SIDRB-4251] adding EU MixPanel routing

devonbush:

We were previously sending all mixpanel events to the US server, which would forward EU project data to EU servers so it wouldn't be stored in the US.  Mixpanel is stopping that forwarding capability in June, so we need to directly send data for EU projects to the Mixpanel EU servers

covered by automated tests

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/logging/MixpanelService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/logging/MixpanelService.java
@@ -21,7 +21,7 @@ import java.util.Map;
 @Service
 @Slf4j
 public class MixpanelService {
-    private static final String EU_DOMAIN_PROJECT = "thehearthive.org";  // if we get a second EU customer, add a useMixpanelEU boolean to PortalEnvironmentConfig
+    private static final String EU_DOMAIN_PROJECT = "thehearthive";  // if we get a second EU customer, add a useMixpanelEU boolean to PortalEnvironmentConfig
     private static final String MIXPANEL_TOKEN_ENV_VAR = "env.mixpanel.token";
     private static final String MIXPANEL_ENABLED_ENV_VAR = "env.mixpanel.enabled";
     private final MixpanelConfig mixpanelConfig;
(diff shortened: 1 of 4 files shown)


devonbush:
CLAUDE:

Branch Review: SIDRB-4251 → development                                                                                                                                
                                                                                                                                                                         
  .github/workflows/require-ticket.yml                                                                                                                                   
                  
  Fixed locally with /(JN|DDO|SIDRB)-\d+/ — correct. Needs to be committed before merging.                                                                               
                  
  MixpanelService.java

  EU routing logic is correct and well-structured. Hardcoded EU_DOMAIN_PROJECT is an acceptable pragmatic decision with the acknowledged comment. No issues.

  MixpanelServiceTests.java

  Tests cover all three delivery paths (global, custom-token, EU). Helper refactoring is clean. No issues.

  CLAUDE.md

  Fine to include.

  ---
  Status: Ready to merge after committing the workflow file fix.

core/src/main/java/bio/terra/pearl/core/service/logging/MixpanelService.java
```diff
@@ -124,6 +146,10 @@ protected String getEventCurrentDomain(JSONObject event) {
         return domain;
     }
 
+    protected boolean isEUProject(String eventDomain) {
+        return EU_DOMAIN_PROJECT.equals(eventDomain);
```
connorlbark: I think this will break  for subdomains, like `sandbox.`, and, perhaps less importantly, for `hearthive.juniper-cmi.org`. Should it be a `eventDomain.contains('hearthive')` instead?

connorlbark (approved):

core/src/main/java/bio/terra/pearl/core/service/logging/MixpanelService.java
devonbush: good catch -- will fix

ashin-broad: Agreed that a `.equals()` might be a bit strict here. Also not sure if just checking for the existence of `hearthive` would be sufficient enough, but I'll leave it to your discretion!

ashin-broad (approved):

PR #1766, opened 2026-04-16
[SIDRB-4361] send reminders for viewed surveys

connorlbark:

reminders weren't being sent properly for hearthive outreach surveys, because they would become viewed when someone opened it. this was being filtered out by the reminder service.


.github/workflows/sonar.yaml
connorlbark, line 1: this was annoying me so I went ahead and just axed it.

(did we get the OK to do this? I remember there being some discussion. can revert this & later work on fixing it, I just remember there being some issues with billing etc...)

ashin-broad (approved):

devonbush:
Cool, this was intentional at the time -- we thought it would be undesirable to send reminders if someone had already looked at the survey.  but it sounds like Rachel wants more reminders

devonbush (approved):

PR #1767, opened 2026-04-16
[SIDRB-4354] allow non-postal matches

connorlbark:

I started to go down the rabbithole of making it possible for participants to override invalid addresses. It was... _not_ simple, as I anticipated.

The key issue is that, wherever used, we had to gate the question via a condition like `suggestedAddress is not null and modalDismissed is false`.

since we have that condition, it won't show the modal unless we have a suggested address.

we _could_ upgrade the modals to handle invalid cases without suggested addresses and change all the conditional expressions, upon release, in all of the forms that use it to show the modal even if no suggested address exists. but.... it started to grow so large in scope, I am opting to move on and consider this "good enough" for 99% of these cases.

Hard to test (unless you know of brand-new developments/neighborhoods off the top of your head). Covered by unit tests.

### DIFF
--- a/scripts/auto_translate.sh
+++ b/scripts/auto_translate.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+if ! command -v trans 2>&1 >/dev/null
+then
+    echo "'trans' does not exist. Please install with 'brew install translate-shell'"
+    exit 1
+fi
+
+if ! command -v jq 2>&1 >/dev/null
+then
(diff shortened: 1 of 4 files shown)


scripts/auto_translate.sh
connorlbark, line 1: sneaking this in from an old pr that never got approved, almost needed it for the frontend since I would have had to add a bunch of new language keys. while I didn't need to, it will be useful in the future.

ashin-broad (approved):

devonbush (approved):

PR #1768, opened 2026-04-19
[JN-1775] allowing survey deactivation

devonbush:

Adds a "decativate" option to all surveys (including irb/sandbox).   This replaces the previous "archive" option in sandbox
image

Deactivation gives the option to either just pause the survey, or to also remove existing participant tasks.
image

Inactive surveys are now also shown in the survey list with an option to activate
image

This also fixes behavior where the participant UX would error if a participant tried to access an inactive survey.  This reflects the realization that admin deactivation of a survey typically is not intended to impact participants who have already completed the survey, but only future enrollees

1. redeploy admin and repopulate demo
2.  go to https://localhost:3000/demo/studies/heartdemo/env/sandbox/forms and deactivate some active surveys
3. confirm deactivated surveys are not offered to new participants


api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConfiguredSurveyController.java
```diff
@@ -126,14 +126,31 @@ public ResponseEntity<Object> create(
...
+      String envName,
+      UUID configuredSurveyId,
+      Boolean cancelTasks) {
+    AdminUser operator = authUtilService.requireAdminUser(request);
+    EnvironmentName environmentName = EnvironmentName.valueOfCaseInsensitive(envName);
+
+    surveyExtService.deactivateSurvey(
+        PortalStudyEnvAuthContext.of(operator, portalShortcode, studyShortcode, environmentName),
+        configuredSurveyId,
+        Boolean.TRUE.equals(cancelTasks));
```
connorlbark: what is this `Boolean.TRUE.equals(` for?

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
```diff
@@ -118,7 +118,13 @@ public SurveyWithResponse findWithActiveResponse(UUID studyEnvId, UUID portalId,
 
         StudyEnvironmentSurvey configSurvey = studyEnvironmentSurveyService
                 .findActiveBySurvey(studyEnvId, stableId)
-                .stream().findFirst().orElseThrow(() -> new NotFoundException("no active survey found"));
+                .stream().findFirst()
+                .orElseGet(() -> studyEnvironmentSurveyService
+                        .findAllWithSurveyNoContent(List.of(studyEnvId), stableId, false)
+                        .stream()
+                        .max(Comparator.comparing(StudyEnvironmentSurvey::getCreatedAt))
+                        .orElseThrow(() -> new NotFoundException("no survey found for stableId: " + stableId)));
```
connorlbark, line 126: curious: what do we need the fallback for? just to make sure if someone goes back to a bookmarked page (or something) that it doesn't break?

I'm a little worried this will interact weirdly with longitudinal surveys, esp. for A-T which has a large number of `REMOVED` responses. I'd want to vet this more deeply before release if we need it. My gut tells me there's some logic in the frontend related to creating new responses from `REMOVED` that this might start hitting edge cases on.

ui-admin/src/study/surveys/DeactivateSurveyModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+import Modal from 'react-bootstrap/Modal'
+import LoadingSpinner from 'util/LoadingSpinner'
+import Api from 'api/api'
+import { Store } from 'react-notifications-component'
+import { failureNotification } from 'util/notifications'
+import { PortalContext, PortalContextT } from 'portal/PortalProvider'
+import { StudyEnvironmentSurveyNamed } from '@juniper/ui-core'
+
+/** renders a modal that allows deactivating a survey, with an option to also cancel existing participant tasks */
+const DeactivateSurveyModal = ({
```
connorlbark, line 12: IMO `Deactivate` makes it sound like this is easily reversible at any time (vs Archive, which is definitely irreversible), but if you cancel the tasks you can't go back from that easily.

What's the use case for canceling tasks? Is there a need for that? 

I wonder if we should just make sure Deactivated survey responses don't end up in the export and don't trigger reminders, and remove the ability to auto-cancel tasks. Then, presumably we could remove the logic about fetching Removed survey responses. It would make this operation feel safer (to me), as you could re-activate and everyone's tasks are still in the same state as they expect.

connorlbark (commented):
Looks great code-wise, but I have a couple thoughts about the approach I wanna talk through first

api-admin/src/main/java/bio/terra/pearl/api/admin/controller/forms/ConfiguredSurveyController.java
devonbush: It's the safest way to convert from Boolean, which the controller takes, to `boolean`--otherwise it will throw an exception on null

core/src/main/java/bio/terra/pearl/core/service/survey/SurveyResponseService.java
devonbush: Yes, right now if a person clicks on a deactivated survey in their dashboard, it will error, even if they filled it out while it was still active.  So this falls back to the latest inactive version if that's the case.  it should have no impact on longitudinal stuff -- this isn't where we're finding the response -- just the survey

ui-admin/src/study/surveys/DeactivateSurveyModal.tsx
devonbush: This is mostly for HeartHive, who regularly creates Outreach surveys advertising other studies, that then they need to pause/deactivate as the other studies close enrollment.  And yes, the idea is that it's easily reversible (if you don't delete the tasks)

MatthewBemis (approved):
:shipit:

connorlbark (approved):

PR #1769, opened 2026-04-21
[SIDRB-4371] fix lang reload.... again

connorlbark:

changing languages doesn't work until you do a full page relaod after changing the language.

### DIFF
--- a/ui-core/src/participant/ParticipantNavbar.tsx
+++ b/ui-core/src/participant/ParticipantNavbar.tsx
@@ -312,7 +312,7 @@ export function LanguageDropdown({ languageOptions, selectedLanguage, changeLang
       changeLanguage(langQueryParam)
       reloadPortal()
     }
-  }, [langQueryParam])
+  }, [langQueryParam, selectedLanguage, languageOptions])
 
   return (
     languageOptions.length > 1 ? (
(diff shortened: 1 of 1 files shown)


ui-core/src/participant/ParticipantNavbar.tsx
```diff
@@ -334,8 +334,6 @@ export function LanguageDropdown({ languageOptions, selectedLanguage, changeLang
             return (
               <button key={index} className="dropdown-item" aria-label={lang.languageName}
                 onClick={() => {
-                  changeLanguage(lang.languageCode)
```
connorlbark, line 337: before I tried to fix this a couple months ago, this was previous paired with a `reloadPortal` (to be honest, I am not sure why I left the `changeLanguage` - I think that was an oversight).

it turns out that both `changeLanguage` and `reloadPortal` need to be called at the same time; calling one without calling the other can lead to weird state issues. in this case, spanish gets set as the selected language in local storage but the portal content uis never loaded. after the navigate, the use effect doesn't trigger the portal reload because the language in local storage matches the query param.

setting `lang` as the key is enough to trigger everything once the page gets reloaded.

ui-core/src/participant/ParticipantNavbar.tsx
```diff
@@ -312,7 +312,7 @@ export function LanguageDropdown({ languageOptions, selectedLanguage, changeLang
       changeLanguage(langQueryParam)
       reloadPortal()
     }
-  }, [langQueryParam])
```
connorlbark, line 315: these are missing from the dependencies list; added them just in case - better to accidentally reload more often than to have stale other languages floating around.

ashin-broad (approved):

devonbush (approved):

PR #1797, opened 2026-06-01
[SIDRB-4453] Fix google translate + make loading language a bit more robust

connorlbark:

1. Stops google translate from touching anything in SurveyJS - which breaks when you change pages with Google trnslate on
2. Makes the i18n provider smart enough to change languages to the user's preferred browser language

### DIFF
--- a/ui-core/src/participant/I18nProvider.tsx
+++ b/ui-core/src/participant/I18nProvider.tsx
@@ -7,6 +7,7 @@ import React, {
 import { useApiContext } from './ApiProvider'
 import { SUPPORT_EMAIL_ADDRESS } from '../util/supportUtils'
 import { EnvironmentName } from 'src/types/study'
+import { PortalEnvironmentLanguage } from 'src/types/portal'
 
 export const I18nContext = createContext<I18nContextT | null>(null)
 
@@ -46,8 +47,20 @@ const SELECTED_LANGUAGE_KEY = 'selectedLanguage'
 /**
  * Provider for the current users i18n context.
  */
-export function I18nProvider({ defaultLanguage, portalShortcode, environmentName, children }: {
-  defaultLanguage: string, portalShortcode?: string, environmentName: EnvironmentName, children: React.ReactNode
+export function I18nProvider({
+  defaultLanguage,
+  portalShortcode,
+  environmentName,
+  loadedPortalContentLang,
+  reloadPortalContent,
+  children
+}: {
+  defaultLanguage: string,
+  portalShortcode?: string,
+  environmentName: EnvironmentName,
+  children: React.ReactNode,
+  loadedPortalContentLang?: string,
+  reloadPortalContent?: (language: string) => void
 }) {
   const Api = useApiContext()
   const [isLoading, setIsLoading] = useState(true)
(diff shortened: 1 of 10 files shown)


ui-core/src/participant/I18nProvider.tsx
```diff
@@ -61,6 +74,12 @@ export function I18nProvider({ defaultLanguage, portalShortcode, environmentName
     localStorage.setItem(SELECTED_LANGUAGE_KEY, language)
   }
 
+  useEffect(() => {
+    if (loadedPortalContentLang !== selectedLanguage) {
+      reloadPortalContent && reloadPortalContent(selectedLanguage)
+    }
+  }, [selectedLanguage, loadedPortalContentLang])
```
connorlbark, line 81: this is the core fix.
forces reload of portal content if it doesn't match selected lanaguage.
this does mean, e.g., if a user's browser defaults to spanish, english portal content will load first then it will detect spanish is an option and will load it second.

ideal solution would mean putting i18n provider above the portalenv provider and moving all language logic (from navbar, etc.) into it - but that would be a bigger refactor

ui-core/src/participant/I18nProvider.tsx
```diff
@@ -130,3 +149,21 @@ const JUNIPER_TO_B2C_LOCALE_MAP: Record<string, string> = {
 export const getB2CLocale = (key: string): string => {
   return JUNIPER_TO_B2C_LOCALE_MAP[key] || 'en'
 }
+
+
+export const inferBrowserDefaultLanguageCode = (languageOptions: PortalEnvironmentLanguage[]) => {
+  const preferredLanguages = [
+    navigator.language || '',
+    ...(navigator.languages || [])
```
connorlbark, line 157: `navigator.language` == user's language preference

ui-participant/src/Navbar.tsx
```diff
@@ -12,14 +12,12 @@ import { UserManager } from 'oidc-client-ts'
...
 import mixpanel from 'mixpanel-browser'
 import { usePortalEnv } from 'providers/PortalProvider'
-import { useActiveUser } from 'providers/ActiveUserProvider'
 
 type NavbarProps = JSX.IntrinsicElements['nav']
 
 /** renders the navbar for participants */
 export default function Navbar(props: NavbarProps) {
-  const { user, logoutUser, proxyRelations } = useUser()
-  const { profile, ppUser } = useActiveUser()
```
connorlbark, line 22: also: instead of reading `preferred language` from the `governed user` account, read it from the proxy account.

this was causing issues when changing participants / extra unnecessary refreshes, and it also makes more sense - if you had >1 proxy, one might know english and the other might not. also, the data migrated from DSM put the preferred language on the proxy - so if those folks logged in, it will find the correct language for them upon getting to their dashboard.

ui-core/src/participant/I18nProvider.tsx
```diff
@@ -130,3 +149,21 @@ const JUNIPER_TO_B2C_LOCALE_MAP: Record<string, string> = {
...
+
+  for (const preferred of preferredLanguages) {
+    const langCode = preferred.split('-')[0]
+
+    if (languageOptions.find(opt => opt.languageCode === langCode)) {
+      return langCode
+    }
+  }
+
+  return 'en'// fallback
```
ashin-broad, line 168: nit

```suggestion
  return 'en' // fallback
```

ashin-broad (approved):

devonbush (approved):
looks good, good work digging in on this

PR #1805, opened 2026-06-17
[SIDRB-4453] fix oauth issues, make preferred language more robust

connorlbark:

We will need to follow this with a SQL update to make sure preferred languages in proxy and governed profiles match.

change your browser settings to make spanish the top language
clear all cookies, session, and local storage for juniper
ensure that the page loads in spanish
register a new account as a self-enrolled user
ensure no oauth errors and app is still in spanish. ensure preferred language matches and emails received in spanish
clear all cookies, session, and local storage
register a new account as a proxy user
ensure no oauth errors, app is still in spanish. ensure preferred language matches for both proxy and governed user and welcome emails recieved in spanish
use language changer to change language to dev
ensure that both proxy and governed user have a preferred language of dev now

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
+++ b/core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
@@ -166,6 +166,7 @@ public class RegistrationService {
         Profile governedProfile =  Profile.builder()
                 .contactEmail(proxyProfile.getContactEmail())
                 .doNotEmail(proxyProfile.isDoNotEmail())
+                .preferredLanguage(proxyProfile.getPreferredLanguage())
                 .givenName(null)
                 .familyName(null)
                 .build();
--- a/ui-participant/src/Navbar.tsx
+++ b/ui-participant/src/Navbar.tsx
@@ -12,12 +12,14 @@ import { UserManager } from 'oidc-client-ts'
 import { getOidcConfig } from 'authConfig'
 import mixpanel from 'mixpanel-browser'
 import { usePortalEnv } from 'providers/PortalProvider'
+import { useActiveUser } from 'providers/ActiveUserProvider'
 
 type NavbarProps = JSX.IntrinsicElements['nav']
 
 /** renders the navbar for participants */
 export default function Navbar(props: NavbarProps) {
   const { user, logoutUser, proxyRelations, profile, ppUsers } = useUser()
+  const { profile: governedProfile, ppUser: governedPpUser } = useActiveUser()
   const { selectedLanguage, changeLanguage } = useI18n()
   const config = useConfig()
   const envSpec = getEnvSpec()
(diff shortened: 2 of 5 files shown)


core/src/main/java/bio/terra/pearl/core/service/workflow/RegistrationService.java
```diff
@@ -166,6 +166,7 @@ public RegistrationResult registerGovernedUser(ParticipantUser proxyUser, Portal
         Profile governedProfile =  Profile.builder()
                 .contactEmail(proxyProfile.getContactEmail())
                 .doNotEmail(proxyProfile.isDoNotEmail())
+                .preferredLanguage(proxyProfile.getPreferredLanguage())
```
connorlbark, line 169: when a proxy user creates a spanish account, the proxy's profile gets set to spanish, but the governed user doesn't.

since emails go through the governed user, this means emails are received in english _until the person manually selects spanish again to trigger an update on the governed user's profile_

this makes sure, on enrollment, the governed user also gets a preferred language of spanish as well, so emails go through as expected

ui-participant/src/providers/PortalProvider.tsx
```diff
@@ -76,7 +78,7 @@ export default function PortalProvider({ children }: { children: React.ReactNode
   }
 
   const reloadPortal = (languageCode?: string) => {
-    const savedLanguage = localStorage.getItem('selectedLanguage')
+    const savedLanguage = localStorage.getItem('selectedLanguage') || returnToLanguage
```
connorlbark, line 81: this is a bandaid of a bandaid

since I started triggering refreshes of portal content when a different language is selected, the initial content loaded upon returning from oauth will be english. if the person's browser is a different language, this will immediately trigger a refresh of portal content. this refresh interferes with the oauth flow in unpredictable ways, and _most of the time_ causes an error.

returnToLanguage is set before we go into the oauth flow with the language the user was browsing in before. thus, upon returning from the oauth flow, it will fetch the correct language initially instead of causing the double-refresh that breaks the oauth flow.

ui-participant/src/Navbar.tsx
```diff
@@ -46,6 +48,14 @@ export default function Navbar(props: NavbarProps) {
...
         ppUserId: ppUser.id
       })
+
+      // if in proxy scenario, also update the governed user
+      if (governedPpUser && governedProfile && governedProfile.id != profile.id) {
+        await Api.updateProfile({
+          profile: { ...governedProfile, preferredLanguage: selectedLanguage },
+          ppUserId: governedPpUser.id
+        })
+      }
```
connorlbark, line 58: finally: when the user changes their language in the frontend, make sure to update proxy and governed user in lockstep so that emails are sent in the right language.

devonbush (approved):

ui-participant/src/login/RedirectFromOAuth.tsx
```diff
@@ -56,6 +56,7 @@ export const RedirectFromOAuth = () => {
       }
 
       if (auth.error) {
+        console.error(auth.error)
```
ashin-broad, line 59: was this intentionally left?

connorlbark: yes. it's impossible to debug otherwise. I'd rather keep this so it's easier to debug in prod/dev

ashin-broad (approved):

PR #1807, opened 2026-07-08
[SIDRB-4528] fix outreach tasks for cross-study proxies

connorlbark:

### DIFF
--- a/ui-participant/src/hub/OutreachTasks.tsx
+++ b/ui-participant/src/hub/OutreachTasks.tsx
@@ -87,6 +87,7 @@ export default function OutreachTasks({ enrollees, studies }: {enrollees: Enroll
   }
 
   useEffect(() => {
+    setOutreachActivities([])
     if (enrollees.length) {
       // the component may get rendered with zero enrollees during login/logout, don't bother fetching tasks then
       loadOutreachActivities()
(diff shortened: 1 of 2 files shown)


ui-participant/src/hub/OutreachTasks.tsx
```diff
@@ -138,5 +144,5 @@ const studyForTask = (task: ParticipantTask, studies: Study[]) => {
 
 /** finds the enrollee that corresponds to the given task */
 const enrolleeForTask = (task: ParticipantTask, enrollees: Enrollee[]) => {
-  return enrollees.find(enrollee => enrollee.studyEnvironmentId === task.studyEnvironmentId)!
+  return enrollees.find(enrollee => enrollee.id === task.enrolleeId)
```
connorlbark, line 148: not sure why this method matched on study env id and not enrollee id, but regardless the governed user was not in all the same study envs so some of these were returning undefined & blowing things up.

also updated to match on enrollee id so we don't accidentally try to mark the outreach tasks as viewed for  enrollees that don't have the task.

ashin-broad (approved):

ui-participant/src/hub/OutreachTasks.tsx
```diff
@@ -54,7 +55,7 @@ export default function OutreachTasks({ enrollees, studies }: {enrollees: Enroll
 
   const sortedOutreachTasks = outreachTasks.sort((a, b) => {
     return a.task.createdAt - b.task.createdAt
-  }).filter(({ task }) => task.status !== 'COMPLETE')
+  }).filter(({ task }) => task.status !== 'COMPLETE' && isTaskVisible(task))
```
connorlbark, line 58: drive by fix: `REMOVED` was not being removed from dashboards.

devonbush (approved):
nice!  and wow, that is a PR title--Juniper continues to surprise me with how complex it got

PR #1812, opened 2026-07-27
[SIDRB-4559] cross-study answer referencing

connorlbark:

allows usage of `{otherSurvey['otherStudy'].question}` in surveys to grab values across study boundaries for hearthive.

### DIFF
--- a/core/src/main/java/bio/terra/pearl/core/model/survey/SurveyWithResponse.java
+++ b/core/src/main/java/bio/terra/pearl/core/model/survey/SurveyWithResponse.java
@@ -1,11 +1,13 @@
 package bio.terra.pearl.core.model.survey;
 
-import java.util.List;
+import java.util.Map;
 
 /** convenience class for grouping together a form and its configuration and most recent/active response */
 public record SurveyWithResponse(StudyEnvironmentSurvey studyEnvironmentSurvey,
                                  SurveyResponse surveyResponse,
-                                 // if the survey uses answers from other surveys, this will be populated
-                                 List<Answer> referencedAnswers) {
+                                 // if the survey uses answers from other surveys, this will be populated.
+                                 // keyed by the full variable name (e.g. "survey1.question1" or "survey1['otherStudy'].question1")
+                                 // so the frontend can set the surveyJS variable directly without needing to reconstruct it
+                                 Map<String, Answer> referencedAnswers) {
 }
 
--- a/ui-participant/src/studies/enroll/PreEnroll.tsx
+++ b/ui-participant/src/studies/enroll/PreEnroll.tsx
@@ -110,7 +110,7 @@ export function EligiblePreEnrollView({ enrollContext, survey, eligibilityResult
       proxyProfile,
       studyEnvParams: { envName: studyEnv.environmentName, studyShortcode, portalShortcode },
       enrolleeShortcode: enrollee?.shortcode || '',
-      referencedAnswers: [],
+      referencedAnswers: {},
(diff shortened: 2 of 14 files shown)


ui-participant/src/studies/enroll/PreEnroll.tsx
```diff
@@ -36,46 +37,59 @@ const ENROLLMENT_QUALIFIED_VARIABLE = 'qualified'
...
+  const [isLoading, setIsLoading] = useState(!!eligibleRule || !!user)
   const [eligibilityResult, setEligibilityResult] = useState<EligibilityResult>({ eligible: !!eligibleRule })
+  const [referencedAnswers, setReferencedAnswers] = useState<Record<string, Answer>>({})
 
-  const checkEligibility = async () => {
-    if (eligibleRule) {
-      setIsLoading(true)
-      try {
+  const loadPreEnrollData = async () => {
+    if (!eligibleRule && !user) {
```
ashin-broad, line 49: if there is no user should we do something like `setReferencedAnswers({})`?

ui-participant/src/studies/enroll/PreEnroll.tsx
```diff
@@ -36,46 +37,59 @@ const ENROLLMENT_QUALIFIED_VARIABLE = 'qualified'
...
+      return
+    }
+    setIsLoading(true)
+    try {
+      if (eligibleRule) {
         const response = await Api.checkEligible(enrollContext.studyShortcode)
         setEligibilityResult(response)
+      }
+      if (user) {
```
ashin-broad, line 58: Same as above, should we act on if a user doesn't exist by doing  `setReferencedAnswers({})`?

ashin-broad (approved):
Approving with a question

ui-participant/src/studies/enroll/PreEnroll.tsx
connorlbark: it's the default value and a user's logged in state should never change during a pre-enroll so it should be safe to rely on the default

core/src/main/java/bio/terra/pearl/core/model/survey/SurveyWithResponse.java
```diff
@@ -1,11 +1,13 @@
...
-import java.util.List;
+import java.util.Map;
 
 /** convenience class for grouping together a form and its configuration and most recent/active response */
 public record SurveyWithResponse(StudyEnvironmentSurvey studyEnvironmentSurvey,
                                  SurveyResponse surveyResponse,
-                                 // if the survey uses answers from other surveys, this will be populated
-                                 List<Answer> referencedAnswers) {
+                                 // if the survey uses answers from other surveys, this will be populated.
+                                 // keyed by the full variable name (e.g. "survey1.question1" or "survey1['otherStudy'].question1")
```
MatthewBemis, line 9: not that it matters a ton, but would `otherStudy['survey1'].question1` read a bit more natural? to me that would read as "accessing `survey1` within `otherStudy`" whereas right now it sort of feels backwards

MatthewBemis (approved):
looks reasonable to me! i left a comment about the syntax but feel free to ignore 😄

core/src/main/java/bio/terra/pearl/core/model/survey/SurveyWithResponse.java
connorlbark: yeah, I totally agree in retrospect, but this is matching the existing search expression syntax that hearthive is already using

PR #1815, opened 2026-08-13
[SIDRB-4611] allow showCompleteButton for outreach studies

connorlbark:

### DIFF
--- a/ui-core/src/surveyUtils.tsx
+++ b/ui-core/src/surveyUtils.tsx
@@ -400,7 +400,8 @@ export function useSurveyJSModel(
     refreshData: SurveyJsResumeData | null,
     pagerPageNumber: number | null,
     readonly?: boolean) {
-    const newSurveyModel = surveyJSModelFromForm(form)
+    const formContent = getFormContent(form)
+    const newSurveyModel = surveyJSModelFromFormContent(formContent)
 
     Object.entries(extraCssClasses).forEach(([elementPath, className]) => {
       set(newSurveyModel.css, elementPath, classNames(get(newSurveyModel.css, elementPath), className))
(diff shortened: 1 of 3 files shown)


ui-core/src/surveyUtils.tsx
```diff
@@ -430,6 +431,10 @@ export function useSurveyJSModel(
     newSurveyModel.completedHtml = '<div></div>'  // the application UX will handle showing any needed messages
     newSurveyModel.onServerValidateQuestions.add(createAddressValidator(addr => Api.validateAddress(addr), i18n))
 
+    if (form.surveyType === 'OUTREACH' && formContent?.showCompleteButton !== true) {
+      newSurveyModel.showCompleteButton = false
+    }
```
connorlbark, line 439: instead of a blanket CSS hide of complete buttons, allows users to set `showCompleteButton = true` to override the default.

ui-core/src/surveyUtils.tsx
```diff
@@ -400,14 +401,17 @@ export function useSurveyJSModel(
...
       set(newSurveyModel.css, elementPath, classNames(get(newSurveyModel.css, elementPath), className))
     })
 
-    if (refreshData) {
-      newSurveyModel.data = refreshData.data
+    if (refreshData && !isEmpty(refreshData.data)) {
+      for (const key in refreshData.data) {
+        newSurveyModel.setValue(key, refreshData.data[key])
+      }
     }
```
connorlbark, line 415: DRIVE BY FIX: this one was a doozy to find, but luckily easy to fix. but we were running into this issue:
https://github.com/surveyjs/survey-library/issues/1149

something about the blanket `data` set was causing some race conditions with certain default expressions. by setting one answer at a time instead, it never conflicts with any of the default values set.

which is weird because it works for most default value expressions - but something about the expression hearthive was setting up was running up against this issue.

devonbush (approved):

ashin-broad (approved):

